PR #1, opened 2015-03-26
Specify python version required in readme

codemac:
Just a small silly change. Ran into this on linux where python3 is the standard install, and ran python scripts/mk_make.py without thinking about it.

Sorry if pull requests are not the desirable collaboration tool. Feel free to close this.

### DIFF
--- a/README
+++ b/README
@@ -8,7 +8,7 @@ Z3 can be built using Visual Studio Command Prompt and make/g++.
    cd build
    nmake
 
-2) Building Z3 using make/g++ and Python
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
updated readme

wintersteiger:
Sorry about the confusion. We do support python 3 but we don't test it often enough. I just checked and for me everything worked fine. 

Of course, pull requests is the right way to submit such changes; in this case we just preferred to fix the problem in a different way.

PR #12, opened 2015-03-28
Cleanup: avoid defining reserved identifiers

pcc:
These can cause problems when building against certain C++ standard
libraries. In particular, some versions of libstdc++ use __out as a variable
name, which conflicts with the SAL __out macro.

Also start using [[clang::fallthrough]] when compiling with Clang.

### DIFF
--- a/doc/z3api.dox
+++ b/doc/z3api.dox
@@ -1567,7 +1567,7 @@ PREDEFINED             = Z3_ast_opt:=Z3_ast \
                          Z3_bool_opt:=Z3_bool \
                          Z3_func_interp_opt:=Z3_func_interp \
                          Z3_model_opt:=Z3_model \
-                         __out_opt:=__out
+                         Z3_out_opt:=Z3_out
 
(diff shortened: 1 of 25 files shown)


nunoplopes:
I'm also seeing build failures with gcc 4.8. The bundled libstd++ uses __in in some headers.
This patch fixes the issues here.

wintersteiger:
This looks good to me. @nunoplopes could you take another look at this to check that it doesn't break anything important? 

@pcc if it's not too much trouble, could I get you to sign a [CLA](https://github.com/Z3Prover/z3/wiki/Contribution-Guidelines)?

nunoplopes:
This patch is obsolete now.  I've already removed all the __in*/__out* stuff.  We are only missing the fallthrough. I will add that part.

wintersteiger:
Great, thanks!

pcc:
Closing since it looks like this pull request has been/is being obsoleted by other changes.

PR #11, opened 2015-03-28
Corrected typo: interger -> integer

Confusion:
Note while reading documentation

PR #24, opened 2015-04-03
Static analysis cleanup and open issues

daniel-j-h:
I used LLVM's (3.6) static analyzer and clang's (3.6) Weverything flag in order to check the source.

There are minor issues I could fix myself, e.g. undefined behavior when doing 1 << 31.
(Interestingly this is not UB anymore in C++14, see: http://stackoverflow.com/q/26331035)

This increment looks suspicious, as it is dead code:
https://github.com/Z3Prover/z3/blob/d01c3491a67d4a06ea1f6e57d385f04914529317/src/ast/used_vars.cpp#L106-L112

Clarify operator precedence: && trumps ||: Is this correct?
https://github.com/Z3Prover/z3/blob/d01c3491a67d4a06ea1f6e57d385f04914529317/src/ast/fpa/fpa2bv_converter.cpp#L1883-L1884

Undefined behavior in reinterpret_cast's, one example is this, but there are more:
https://github.com/Z3Prover/z3/blob/d01c3491a67d4a06ea1f6e57d385f04914529317/src/util/hwf.cpp#L161
> "dereference of type 'const uint64 *' (aka 'const unsigned long long *') that was reinterpret_cast from type 'double *' has undefined behavior"

-Wundefined-reinterpret-cast finds those.

This loop is executed only a single time. Always. Is this correct?
https://github.com/Z3Prover/z3/blob/d01c3491a67d4a06ea1f6e57d385f04914529317/src/util/prime_generator.cpp#L70

Note: based on the comment by wintersteiger in https://github.com/Z3Prover/z3/pull/11 this pull request is against the unstable branch.

Would you please be so kind and take a look at those issues -- maybe if you have the time do a static analysis and Weverything build yourself and check the left over warnings.

Thanks,
Daniel


wintersteiger:
Thanks for the patches, those all look good to me. I've moved some of the other issues into separate issues to be looked at later.

wintersteiger:
@daniel-j-h if it's not too much trouble, could I get you to sign a [CLA](https://github.com/Z3Prover/z3/wiki/Contribution-Guidelines) before we merge this?

daniel-j-h:
@wintersteiger signed.

wintersteiger:
Sorry for the premature closure of this. I'm now merging it into master.

wintersteiger:
Merged as of 6749c19ab11be85fe0f00dba3544e63dbdbf5c77

Issue #27, opened 2015-04-07
Questions about and suggestions for build system


daniel-j-h:
The build system currently fails with python3 as the default python interpreter (which is the case e.g. for debian jessie, gentoo, arch, ..), since the mk/* scripts use deprecated python2 syntax.

I saw comments about the python2 requirement being added and then again removed from the readme; but as far as I see it, python2 is still required.  Is this intended? What are the assumptions about the target platforms? python2 and optionally python3? If so, which versions do you want to support?

There is also at least one argument mismatch that I encountered: --help lists the "--makefiles" flag, but if you specify it, the script fails. Looking through the source reveals why: "--onlymakefiles" is the flag's name and not "--makefiles".

Now after trying "--onlymakefiles" the script still fails, this time because it's unable to find a c++ file, rendering this flag completely useless.

Is there a reason for not using e.g. python's argparse and calling it a day?

Please don't see this as criticism, I just want to understand the reasons behind the build system's design. I'm more than glad to help out with the Linux specific parts, if you need any help -- but for this I need a more precise specification of what you actually want :)

wintersteiger:
We have always supported Python 3, but we test it very rarely. We already have far too many combinations of systems and configurations to test every night and for some reason we all ended up using Python 2. I just committed a couple fixes that should make it work with 3 again.

I fixed the --onlymakefiles problem (it's now --makefiles as advertised). I didn't see any errors when running it. Could it be that you or me had some stale files around?

No reason for not using argparse as far as I know. 

We haven't discussed anything that we would want and don't have in the build system. One of the important reasons behind many design decisions was that the primary target up until now was Windows, 32 bit. Python scripts seemed like the best solution to get some sort of unified cross-platform system. Previously, configure scripts and Makefiles would never be in sync with the Visual Studio solution files, so now we generate both from the same Python scripts.

daniel-j-h:
Thanks for the explanations. Regarding your idea about stale files, I re-checked the issue:
With a freshly cleaned repository, unstable branch as of b7bb53406f7807daddf025d7849b2cf7b49c1fb4:

    ./configure --githash=$(git rev-parse HEAD) --makefiles --staticlib

```
Writing build/Makefile
Traceback (most recent call last):
  File "scripts/mk_make.py", line 20, in <module>
    mk_makefile()
  File "/tmp/z3-buildsystem/scripts/mk_util.py", line 1974, in mk_makefile
    c.mk_makefile(out)
[...]

PR #33, opened 2015-04-09
Minor python-related style fixes

daniel-j-h:
Initially I wanted to improve the build system; but then I realized it's a > 5k loc subproject in itself..

So here it goes; with a few minor issues, such as indentation with 3 spaces mixed in and a lot of un-pythonic style, such as semi-colon at end of line.

See #27

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -248,7 +248,7 @@ def check_java():
     if is_verbose():
         print("Finding javac ...")
 
-    if JDK_HOME != None:
+    if JDK_HOME is not None:
         if IS_WINDOWS:
             JAVAC = os.path.join(JDK_HOME, 'bin', 'javac.exe')
         else:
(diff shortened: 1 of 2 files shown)


wintersteiger:
In Python, is there any difference in semantics between `if q != None ...` and ` if q is not None ...`? If not, then I'd prefer to keep `!=`. You're definitely right about the semicolons though, those should indeed be removed.

daniel-j-h:
`if q is not None` is the idiomatic way of doing it, see also: http://stackoverflow.com/a/2209781

It does not make a difference here from a quick glance (since the changes are a bit dated by now).

cheshire:
`!=` will be probably slower, as it will check for overloaded equality method.

wintersteiger:
That sounds good, then we'll switch those to the preferred `is not`. Since this touches a lot of lines of code, I might have to run some tests before I can get that merged. 

Also, if it's not too much trouble, could I get you to sign a [CLA](https://github.com/Z3Prover/z3/wiki/Contribution-Guidelines)?

daniel-j-h:
@wintersteiger signed.

wintersteiger:
Sorry for the premature closer; this is being merged now.

wintersteiger:
Merged into master as of df1f60062498e43c8924575641f5f326d3c17cfa

PR #37, opened 2015-04-12
Add the Optimize module for the ml api.

Drup:
I tested, it seems to work very nicely!

Some remarks:

- I'm not fond of weight as a string. I'm considering replacing by an int.
- The functions to get a proof (SAT) and a reason (UNKNOWN) are lacking.
- I abstracted the fact that optimization objectives were int.

### DIFF
--- a/src/api/ml/z3.ml
+++ b/src/api/ml/z3.ml
@@ -2782,6 +2782,80 @@ struct
   let mk_fixedpoint ( ctx : context ) = create ctx
 end
 
+module Optimize =
+struct
+  type optimize = z3_native_object
+
+  type objective = int
+
+  let create ( ctx : context ) =
+    let res : optimize = { m_ctx = ctx ;
+			     m_n_obj = null ;
+			     inc_ref = Z3native.fixedpoint_inc_ref ;
+			     dec_ref = Z3native.fixedpoint_dec_ref } in
+    (z3obj_sno res ctx (Z3native.mk_optimize (context_gno ctx))) ;
+    (z3obj_create res) ;
+    res
+
+  let get_help ( x : optimize ) = Z3native.optimize_get_help (z3obj_gnc x) (z3obj_gno x)
+
+  let set_parameters ( x : optimize ) ( p : Params.params )=
+    Z3native.optimize_set_params (z3obj_gnc x) (z3obj_gno x) (z3obj_gno p)
+
+  let get_param_descrs ( x : optimize ) =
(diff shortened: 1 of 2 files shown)


Drup:
I get a bug when trying to retrieve the model after `Z3_optimize_check` when the optimization objective is unbounded. 

My example is `assert(x ≥ 0 && y ≥ 0)` and `maximize(x+y)`. I get an `Z3native.Exception("invalid argument")`.

Am I doing something wrong on the ml binding side ? I don't understand the issue.

It seems to work in the python API.

Drup:
I found the issue ... The issue was not the model extraction, but the print of the value of the upper bound of the optimization objective just after. It seems none of the specialized printer (like get_numeral_string) accept to print infinity except ast_to_string.

Can we get a function in the API to test if a number is infinite ?
Also, what is the expected sort of the expression taken by minimize/maximize ? Int or real ?

NikolajBjorner:
the ast returned may be of one of the following forms:
- infinity
- -infinity
- constant
- numeric constant +/- numeric constant * epsilon

epsilon and infinity are fresh constants. 
For example if you maximize 
(assert (< x 0))
(assert (< y x))
then x set to -epsilon and y is -2*epsilon (or something like this)

you can maximize, minimize linear integer, real or bit-vector expressions.
Bit-vector objectives are treated as syntactic sugar for weighted soft constraints.

Drup:
I see. 

The sort of infinity seems to always be int, which is a bit inconvenient (infinity is not representable as an int in most big int libraries, as opposed to as a rational). I think the sort of the value of optimization objective's value should be the same of the expression maximized/minimized.

 Also, I'm still not sure how to test that the result is, indeed, infinity (or one of the other possibility) using the C API (and not [look at the string](https://github.com/termite-analyser/termite/blob/master/src/monodimMultiPc.ml#L195)....)

Drup:
Anything blocking this to be merged ?

wintersteiger:
Yes, the license change hasn't fully propagated yet, we need to resolve a couple more issues before we can accept pull requests.

Drup:
Oh, ok!

wintersteiger:
Sorry about the early closure, the opt branch has been removed while the pull request hadn't been merged yet. Could you perhaps submit a new PR based on unstable? Thanks!

wintersteiger:
Superseded by eb5af100bda2a799669993dbd32c36dcc3d12260

PR #49, opened 2015-04-21
use a base iz3_exception class for exceptions raised during interpolation

agriggio:
Using a base exception class, derived from z3_exception, makes it possible to
recover gracefully if something goes wrong during the computation of
interpolants.

### DIFF
--- a/src/interp/iz3base.h
+++ b/src/interp/iz3base.h
@@ -117,7 +117,7 @@ class iz3base : public iz3mgr, public scopes {
 
     /** Interpolator for clauses, to be implemented */
     virtual void interpolate_clause(std::vector<ast> &lits, std::vector<ast> &itps){
-        throw "no interpolator";
+        throw iz3_exception("no interpolator");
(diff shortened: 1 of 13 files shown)


wintersteiger:
That sounds like a good idea; @kenmcmil can you take a look to check this doesn't break something improtant?

wintersteiger:
@agriggio if it's not too much trouble, could I get you to sign a [CLA](https://github.com/Z3Prover/z3/wiki/Contribution-Guidelines) before we merge this?

agriggio:
I don't think it's necessary, I worked on this while I was visiting MSR, when I was an employee.

wintersteiger:
Aha! Of course, in that case we won't need the CLA.

wintersteiger:
CLA not required, MSFT employee at the time.

wintersteiger:
Merged as of b7387ed7ac5646717a35d92e40d23b3cc8548cb4

PR #55, opened 2015-04-23
support openbsd

zshipko:


wintersteiger:
Looks good to me, thanks for the patch, will be merged soon!

zshipko:
Great, thank you!

hogand:
I don't think this is quite right.  It will let you run `make`.  But if you try to run `make examples`, it rebuilds a lot of z3 along with the examples.  It takes about 1 minute to make the examples with this diff.

Here's what I have been using: https://github.com/hogand/z3/commit/ab4b72b94bd368fbceafc503b7302663b0dd6b62

OpenBSD replaced "sudo" with "doas" in the base system, but I don't think that's worth updating the READMEs for.  The users should already be used to that change.

zshipko:
Thanks, @hogand. That does look like a much more complete solution.

wintersteiger:
Thanks, this now merged.

Hogand, if you don't mind your more complete fix to be integrated, could you submit a separate pull request for that? Thanks!

Issue #72, opened 2015-05-05
Z3 isn't pip-installable


zardus:
Most python packages nowadays can be installed via pip. For z3, this would look like:

```ShellSession
# pip install z3
```

This is super awesome not only because it makes it super easy to install z3, but also because it lets people easily put z3 in the requirements.txt of their packages, so that pip will automatically install it as a dependency, which will be cool beyond belief!

There are two steps required to support this:

1. Create a `setup.py` that installs the package.
2. Register and upload a build to https://pypi.python.org

I'll submit a pull request for 1 shortly. I don't think it's my place to do 2, but once 1 is done, it's a matter of a few commands that you can run to register and upload to pypi.

zardus:
Opened PR #73 that implement part 1.

NikolajBjorner:
This issue is super-seeded by #288

PR #73, opened 2015-05-05
Pythonic installation support.

zardus:
This pull request adds a `setup.py` file for easy installation in python. The setup.py file reuses the current `./configure; cd build; make` workflow and mostly replaces `make install`. Currently, the version number is manually set, but it can probably be changed to get fetched automatically from the `mk_*` scripts. Also, I wasn't sure what to put for "author" and "maintainer", so I left those out.

As noted in #72, it's the first step toward `pip install z3`. Right now, one can do:

```ShellSession
# pip install git+https://github.com/zardus/z3@py-setup
```

If this PR is merged (eventually into master, I guess?), the following will be possible:

```ShellSession
# pip install git+https://github.com/Z3Prover/z3
```

The ultimate goal of `pip install z3` will require registering the package with pypi.

Some other bonuses:

Binary distributions can be built and installed with:

```ShellSession
# pip wheel git+https://github.com/zardus/z3@py-setup
# pip install /path/to/z3-4.4-py2-none-any.whl
```

Once z3 is registered to pypi, (probably hackish) deb packages can be created/installed with:

```ShellSession
# sudo apt-get install python-stdeb
# pypi-install z3
```


zardus:
I should say that the setup.py isn't pip-specific. `python setup.py install` works if you've already cloned the repo and want to avoid pip magic.

cao:
+1

wintersteiger:
Thanks for the contribution! You should put your own name into the author/maintainer fields, if you would actually like to do that, otherwise we should someone else to maintain it.

zardus:
I'm totally down to do maintain the `setup.py`. My only qualm was that the description of those fields (https://docs.python.org/2/distutils/apiref.html#distutils.core.setup) are "The name of the package author" and "The name of the current maintainer", and I didn't want to give the impression that Z3 is somehow my work.

How about putting something like "The Z3 Project" for the author field and me for the maintainer field? I'll push a commit with that if that sounds good to you!

wintersteiger:
Oh, I see! Yes, I suppose "The Z3 Theorem Prover Project" or a subset of those words would be the right thing to use for the author field. There are other projects/products out there called Z3 (e.g., BMW, SONY, Blackberry), so maybe we should include the "Theorem Prover" to make it easier to identify what it is.

zardus:
Alright, added! I like the author name; it's consistent with the project name in `setup.py`, too.

setup.py
```diff
@@ -0,0 +1,51 @@
...
+import subprocess
+import multiprocessing
+from distutils.core import setup
+from distutils.errors import LibError
+from distutils.command.build import build as _build
+
+class build(_build):
+    @staticmethod
+    def _configure():
+        if subprocess.call(['./configure']) != 0:
```
asmeurer: Shouldn't this do `['./configure', '--prefix', sys.prefix]` or something similar?

asmeurer:
Does this work on Windows?

setup.py
zardus: If `--prefix` isn't specified, z3 grabs it from distutils, in mk_util.py (`PYTHON_PACKAGE_DIR=distutils.sysconfig.get_python_lib()`). I was torn between being more explicit about it (and passing the prefix option) vs having less to track in `setup.py`. In the end, I went with the exact procedure I used to do manual installs, but I could see it maybe being better to be explicit.

zardus:
As-is, I don't think it'll work on Windows. The docs say that `configure` followed by `nmake` should work, as long as your environment is properly set up (with Visual Studio and the like), so it should be a pretty minor change. I don't have a Windows VM with VS to test with at the moment, though.

It *should* work as-is in a cygwin environment.

setup.py
asmeurer: Also it would probably be better to use `cwd=dirname(__file__)` rather than assuming the setup.py is run from its containing directory.

asmeurer: But what's to guarantee that `mk_util.py` is run with the same Python as setup.py?

asmeurer:
FYI I'm having difficulty installing the latest release on Windows because it seems to be having trouble with the space in my path where Visual Studio is installed.

asmeurer:
Regarding Windows, I think it's important to use distutils so that the right version of Visual Studio is used, i.e., the same one that was used to compile Python (typically 2008 for Python 2 and 2010 for Python 3).

setup.py
zardus: My understanding is that `setup.py` is always run with its directory as the current working directory. Most of the setup.pies I've seen (not that many, tbh) make this assumption. That being said, it's not too hard to fix.

Good point about the Python executable!

I made `setup.py` path-aware and custom Python-aware (in terms of `mk_util.py`) in commit a7ab040cdf1865585bdad6f2d294d744db94b63d.

zardus:
Is the space causing issues in `setup.py` or in `configure`? I don't see where it'd choke in `setup.py`, but I don't have a Windows+VS install anywhere :-(

asmeurer:
The space thing is in the current tagged release (I don't remember what step in the install failed; I'll check later and post an error log).

asmeurer:
By the way, different Python interpreters aren't so exotic. My use-case is to build a [conda package for z3](https://github.com/conda/conda-recipes/tree/master/z3) but for conda to work the Python that runs setup.py must be the same prefix where everything gets installed. But in general, every conda environment has an "exotic" Python interpreter in it.

zardus:
Oh, that was just tongue in cheek :-). I personally use pypy quite a bit.
My setup usually has pypy linked as `python` in the virtualenv (and, thus,
in my PATH), and z3 uses the python in the PATH by default, so your
use-case (of, I guess, a different python than the one in the path?) hadn't
occurred to me.

asmeurer:
Yes ideally everything should work without being in the PATH. It should all just reference things relative to the install prefix.

zachriggle:
Does z3 support Python3?  Currently zardus' script fails for `pip3 install git+https://github.com/zardus/z3.git@py-setup`.

zardus:
Python3 doesn't support mixed use of tabs and spaces in indentation, which, unfortunately, happens in `mk_util.py`. There are a few pull requests to fix it, including #75.

On top of that, it looks like python3 chokes on the binary `z3` script. I pushed 80d1ff8 to skip it on python3.

Until these #75 gets merged in, I made a combined branch of them that you can install with:

```bash
pip3 install git+https://github.com/zardus/z3.git@pypy-and-python3-setup
```

(I also threw in some pypy fixes, which is another PR around here somewhere).

Works on python3 for me!

zachriggle:
Ping on this :)

NikolajBjorner:
Thanks for the reminder and sorry for the delay.

zardus:
FYI, we recently released a binary analysis framework (http://angr.io) that uses Z3 to do constraint solves. In order to make the framework pip-installable, I pushed our pip-installable branch of Z3 to pypi as `angr-z3` so that we can list it in the various `requirements.txt` files. We'll switch over to upstream once this is merged in, of course. Just wanted to give you a heads up in case you see people doing `pip install angr-z3` and so forth!

musically-ut:
Ping on this. :)

zardus:
If there's any paperwork I need to fill out (I heard the term `CLA` being thrown around by a friend at Microsoft?) to speed up this merge, please let me know!

musically-ut:
@zardus This PR doesn't contain changes from https://github.com/angr/angr-z3/pull/1

wintersteiger:
@zardus Indeed, if you're happy to sign a CLA then we won't have to think about making exceptions etc. Take a look at our [contribution guidelines](https://github.com/Z3Prover/z3/wiki/Contribution-Guidelines) which also contain the link to the CLA electronic signature website.

zardus:
Cool, I've now signed the CLA :-)

I also pushed the commit from https://github.com/angr/angr-z3/pull/1, but I guess this means @cao has to sign the CLA as well?

Please let me know if there's anything else we can do to speed up merging this PR!

cao:
I signed the CLA also.

It would be good to add Windows support (use the .dll instead of the .so), so that we don't need to open another pull request for it.

zardus:
I agree. No windows here, though :-(

zardus:
Okay, thanks go @cao, we now have Windows pip installation support!

scripts/mk_util.py
```diff
@@ -2177,6 +2189,8 @@ def mk_pat_db():
     for line in fin:
         fout.write('"%s\\n"\n' % line.strip('\n'))
     fout.write(';\n')
+    fin.close()
```
asmeurer: Why don't you use `with open` in all these places?

NikolajBjorner:
' with open' - thanks, I will use this too.
BTW, the header files are listed also in mk_project.py (but in 3 places and in local variables). 
I would factor out these header files now that other modules use them.

wintersteiger:
Could you perhaps move all pip related scripts into scripts/packages/pip instead of the main directory? I expect that we will end up with packaging scripts for many other systems, and we should get those organized right from the start.

zardus:
It looks like that's actually pretty tricky. There seems to be an assumption (in pip and the like) that setup.py only needs stuff lower in the directory structure. For example, if I do `pip install /path/to/z3/scripts/packages/pip`, pip creates the build directory from that directory, so the rest of Z3 is nowhere to be found.

One possible solution would be to keep the setup scripts in another repo, and have them clone the main repo if needed. From personal experience, this is a bit of a pain in terms of maintenance, usability, and support.

wintersteiger:
Sure, a separate repository would do the trick. But, maybe we can also just create a symlink to the upper directory, which may be a bit tricky on Windows. Perhaps just make a copy of the src directory, that would at least be quicker than checking out the repository?

zardus:
I did try the symlinks, but it's not quite that simple. Basically, when installing from the filesystem, pip will actually copy the directory containing setup.py to a temporary directory in tmp. After that, the relative symlinks are broken. Before that, as far as I can tell, we don't get an opportunity to run any code.

The best I've gotten now is something that works only for "editable installs" (essentially, a link to the Z3 python modules from python's site-packages) and for pre-packaged pip distribution -- installing the git version with `pip install git+https://github.com/Z3Prover/z3` won't work. That's not *too* bad, I guess -- if people want to stay on the bleeding edge, they'll have to clone the git repo -- but it's not as clean as other (more python-specific) projects tend to be.

That might still be better than the separate repo just for setup.py. As I said, in my experience that comes with its own set of issues in regards to keeping two weirdly-related repositories synchronized. Even if you just download a src zip, you gotta track *which* one to download, track version numbers across both repos, deal with supporting people for whom the download or extraction fails, and so forth.

IMO, of course, the best way to go is to have the `setup.py` in the repo root. But this is all your call, of course :-)

wintersteiger:
Great, thanks for looking into this! Another way would be to copy the .py when the mk_make.py creates the build directory. 

In the end, who has to run this script? Is it only when a new package is created and added to the pip world of packages, or will it run every time a user installs the package?

wintersteiger:
Sorry for the premature closer, reopening.

wintersteiger:
This one wasn't resolved yet. Can you clarify whether all users who want to install Z3 will have to run this script, or whether this only needs to be done once by whoever publishes the package? Do I understand correctly, that setup.py will not be needed when the package is finally registered in pypi? Also, what is the difference between pip, pypy, and pypi, will all of those require different setup.py files in different directories? Is https://www.pythonpackages.com/ involved in any way and possibly others as well?

asmeurer:
PyPI is the official index (https://pypi.python.org/pypi). It's where you should upload the package. The typical way to do this is to use `python setup.py sdist`, `python setup.py register`, and `python setup.py upload` (sdist means source tarball; it's also possible to build binary distributions called wheels, but I would worry about figuring that out after you get everything else working). 

`pip` is the tool that installs packages from PyPI. So most people will install the Python bindings with `pip install z3`. They can also download the source from PyPI manually and run `python setup.py install`. 

PyPy is an alternative implementation of Python that is focused on performance. The only relationship it has to PyPI is that people often (mis)pronounce PyPI as "PyPy". 

https://www.pythonpackages.com/ looks like it is just a little tool that gathers data from PyPI. It isn't necessary to use (I've never even heard of it before). 

What you want to do here is make sure that `python setup.py sdist` works and creates a tarball that can be installed on its own.

wintersteiger:
Alright, thanks for the quick answers!

Am I right about setup.py being only required for the person who creates the package and after that the package is uploaded to PyPI, where everybody can download it, but they won't have to run setup.py? 

If end-users have to run setup.py or any other files and scripts, can we save those in scripts/packages/pypi/* and only the person packaging the destribution will then copy them to the root directory?

Also, which of the other files in [Packaging and Distributing Projects](https://packaging.python.org/en/latest/distributing/?highlight=sdist#initial-files) are now or will later be required?

asmeurer:
setup.py is used both by you and the people who install the package. It includes itself in the source tarball. What `pip install z3` is effectively downloads z3 from PyPI, unpacks it, and runs `setup.py install`. Or users can do that manually. 

I'm no expert on writing setup.py files, and this looks like a tricky scenerio. I think as long as you can figure out how to make the sdist contain the files in the normal way (setup.py at the root, python package files at the root), then you can put them in the git repo wherever you want. I have no idea how to make it do that, though. Unless someone else on this thread knows how to do this, you may want to solicit help from the PyPA (Python Packaging Authority; they maintain pip and a bunch of other Python packaging tools) on the "best" way to do this. Or if you know of some other library with Python bindings that works similar to z3 you can see what they do.

wintersteiger:
What about the solution proposed at [Pip install fails to install dependency when setup.py not in root directory](http://stackoverflow.com/questions/26542288/pip-install-fails-to-install-dependency-when-setup-py-not-in-root-directory). Would the same trick work for us as well, so that we can have

     pip install git+https://github.com/Z3Prover/z3@egg=subdir&subdirectory=scripts/packages/pypy/

wintersteiger:
Apparently a pull request can't exist on github if the base branch doesn't exist anymore. I have removed the unstable branch because everything will go directly into master from now on. Could you perhaps submit a new request based on master? Thanks!

PR #76, opened 2015-05-05
Fix ./configure under pypy

zardus:
`mk_util.py` has a heavy implicit reliance on cpython's reference counting garbage collector in the way that files are used. They're very rarely explicitly closed and are expected to automatically close when they go out of scope.

Unfortunately, pypy breaks this assumption, as it only closes files when its (non-reference-counting) GC runs. Because of that, when using pypy, `./configure` excepts out with "too many open files".

This PR adds explicit `close()` calls for all of the file operations (that I could find) in `mk_util.py`, making `./configure` work in pypy virtualenvs!

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -180,6 +180,8 @@ def exec_cmd(cmd):
     except:
         # Failed to create process
         return 1
+    finally:
+        null.close()
 
 # rm -f fname
 def rmf(fname):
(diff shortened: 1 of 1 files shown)


wintersteiger:
Thanks for submitting a patch for this, this will be merged soon.

wintersteiger:
If it's not too much trouble, could I get you to sign a [CLA](https://github.com/Z3Prover/z3/wiki/Contribution-Guidelines) before we merge this?

zardus:
I signed the CLA for the other PR (the setup.py). I think that should cover this one as well?

wintersteiger:
Great, thanks! Yes, we need only one CLA in place for all contributions (and even across projects I think).

wintersteiger:
Merged into master as of 6f3b957300dce1c29c6589711b0df15f0fca1ead

PR #107, opened 2015-05-25
Fix the Python FPRef.__lt__ implementation

jmgrosen:
This looks like it was a typo. I imagine this could have some rather bad consequences, though!


wintersteiger:
Thanks for the patch, this look good to me, will be merged soon.

wintersteiger:
Merged as of 9d505ec7ffc54fdf890414945bb7778f10ce8e65

PR #108, opened 2015-05-26
Fix ClassCastException from straight array cast

MarcusVoelker:
Switched cast from Expr[] to BoolExpr[] to loop with casts for each member, because the current alternative gives a ClassCastException (in InterpolationContext.java)


wintersteiger:
That's exactly what I wanted to avoid, but I guess it just won't work. I have now added all conversion functions for all the other *Expr types.

PR #126, opened 2015-06-10
Basic infrastructure for minimum aggregation function

ahorn:
This patch contains the basic infrastructure for supporting minimum aggregation functions in datalog. This includes changes to the front-end and back-end, and a simple unit test.

Note: the patch aims at features that hopefully allow us to better understand our benchmarks. In particular, we would like to find out whether recursive rules mix negation and min aggregation functions. This initial question explains why I have not included any changes to the compiler yet; I only instructed the compiler to ignore min aggregation functions.

### DIFF
--- a/src/ast/dl_decl_plugin.cpp
+++ b/src/ast/dl_decl_plugin.cpp
@@ -44,7 +44,8 @@ namespace datalog {
         m_num_sym("N"),
         m_lt_sym("<"),
         m_le_sym("<="),
-        m_rule_sym("R")
+        m_rule_sym("R"),
+        m_min_sym("min")
     {
     }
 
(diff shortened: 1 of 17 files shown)


NikolajBjorner:
In the reference implementation in src/muz/rel/dl_base.cpp, is the memory management for the functors checked?

ahorn:
A bug indeed: [diff_fn is a leak](https://github.com/ahorn/z3/commit/140fb7942dd051f4e8db608ef2d9b260353e7416#diff-6f0555c2d476fc0366442c8af51a0408R557) Thank you. I have a look tomorrow morning (just got home).

ahorn:
I've fixed the memory leaks in the (unused) reference implementation and updated some of the documentation for relation_table concerning pointer ownership transfer.

PR #134, opened 2015-06-18
Compile min aggregation function

ahorn:
For the datalog plugin to handle min aggregation functions, `-D_MIN_WORK_IN_PROGRESS_` must be explicitly set at compile-time. This disables all transformation rules. Otherwise, if `_MIN_WORK_IN_PROGRESS_` is not set (which is the default), all transformers remain enabled.

On encountering min aggregation functions, the compiler generates code which behaves similarly to `all_or_nothing_deltas` (see fixed-point iteration in commit b67bdc24bb2373fd5a1ca8e5e7606ca7e719f12f). This is illustrated by the `min_entire_table.smt2` integration test. There are around a dozen other integration tests, which pinpoint various corner cases and therefore hopefully help us avoid regressions and track changes to the semantics as we dive deeper into the problem.

We should check with Nuno before merging (he is currently traveling) but I wanted to share the patch (and the examples) because I suspect it will come in handy for discussions etc.


NikolajBjorner:
Instead of using compile time flags, I suggest using a configuration parameter that you later remove.
It is much easier to check for regressions by using the same binary and two different settings of the flag.
The default value of the configuration parameter (and the default value of any flag) should be to disable the new functionality. It should not be to enable the new functionality. In this way it is much less intrusive to merge changes early into unstable as they would be orthogonal to other functionality.

ahorn:
I like that and I look into it.

ahorn:
I've incorporated your feedback and improved a few other things:

1. To avoid the compile-time flag, I've added `fixedpoint.datalog.transform_rules` as a parameter, which should be also useful to @mschlaipfer who needs to disable transformation rules.
2. I've added an example that further illustrates the semantics of the min implementation (see commit 94f34a293daddb3e45352733f02be0bc6710eb32). This goes back to early discussions with @nunoplopes.
3. I've fixed a problem with the handling of ASTs due to a bodged loop condition.

ahorn:
In case it helps reviewing the patch, I wanted to mention that this [example](https://github.com/ahorn/z3/commit/5eca490665aa6b53ede7149ebac04dfe72249c1b#diff-1bedd02bf71f42e630f95ce9e05c0739) illustrates min over joins.

ahorn:
The following two dependency graphs illustrate the last commit (bold arrows denote negation):

g6
g5

wintersteiger:
This looks like a substantial code change/addition. @NikolajBjorner could you take a look at this and see whether this is still OK or whether it is obsolete by now?

wintersteiger:
@ahorn if it's not too much trouble, could I get you to sign a [CLA](https://github.com/Z3Prover/z3/wiki/Contribution-Guidelines) before we merge this?

ahorn:
Hi there, I am happy to sign the CLA but are you sure it's needed? All the work was done while I was an interning with you guys. Either way, let me know and we can take it from there. If you have any questions about the code, let me know.

wintersteiger:
Right, in that case we won't need a CLA. I thought this might have been produced afterwards.

wintersteiger:
@NikolajBjorner could you check whether this wasn't superseded by other changes before I merge this? Thanks!

NikolajBjorner:
or Nuno Lopes?

wintersteiger:
Apparently a pull request can't exist on github if the base branch doesn't exist anymore. I have removed the unstable branch because everything will go directly into master from now on. Could you perhaps submit a new request based on master? Thanks!

nunoplopes:
This feature still needs considerable work before being production ready, so does this patch.  I don't have time at the moment to work on this, but once I do I can recover this patch.

PR #147, opened 2015-06-30
Remove undefined behaviour from bit_vector

hguenther:
When creating a bitvector of zero capacity, the result of the allocation is a null-pointer, which is then passed to memset which makes the compiler assume that it is valid. It can then optimize away the check at the destructor, whether the pointer is null, triggering a segfault.

Signed-off-by: Henning Guenther <t-hennig@microsoft.com>


nunoplopes:
In theory, commit https://github.com/Z3Prover/z3/commit/eeef4d29d68285a505468715bccc892db76f1267 should fix things, but let's way for test results to come back before closing this report.

nunoplopes:
Build bots were fine; closing this.

PR #152, opened 2015-07-08
Avoid unnecessary calls to check_rules()

hguenther:
flush_add_rules was calling check_rules even when no new rules were added.
This could make repeated calls to get_rules() expensive.

Signed-off-by: Henning Guenther <t-hennig@microsoft.com>

### DIFF
--- a/src/muz/base/dl_context.cpp
+++ b/src/muz/base/dl_context.cpp
@@ -457,15 +457,17 @@ namespace datalog {
     }
 
     void context::flush_add_rules() {
-        datalog::rule_manager& rm = get_rule_manager();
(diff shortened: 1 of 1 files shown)


nunoplopes:
I don't think this is correct, since there are other ways of adding rules to m_rule_set, like add_rule(), and therefore it's not correct to dismiss the call to check_rules() even if flush_add_rule() didn't do anything by itself.
I commited a patch that reduces the number of calls to get_rules() in a few places; should be enough for now: https://github.com/Z3Prover/z3/commit/8edd551f20b251e56fa84c124cd2493daf124af7

Issue #168, opened 2015-07-24
Python Interface Does not Support Array Default operator


npricci:
There doesn't seem to be a way to use the "default" operator on arrays using the python interface. I.E., it would be nice to do something like this:

(declare-const FPolicy (Array Int Int Int Bool))

(assert (select FPolicy 0 1 30))
(assert (select FPolicy 0 2 20))
(assert (not (default FPolicy)))

Work in the python interface, but there is no way to use the "default" operator.

NikolajBjorner:
added to unstable

PR #169, opened 2015-07-24
Added default operator to array interface.

npricci:
This patch adds the "default" array operator to the python interface, resolving this issue:  https://github.com/Z3Prover/z3/issues/168


wintersteiger:
This looks good to me, thanks for the patch, will be merged soon.

PR #170, opened 2015-07-25
Remove extra semicolons in C++ headers. 

Dmitriy403:
We don't need semicolons after `extern "C"{...}` or `namespace {...}`. So when i use z3 in project builded with `-Wpedantic` I got this warnings from z3 headers (in my case this warnings treated as errors because of `-Werror`). That would be nice if z3 headers doesn't force to use nasty pragmas for disabling warnings for z3 headers in projects with pretty strict codestyle.


wintersteiger:
This looks good to me, thanks for the patch, will be merged soon!

wintersteiger:
Merged as of 0c774e59c33d04bf17cd602932cea2ff400d480d.

CLA not required, de minimis.

Issue #34, opened 2015-04-10
OCaml installation is broken


Drup:
(follow up of https://z3.codeplex.com/discussions/473552 since codeplex doesn't want to let me add messages anymore)

So, the issue is that it was trying to use an outdated system z3 instead of the libz3.so installed through `make ocamlfind_install`. If I remove the system z3, it doesn't work anymore at all and can't find the libz3.so installed by `make ocamlfind_install`.

I wonder if you could produce only one .so containing everything, That would avoid the issue entirely.

wintersteiger:
That could work, but only in theory. It would require a major rewrite of the ML API and some other parts of Z3 as well. Also, there would be no consistency between multiple different APIs installed on the same system, because they would use different copies of Z3 (compiled with different flags, static vs dynamic, etc). If it's really necessary, we could try to think about linking libz3 statically into libz3ml.so, only when the --staticlib option is on to keep it local for now (the build is probably broken in that case anyways). 

If it can't find libz3.so which is properly installed into your ocamlfind directory, it probably means that your (DY)LD_LIBRARY_PATH isn't set up properly or ldconfig needs to be run? The directory reported by echo `ocamlfind printconf destdir`/Z3 should contain libz3.so/.dylib.

Drup:
> The directory reported by echo ocamlfind printconf destdir/Z3 should contain libz3.so/.dylib. 

That's not true for me, they are installed in `` ` ocamlfind printconf destdir`/stublibs``.

wintersteiger:
Interesting... did they change the default paths yet again? What version of ocamlfind are you using?

Drup:
I don't think they changed it. I'm using version 1.5.5. (and opam, but not sure if it should change anything, @Altgr ?)

Drup:
Note that changing the directory doesn't really help. OCamlers don't usually have their LD_LIBRARY_PATH setup that way, so it would need additional setup from the user, which is sub-optimal.

Ok so, the situation from the ocaml perspective (and in particular, the desire to package z3 and its API in opam, the usual package manager for ocaml which is a *source* package manager) is, as far as I can see the following:
- if distribs don't package z3 (and they don't for now, but I expect it to change) a z3 opam package have to recompile z3 in a whole and so the package will contains its own version of libz3.so, hence linking statically in libz3ml seems acceptable to me. This means long-ish compilation time, but is very flexible.
[...]

PR #172, opened 2015-07-27
If ocamlfind is installed, add destdir/stublibs to rpath

zkincaid:
Fixes the LD_LIBRARY_PATH issue mentioned [here](https://github.com/Z3Prover/z3/issues/34#issuecomment-91525424).

This is the same patch proposed [here](https://github.com/Z3Prover/z3/pull/128).

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -1478,6 +1478,28 @@ class MLComponent(Component):
                     get_component(Z3_DLL_COMPONENT).dll_name))
             out.write(' %s\n' % (os.path.join(sub_dir, 'z3native_stubs$(OBJ_EXT)')))
             out.write('\tocamlmklib -o %s -I %s -ldopt \"-L. -lz3\" ' % (os.path.join(sub_dir, 'z3ml'), sub_dir))
+
+            # Add ocamlfind destdir to rpath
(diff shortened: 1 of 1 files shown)


Drup:
Can we get this merged ?

wintersteiger:
Yes we are very close to getting this merged. Could you review the discussion for #34 again? Perhaps we should always rely on a system installed shared library for Z3?

The proposed patch does not work for me. I'm using Windows 10, OCaml 4.02.1 (win32 native, cl, no cygwin involved). `ocamlfind printconf destdir` reports `C:\ocamlw32\lib\site-lib` and the latter does not contain a directory called `stublibs`.

zkincaid:
Apparently, ocamlfind installs DLLs into `stublibs` if it exists in `ocamlfind printconf destdir`, and otherwise installs them in the package directory.  I believe that OPAM creates the stublibs directory by default.  This should be fixed now, but I don't have a windows machine to test it on.

wintersteiger:
Tested, works on Windows too, so I'm merging this into master now.

wintersteiger:
Merged into master as of 3aedcf0a7e38d842dc0d3e43ec7343a67206a9b0

PR #173, opened 2015-07-27
Z3_lbool should be signed in API

zkincaid:
Fixes a bug in the OCaml bindings.
bug.ml:

    open Z3
    let _ =
      let ctx = mk_context [] in
      ignore (Boolean.get_bool_value (Boolean.mk_false ctx))

Compile with

    ocamlfind ocamlopt -linkpkg -package Z3 bug.ml -o bug

Running ./bug gives

    Fatal error: exception Failure("undefined enum value")


wintersteiger:
Thanks for the patch, this looks good to me, will be merged soon.

wintersteiger:
Merged as of 954a629296765c85cd4752afe93c07da4e23a1f9

PR #188, opened 2015-08-09
Optimize module for ML bindings, second edition.

Drup:
As noted last time, there are no functions in the C API to
- Get the reason for UNKNOWN
- Get the unsat core

### DIFF
--- a/src/api/ml/z3.ml
+++ b/src/api/ml/z3.ml
@@ -14,19 +14,19 @@ let null = Z3native.mk_null()
 let is_null o = (Z3native.is_null o)
 
 (* Internal types *)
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
we have to clear the process for merging pull requests before we can do it, sorry.
I added ML bindings in the opt branch two weeks ago and took them now to unstable.
Added get_reason_unknown. Not adding core support yet. It can so far be emulated by using a solver when retrieving unsat, so it depends if it is an important user scenario to add this functionality.

Drup:
Ok, thanks. At least it's in now.

wintersteiger:
Am I correct in assuming that this PR is obsolete after @NikolajBjorner added an Optimize module to the ML API?

NikolajBjorner:
Obsolete

PR #218, opened 2015-09-13
fix implies(expr const &, expr const &) in z3++.h

cgcgbcbc:
currently it seems that friend expr implies(expr const & a, expr const & b) is not accessible .

### DIFF
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -642,13 +642,7 @@ namespace z3 {
         */
         friend expr operator||(bool a, expr const & b) { return b.ctx().bool_val(a) || b; }
         
-        friend expr implies(expr const & a, expr const & b) {
-            check_context(a, b);
-            assert(a.is_bool() && b.is_bool());
-            Z3_ast r = Z3_mk_implies(a.ctx(), a, b);
-            a.check_error();
-            return expr(a.ctx(), r);
-        }
+        friend expr implies(expr const & a, expr const & b);
         friend expr implies(expr const & a, bool b);
         friend expr implies(bool a, expr const & b);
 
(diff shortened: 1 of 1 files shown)


wintersteiger:
Thanks for the patch, this looks good to me, will be merged soon.

NikolajBjorner:
When i reviewed this I  didn't see how suggested patch changed accessibility.

wintersteiger:
That's a good point. This change moves the definition of implies outside of class expr, but in fact everything in class expr is clearly labelled as public. I can imagine though that some compilers might not like the "friend" declaration on an inline function definition that's not an operator. The definition of pw is also outside of class expr, but all the other operators are inside. 

@cgcgbcbc could you tell us which system/compiler this was a problem with?

@nunoplopes can you answer this mystery quicker than I can look it up?

wintersteiger:
@NikolajBjorner Actually, it was you who previously changed this for all non-operator friends in commit 7ef1e8a3de0ee0ed6ff146710e75041d7542d463, but you missed implies.

NikolajBjorner:
OK, so something with namespaces, thanks.

cgcgbcbc:
@wintersteiger
the problem occurs when I was using osx 10.10，Xcode 6.4（not sure） . I cannot reproduce the problem in OS X 10.11 with Xcode 7，but I'm sure that when I opened this pr，my complier did fail with this. Sorry that I can't downgrade my os，but perhaps I can check the problem on travisci.

wintersteiger:
@cgcgbcbc no worries, it's not super critical. We should move that definition anyway because of those namespace issues and to keep it all in line with the other friend functions defined there.

wintersteiger:
CLA not required, de minimis.

dkim:
This was due to the visibility of `friend` declarations within a class definition:

> http://en.cppreference.com/w/cpp/language/namespace
>
> Names introduced by friend declarations within a non-local class X become members of the innermost enclosing namespace of X, but they do not become visible to lookup (neither unqualified nor qualified) unless a matching declaration is provided at namespace scope, either before or after the class definition. Such name may be found through ADL which considers both namespaces and classes.

Developers can work around it by writing the following declaration in their own code, until the fixed version is released:
```cpp
namespace z3 {

expr implies(expr const & a, expr const & b);

}
```

PR #234, opened 2015-10-07
Fixed typo that accidentally prints warning message if a Z3 context i…

martin-neuhaeusser:
…s created with the 'timeout' parameter


wintersteiger:
Thanks for this fix! CLA not required, de minimis.

PR #242, opened 2015-10-13
scripts: mk_util.py, support DESTDIR

jirislaby:
To be able to package z3 on distros, we need to install to a build
root. For that purpose, DESTDIR usually exists. So add this
functioanlity to the install script.


wintersteiger:
Can you elaborate on why this is needed and why $(PREFIX) isn't enough for this purpose?

jirislaby:
Sure:
https://www.freebsd.org/doc/en/books/porters-handbook/porting-prefix.html

In short:
PREFIX is sometimes used in binaries/scripts to reference other parts of the package. E.g. 'exec("$PREFIX/bin/z3")', while DESTDIR shall not appear in the binaries at all. It is used just to temporarily install (DESTDIR=/home/XXX/ and cp -r /home/XXX/* $PREFIX).

wintersteiger:
Still seems a bit unnecessary to me. There will have to be some script to generate the package anyways and it would be enough if that script supports DESTDIR instead of threading it all the way through the build infrastructure (e.g., wherever python scripts/mk_make.py is called, it can get a custom PREFIX). 

Can you check whether other systems use the same variable name for a different purpose? E.g., OSX is a likely candidate but also other Linux's that are perhaps not a BSD.

jirislaby:
Of course, DESTDIR is used by autotools, cmake, linux kernel, and any random (sane) project.

https://www.gnu.org/software/automake/manual/html_node/DESTDIR.html
https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html

Really, if you are not familiar with this kind of details, I suggest you to use some standardized approach like cmake. It will make sure to do it right and will it do it for free (and cross-platform). As inspiration, what is wrong with the build system, you can take a look what magic they have to do in fedora in order to make a z3 package:
http://pkgs.fedoraproject.org/cgit/z3.git/tree/z3.spec#n99

And yet it is not perfect, as the libs are installed to a bad dir (/usr/lib instead of /usr/lib64 -- mv on line 168), the binary is linked in a completely bad manner (line 149), the lib is linked bad (lines with -Wl), non-safe -O3 cannot be turned off (line 102) etc.

Now I am packaging it for openSUSE and I can confirm the current build system is pure evil :(.

wintersteiger:
Yes, most package managers and most systems are pure evil when you're not used to them. We can't make it easy for everybody and I (not we) fully expect package maintainers to jump through some hoops. This is exactly the reason why the core team of Z3 decided not to provide any packages for any systems; that will have to be done by the system experts with most minimal changes to Z3. The latter simply because any change of that flavor is likely to break an assumption that was made by someone else for a different system...

Z3 does not use/require/support automake or cmake, it is not trying to emulate the Linux kernel, and many might call it insane. Your proposed changes are therefore rather system specific from my point of view. If we are going to modify the build scripts to that extent, I'd rather add a general-purpose feature that works for all systems, so that we have to test it only once. For instance, we could think about custom commands after every cp/mv/install. 

However, for the purpose of publishing ports packages, I think all of this is really not necessary; we have scripts/mk_unix_dist.py to generate a unix distribution file which should be more than sufficient.

jirislaby:
> Z3 does not use/require/support automake or cmake

Ok, you wrote a custom build system which works on Windows instead of using some cross-platform and easy-to-use framework like cmake. I would be perfectly fine with that and won't actually object at all. If it actually worked (on Linux and BSD). But it is not. So I fixed at least 'make install' to work properly here. The rest can be worked around, mostly.

Well, now I persuaded myself to write the cmake rules and push them here. Maybe you will fetch them and try. Or maybe not and I will hold them in the fork 4ever :).

nunoplopes:
BTW, any reason why you remove -fvisibility=hidden from CFLAGS?  Just curious, since that reduces the .so size considerably.

jirislaby:
-fvisibility=hidden from CFLAGS -- that's not me, that's feodra pkg maintainer -- ask Jerry James <loganjerry@gmail.com>

wintersteiger:
> If it actually worked (on Linux and BSD).  

Can you clarify again what exactly was wrong with `make install`? So far it has always done the job of installing our binaries to where they need to go, i.e. into $(PREFIX), on all our Linux and BSDs.

Also, why is `python scripts/mk_unix_dist.py ; cp dist/* $(DESTDIR)` not a satisfactory solution for your packages?

arrowdodger:
@wintersteiger I'm a FreeBSD mainainter for z3 port and i want to +1 what @jirislaby says. This is what we have to do on our side: https://svnweb.freebsd.org/ports/head/math/z3/files/patch-scripts_mk__util.py?view=markup

Having this patch is not a big deal for us, but since @jirislaby done all the hard work, from which we all can benefit, i don't see why not to accept it.

wintersteiger:
Can you refactor this in such a way that mk_util contains only 1 line of code in mk_util.py that refers to $(DESTDIR)? E.g., introduce a function for copying files to $(PREFIX) and only that function adds $(DESTDIR) to the copy command if necessary. This way there is no secret invariant that future implementors of mk_install have to comply with.

OSX is also a flavour of BSD. Did you test your changes there to see whether they use $(DESTDIR) as well, or whether they renamed it to something else, like they did for LD_LIBRARY_PATH?

NikolajBjorner:
Do #338, #340 address the DESTDIR request?

delcypher:
@NikolajBjorner #338 does. This PR can be closed now that has been merged into master.

NikolajBjorner:
Thanks!

PR #275, opened 2015-10-29
Further OpenBSD build support

hogand:
This is an enhancement of the previous commit to support building on OpenBSD.

There are a few things to note:

* OpenBSD replaced base sudo with doas.  I don't think this is necessary to mention for users though.  They should be used to it by now.
* This compiles with BSD make and base gcc 4.2.1 on amd64.


wintersteiger:
Thanks for the contribution! Could you rebase your changes onto the current master so it's easier to see what exactly changed? 

Also, to properly support OpenBSD, we need one more thing implemented: [scoped_timer.cpp](https://github.com/Z3Prover/z3/blob/6e852762baf568af2aad1e35019fdf41189e4e12/src/util/scoped_timer.cpp) currently doesn't select any implementation for timers, i.e., all timeout commands and timeouts used by the tactics (lots of them), will not run correctly. Could you take a look at that and see whether the FreeBSD implementation also works on OpenBSD?

wintersteiger:
Bumping this. Is there a problem with the CLA?

NikolajBjorner:
@hogand I would be happy to merge, but our system says a CLA is required even though these are quite minor updates to enable OpenBSD. If this is a problem, I am inclined to manually add the openBSD case and then take bug reports. Let us know.

NikolajBjorner:
Since I can't merge without the CLA I have added OpenBSD support in separate changes. File bugs against these changes if they don't work.

wintersteiger:
@NikolajBjorner You didn't test whether timeouts work on OpenBSD, did you?

PR #284, opened 2015-10-30
Provide `unsupported` location when verbose is set

zv:
While writing `z3-mode` (an interactive development environment for z3), I realized that `unsupported symbol` did not provide location information. This is a simple patch that throws a location (column & line) information when z3 has verbose set.

If you would like this behavior to be conditional on some other `param`, I would be happy to make the change.

### DIFF
--- a/src/parsers/smt2/smt2parser.cpp
+++ b/src/parsers/smt2/smt2parser.cpp
@@ -2265,6 +2265,7 @@ namespace smt2 {
                 consume_sexpr();
             }
             m_ctx.print_unsupported(s);
+            IF_VERBOSE(1, error("unsupported symbol"););
             next();
             return;
         }


NikolajBjorner:
I am going to add location information to the call above if-verbose, but doesn't seem right to use verbose stream as non console based parser uses would not get the same result.

zv:
I agree with that sentiment but I'm not aware of any existing conventions surrounding the 'appropriateness' of introducing a new top level flag.

NikolajBjorner:
I don't see why a flag is in place here. The default behavior should be more informative. It is now pushed to master branch and we will see how regression tests check out (some will fail because they do string matching)

zv:
Thank you Nikolaj, I will start using this in `z3-mode`!

NikolajBjorner:
Pull request super-seeded by update

zv:
you make some of the funniest spelling mistakes @NikolajBjorner :)

it is superseded.

PR #285, opened 2015-10-31
add line/position information to unsupported command reports per zeph…

NikolajBjorner:
… pull request

Signed-off-by: Nikolaj Bjorner <nbjorner@microsoft.com>


wintersteiger:
This is in response to PR #284.

Just adding that so we have a link between the two.

PR #306, opened 2015-11-12
deploy script for z3 and z3-native

cristipp:


cristipp:
Oops, intended to create this just for our local setup. Sorry for the noise!

PR #338, opened 2015-11-24
Better install under non Windows systems

delcypher:
This series of patches fixes a few issues with the current build system when running ``make install``.

* Adds support for ``$(DESTDIR)`` (see https://www.gnu.org/prep/standards/html_node/DESTDIR.html )
* Makes ``libz3.so`` in the Python package directory a symlink (rather than a copy) if possible, to save space when packaging.

There will be other packaging issues that I'd like to fix in the future but I thought I'd drop some of my initial work so it can be reviewed.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -2087,7 +2087,9 @@ def mk_makefile():
     out.write("\t@echo \"Z3Py scripts stored in arbitrary directories can be also executed if \'%s\' directory is added to the PYTHONPATH environment variable.\"\n" % BUILD_DIR)
     if not IS_WINDOWS:
         out.write("\t@echo Use the following command to install Z3 at prefix $(PREFIX).\n")
-        out.write('\t@echo "    sudo make install"\n')
+        out.write('\t@echo "    sudo make install"\n\n')
+        out.write("\t@echo If you are doing a staged install you can use DESTDIR.\n")
+        out.write('\t@echo "    make DESTDIR=/some/temp/directory install"\n')
     # Generate :examples rule
     out.write('examples:')
     for c in get_components():
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -1017,11 +1017,15 @@ class LibComponent(Component):
 
     def mk_install(self, out):
         for include in self.includes2install:
-            out.write('\t@cp %s %s\n' % (os.path.join(self.to_src_dir, include), os.path.join('$(DESTDIR)$(PREFIX)', 'include', include)))
+            MakeRuleCmd.install_files(
+                out,
+                os.path.join(self.to_src_dir, include),
+                os.path.join('include', include)
+            )
 
     def mk_uninstall(self, out):
         for include in self.includes2install:
-            out.write('\t@rm -f %s\n' % os.path.join('$(DESTDIR)$(PREFIX)', 'include', include))
+            MakeRuleCmd.remove_installed_files(out, os.path.join('include', include))
 
     def mk_win_dist(self, build_path, dist_path):
         mk_dir(os.path.join(dist_path, 'include'))
@@ -1103,11 +1107,11 @@ class ExeComponent(Component):
     def mk_install(self, out):
         if self.install:
             exefile = '%s$(EXE_EXT)' % self.exe_name
-            out.write('\t@cp %s %s\n' % (exefile, os.path.join('$(DESTDIR)$(PREFIX)', 'bin', exefile)))
+            MakeRuleCmd.install_files(out, exefile, os.path.join('bin', exefile))
 
     def mk_uninstall(self, out):
         exefile = '%s$(EXE_EXT)' % self.exe_name
-        out.write('\t@rm -f %s\n' % os.path.join('$(DESTDIR)$(PREFIX)', 'bin', exefile))
+        MakeRuleCmd.remove_installed_files(out, os.path.join('bin', exefile))
 
     def mk_win_dist(self, build_path, dist_path):
         if self.install:
@@ -1283,9 +1287,11 @@ class DLLComponent(Component):
     def mk_install(self, out):
         if self.install:
             dllfile = '%s$(SO_EXT)' % self.dll_name
+            MakeRuleCmd.install_files(out, dllfile, os.path.join('lib', dllfile))
+            # FIXME: All this stuff to do with the Python bindings needs to
+            # be refactored to use MakeRuleCmd.
             dllInstallPath = os.path.join('$(PREFIX)', 'lib', dllfile)
             pythonPkgDirSubst = PYTHON_PACKAGE_DIR.replace('$(PREFIX)', PREFIX, 1)
-            out.write('\t@cp %s $(DESTDIR)%s\n' % (dllfile, dllInstallPath))
             if IS_WINDOWS or not pythonPkgDirSubst.startswith(PREFIX):
                 out.write('\t@cp %s $(DESTDIR)%s\n' % (dllfile, os.path.join(PYTHON_PACKAGE_DIR, dllfile)))
             else:
@@ -1299,15 +1305,15 @@ class DLLComponent(Component):
                 out.write('\t@ln -s %s $(DESTDIR)%s\n' % (relativePath, os.path.join(PYTHON_PACKAGE_DIR, dllfile)))
             if self.static:
                 libfile = '%s$(LIB_EXT)' % self.dll_name
-                out.write('\t@cp %s %s\n' % (libfile, os.path.join('$(DESTDIR)$(PREFIX)', 'lib', libfile)))
-
+                MakeRuleCmd.install_files(out, libfile, os.path.join('lib', libfile))
 
     def mk_uninstall(self, out):
         dllfile = '%s$(SO_EXT)' % self.dll_name
-        out.write('\t@rm -f %s\n' % os.path.join('$(DESTDIR)$(PREFIX)', 'lib', dllfile))
+        MakeRuleCmd.remove_installed_files(out, os.path.join('lib', dllfile))
+        # FIXME: Use MakeRuleCmd
         out.write('\t@rm -f $(DESTDIR)%s\n' % os.path.join(PYTHON_PACKAGE_DIR, dllfile))
         libfile = '%s$(LIB_EXT)' % self.dll_name
-        out.write('\t@rm -f %s\n' % os.path.join('$(DESTDIR)$(PREFIX)', 'lib', libfile))
+        MakeRuleCmd.remove_installed_files(out, os.path.join('lib', libfile))
 
     def mk_win_dist(self, build_path, dist_path):
         if self.install:
@@ -1474,14 +1480,16 @@ class JavaDLLComponent(Component):
     def mk_install(self, out):
         if is_java_enabled() and self.install:
             dllfile = '%s$(SO_EXT)' % self.dll_name
-            out.write('\t@cp %s %s\n' % (dllfile, os.path.join('$(PREFIX)', 'lib', dllfile)))
-            out.write('\t@cp %s.jar %s.jar\n' % (self.package_name, os.path.join('$(PREFIX)', 'lib', self.package_name)))
+            MakeRuleCmd.install_files(out, dllfile, os.path.join('lib', dllfile))
+            jarfile = '{}.jar'.format(self.package_name)
+            MakeRuleCmd.install_files(out, jarfile, os.path.join('lib', jarfile))
 
     def mk_uninstall(self, out):
         if is_java_enabled() and self.install:
             dllfile = '%s$(SO_EXT)' % self.dll_name
-            out.write('\t@rm %s\n' % (os.path.join('$(PREFIX)', 'lib', dllfile)))
-            out.write('\t@rm %s.jar\n' % (os.path.join('$(PREFIX)', 'lib', self.package_name)))
+            MakeRuleCmd.remove_installed_files(out, os.path.join('lib', dllfile))
+            jarfile = '{}.jar'.format(self.package_name)
+            MakeRuleCmd.remove_installed_files(out, os.path.join('lib', jarfile))
 
 class MLComponent(Component):
     def __init__(self, name, lib_name, path, deps):
@@ -2080,9 +2088,10 @@ def mk_install(out):
     if is_ml_enabled() and OCAMLFIND != '':
         out.write('ocamlfind_install')
     out.write('\n')
-    out.write('\t@mkdir -p %s\n' % os.path.join('$(DESTDIR)$(PREFIX)', 'bin'))
-    out.write('\t@mkdir -p %s\n' % os.path.join('$(DESTDIR)$(PREFIX)', 'include'))
-    out.write('\t@mkdir -p %s\n' % os.path.join('$(DESTDIR)$(PREFIX)', 'lib'))
+    MakeRuleCmd.make_install_directory(out, 'bin')
+    MakeRuleCmd.make_install_directory(out, 'include')
+    MakeRuleCmd.make_install_directory(out, 'lib')
+    # FIXME: use MakeRuleCmd for the Python stuff
     out.write('\t@mkdir -p $(DESTDIR)%s\n' % PYTHON_PACKAGE_DIR)
     for c in get_components():
         c.mk_install(out)
@@ -2108,6 +2117,7 @@ def mk_uninstall(out):
     out.write('uninstall:\n')
     for c in get_components():
         c.mk_uninstall(out)
+    # FIXME: use MakeRuleCmd for the Python stuff
     out.write('\t@rm -f $(DESTDIR)%s*.py\n' % os.path.join(PYTHON_PACKAGE_DIR, 'z3'))
     out.write('\t@rm -f $(DESTDIR)%s*.pyc\n' % os.path.join(PYTHON_PACKAGE_DIR, 'z3'))
     out.write('\t@rm -f $(DESTDIR)%s*.pyc\n' % os.path.join(PYTHON_PACKAGE_DIR, '__pycache__', 'z3'))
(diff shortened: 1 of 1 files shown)


wintersteiger:
As suggested in #242, can you refactor this so that there is only 1 function that refers to the string "$(DESTDIR)"? Otherwise this will break frequently because others will not be aware of the fact that some systems want/require this functionality.

delcypher:
Note you can now do this...

```bash
$ python scripts/mk_make.py --prefix /usr
$ cd build
$ make -j8
$ mkdir staged
$ make DESTDIR=/home/dsl11/dev/z3-upstream/build/staged install
tree staged
staged/
└── usr
    ├── bin
    │   └── z3
    ├── include
    │   ├── z3_algebraic.h
    │   ├── z3_api.h
    │   ├── z3_fpa.h
    │   ├── z3.h
    │   ├── z3++.h
    │   ├── z3_interp.h
    │   ├── z3_macros.h
    │   ├── z3_polynomial.h
    │   ├── z3_rcf.h
    │   └── z3_v1.h
    └── lib
        ├── libz3.so
        └── python3.5
            └── site-packages
                ├── libz3.so -> ../../libz3.so
                ├── __pycache__
                │   ├── z3consts.cpython-35.pyc
                │   ├── z3core.cpython-35.pyc
                │   ├── z3.cpython-35.pyc
                │   ├── z3num.cpython-35.pyc
                │   ├── z3poly.cpython-35.pyc
                │   ├── z3printer.cpython-35.pyc
                │   ├── z3rcf.cpython-35.pyc
                │   ├── z3test.cpython-35.pyc
                │   ├── z3types.cpython-35.pyc
                │   └── z3util.cpython-35.pyc
                ├── z3consts.py
                ├── z3core.py
                ├── z3num.py
                ├── z3poly.py
                ├── z3printer.py
                ├── z3.py
                ├── z3rcf.py
                ├── z3test.py
                ├── z3types.py
                └── z3util.py

7 directories, 33 files

```

The contents of the staged directory are now suitable for packaging on a Linux distro (and probably other UNIX like operating systems).

delcypher:
@wintersteiger 
> As suggested in #242, can you refactor this so that there is only 1 function that refers to the string "$(DESTDIR)"? Otherwise this will break frequently because others will not be aware of the fact that some systems want/require this functionality.

There isn't really presidence for this as ``$(PREFIX)`` (which is written everywhere ``$(DESTDIR)`` needs to be used) is written all over the place. I could obviously write

```python
def DestDir():
    return "$(DESTDIR)"
```

and call it but I don't think that would achieve anything other than making the code hard to read.

If you really want this refactored then arguable use of ``$(PREFIX)`` must be refactored as well.

delcypher:
> If you really want this refactored then arguably the use of $(PREFIX) must be refactored as well.

Would some sort of function that is used to automatically generate the ``cp`` lines for make files be acceptable (I would also need to write a similar one to setup sym links)? That's the only way I can see to refactor this.

wintersteiger:
Yes, that's exactly what I meant, introduce a function for "copy to install directory", then its clear what needs to be done for new systems/components and we can't forget one or the other part when adding new files to copy.

C

Sent from my Windows Phone

delcypher:
@wintersteiger 

I've addressed your comments. That was a lot more work than I was intending to do (refactoring code that clearly needed refactoring that I didn't write) but I guess it will help in the long run.

wintersteiger:
Fantastic, thanks for all the work! This looks a lot cleaner now and it's easy to read too. 

We will have to do some testing on various platforms and Python's to see whether this breaks any existing things, which will need some time. Could you tell us which systems you've already tested it on?

delcypher:
> We will have to do some testing on various platforms and Python's to see whether this breaks any existing things, which will need some time. Could you tell us which systems you've already tested it on?

I've tested that Z3 builds and installs on Arch Linux (Python ``3.5.0`` and Python ``2.7.10``). I assumed that Python ``2.7`` was the minimum when writing the code, it ought to work on some older versions of Python (although no one should be using a version of Python older than ``2.7``. ``2.7.0`` was released 5 years ago).

It is probably worth noting that I am also going to spend some time teaching the build system to build and install the .NET bindings to Z3 under Linux (using mono). That work is going to be based on top of this work but I'd like to submit that as a separate pull request (PR).

However if it going to take a long time (i.e. month(s)) to test this PR would it be better for me to add my commits for .NET support under Linux to this PR first (even though they aren't conceptionally really related)?

wintersteiger:
That would be fantastic if you could add the .NET API to the Linux build. I've built with mono manually before so no major obstacles should be in the way code-wise, but the build will need some tinkering. As you say, this can go into a separate PR.

Python 2.7 is considered the minimum everywhere else too, and everything should work with 3.5. In the last couple of weeks I fixed a few problems with string encoding in 3.5. Annoyingly the default encoding changes when 3.5 runs scripts when piping to a file instead of writing to the console. I don't think there should be any more problems, but if you happen to know, it would be good to get it confirmed that 3.5 works in both settings.

delcypher:
> Python 2.7 is considered the minimum everywhere else too, and everything should work with 3.5. In the last couple of weeks I fixed a few problems with string encoding in 3.5. Annoyingly the default encoding changes when 3.5 runs scripts when piping to a file instead of writing to the console. I don't think there should be any more problems, but if you happen to know, it would be good to get it confirmed that 3.5 works in both settings.

I just tried using Python ``3.5`` and piped to a file when running ``scripts/mk_make.py``. It seemed to work fine.

delcypher:
@wintersteiger Sorry I had to do a small refactor (25d4cab) so that ``$(DESTDIR)$(PREFIX)`` is available outside the ``MakeRuleCmd`` class. I need this to install the ``.NET`` bindings.

I hope that hasn't disrupted your testing too much.

wintersteiger:
I suppose by "outside" you mean via `install_root()`?

delcypher:
> I suppose by "outside" you mean via install_root()?
Yes I do.

delcypher:
@wintersteiger Sorry. I found a bug (fixed in d205b17). I also allowed the names of binary, lib and include directories to be customised in d2ba6f0. Seeing as master has changed I also rebased.

delcypher:
@wintersteiger By the way, I realise the fact I keep rebasing and adding stuff really doesn't help your testing. Sorry about this. I'm currently preparing packages for Z3 and I seem to keep findings problems with this PR that I have to go back and fix. Hopefully once I've successfully made packages this PR will stabilise.

delcypher:
@wintersteiger Just to let you know I've successfully built linux packages using this PR so I won't be modifying it any more so testing this PR should be fine now.

delcypher:
@wintersteiger @NikolajBjorner Is there anything I can do to help get this merged? #340 can't be reviewed properly until this happens.

wintersteiger:
Yes, test it on as many system as you possibly can, using python 2.x and 3.x and check whether all files go into the right directories and whether they are removed upon uninstall. Arch Linux is not a good reference for this at all, at the very least it needs to be tested on Ubuntu and on OSX; the latter of which uses different directory structures than the various linuxes. The $DESTDIR issue first came up on connection with ports on FreeBSD, so I suppose that has been tested thoroughly?

delcypher:
> Yes, test it on as many system as you possibly can, using python 2.x and 3.x and check whether all files go into the right directories and whether they are removed upon uninstall. Arch Linux is not a good reference for this at all, at the very least it needs to be tested on Ubuntu and on OSX;

I have already tested this on Ubuntu 12.04LTS and Ubuntu 14.04LTS (see #345) (along with a few other patches). I do not have access to an OSX machine so I cannot test this.

I do not have a FreeBSD set up either, although I could set one up I'm not very experienced in this area.

>  The $DESTDIR issue first came up on connection with ports on FreeBSD, so I suppose that has been tested thoroughly?

Do you mean #242? If so that's OpenSUSE, not FreeBSD.

``$DESTDIR`` is not a FreeBSD specific thing anyway, this is something that is generally expected of **all build systems** under non-windows. If it works under Linux it's very likely that it will also work under UNIX like operating systems.

``$DESTDIR`` only matters if you set it and its not set by default so the installation locations are not changed from where were before this PR. The only real change for the installation locations in this PR is that the python site-packages directory must now live somewhere underneath the prefix.

Anyway, if you really care about building on these platforms why don't you have continuous integration builds set up for them (even better, have them build PRs and have them be publicly viewable)? It is not realistic to expect every contributor to test their changes under many platforms because

* They might not have access to all of the platforms you care about.
* Even if they do have access they might have the time to run on all the different platforms.

delcypher:
@arrowdodger I noticed you mentioned you were the maintainer of a Z3 port for FreeBSD in #242. Any chance you could test this PR?

This adds ``DESTDIR`` support like the other PR but in a way the @wintersteiger seems to be slightly happier with?

arrowdodger:
@delcypher With these changes i was able to package z3 without any patches on our side. Great work!

I've tested this with python 2.7, but if you wish, i can try python 3 too.

delcypher:
@arrowdodger Thanks for testing. If you could try with python 3 too that would be great.

NikolajBjorner:
thanks, I am still having issues with our Mac. It is configured without mono and also does not have the python library path under /usr/local. The patch to the missing mono is easy: I simply disable DOTNET_ENABLED if the compiler isn't found. In this way users don't have to explicitly toggle this option. We might to something similar for other front-ends, but that can wait. 
For the prefix mismatch, it doesn't suffice to treat this as a warning condition: there is a lot of state later on that depends on the assumption now that the paths are aligned. I would merge the pull request if I can get the mac version to run, but it will take a bit extra work. I don't want to break too many installations up front.

delcypher:
> Thanks, I am still having issues with our Mac. It is configured without mono and also does not have the python library path under /usr/local. The patch to the missing mono is easy: I simply disable DOTNET_ENABLED if the compiler isn't found. In this way users don't have to explicitly toggle this option. We might to something similar for other front-ends, but that can wait. 

@NikolajBjorner Thanks for taking another look. I think you may be confusing this PR with #340. This (#338) PR doesn't add support for mono. It may have been a mistake to have submitted both PRs at the same time given that one is dependant on other.

For now it would be best to concentrate on this PR. We can look at the ``.NET`` binding support once this PR is merged.

> For the prefix mismatch, it doesn't suffice to treat this as a warning condition: there is a lot of state later on that depends on the assumption now that the paths are aligned. I would merge the pull request if I can get the mac version to run, but it will take a bit extra work. I don't want to break too many installations up front.

I believe you are talking about the logic regarding where Python bindings should be installed. I made it a requirement that the location of the python bindings live under the prefix. This is required for the install  and staged installs (using DESTDIR) to work correctly under Linux/FreeBSD.

I had a quick google and looks like the site-packages directory lives under ``/Library/Python/2.7/site-packages`` under OSX which is a bit of problem as that obviously isn't underneath ``/usr/local``. However it looks like if you use a homebrew version of Python you will find it at ``/usr/local/lib/python2.7/site-package`` and for mac ports it will be at ``/opt/local/Library/Frameworks/Python.framework/2.7/site-packages``.

On your mac where is the site-packages folder? What do you think the behaviour should be of Z3's build system when the python ``site-packages`` directory is not under the install prefix? It looks like things are okay for a homebrew install of python (it's under ``/usr/local``) but it looks like its a problem for other Python installs on OSX.

One approach would be to change the logic so that doesn't assume that the Python install lives under the install prefix and will just do an install to ``$(DESTDIR)/path/to/site-packages`` is but emit some sort of warning when this happens. I really don't like this though, the fact this would install something that isn't under the install prefix is **very misleading** and gross.

Perhaps we could when picking the ``site-packages`` directory, iterate through the contents of ``sys.path`` until we find one under the install prefix and give up if there isn't one.

What do you think?

NikolajBjorner:
My local (crude) patches are as follows:


<code>
<pre>
     else:
         monoCompilerExecutable = CSC
     monoCompilerPath = which(monoCompilerExecutable)
+    if CSC == None and monoCompilerPath == None:
+        disable_dotnet()
+        return

                    ". This would lead to a broken installation."
                    "Use --pypkgdir= to change the Python package directory") %
                   (PYTHON_PACKAGE_DIR, PREFIX))
-            sys.exit(1)
+            #sys.exit(1)


 def is_dotnet_enabled():
     return DOTNET_ENABLED

+def disable_dotnet():
+    global DOTNET_ENABLED
+    DOTNET_ENABLED = False
+
 def is_compiler(given, expected):
     """
     Return True if the 'given' compiler is the expected one.
@@ -3418,7 +3425,7 @@ class MakeRuleCmd(object):
         assert isinstance(dest, str)
         assert not ' ' in dest
         assert not os.path.isabs(src_pattern)
-        assert not os.path.isabs(dest)
+        # assert not os.path.isabs(dest)
         cls.write_cmd(out, "cp {src_pattern} {install_root}{dest}".format(
             src_pattern=src_pattern,
             install_root=cls.install_root(),
@@ -3429,7 +3436,7 @@ class MakeRuleCmd(object):
         assert len(pattern) > 0
         assert isinstance(dir, str)
         assert not ' ' in dir
-        assert not os.path.isabs(dir)
+        #assert not os.path.isabs(dir)
         cls.write_cmd(out, "mkdir -p {install_root}{dir}".format(
             install_root=cls.install_root(),
             dir=dir))
@@ -3472,6 +3479,8 @@ class MakeRuleCmd(object):
         assert len(target) > 0
         assert len(link_name) > 0
         assert not os.path.isabs(target)
+        if os.path.isabs(link_name):
+           return
         assert not os.path.isabs(link_name)

         # We can't test to see if link_name is a file or directory
@@ -3523,6 +3532,8 @@ class MakeRuleCmd(object):
         out.write("\t@{}\n".format(line))

 def strip_path_prefix(path, prefix):
+    if not path.startswith(prefix):
+        return path
     assert path.startswith(prefix)
     stripped_path = path[len(prefix):]
     stripped_path.replace('//','/')

</pre>
</code>

This might get the OSX version back to parity before the update. If so, then I suggest to merge your pull request and polish the mac version next.

delcypher:
@NikolajBjorner Crude indeed :)

I'd be very happy to see this PR merged but I think it's worth thinking about what we want to do for OSX. I designed the ``MakeRuleCmd`` carefully to make it very difficult to accidently install stuff outside the install prefix because doing so is likely a mistake. This is why you are having to remove those asserts.

I don't think your patch works though because you'll still end up with ``$(DESTDIR)$(PREFIX)`` in front of any absolute path you try to force into the ``install_files()``, ``remove_installed_files()``, etc... methods. The install won't end up where you want.

NikolajBjorner:
BTW, I am testing whatever "git pull https://github.com/delcypher/z3-1.git better_packaging" gives me.

delcypher:
> BTW, I am testing whatever "git pull https://github.com/delcypher/z3-1.git better_packaging" gives me.

That might not necessarily give you the correct contents. It will merge in my commits into whatever branch you are already on. I personally prefer doing something like this

```
$ git remote add delcypher https://github.com/delcypher/z3-1.git
$ git fetch delcypher
$ git checkout -b test_better_packaging delcypher/better_packaging
```

This will add a remote for my repository and then creates a local branch ``test_better_packaging`` that is the same as my ``better_packaging`` branch.

delcypher:
@NikolajBjorner Maybe you're right though. It might be easier to merge this PR into master (it's already pretty large) and then address the OSX issue.

Then if you want I can implement a special case for the Python bindings under OSX. I won't have anyway to really test them properly though.

NikolajBjorner:
Seems I would also have to modularize the python dependencies: currently other components follow a methodology with components, but python does not. It will make it much easier to disable the installation of the python component when detecting that the paths are not aligned. 

To show what I mean: we have the function

<code>
<pre>
def mk_uninstall(out):
    out.write('uninstall:\n')
    for c in get_components():
        c.mk_uninstall(out)
    pythonPkgDirWithoutPrefix = strip_path_prefix(PYTHON_PACKAGE_DIR, PREFIX)
    MakeRuleCmd.remove_installed_files(out, '{}*.py'.format(os.path.join(pythonPkgDirWithoutPrefix, 'z3')))
    MakeRuleCmd.remove_installed_files(out, '{}*.pyc'.format(os.path.join(pythonPkgDirWithoutPrefix, 'z3')))
    MakeRuleCmd.remove_installed_files(out, '{}*.pyc'.format(os.path.join(pythonPkgDirWithoutPrefix, '__pycache__', 'z3')))
    out.write('\t@echo Z3 was successfully uninstalled.\n')
    out.write('\n')
</pre>
</code>

It loops over components, except for python, which is special.
For the other components, it is possible to selectively enable/disable them based on the state of the local machine. I will do this update and merge the pull tomorrow unless Christoph objects.

NikolajBjorner:
It is now in the master branch. Some of our build machines may react, but at least we can move from this point onwards.

arrowdodger:
@delcypher Building with python 3.4 was also success. Thanks for fixing this.

delcypher:
@arrowdodger No problem. By the way I'm not sure if FreeBSD builds mono (the ".NET" implementation) but if that is installed when building Z3 it will also build and install the ``.NET`` bindings. Those changes are in ``master`` but some additional bugs have been introduced so the ``master`` branch doesn't build properly any more (#357 fixes this and hopefully will be merged in the not to distant future).

wintersteiger:
One comment on continuous integration: We automatically build and test every night (using scripts and tests from Z2Prover/z3test), but we don't have test scripts for testing the installation targets. Contributions are most welcome though!

Thanks for the addtional tests @arrowdodger, that saved us a lot of time!

Regarding python installation: On our test/build mac Python 2.7.10 lives in `/Library/Frameworks/Python.framework/Versions/2.7/bin/` it's been way too long since I installed that, so I can't really remember who put it there, but I suspect it was an original OSX installer package from python.org.

arrowdodger:
@delcypher Is there a flag to turn off build of mono bindings even if mono is present in the system? If it is not, it'd be nice to have it.

delcypher:
@arrowdodger You can use ``--nodotnet`` to disable building the bindings right now. I'm not sure if it will remain that way when the next stable release is made. It's quite confusing right now because the Java bindings are off by default but enabled by ``--java`` and the ".NET" bindings are on by default and switched off by ``--nodotnet``. I really don't like this. I think the ".NET" bindings should be off by default and enabled with a ``--dotnet`` flag so it matches the behaviour for the Java bindings.

PR #340, opened 2015-11-26
Mono dotnet bindings

delcypher:
This is an early preview of my work to get the ".NET bindings" building and working under Linux.

Please do not merge it yet!

This work is dependent on PR on #338 which hasn't been merged to ``master`` yet so a lot of commits that you see here are in that PR. Once that PR is merged I will rebase so you can see the commit(s) that matters.

The commits of interest are 9e288193fd8d707fed26f6589fa995953798f67a and 2514042c2de04614527fca6ef9e449e3f433e186

So far the ".NET bindings" seem to install and I can build the example but I'm having trouble with Monodevelop detecting them. I believe it is because I haven't provided a PkgConfig file (apparently Monodevelop doesn't use the GAC for finding packages).


delcypher:
The build system now generates a pkg-config file too so that Mono develop can find the Z3 bindings. See 65294bc958c70732d1b163ff820f173002239a13

delcypher:
@wintersteiger I've tested under Arch Linux (with Mono 4.0.4 and Monodevelop 5.9.7). The bindings install correctly and Monodevelop is able to find ``Microsoft.Z3.Sharp`` in its list of packages.  For reference I built "split packages" for this to test installation on my system. The PKGBUILD file can be found at

https://gist.github.com/delcypher/a929b1d289270a04040c

When you have time would you be able to take a look? The commits that need reviewing are

9e288193fd8d707fed26f6589fa995953798f67a
2514042c2de04614527fca6ef9e449e3f433e186

The rest are part of PR #338 and shouldn't be reviewed here.

NikolajBjorner:
Seems to work OK on Windows. 
There is a hiccup on the Mac:

z3-mac:z3 nbjorner$ python scripts/mk_make.py
  File "scripts/mk_make.py", line 12, in <module>
    parse_options()
  File "/Users/nbjorner/z3/scripts/mk_util.py", line 715, in parse_options
    printPythonPackageMessage("ERROR")

NikolajBjorner:
The issues seem related to machines that don't have .NET enabled.
The scripts set DOTNET_ENABLED=True by default. So if you invoke the script without any command-line parameters on machines where .NET isn't present, then it ends up in the error paths.
The "printPythonPackageMessage seems to rely on dependencies that are not in scope. The format string passed to the next line has three arguments, the first one is not defined.
Previous behavior was to not die on the python package directories not being under /usr/local, so 
I also wonder if the script should really exit at this point.
When disabling exit, the script goes on, but now assumes that .NET must be on the machine because the default value of DOTNET_ENABLED is True. It seems to me, it should just diasable .NET compilation in default mode.

delcypher:
@NikolajBjorner 
> z3-mac:z3 nbjorner$ python scripts/mk_make.py
File "scripts/mk_make.py", line 12, in 
parse_options()
File "/Users/nbjorner/z3/scripts/mk_util.py", line 715, in parse_options
printPythonPackageMessage("ERROR")

The problem here is that you are reviewing this PR before reviewing #338 . This PR contains all the commits in #338 because it won't work without them. So #338 really needs to be reviewed and merged first before tackling this PR.

The problem you report is part of #338 . I've fixed the issue and rebased this PR on it.

delcypher:
> When disabling exit, the script goes on, but now assumes that .NET must be on the machine because the default value of DOTNET_ENABLED is True. The script then runs into another unrecoverable error. It seems to me, it should just diasable .NET compilation in default mode.

Regarding this, there is a comment on this in 61d1cd524ea1cfd73b0554071ebdb7b891701b4a

```
...

    Building these bindings is unfortunately on by default because
    I didn't want to change the command line interface (i.e. ``--nodotnet``)
    which people might be relying on. This should really be changed to
    match the other binding flags (e.g. ``--java``) but I will leave
    this for now.
....
```

PR #344, opened 2015-11-30
Fixed stray UTF-8 Byte order mark in ``InterpolationContext.cs``.

delcypher:
Old versions of the mono compiler don't like it.


NikolajBjorner:
Thanks

PR #350, opened 2015-12-02
Fix error under Python 3.5

delcypher:
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xef in position
0: ordinal not in range(128)
```

It seems to occur when running in the OpenSUSE build environment but
it doesn't when I run outside. The fix here is to read the file as
binary as explicitly decode and encode as utf8. In Python3 we could
pass an encoding parameter to the call to ``open()`` but that option
isn't available for Python 2.

Whilst I'm here also switch to using ``with`` statements for the calls
to ``open()``. The previous code with explicit ``close()`` calls is very
bad style and also unsafe if an exception gets thrown as the files might
not get closed.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -2276,24 +2276,26 @@ def mk_all_assembly_infos(major, minor, build, revision):
 def mk_assembly_info_version(assemblyinfo, major, minor, build, revision):
     ver_pat   = re.compile('[assembly: AssemblyVersion\("[\.\d]*"\) *')
     fver_pat  = re.compile('[assembly: AssemblyFileVersion\("[\.\d]*"\) *')
-    fin  = open(assemblyinfo, 'r')
-    tmp  = '%s.cs' % assemblyinfo
-    fout = open(tmp, 'w')
-    num_updates = 0
-    for line in fin:
-        if ver_pat.match(line):
-            fout.write('[assembly: AssemblyVersion("%s.%s.%s.%s")]\n' % (major, minor, build, revision))
-            num_updates = num_updates + 1
-        elif fver_pat.match(line):
-            fout.write('[assembly: AssemblyFileVersion("%s.%s.%s.%s")]\n' % (major, minor, build, revision))
-            num_updates = num_updates + 1
-        else:
-            fout.write(line)
-    # if VERBOSE:
-    #    print("%s version numbers updated at '%s'" % (num_updates, assemblyinfo))
-    assert num_updates == 2, "unexpected number of version number updates"
-    fin.close()
-    fout.close()
+    with open(assemblyinfo, 'rb') as fin:
+        tmp  = '%s.cs' % assemblyinfo
+        with open(tmp, 'wb') as fout:
+            num_updates = 0
+            for line in fin:
+                line_decoded = line.decode(encoding='utf8')
+                line_to_write = None
+                if ver_pat.match(line_decoded):
+                    line_to_write = ('[assembly: AssemblyVersion("%s.%s.%s.%s")]\n' % (major, minor, build, revision)).encode(encoding='utf8')
+                    num_updates = num_updates + 1
+                elif fver_pat.match(line_decoded):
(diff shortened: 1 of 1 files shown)


wintersteiger:
The change to with-statements is definitely a good idea, thanks for that fix! 

Regarding the string decoding, this needs a lot more testing. I've recently added some other calls to decode, because Python 3.5 is always strictly unicode everywhere, but which encoding is in fact used depends on the environment the program is run in; for instance, piping the output into a file changes the global string encoding that is used. At one point I also had an explicit `decode('utf-8')` somewhere in the scripts, but I had to remove it because it didn't work in all circumstances, so I suspect we might find a similar problem with this change.

delcypher:
@wintersteiger Okay. 

Thinking about this though this code though you should probably just use the ``configure_file()`` function that is part of #340.

Then you can change the ``AssemblyInfo`` file to

```csharp
using System;
using System.Reflection;
using System.Runtime.CompilerServices;
using System.Runtime.InteropServices;
using System.Security.Permissions;

// General Information about an assembly is controlled through the following 
// set of attributes. Change these attribute values to modify the information
// associated with an assembly.
[assembly: AssemblyTitle("Z3 .NET Interface")]
[assembly: AssemblyDescription(".NET Interface to the Z3 Theorem Prover")]
[assembly: AssemblyConfiguration("")]
[assembly: AssemblyCompany("Microsoft Corporation")]
[assembly: AssemblyProduct("Z3")]
[assembly: AssemblyCopyright("Copyright (C) 2006-2014 Microsoft Corporation")]
[assembly: AssemblyTrademark("")]
[assembly: AssemblyCulture("")]

// Setting ComVisible to false makes the types in this assembly not visible 
// to COM components.  If you need to access a type in this assembly from 
// COM, set the ComVisible attribute to true on that type.
[assembly: ComVisible(false)]

// The following GUID is for the ID of the typelib if this project is exposed to COM
[assembly: Guid("4853ed71-2078-40f4-8117-bc46646bce0e")]

// Version information for an assembly consists of the following four values:
//
//      Major Version
//      Minor Version 
//      Build Number
//      Revision
//
// You can specify all the values or you can default the Build and Revision Numbers 
// by using the '*' as shown below:
// [assembly: AssemblyVersion("x.x.x.x")]
[assembly: AssemblyVersion("@VER_MAJOR@.@VER_MINOR@.@VER_BUILD@.@VER_REVISION@")]
[assembly: AssemblyFileVersion("@VER_MAJOR@.@VER_MINOR@.@VER_BUILD@.@VER_REVISION@")]
```

I would also rename it ``AssemblyInfo.cs.in`` (that is a common convention in build systems for template files)

Then you would simply call the following in the python script.

```python
configure_file('/path/to/AssemblyInfo.cs.in', '/path/to/AssemblyInfo.cs',
  {
     'VER_MAJOR': VER_MAJOR,
     'VER_MINOR': VER_MINOR,
     'VER_BUILD': VER_BUILD,
     'VER_REVISION': VER_REVISION,
  }
)
```

This approach has several pros

* A single implementation for generating build files that need to templated in some way. This avoids reimplementing "yet another function to do text substitution in a file".
* This approach of handling templated files is standard practise for build systems (e.g. autoconf, CMake)

wintersteiger:
Sure, all good ideas. First we should make sure that #340 has no string decoding type of problems then we can conveniently re-use that.

delcypher:
@wintersteiger #340 is still blocked on #338 . Once #338 is merged I can rebase #340 and it can be tested.

NikolajBjorner:
what is the action on this request now that #338, #340 are merged?

delcypher:
I'd like to reimplement is using the ``configure_file()`` function that is now available in master. Once I've done that I'll test it in the environment where I observed the Python 3.5 bug and see if the function ``configure_file()`` has the same encoding bug.

So I'll close this PR.

wintersteiger:
Great, thanks for taking another go at this!

Issue #124, opened 2015-06-07
OSX install wrong locations


skaller:
On OSX, the default install is to /Library, but this is used like a Unix prefix.
This is incorrect. If you install into Library you have to install following
Apple's framework specification. Otherwise please use /usr or /usr/local
exactly as on Linux (I always use /usr/local to separate stuff I install from
unix system stuff and stuff made by or specially for Apple).

Please remove the @ in front of the install commands. The user
needs to SEE where stuff is going. (eg there is no man page).

PR #357, opened 2015-12-07
Fix recent build system breakages

delcypher:
@NikolajBjorner 's recent changes for the python binding support in the build system broke the build for me. Here are a bunch of a fixes and also a small refactor

### DIFF
--- a/scripts/mk_project.py
+++ b/scripts/mk_project.py
@@ -81,17 +81,17 @@ def init_project_def():
             includes2install=['z3.h', 'z3_v1.h', 'z3_macros.h'] + API_files)
     add_exe('shell', ['api', 'sat', 'extra_cmds','opt'], exe_name='z3')
     add_exe('test', ['api', 'fuzzing', 'simplex'], exe_name='test-z3', install=False)
-    add_dll('api_dll', ['api', 'sat', 'extra_cmds'], 'api/dll', 
-            reexports=['api'], 
-            dll_name='libz3', 
-            static=build_static_lib(),
-            export_files=API_files)
+    _libz3Component = add_dll('api_dll', ['api', 'sat', 'extra_cmds'], 'api/dll',
+                              reexports=['api'],
+                              dll_name='libz3',
+                              static=build_static_lib(),
+                              export_files=API_files)
     add_dot_net_dll('dotnet', ['api_dll'], 'api/dotnet', dll_name='Microsoft.Z3', assembly_info_dir='Properties')
     add_java_dll('java', ['api_dll'], 'api/java', dll_name='libz3java', package_name="com.microsoft.z3", manifest_file='manifest')
     add_ml_lib('ml', ['api_dll'], 'api/ml', lib_name='libz3ml')
     add_hlib('cpp', 'api/c++', includes2install=['z3++.h'])
     set_z3py_dir('api/python')
-    add_python_install()
+    add_python_install(_libz3Component)
     # Examples
     add_cpp_example('cpp_example', 'c++') 
     add_cpp_example('z3_tptp', 'tptp') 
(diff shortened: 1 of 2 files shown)


wintersteiger:
Testing this atm.

delcypher:
Great. Thanks :)

wintersteiger:
On our OSX test machine the assertion at line 3584 failes because the Python `path` is not in `prefix`. See also #124. We should at the very least make sure that this works when users install the default packages from python.org. I think it's fine if homebrew, macports, and other managed installation throw errors or require another install directory setting.

delcypher:
@wintersteiger Sorry about that. This was an oversight on my part 753f6d5 should hopefully fix that.

delcypher:
@wintersteiger It might be worth noting as well that @NikolajBjorner changed the behaviour of the build system such that if the Python package directory is not under the install prefix the generated ``Makefile`` will not install the Python bindings. This is better than before because my PR (#338) would refuse to build if it detected this scenario. This might not be ideal.

If we do decide to make an exception for Python under OSX then we need to be very careful about how we do it because installing files outside the installation prefix is counter-intuitive and dangerous.

wintersteiger:
So, the default intstallers from python.org install into `/Library/Frameworks/Python.framework/` (both 2.7 and 3.5) and the installer does not provide an option for installing into a different directory. I think the best way to handle this would be to add a conditional rule that installs into `/Library/...` when DESTDIR is not set. 

At the end of the installation we print "Z3py was installed at ..." and the configuration summary also prints the pkg dir. I think it would be fine to print a warning message if the python pkg dir is not under the prefix to make it clear to the user that they might break something if they go on.

delcypher:
>  I think the best way to handle this would be to add a conditional rule that installs into /Library/... when DESTDIR is not set.

I was actually thinking of doing it slightly differently. I was thinking of providing an extra method(s) in ``MakeRuleCmd`` class that allows things to be installed outside the prefix and only allow it be used on OSX when the word ``python`` appears somewhere in the path. My reasoning here is that ``DESTDIR`` must always be respected. Doing this doesn't prevent a direct install on OSX because ``DESTDIR`` is not normally set so running ``make install`` would install into ``/Library/Frameworks/Python.framework/...``. If ``DESTDIR`` was set then it would end up installing to ``$(DESTDIR)/Library/Frameworks/Python.framework/...``.

Although branching on the value of ``DESTDIR`` inside the Makefile is possible I don't think that's a good route to go down.

wintersteiger:
I see your reasoning, but at the time of running mk_make.py we don't really know whether DESTDIR will be used later or not, so that branch would have to be in the Makefile. If we detect an empty DESTDIR during mk_make.py then we would install into `/Library/...` even if DESTDIR is non-empty at the time of `make install`, right?

delcypher:
@wintersteiger 
> I see your reasoning, but at the time of running mk_make.py we don't really know whether DESTDIR will be used later or not, so that branch would have to be in the Makefile. If we detect an empty DESTDIR during mk_make.py then we would install into /Library/... even if DESTDIR is non-empty at the time of make install, right?

No, I'm proposing something far more simple. Basically when generating the install command in the makefile rule for the Python bindings on OSX we will detect if the python package directory is in the prefix.

If it is then we emit the normal that we do normally.

```
	@cp z3*.py $(DESTDIR)$(PREFIX)/lib/python3.5/site-packages

```

if the python package directory is not in the prefix (and we are on OSX) then we emit this

```
        @echo WARNING python bindings are being installed outside of $(PREFIX)
	@cp z3*.py $(DESTDIR)/Library/Python/3.5/site-packages/ 
```

If we are not on OSX emission of the above is not allowed and running the python script will fail with an error.

If we do it this way

* We are still respecting ``DESTDIR`` in all cases
* If DESTDIR is empty (which it normally is) the bindings will still be installed on OSX to the right location for them to work.
* We emit a warning at install time about the dodgy install that we are doing. We could (and should) emit a warning at configure time too.
* On other platforms we can still ban installing outside of ``PREFIX``.

Does that make sense?

wintersteiger:
Yeah, that sounds great, that way everybody should be happy. Are you ok making those changes to the current PR or should we do the merge first?

delcypher:
I'll make the changes to this PR.

delcypher:
@wintersteiger @NikolajBjorner 

I've added a patch that "should" mean that installing python bindings works under OSX.

Sorry this took longer than intended, I had to redo a lot of my changes when I rebase because they build system had already been changed.

PR #363, opened 2015-12-09
Refactor ``mk_all_assembly_infos()`` to use the ``configure_file()`` and misc fixes for dotnet bindings

delcypher:
Refactor ``mk_all_assembly_infos()`` to use the ``configure_file()``
function. The old implementation was buggy under Python 3.5 and
unsafe (not using with statements on calls to ``open()``).

This is a rework of #350 .

I quickly tested under Python 3.5 in the environment where I previously had crashes (``UnicodeDecodeError: 'ascii' codec can't decode byte 0xef in position
0: ordinal not in range(128)``)

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -2582,42 +2582,23 @@ def mk_version_dot_h(major, minor, build, revision):
     if VERBOSE:
         print("Generated '%s'" % os.path.join(c.src_dir, 'version.h'))
 
-# Generate AssemblyInfo.cs files with the right version numbers by using AssemblyInfo files as a template
+# Generate AssemblyInfo.cs files with the right version numbers by using ``AssemblyInfo.cs.in`` files as a template
 def mk_all_assembly_infos(major, minor, build, revision):
     for c in get_components():
         if c.has_assembly_info():
-            assembly = os.path.join(c.src_dir, c.assembly_info_dir, 'AssemblyInfo')
-            if os.path.exists(assembly):
-                # It is a CS file
-                mk_assembly_info_version(assembly, major, minor, build, revision)
+            assembly_info_template = os.path.join(c.src_dir, c.assembly_info_dir, 'AssemblyInfo.cs.in')
+            assembly_info_output = assembly_info_template[:-3]
+            assert assembly_info_output.endswith('.cs')
+            if os.path.exists(assembly_info_template):
+                configure_file(assembly_info_template, assembly_info_output,
+                               { 'VER_MAJOR': str(major),
+                                 'VER_MINOR': str(minor),
+                                 'VER_BUILD': str(build),
+                                 'VER_REVISION': str(revision),
+                               }
+                              )
             else:
-                raise MKException("Failed to find assembly info file 'AssemblyInfo' at '%s'" % os.path.join(c.src_dir, c.assembly_info_dir))
-
-
-# Generate version number in the given 'AssemblyInfo.cs' file using 'AssemblyInfo' as a template.
-def mk_assembly_info_version(assemblyinfo, major, minor, build, revision):
-    ver_pat   = re.compile('[assembly: AssemblyVersion\("[\.\d]*"\) *')
-    fver_pat  = re.compile('[assembly: AssemblyFileVersion\("[\.\d]*"\) *')
-    fin  = open(assemblyinfo, 'r')
-    tmp  = '%s.cs' % assemblyinfo
-    fout = open(tmp, 'w')
-    num_updates = 0
-    for line in fin:
-        if ver_pat.match(line):
-            fout.write('[assembly: AssemblyVersion("%s.%s.%s.%s")]\n' % (major, minor, build, revision))
-            num_updates = num_updates + 1
-        elif fver_pat.match(line):
-            fout.write('[assembly: AssemblyFileVersion("%s.%s.%s.%s")]\n' % (major, minor, build, revision))
-            num_updates = num_updates + 1
-        else:
-            fout.write(line)
-    # if VERBOSE:
-    #    print("%s version numbers updated at '%s'" % (num_updates, assemblyinfo))
-    assert num_updates == 2, "unexpected number of version number updates"
-    fin.close()
-    fout.close()
-    if VERBOSE:
-        print("Updated '%s'" % assemblyinfo)
+                raise MKException("Failed to find assembly template info file '%s'" % assembly_info_template)
 
 ADD_TACTIC_DATA=[]
 ADD_PROBE_DATA=[]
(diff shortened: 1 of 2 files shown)


wintersteiger:
Is the string decode error what you fixed, or is it that the new behaviour?

delcypher:
@wintersteiger Well I replaced the code that was broken under Python 3.5 (raised the ``UnicodeDecodeError`` in some circumstances) with a call to ``configure_file()`` and that function doesn't seem to exhibit the same problem.

So it's kind of two things in one.

delcypher:
> So it's kind of two things in one.

That wasn't intentional. It's just that it seems like ``configure_file()`` doesn't need fixing (at least on my platform).

wintersteiger:
Aah, I see! The thing to keep in mind is that Python 3.x is always fully unicode-enabled and some command line tools produce more than just ASCII (e.g., the next version of gcc), so sometimes we need to be careful with the decoding. In this case it may be that AsssemblyInfo.cs(.in) contains unicode characters, but I wouldn't have thought that it does.

delcypher:
>  In this case it may be that AsssemblyInfo.cs(.in) contains unicode characters, but I wouldn't have thought that it does.

I checked there are no utf-8 characters in ``AssemblyInfo.cs.in``.

Is there anything I need to do that is blocking this PR being merged?

wintersteiger:
Yes, testing.

wintersteiger:
On my Ubuntu 14.04 with the mono-complete package installed I get

    error CS2008: No files to compile were specified

(but they are specified on the command line), with mcs version 3.2.8.0. The issue may also exist with the old AssemblyInfo (old as in before this pull request). 

To be consistent we should also make the .NET API optional on all platforms (including WIndows). This is already the second incident within as many days, where the mono bindings don't work because of an old (I think) version of mcs. Perhaps we should even add a check for the version of mcs instead of just testing existence. 

Does mcs version 3.x mean that it supports only .NET 3.x or are those numbers unrelated?

Full command line and error see [here](https://gist.github.com/wintersteiger/c04902bdbd82a74e95b8).

delcypher:
> On my Ubuntu 14.04 with the mono-complete package installed I get
> ```
> error CS2008: No files to compile were specified
>(but they are specified on the command line), with mcs version 3.2.8.0. The issue may also exist with the old AssemblyInfo (old as in before this pull request).
> ```

I didn't have any problems when building my package for Ubuntu 14.04LTS. I used the ``mono-devel`` package for building (I'm not sure if ``mono-complete`` includes it). I will try spinning up an Ubuntu 14.04LTS container on my local machine to see if I can reproduce this.

> To be consistent we should also make the .NET API optional on all platforms (including WIndows). This is already the second incident within as many days, where the mono bindings don't work because of an old (I think) version of mcs. Perhaps we should even add a check for the version of mcs instead of just testing existence.

We are in agreement here. I think it's really inconsistent to have ``.NET`` enabled by default but other bindings like the Java bindings off by default. I can knock together a patch to fix this if you like. It will break the command line interface to the ``mk_make.py`` script though.

> Does mcs version 3.x mean that it supports only .NET 3.x or are those numbers unrelated?

The version numbers are unrelated. For mono 2.11 and greater ``mcs`` is compiler for all .NET frameworks. Prior to this ``dmcs`` was the compiler for .NET 4, see http://www.mono-project.com/docs/about-mono/languages/csharp/

wintersteiger:
Great, thanks for taking another look at this! By "breaking the command line interface" I suppose you mean that the option name/flag is going to change? (Fine with me.)

I think all other APIs are consistently either on everywhere or off everywhere, so those should be fine for now.

delcypher:
> I didn't have any problems when building my package for Ubuntu 14.04LTS. I used the mono-devel package for building (I'm not sure if mono-complete includes it). I will try spinning up an Ubuntu 14.04LTS container on my local machine to see if I can reproduce this.

I can reproduce your issue. I built in a clean container (installed ``build-essential``, ``git``, ``mono-complete`` and ``python``). I didn't see this before because I didn't build a debug build before. It looks like one of the debug flags passed ``mcs`` breaks compilation...weird

If I do a release build the ".NET" bindings built fine apart from a warning

```
../src/api/dotnet/InterpolationContext.cs(34,16): warning CS1574: XML comment on `Microsoft.Z3.InterpolationContext.InterpolationContext(System.Collections.Generic.Dictionary<string,string>)' has cref attribute `Context.Context(Dictionary<string, string>)' that could not be resolved
```

I'll take a look at this and see if I can fix it.

delcypher:
@wintersteiger d3d1049 should fix the issue you have seen. Please try my branch again. Note I also rebased on ``master`` so you'll probably need to delete your copy of my branch and fetch it again.

wintersteiger:
Indeed, that particular problem is fixed, but now I see a couple of others:

- the unresolved XML comment cref is strange, I don't get that on Windows (but also not critical...)
- the dotnet_example has the right build rules, but the `examples:` target is missing `_ex_dotnet_example`
- the dotnet_example throws a build error because it doesn't like backslashes (at least on OSX): 

    /usr/bin/mcs /out:dotnet_example /reference:Microsoft.Z3.dll /debug:full /reference:System.Numerics.dll /platform:x86 ..\examples\dotnet/Program.cs
    error CS2001: Source file '..examplesdotnet/Program.cs' could not be found

delcypher:
> * the unresolved XML comment cref is strange, I don't get that on Windows (but also not critical...)

This may well be a bug in mono. I may take a look if I have time but I agree not a critical issue.

> * the dotnet_example has the right build rules, but the examples: target is missing _ex_dotnet_example

I have fixed this in 44393b37debda8627f69249595ce5540e2939ac4

> the dotnet_example throws a build error because it doesn't like backslashes (at least on OSX):
>
> ```
>/usr/bin/mcs /out:dotnet_example /reference:Microsoft.Z3.dll /debug:full /reference:System.Numerics.dll /platform:x86 ..\examples\dotnet/Program.cs
> error CS2001: Source file '..examplesdotnet/Program.cs' could not be found
> ```

Sorry I never built the examples this way before. I only tested building against the installed ``Microsoft.Z3.dll``. I've fixed this issue in 439d8d8afe7c1e9c4f6d38613177965e06d2d82c

wintersteiger:
Alright this looks good now. There are a number of warnings about unused variables, but we can fix that fter this PR is merged.

delcypher:
@wintersteiger 
> Great, thanks for taking another look at this! By "breaking the command line interface" I suppose you mean that the option name/flag is going to change? (Fine with me.)

I forgot to reply to this bit. Yes I meant this would the option name would change. I've not implemented this. Would you still like a PR for this?

wintersteiger:
No worries, I've changed a few things around earlier today, now I'm waiting for the automatic builds to throw errors, but I don't expect too many of them.

PR #375, opened 2015-12-12
Tidy up readme

delcypher:


delcypher:
If you'd like to see how this looks when rendered on GitHub see

https://github.com/delcypher/z3-1/tree/tidy_up_readme

wintersteiger:
This will take some time to review. We were not planning to turn README into a Github frontpage. If we want to revamp this we should separate the information into README/INSTALL/CONTRIBUTING/MAINTAINERS/etc. Also, as of yet I'm not sure that markdown is the right format for that; we may want to have README _and_ a README.md.

delcypher:
@wintersteiger Okay. The markdown is one part of it but there also some meaningful changes here that correct bits of the README that are incorrect.

NikolajBjorner:
Thanks!!!

PR #385, opened 2015-12-22
Fixing building under OS X

tigertoes:
Previous method `is_osx_hack` didn't exist, possibly wasn't cleaned up. This change has been tested and Z3 compiles successfully within a 2.7 virtualenv running on OS X 10.10.5.


wintersteiger:
Good catch, thanks for the fix!

Issue #387, opened 2015-12-25
segfault in datalog join planner


seanmcl:
I'm regularly hitting a segfault in the following instruction 

    Stack: [0x00000001267fa000,0x00000001268fa000],  sp=0x00000001268f8400,  free space=1017k
    Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
    C  [libz3.dylib+0x1f431c]  _ZN7datalog12join_planner21remove_rule_from_pairENSt3__14pairIP3appS4_EEPNS_4ruleEj+0xbc

    Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
    j  com.microsoft.z3.Native.INTERNALfixedpointQuery(JJJ)I+0
    j  com.microsoft.z3.Native.fixedpointQuery(JJJ)I+4
    j  com.microsoft.z3.Fixedpoint.query(Lcom/microsoft/z3/BoolExpr;)Lcom/microsoft/z3/Status;+23

I have a 254M gzipped (3G unzipped) core dump file if you'd like me to send it.

seanmcl:
It's difficult to produce a SMT-LIB2 format file for this bug, since I don't think it's possible to define finite domain sorts, which my segfaulting example uses.  Is there a workaround for this?

seanmcl:
Fixed by https://github.com/Z3Prover/z3/commit/38865ffe0daea832693046e928a9919cf88b2ed5

PR #389, opened 2015-12-31
Bandaid for segfault in datalog planner

seanmcl:
This change avoids a deallocation that was crashing z3.  The underlying
problem is unresolved.

Fixes (at least temporarily) https://github.com/Z3Prover/z3/issues/387


NikolajBjorner:
I added a different patch. It should assert in debug mode, but not crash in release mode.

PR #390, opened 2015-12-31
CMake/CLion support

seanmcl:
To use the CLion IDE from Jetbrains, we need to build
Z3 with CMake.  This change adds CMake support to Z3
and allows developement with CLion with minimal effort.

I'm not sure if the Z3 team is interested in having this on the master branch.  But it took enough time to figure out that I figured others may benefit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,10 @@ src/api/ml/z3.mllib
 *.bak
 doc/api
 doc/code
+.idea/
+/CMakeLists.txt
+/cmake/bin/*
+/cmake/lib/*
+/cmake/include-dirs.txt
+/cmake/sources-libz3.txt
+/cmake/sources-z3.txt
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1,0 +1,33 @@
+cmake_minimum_required(VERSION 3.3)
+
+project(z3)
+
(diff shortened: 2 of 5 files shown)


seanmcl:
Will re-submit when I have the CLA worked out.  Anyone wanting to work with CLion contact me privately until then.

wintersteiger:
Sounds good. In general I'd say we are interested in this addition, but only if it is maintained by someone with CMake experience and in a timely manner. 

In the pull request, the Makefile itself seems rather simple though. It seems to me that there is only one particular combination of `CMAKE_CXX_FLAGS`, while the default configuration scripts will run a multitude of tests to figure out which flags need to be enabeld. So, this is probably still a lot of work until it's more widely usable.

delcypher:
@wintersteiger @seanmcl 
I didn't realise a CMake build system would be accepted based on the discussion I saw in https://github.com/Z3Prover/z3/pull/242

If I had known that I probably would have started implementing one. I'd happily be the maintainer for it but it needs a lot more work that what I saw in @seanmcl 's PR. That implementation is missing **a lot of things**.

seanmcl:
If it wasn't clear, this PR was not intended to replace the python
scripts.  It was intended to develop and debug in CLion, for which it is
sufficient on OSX.

PR #401, opened 2016-01-06
Codesmells in Java API

cheshire:
Hi,

Z3 Java API is quite messy, and quite unidiomatic (I assume many of those things are C++-isms).
This change attempts to clean it up.
As a bonus, two bugs were uncovered in a process as well.

### DIFF
--- a/src/api/java/AST.java
+++ b/src/api/java/AST.java
@@ -29,6 +29,7 @@ public class AST extends Z3Object implements Comparable<AST>
      * 
      * @param o another AST
      **/    
+    @Override
     public boolean equals(Object o)
     {
         AST casted = null;
@@ -56,6 +57,7 @@ public class AST extends Z3Object implements Comparable<AST>
      * positive if after else zero.
      * @throws Z3Exception on error
      **/
+    @Override
     public int compareTo(AST other)
     {
         if (other == null)
@@ -74,6 +76,7 @@ public class AST extends Z3Object implements Comparable<AST>
      * 
      * @return A hash code
      **/
+    @Override
     public int hashCode()
     {
         int r = 0;
@@ -188,6 +191,7 @@ public class AST extends Z3Object implements Comparable<AST>
     /**
      * A string representation of the AST.
      **/
+    @Override
     public String toString()
     {
         try
@@ -217,6 +221,7 @@ public class AST extends Z3Object implements Comparable<AST>
         super(ctx, obj);
     }
 
+    @Override
     void incRef(long o)
     {
         // Console.WriteLine("AST IncRef()");
@@ -226,6 +231,7 @@ public class AST extends Z3Object implements Comparable<AST>
         super.incRef(o);
     }
 
+    @Override
     void decRef(long o)
     {
         // Console.WriteLine("AST DecRef()");
--- a/src/api/java/AST.java
+++ b/src/api/java/AST.java
@@ -53,15 +53,10 @@ public class AST extends Z3Object implements Comparable<AST>
     @Override
     public int compareTo(AST other)
     {
-        if (other == null)
+        if (other == null) {
             return 1;
-
-        if (getId() < other.getId())
-            return -1;
-        else if (getId() > other.getId())
-            return +1;
-        else
-            return 0;
+        }
+        return Integer.compare(getId(), other.getId());
     }
 
     /**
@@ -94,11 +89,12 @@ public class AST extends Z3Object implements Comparable<AST>
     public AST translate(Context ctx)
     {
 
-        if (getContext() == ctx)
+        if (getContext() == ctx) {
             return this;
-        else
+        } else {
             return new AST(ctx, Native.translate(getContext().nCtx(),
-                    getNativeObject(), ctx.nCtx()));
+                getNativeObject(), ctx.nCtx()));
+        }
     }
 
     /**
--- a/src/api/java/AST.java
+++ b/src/api/java/AST.java
@@ -72,12 +72,7 @@ public class AST extends Z3Object implements Comparable<AST>
     @Override
     public int hashCode()
     {
-        int r = 0;
-        try {
-            Native.getAstHash(getContext().nCtx(), getNativeObject());
-        }
-        catch (Z3Exception ex) {}
-        return r;
+        return Native.getAstHash(getContext().nCtx(), getNativeObject());
     }
 
     /**
--- a/src/api/java/AST.java
+++ b/src/api/java/AST.java
@@ -28,24 +28,17 @@ public class AST extends Z3Object implements Comparable<AST>
      * Object comparison.
      * 
      * @param o another AST
-     **/    
+     **/
     @Override
     public boolean equals(Object o)
     {
-        AST casted = null;
+        if (o == null) return false;
+        if (o == this) return true;
+        if (o.getClass() != this.getClass()) return false;
+        AST casted = (AST) o;
 
-        try
-        {
-            casted = AST.class.cast(o);
-        } catch (ClassCastException e)
-        {
-            return false;
-        }
-
-        return  (this == casted) ||
-                (this != null) &&
-                (casted != null) &&
-                (getContext().nCtx() == casted.getContext().nCtx()) &&
+        return
+            (getContext().nCtx() == casted.getContext().nCtx()) &&
                 (Native.isEqAst(getContext().nCtx(), getNativeObject(), casted.getNativeObject()));
     }
 
--- a/src/api/java/AST.java
+++ b/src/api/java/AST.java
@@ -32,9 +32,8 @@ public class AST extends Z3Object implements Comparable<AST>
     @Override
     public boolean equals(Object o)
     {
-        if (o == null) return false;
         if (o == this) return true;
-        if (o.getClass() != this.getClass()) return false;
+        if (!(o instanceof AST)) return false;
         AST casted = (AST) o;
 
         return
--- a/src/api/java/Context.java
+++ b/src/api/java/Context.java
@@ -373,7 +373,7 @@ public class Context extends IDisposable
            
     {
         checkContextMatch(names);
-        int n = (int) names.length;
+        int n = names.length;
         ConstructorList[] cla = new ConstructorList[n];
         long[] n_constr = new long[n];
         for (int i = 0; i < n; i++)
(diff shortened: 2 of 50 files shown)


wintersteiger:
Fantastic, thanks for all the hard work! The Java API can definitely do with some polishing. 

It will take me some time to go through that to see what I have to expect to be broken after this. For instance, right away the first change in AST::equals will break applications that use older versions of Java (pre 1.7), so some serious testing will be necessary before we merge this. Can you tell me what systems and flavours of Java you tested all of this on? Also: did you actually test everything or just a small subset that you use in your application?

src/api/java/AST.java
```diff
@@ -28,23 +28,16 @@
...
-            casted = AST.class.cast(o);
-        } catch (ClassCastException e)
-        {
-            return false;
-        }
-
-        return  (this == casted) ||
-                (this != null) &&
-                (casted != null) &&
-                (getContext().nCtx() == casted.getContext().nCtx()) &&
```
wintersteiger: `o` can be an object other than an AST (at least in older versions of Java, but perhaps not in 1.8). I remember adding those casts for that particular reason. What is the difference between `o instanceof AST` and `try {AST.class.cast(o)} catch ...`? Will this behave the same on all platforms and through all versions of Java? If not, then I'd rather keep the old code instead of spending a lot of time on testing the new code.

cheshire:
@wintersteiger TBH I have done very minimal amount of testing, and I should probably do more. I assume there are no hidden unit-tests anywhere?
I am using Java 7. However I don't quite understand what change do you mean --- is it the `@Override` annotation? That should work since at least Java 1.6. 

Currently we are still using Z3 C API in our JavaSMT project, if we switch to Z3 Java API, we could leverage JavaSMT's unit-tests as well, but we are mostly blocked by the finalizers issue I have mentioned before. Maybe if I get some hard benchmarking numbers and do the pull request we can negotiate about changing the garbage-collecting strategy?

src/api/java/AST.java
```diff
@@ -74,14 +63,10 @@ else if (getId() > other.getId())
...
+    @Override
     public int hashCode()
     {
-        int r = 0;
-        try {
-            Native.getAstHash(getContext().nCtx(), getNativeObject());
-        }
-        catch (Z3Exception ex) {}
-        return r;
+        return Native.getAstHash(getContext().nCtx(), getNativeObject());
```
wintersteiger: By declaration, getAstHash can throw Z3Exception (see Native.java), but hashCode() is not supposed to throw exceptions, so we have to catch it here.

src/api/java/Context.java
```diff
@@ -322,13 +326,6 @@ public Constructor mkConstructor(Symbol name, Symbol recognizer,
 
     /**
      * Create a datatype constructor. 
-     * @param name  
-     * @param recognizer  
-     * @param fieldNames  
-     * @param sorts  
-     * @param sortRefs 
-     * 
-     * @return
```
wintersteiger: Instead of removing these, can you copy over the parameter descriptions from the previous function?

src/api/java/AST.java
cheshire: If class is known at a compile time, there's no difference between the two.
I have been using Java extensively since about 1.5, and TBH I have never even seen the `class.cast` form.
It looks highly un-idiomatic to me, people either use `getClass() == other.getClass()` to verify the exact class match, or `instanceof` if inheritance is allowed.

src/api/java/Context.java
```diff
@@ -391,10 +388,6 @@ public DatatypeSort mkDatatypeSort(String name, Constructor[] constructors)
 
     /**
      * Create mutually recursive data-types. 
-     * @param names  
-     * @param c 
-     * 
-     * @return
```
wintersteiger: Same here, copy description from function before this one.

src/api/java/AST.java
cheshire: Well yes, but the code above always returns zero.
Do you want to cache the hash value? Or just catch the exception?

wintersteiger:
No, the override annotation is fine, that shouldn't cause any trouble. It's the exception coming out of AST.class.cast, that was there for a particular reason (that escapes me at the moment), but the semantics of equals changed in Java 1.7, because the old behaviour was too complicated.

wintersteiger:
The garbage collection strategy is fine for now, we shouldn't worry about that too much. At this point, the performance of that code in the Java API isn't really a concern. If the API is used to set up millions of new problems per second, then it's probably best to switch over to a C level layer anyways. 

Which finalizers issue are you talking about? As far as I am concerned, all the finalization is just fine at the moment.

wintersteiger:
Regarding testing: Without testing or an active application to drive it, we can't merge changes to thousands of lines of code, so this PR will have to be blocked until that is done.

src/api/java/FuncDecl.java
```diff
@@ -370,9 +358,6 @@ void checkNativeObject(long obj)
 
     /**
      * Create expression that applies function to arguments. 
-     * @param args 
-     * 
-     * @return
```
wintersteiger: Fill in instead of removing the comments.

src/api/java/FuncDecl.java
```diff
@@ -29,35 +29,22 @@
...
+            (Native.isEqFuncDecl(
+                getContext().nCtx(),
+                getNativeObject(),
+                other.getNativeObject()));
     }
 
-    /**
-     * A string representations of the function declaration.
-     **/
+    @Override
```
wintersteiger: Keep documentation, fill in the missing parts instead of removing it.

cheshire: `FuncDecl#toString` returns a string representation of the function declaration just based on the function name, doesn't it?

src/api/java/Context.java
```diff
@@ -3878,7 +3870,7 @@ public void updateParamValue(String id, String value)
     }
 
     protected long m_ctx = 0;
-    protected static Object creation_lock = new Object();
+    protected static final Object creation_lock = new Object();
```
wintersteiger: What are the consequences of this `final`? Will this break at runtime when someone tries to create multiple Context objects at the same time?

src/api/java/IDecRefQueue.java
```diff
@@ -21,7 +21,7 @@
 
 public abstract class IDecRefQueue
 {
-    protected Object m_lock = new Object();
+    protected final Object m_lock = new Object();
```
wintersteiger: Like Context::creation_lock: Will this break at runtime in parallel programs, or when a parallel garbage collector is used?

src/api/java/Sort.java
```diff
@@ -27,25 +27,20 @@
 {
     /**
      * Equality operator for objects of type Sort. 
-     * @param o
-     * @return
```
wintersteiger: Fill in instead of removing the comments.

src/api/java/Context.java
cheshire: It's a compile-time annotation.
It will not allow to compile a code which changes the creation lock.

src/api/java/IDecRefQueue.java
cheshire: Same as above.

src/api/java/Status.java
```diff
@@ -38,7 +38,7 @@
         this.intValue = v;
     }
 
-    public static final Status fromInt(int v)
+    public static Status fromInt(int v)
```
wintersteiger: What will be the consequences of this change?

wintersteiger:
Alright, I've added a bunch of comments to the commits, would be nice if you could take a look at those. 

It might make sense to split this PR into two, separating the purly syntactic changes from the others. Then we can at least merge all those annotations (do they have any consequences at all?), curly braces and indentation fixes right away.

src/api/java/Status.java
cheshire: None whatsoever, "final" does not make sense when added to static methods (as child classes can override the method anyway).
Annotations should only appear when they have meaningful semantics, and this one does not.

cheshire:
Hi, yes, sure.
I'll try to connect the Z3 Java API to JavaSMT, then I can use the existing unit tests.

> do they have any consequences at all?
Not at runtime. They act at compile time only.

The point of this changelist was not to be exactly the same at runtime, but it got a bit unwieldy.

wintersteiger:
That would be great if you could hook up JavaSMT!

Sure, just makes code review a lot easier if you get rid of all the syntactic changes first.

NikolajBjorner:
Thanks! These contain a lot of improvements!

src/api/java/Sort.java
cheshire: Can it be done in a separate pull request/commit? No information is being removed here.

wintersteiger: It's not about not removing information, but about adding useful information. It's fine though, just remind me when you're done and I'll do it myself.

cheshire:
Hi, we have tested this branch with our rather extensive https://github.com/sosy-lab/java-smt/ test coverage. Everything appears to be working fine.

What should I do to get this merged? Add additional Javadoc comments to the cases where it was removed?

cheshire:
Personally I would prefer to merge it as-is, because no documentation was _removed_ in this PR (empty Javadoc was not conveying any information). In my opinion adding documentation is a separate problem, but of course it is up to you.

wintersteiger:
Yes, the documentation should have been filled in. Does javadoc report any warnings or other issues that weren't there before?

cheshire:
> Does javadoc report any warnings or other issues that weren't there before?

Nothing new I can see.
Speaking of Javadoc, how is one supposed to get it / view it?
The generated JAR files doesn't have the source code attached (it is often customary to ship two JARs for readability, with and without the source code).
The full Java code isn't available until the `configure` was run, and it's not clear where the final Java code is / how to get the HTML output from it.

wintersteiger:
For the API doc website that's generated by doxygen, see [z3api.dox](https://github.com/Z3Prover/z3/blob/master/doc/z3api.dox), but you can also run javadoc directly, e.g., in src\api\java do

    javadoc *.java enumerations\*.java

It currently reports 8 errors and 100 warnings. (I'll fix at least the errors later today.)

cheshire:
Hi @wintersteiger I am somewhat confused --- am I supposed to fix the warnings before this commit was merged? Weren't most of the warnings there before?

wintersteiger:
No, you should have filled in the documentation instead of removing the macros. At the time you had your cursor on them it would have been zero effort. Now someone else (probably myself) has to go and figure out what's missing. 

So, yes, OK to merge. I expect Java 1.6 was not tested, so if there are any users of that era left, we may get complaints about those equality functions after merging this.

PR #409, opened 2016-01-10
Respect DESTDIR when installing OCaml bindings

delcypher:
Please note I don't really use the OCaml bindings. I was just testing to see what files would be installed when running ``make install`` and I noticed that it was broken because it would try to install files outside ``DESTDIR``.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -3593,6 +3593,15 @@ class MakeRuleCmd(object):
         return "$(DESTDIR)$(PREFIX)/"
 
     @classmethod
+    def _is_str(cls, obj):
+        if sys.version_info.major > 2:
+            # Python 3 or newer. Strings are always unicode and of type str
+            return isinstance(obj, str)
+        else:
+            # Python 2. Has byte-string and unicode representation, allow both
+            return isinstance(obj, str) or isinstance(obj, unicode)
+
+    @classmethod
     def _install_root(cls, path, in_prefix, out, is_install=True):
         if not in_prefix:
             # The Python bindings on OSX are sometimes not installed inside the prefix.
@@ -3610,9 +3619,9 @@ class MakeRuleCmd(object):
     @classmethod
     def install_files(cls, out, src_pattern, dest, in_prefix=True):
         assert len(dest) > 0
-        assert isinstance(src_pattern, str)
+        assert cls._is_str(src_pattern)
         assert not ' ' in src_pattern
-        assert isinstance(dest, str)
+        assert cls._is_str(dest)
         assert not ' ' in dest
         assert not os.path.isabs(src_pattern)
         install_root = cls._install_root(dest, in_prefix, out)
@@ -3625,7 +3634,7 @@ class MakeRuleCmd(object):
     @classmethod
     def remove_installed_files(cls, out, pattern, in_prefix=True):
         assert len(pattern) > 0
-        assert isinstance(pattern, str)
+        assert cls._is_str(pattern)
         assert not ' ' in pattern
         install_root = cls._install_root(pattern, in_prefix, out, is_install=False)
 
@@ -3636,7 +3645,7 @@ class MakeRuleCmd(object):
     @classmethod
     def make_install_directory(cls, out, dir, in_prefix=True):
         assert len(dir) > 0
-        assert isinstance(dir, str)
+        assert cls._is_str(dir)
         assert not ' ' in dir
         install_root = cls._install_root(dir, in_prefix, out)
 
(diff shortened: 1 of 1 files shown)


wintersteiger:
Haven't had time to test any of this yet, but a crucial piece of additional information is that OCaml being installed and found does not mean that ocamlfind is also installed and found. It's a separate tool and we want to be able to build the API even if it can't be installed via ocamlfind. Could you check whether that is still the case after your changes?

delcypher:
@wintersteiger 
> Haven't had time to test any of this yet, but a crucial piece of additional information is that OCaml being installed and found does not mean that ocamlfind is also installed and found. It's a separate tool and we want to be able to build the API even if it can't be installed via ocamlfind. Could you check whether that is still the case after your changes?

Ah. I assumed that if ``--ml`` is passed to the configure then it would be a reasonable requirement to also have ``ocamlfind`` so my changes raise an exception if ``ocamlfind`` is missing (previously it would just silently not install the bindings). I could change this to a warning if you like? It seems a little strange though. I don't think any of the other bindings can built but not installed.

delcypher:
@wintersteiger 

I thought I might give a concrete example in an attempt to justify my reasoning.

The ".NET" bindings require the ``gacutil`` utility in order to install and if the ``gacutil`` is not found during the configure step then the configure script fails. I consider this to be the right behaviour here because.

* The ``gacutil`` utility is usually available if mono is installed
* Enabling bindings on the command line (i.e.``--dotnet``) and having them built but not installed is strange.

The behaviour I've switched too for OCaml is very similar. ``ocamlfind`` is required to install the Z3 bindings and with my changes if it is not found the configure step fails.

wintersteiger:
Like I said, the difference is that OCaml does not necessarily come with ocamlfind. See [here](http://ocaml.org/docs/install.html), look for findlib. That's why there were checks like `if ... OCAMLFIND=''`. Most users will have install it, but some don't, nothing 'strange' about that. 

Regardless of this, I think all APIs should build without error, but the installation can fail if some of those tools only relevant for installation are missing. I think a warning as you suggest is a much better idea than refusing to build them and it doesn't hurt us in any way.

delcypher:
@wintersteiger I've redone the commits so that failing to find ``ocamlfind`` just gives a warning.

wintersteiger:
Looks good, thanks for the fixes! Did you test this to check whether it actually works with non-empty $(DESTDIR), or is this pure conjecture so far?

delcypher:
@wintersteiger Conjecture. I will test now.

wintersteiger:
Great, thanks!

delcypher:
Okay so I think installation works properly when using Python3. On Ubuntu 14.04 the ocaml files end up in

```
tree /usr/local/lib/ocaml/
/usr/local/lib/ocaml/
└── 4.01.0
    ├── stublibs
    │   ├── dllz3ml.so
    │   └── dllz3ml.so.owner
    └── Z3
        ├── libz3ml.a
        ├── META
        ├── z3.cmi
        ├── z3enums.cmi
        ├── z3enums.mli
        ├── z3ml.a
        ├── z3ml.cma
        ├── z3ml.cmxa
        ├── z3ml.cmxs
        ├── z3.mli
        ├── z3native.cmi
        └── z3native.mli

3 directories, 14 files
```

I don't know how to use OCaml so I don't know if that's right

But unfortunately there's a bug when using Python 2. I assumed that the ``check_output()`` function always returns a ``str`` but with Python 2 this does not seem to be the case (I get back an object of type ``Unicode``) which later leads to an assertion failure in my code (``MakeRuleCmd.make_install_directory()``). How would you like me to handle this?

wintersteiger:
That looks alright to me. Not sure about libz3ml.a, might have to go into stublibs as well, but can't check right now. 

Regarding strings, the crucial difference is that Python 3.x is unicode, only and always. In 2.x you can have unicode if you want, but it's not necessary to always use it. There are a couple calls to `decode` in mk_util.py and update_api.py, it's probably one of those that don't behave right, or we need to add one.

delcypher:
@wintersteiger I decided to go for the simplest change (250c8d0) (I really didn't want to change how ``check_output()`` behaves just in case you need to support Unicode paths). I've made the asserts weaker in the ``MakeRuleCmd`` class to allow ``unicode`` strings.

Hopefully you'll be happy with this change.

OCaml is very unfamiliar territory for me so I don't know if the ``ocamlfind`` tool is behaving correctly here.

wintersteiger:
Sure, totally happy with that change. There may well be a more elegant way, but for now that should work.

PR #427, opened 2016-02-02
Convert stream to bool explicitly

jwakely:
In C++11 there is no implicit conversion from iostream classes to `void*`, just an explicit conversion to bool.


wintersteiger:
Look good to me, thanks for the fix!

PR #441, opened 2016-02-08
Lackr

MikolasJanota:
Adding code for explicit Ackermannization. This includes a qfufbv solver based on Ackermannization and a tactic to remove UFs by Ackermannization.


delcypher:
@MikolasJanota If you want someone to review this you should rebase on master (Merge commits aren't helpful for reviewing) and squash commits so there are as few commits as possible.

wintersteiger:
Good suggestion from @delcypher. In this particular case, the PR has already been reviewed outside of github.

PR #444, opened 2016-02-10
Explicitly state what the units of the timeout parameter for the "smt" module are.

delcypher:
Note I guessed that the unit are in milliseconds because that's what the units of the other timeout parameters seem to be. If this is incorrect please let me know what the units are and I will correct this PR.


NikolajBjorner:
Thanks!

PR #446, opened 2016-02-10
Fix gcc build failure on ARM caused by including <emmintrin.h>

msullivan:
src/util/hwf.cpp tries to use <emmintrin.h> to directly use SSE
intrinsics. Make sure to only use those when actually on x86 or
x86_64.

### DIFF
--- a/src/util/hwf.cpp
+++ b/src/util/hwf.cpp
@@ -29,7 +29,8 @@ Revision History:
 #include<fenv.h>
 #endif
 
-#ifndef _M_IA64
+#if defined(__x86_64__) || defined(_M_X64) ||	\
+    defined(__i386) || defined(_M_IX86)
 #define USE_INTRINSICS
 #endif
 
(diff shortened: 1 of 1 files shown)


wintersteiger:
IA64 machines are the only ones to natively support the FMA operation, which we want intrinsics for. Did you remove the `_M_IA64` macro no purpose?

msullivan:
That `#ifndef` turns on intrinsics if `_M_IA64` *isn't* defined. My read of that was that it meant to disable them for all non-x86 platforms and was written when IA64 was the only such one. 

It looks like the way the IA64 FMA instruction is used is by setting the `fp_contract` pragma (https://github.com/Z3Prover/z3/blob/94453033b63db76647ae2fad2e14878277f6c10b/src/util/hwf.cpp#L301-L303), just doing the operation (https://github.com/Z3Prover/z3/blob/94453033b63db76647ae2fad2e14878277f6c10b/src/util/hwf.cpp#L311-L314), and then clearing the pragma (https://github.com/Z3Prover/z3/blob/94453033b63db76647ae2fad2e14878277f6c10b/src/util/hwf.cpp#L333-L335).

wintersteiger:
Thanks for taking another look! You're right, I totally forgot about the fpcontract, that should be fine then.

PR #453, opened 2016-02-16
Fix incorrect (off by one) bound check in clause_allocator

delcypher:
Fix incorrect (off by one) bound check. Also assert that we don't
increment ``m_num_segments`` beyond the maximum value
(``c_max_segments``).

This is related to #436.

When doing an AddressSanitized build and running the ``c_example``
it looks like Z3 tries to create too many segments and index out of
bounds. Fixing the checks here causes them to fail which should help
us narrow down the problem.


NikolajBjorner:
Thanks! yes

PR #459, opened 2016-02-25
Implement a CMake build system.

delcypher:
This build system isn't entirely independent from the old build system yet because it doesn't know how to do

* exec_pyg_scripts()
* mk_pat_db()
* mk_all_install_tactic_cpps()
* mk_all_mem_initializer_cpps()
* mk_all_gparams_register_modules()

Because of this in order for this build system to work you need to
manually invoke ``python scripts/mk_make.py --build /some/dummy/path``
first so that the generated source files get created. The fact that the
old build system pollutes the source tree with generated files is bad
design but I've abused that here just so I can get the CMake build
system up and running. I will fix this in the future but I haven't done
it yet because it requires that I refactor the above python functions so

* they can be called from CMake
* they emit files into the build tree and not the source tree.

Currently the build succeeds on Linux using CMake 3.4.3 using
the "Unix Makefiles" generator. It is able to install files (``make
install``) but not uninstall (I will implement this later). The build
also builds the C API example successfully.

The CMake build should work on other platforms but I have not really paid attention to the compiler
flags needed by MSVC so future work will need to fix that.

This implementation is substantially superior to #390 in that I've spent a lot of effort to either replicate the behaviour of the existing build system or I've made a note of what needs fixing in the future.

@wintersteiger @NikolajBjorner @nunoplopes 
My hope is you will be able to merge this fairly quickly as I've not modified any existing files, I've just added new ones. Future work to improve the CMake build system will require that I refactor the old build system which would then require a thorough review and testing.

I'd also like to say that I am happy to be the maintainer for the CMake build system. My hope is that eventually the CMake build system will get to feature parity with the old build system at which point we can retire the old the build system.


delcypher:
@jirislaby @arrowd This PR may interest you. Note as it says in the commit message the build system isn't quite independent from the python build system yet but provided this PR gets accepted it will be in the not too distant future.

NikolajBjorner:
So your take is that cmake should replace the scripts we use. 
Could you tune me into what it would now enable?

Each directory in the pull request contains txt files that list dependencies. 
Your message above indicates that these are auto-generated, which would 
be rather useful.

[(BTW, the TSE group is building cloud-scale build systems and is hiring :-) )](http://research.microsoft.com/en-us/projects/tse/)

delcypher:
@NikolajBjorner 
> So your take is that cmake should replace the scripts we use. 
> Could you tune me into what it would now enable?

CMake would **eventually** be able to replace those scripts but what is in this PR cannot replace those python scripts yet. This PR is a step in that direction however.

Using CMake rather than your own build scripts offers many advantages

* Support for various generators like Makefiles, Ninja (gives you superfast incremental builds) and a bunch of IDEs like Visual Studio, Xcode...

* CMake makes it easy to sort out issues like support ``DESTDIR``, rpath for binaries (means binaries in the build directory know where their required shared libraries are so no need for LD_LIBRARY_PATH), setting SOVERSION, creating source distributions

* Using CMake opens up the possibility of building Z3 with a cross compiler in the future.

> Each directory in the pull request contains txt files that list dependencies. 
Your message above indicates that these are auto-generated, which would 
be rather useful.

I think I miscommunicated because that is **not what I meant**. The ``CMakeLists.txt`` files are **not automatically** generated they must be in the source tree. They currently contain declarations like this...

```
z3_add_component(fpa
  SOURCES
    fpa2bv_converter.cpp
  INCLUDE_DIRS
    "ast"
    "ast/rewriter"
    "ast/simplifier"
    "math/polynomial"
    "math/automata"
    "util"
)
```

``z3_add_component()`` is a convenience macro I wrote.

* file names that come after ``SOURCES`` specify the source files used to build a component. This **cannot** be automatic. I've seen people do things with ``(file GLOB ...)`` so they don't have to list source files which is **really bad**. In CMake the source files **must be** explicit otherwise CMake doesn't know when it needs to reconfigure.

* paths under ``INCLUDE_DIRS`` set the include paths needed for the component. I hope one day I can remove this but I can't without changing the source files. The core of the problem is Z3's source files include header files like this

```
#include "fpa2bv_rewriter_params.hpp"
```

Doing this style of include really complicates the build system but things can be made **much simpler** if you simply write your includes rooted in the ``src`` directory. 

```
#include "ast/fpa/fpa2bv_rewriter_params.hpp"
```

That way the compiler only needs to be passed a single include path, the path to the ``src/`` directory. Not only does this make the build system **much simpler** it also makes your code much clearer because it is more obvious where the include file actually lives. It would also prevent problems where two components have a header file with the same name. Although this doesn't look like a problem right now it looks like an unnecessary risk to me.

Anyway my main focus was to get a CMake build system set up without changing your source files so this why the ``z3_add_component()`` macro requires a list of include directories.

NikolajBjorner:
I used a build environment with such .txt files a while ago and there 
is some aggregate hassle to maintain such.txt files manually.

It seems to me that fixing the include paths to be relative to src is scriptable and can be enforced using scripts (the python build scripts already check include paths). This is done practically once (and occasionally if new includes are added without paths). Then the .txt files need only depend on the set of .cpp files in the current directories, so their format becomes mostly uniform and therefore also scriptable using a script that walks the source directories, collects the .cpp files and lists them as sources.

delcypher:
@NikolajBjorner 
> I used a build environment with such .txt files a while ago and there 
is some aggregate hassle to maintain such.txt files manually.

In my experience the hassle with CMake is fairly minimal. Adding a source file now to text file now and then is a small price to pay for a reliable build system that does incremental builds properly.

> It seems to me that fixing the include paths to be relative to src is scriptable and can be enforced using scripts (the python build scripts already check include paths).

Sure this could be fixed at some point. But that is orthogonal to getting this PR accepted.

> This is done practically once (and occasionally if new includes are added without paths). Then the .txt files need only depend on the set of .cpp files in the current directories, so their format becomes mostly uniform and therefore also scriptable using a script that walks the source directories, collects the .cpp files and lists them as sources.

Given my experience with CMake I would say this is a bad idea. These ``CMakeLists.txt`` files will not be uniform and they are meant to be under source control. They may look uniform right now but I will need add extra code to some of them to support generating some the automatically generated header files and source files in the future.

I find it hard to believe that the Z3 devs are renaming/removing or adding new source files so frequently that modifying text files becomes a serious drain on productivity. I would say that the current build system (i.e. the python/makefile build system) is a much more serious drain on productivity. It is incredibly unreliable (I basically have to through away the entire build tree everytime I do a ``git pull`` and build from scratch) and the code itself is quite messy making it hard to maintain.

delcypher:
@NikolajBjorner Just to note I'm not saying get rid of the python/Makefile build system right away. The CMake build system can live along side the existing build system for now. They can co-exist for as long as necessary. I am happy to take on the work on maintaining the CMake build system.

It is perfectly reasonable to sceptical of new build system if you've never used but I would suggest just giving it a try once I get it in good working order.

nunoplopes:
Dan: the question is what does cmake really buys us?  The set of people that work on Z3 is fairly small, and I don't think anyone has trouble with the current system.
Also, Z3 has a bunch of files that are automatically generated, so not all python scripts can co away.

Finally, I'm not supportive of merging any solution that is not complete, because the risk of then having to live with something half-backed is high and then *we* will pay the price.

While I'm not against cmake in principle (I do use it in other projects), I fail to see value in changing the build system at this point, and even less in merging a non-complete solution.

delcypher:
> Dan: the question is what does cmake really buys us? The set of people that work on Z3 is fairly small, and I don't think anyone has trouble with the current system.

I've already mentioned some the advantages that CMake gives. Apart from those I've already mentioned the reduced maintenance (maintaining a completely custom build system is not something I would ever want to be doing) seems like a big advantage to me. As for people having trouble with the current system I certainly have problems with it and some other downstream users of Z3 do too.

> Finally, I'm not supportive of merging any solution that is not complete, because the risk of then having to live with something half-backed is high and then we will pay the price.

I don't really get your logic here. There is currently no price to pay at all for merging this

* It touches none of the existing code. The existing build system is completely unaffected.
* I am offering to maintain it so you don't even need to care about the CMake build system. If I ever disappear completely and no one steps up to maintain the build system you can remove it without breaking anything.

Partially implemented stuff gets merged into projects all the time and then gets gradually improved on. How is this any different? I cannot replicate all the functionality of the existing build system immediately. That is too much to do in one go so I want to do it incrementally.

> While I'm not against cmake in principle (I do use it in other projects), I fail to see value in changing the build system at this point, and even less in merging a non-complete solution.

The only reason I embarked on doing this work (apart from my dissatisfaction with the current build system) is that the reception to #390 seemed reasonably warm. I was not expecting such a negative reaction to this PR.

If the Z3 team really want to keep the python build system and never have the CMake build system replace it then that is the team's choice and there's nothing I can do about that. But that doesn't mean you shouldn't accept this contribution. I am willing to maintain the CMake build system (on my own if I have to).

NikolajBjorner:
It is really wonderful that you are doing this.
The rest is a question of available bandwidth, mechanics, sustainability, etc..

delcypher:
@NikolajBjorner @nunoplopes 
If it makes things easier I can implement more of the CMake build system in this PR but that would involve touching the existing build system and I wanted to keep my PRs as small as possible so they are easier to review.

arrowd:
Since i can't comment on individual files of the pull request, i'll do this there.

@delcypher 
Regarding `cmake/target_arch_detect.cmake` - do you really need this? Wouldn't [CMAKE_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/v3.4/variable/CMAKE_SYSTEM_PROCESSOR) or [CMAKE_SIZEOF_VOID_P](https://cmake.org/cmake/help/v3.4/variable/CMAKE_SIZEOF_VOID_P) suffice in this case? There is also [TargetArch module](https://github.com/petroules/solar-cmake/blob/master/TargetArch.cmake) if we really care about cross-compiling.

```
+  list(APPEND Z3_COMPONENT_CXX_DEFINES "-D_FREEBSD_")
```

AFAIK, every OS have a define that is set automatically. At least for FreeBSD there is `__FreeBSD__` which is always defined, so i'd use it, instead of introducing another one.

```
+  else()
+    message(FATAL_ERROR "USE_OPENMP was enabled but a working OpenMP install"
+            " could not be found. Try disabling it by passing "
+            "-DUSE_OPENMP:BOOL=FALSE to CMake")
+  endif()
```
From my personal experience, the more canonical thing to do there is to just turn off a feature and issue a warning to let the user know. It is a good tone when configuring with default options always succeeds even if optional stuff is missing.

```
+if (BUILD_LIBZ3_SHARED)
+  if (NOT MSVC)
+    z3_add_cxx_flag("-fPIC" REQUIRED)
+  endif()
+endif()
```
Are sure this is needed? Adding `SHARED` to `add_library` does this automatically.

Haven't tried out this yet, will do on weekend.

delcypher:
@arrowd 
> Since i can't comment on individual files of the pull request, i'll do this there.
>
> Regarding cmake/target_arch_detect.cmake - do you really need this? Wouldn't > CMAKE_SYSTEM_PROCESSOR or CMAKE_SIZEOF_VOID_P suffice in this case? There is also TargetArch module if we really care about cross-compiling.

CMAKE_SYSTEM_PROCESSOR could be used but I like the approach of ``TargetArch`` more so I reimplemented a small subset of it. I didn't want to bring in external code because I'm not sure if I can contribute that.

>  list(APPEND Z3_COMPONENT_CXX_DEFINES "-D_FREEBSD_")
> AFAIK, every OS have a define that is set automatically. At least for FreeBSD there is __FreeBSD__ which is always defined, so i'd use it, instead of introducing another one.

I'm just reimplementing what the existing build system does. The source code currently uses ``_FREEBSD_``.

>  else()
>    message(FATAL_ERROR "USE_OPENMP was enabled but a working OpenMP install"
>            " could not be found. Try disabling it by passing "
>            "-DUSE_OPENMP:BOOL=FALSE to CMake")
>  endif()
>
> From my personal experience, the more canonical thing to do there is to just turn off a feature and  issue a warning to let the user know. It is a good tone when configuring with default options always  succeeds even if optional stuff is missing.

I see your point. My concern with just switching off OpenMP support if it is not found is that if someone passes ``-DUSE_OPENMP:BOOL=TRUE`` explicitly they might not get an OpenMP build of Z3 is their OpenMP install is broken.

+if (BUILD_LIBZ3_SHARED)
+  if (NOT MSVC)
+    z3_add_cxx_flag("-fPIC" REQUIRED)
+  endif()
+endif()
Are sure this is needed? Adding SHARED to add_library does this automatically.

I think this is because I'm not building libz3 in a conventional way.

> Haven't tried out this yet, will do on weekend.

You might want to hold off on that. I've been trying to implement some of the missing functionality (dependence on the existing building system to generate some ``.h`` and ``.cpp`` files) and I've realised I need to redo quite a few things.

delcypher:
I'm closing this PR as I've realised I need to change the design quite a bit. I'll resubmit when it's in better shape.

PR #467, opened 2016-03-01
Fix build

4tXJ7f:
Previous commits seem to have removed some variable declarations, so the
build did not work. This patch reintroduces the variables.


wintersteiger:
This was already fixed in master, but thanks for the fix anyways!

Issue #464, opened 2016-02-29
Z3py: FPVal(-0.0, Float32()) returns +0.0


4tXJ7f:
Creating an FPVal from -0.0 results in +0.0:

```
>>> get_version_string()
'4.4.1'
>>> y = FPVal(-0.0, Float32())
>>> y
+0.0
>>> y = FPVal("-0.0", Float32())
>>> y
+0.0
```

wintersteiger:
Thanks for the report! Those conversion were indeed buggy, but that should all be fixed now.

4tXJ7f:
Awesome, thanks a lot!

PR #470, opened 2016-03-01
[Z3py] Fix conversion of float -0.0

4tXJ7f:
Issue #464 was partially fixed by c171170 but there was still a problem
when creating an FPVal from the float -0.0 (from string '-0.0' worked
fine). Reason for the problem was the use of
float.as_integer_ratio() which does not return an accurate
representation of -0.0 because the denominator is always positive. This
fix explicitly handles -0.0.

### DIFF
--- a/src/api/python/z3.py
+++ b/src/api/python/z3.py
@@ -8409,11 +8409,16 @@ def FPSort(ebits, sbits, ctx=None):
 
 def _to_float_str(val, exp=0):
     if isinstance(val, float):
-        v = val.as_integer_ratio()
-        num = v[0]
-        den = v[1]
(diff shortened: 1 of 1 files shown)


4tXJ7f:
Sounds good :) FWIW `float.as_integer_ratio()` works for other negative numbers.

wintersteiger:
Not for -oo! Turns out this is all a bit more complicated in Python, but I hope I caught all relevant cases now (see [this](https://github.com/Z3Prover/z3/commit/59e695f2be96e586ebb7c9c7560ec12c7436f970)).

4tXJ7f:
Oh yeah, I did not think about that.

PR #483, opened 2016-03-05
Bugfix for arith_rewriter single operand division

### LINKED ISSUE #481: `is_irrational_algebraic_numeral` et al. may receive null arguments
Here is a simple demonstration of what could trigger this bug
```
(set-option :numeral-as-real true)
(declare-const x Real)

(simplify (acos (/ 2)))
```

I would fix this but I am not sure if the issue should be tackled at the `simplify_cmd` level or if the arithmetic utility functions should be changed to return an errno. Much of the simplification code receives malformed input happily, so fixing this somewhat minor bug could be a fairly large undertaking unless z3 just panics when processing empty literals (which would probably be best anyway)

zv:
fixes #481


NikolajBjorner:
Thanks, I have a bunch of other local changes right now that I am testing, so thanks for submitting this one.

PR #498, opened 2016-03-09
Refactor generated file code out of ``mk_util.py`` and into ``mk_genfile_common.py``

delcypher:
Refactor generated file code out of ``mk_util.py`` and into ``mk_genfile_common.py``.

This an attempt at work discussed by myself and @wintersteiger  in #461 

The idea was to pull all the code that is used by both the CMake and Python build system out of ``mk_util.py`` and decouple that code from the Python build system. This offers several advantages

* Having the shared Python and CMake build system code in a separate file makes it clear which code is only used by the Python build system (i.e. ``mk_util.py``) and which code is used by both build systems. 

* When pulling down new code the CMake build system is less likely to unnecessarily rebuild code due to ``mk_util.py`` changing.

Whilst I was doing this I also tidied up a few things

* Removed some unnecessary use of global state
* Fixed some bad indentation I had introduced in #461

I'm reasonably confident that I didn't break anything but of course this will require more testing. The existing build system seems to work (both Python3 and Python2 under Arch Linux with gcc) and the CMake build system seems to be still working (Arch Linux with gcc and Windows with MSVC).

Something that still needs fixing is ``update_api.py`` but that's a much bigger problem to be tackled in the future.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,6 +343,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 # FIXME: We should drop the dependency on ``mk_util.py``
 set(Z3_GENERATED_FILE_EXTRA_DEPENDENCIES
   "${CMAKE_SOURCE_DIR}/scripts/mk_util.py"
+  "${CMAKE_SOURCE_DIR}/scripts/mk_genfile_common.py"
 )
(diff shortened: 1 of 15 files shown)


wintersteiger:
Looks good after some preliminary tests. Please run some more tests, especially with non-Python APIs just to make sure nothing that we don't have regression tests for breaks.

Issue #477, opened 2016-03-03
Z3 crashes under high memory load


kirschju:
Hi z3 folks!

I've recently observed a crash in libz3.so when I run z3 under very high memory pressure. The crash is in [small_object_allocator.cpp in Line 111](https://github.com/Z3Prover/z3/blob/master/src/util/small_object_allocator.cpp#L111). It apparently happens because the free list contains an invalid element that is clearly not a pointer (such that dereferencing r crashes). This is the dump of the `m_free_list` array at the time of the crash:

	(gdb) x/200xg $rdi
	0x7f31680eee00:	0x0000000000000000	0x0000000000000000
	0x7f31680eee10:	0x0000000000000000	0x0000000000000000
	0x7f31680eee20:	0x00007f2c40776748	0x0000000000000000
	0x7f31680eee30:	0x00007f25a0d0f858	0x00007f2532156ce8
	0x7f31680eee40:	0x00007f289a849ab8	0x00007f3168104678
	0x7f31680eee50:	0x00007f2beb8e86c8	0x00007f312c8a75f8
	0x7f31680eee60:	0x00007f312c8a55e8	0x0000000000000000
	0x7f31680eee70:	0x0000000000000000	0x0000000000000000
	0x7f31680eee80:	0x0000000000000000	0x0000000000000000
	0x7f31680eee90:	0x0000000000000000	0x0000000000000000
	0x7f31680eeea0:	0x0000000000000000	0x0000000000000000
	0x7f31680eeeb0:	0x0000000000000000	0x0000000000000000
	0x7f31680eeec0:	0x0000000000000000	0x0000000000000000
	0x7f31680eeed0:	0x0000000000000000	0x0000000000000000
	0x7f31680eeee0:	0x0000000000000000	0x0000000000000000
	0x7f31680eeef0:	0x0000000000000000	0x0000000000000000
	0x7f31680eef00:	0x0000000000000000	0x0000000000000000
	0x7f31680eef10:	0x0000000000000000	0x0000000000000000
	0x7f31680eef20:	0x00007f2c40777d38	0x0000000000000000
	0x7f31680eef30:	0x00007f282835cbc8	0x0000000000007f01
	0x7f31680eef40:	0x00007f289a849b08	0x00007f3168104838
	0x7f31680eef50:	0x00007f2beb8e8778	0x00007f312c8a7608
	0x7f31680eef60:	0x00007f312c8a55f8	0x0000000000000000
[...]

PR #500, opened 2016-03-09
In ``small_object_allocator`` remove early return for Debug builds

delcypher:
In ``small_object_allocator`` remove early return for Debug builds on non-windows platforms.

The existing code meant that the ``SASSERT()`` calls were never checked
on non-windows platforms because they were only "syntactically"
reachable in Release builds (in which they are no-ops). The comments
suggest that this early return was for Valgrind but this doesn't seem
like a good idea. If false positives occur when running Z3 under
valgrind then a suppression file should be used rather than trying to
change Z3's behaviour. This is related to #477 .


nunoplopes:
It's very useful to be able to use memory debugging tools.  Pool alocators break all of them, so a jail break seems a good trade-off.

PR #503, opened 2016-03-09
Fix bug in ufbv tactic that enabled ackermannization even if unsat co…

martin-neuhaeusser:
…re or proof generation are requested

There probably is a bug in the UFBV tactic that enables Ackermannization even if unsat core or proof generation are enabled. Such a configuration aborts as soon as the first unsat core is requested.


wintersteiger:
Yep, that's the right thing to do (at least for now). Thanks for the fix!

wintersteiger:
@MikolasJanota: If you have some cycles, it would be nice to get at least unsat cores working.

PR #504, opened 2016-03-09
[CMake] Provide a way to customise the install directories

delcypher:
[CMake] Provide a way to customise the install directories used for
executables, include files and libraries. We use
``GNUInstallDirs.cmake`` which ships with CMake to do the difficult work
of setting a sensible default and setting up CMake cache variables.
These can be overriden when running CMake by setting the
``CMAKE_INSTALL_BINDIR``, ``CMAKE_INSTALL_INCLUDEDIR`` and
``CMAKE_INSTALL_LIBDIR`` cache variables.

@jirislaby Could you please test this? For the use case you mentioned to me in #461 you can do ``cmake -DCMAKE_INSTALL_LIBDIR=lib64 /path/to/z3/root`` if you really need to force the library installation directory it but I think ``GNUInstallDirs.cmake`` should pick ``lib64`` by default.

@wintersteiger Could you hold off merging this until @jirislaby reports everything is okay?


jirislaby:
On 03/09/2016, 03:23 PM, Dan Liew wrote:
> @jirislaby <https://github.com/jirislaby> Could you please test this?
> For the use case you mentioned to me in #461
> <https://github.com/Z3Prover/z3/pull/461> you can do |cmake
> -DCMAKE_INSTALL_LIBDIR=lib64 /path/to/z3/root| if you really need to
> force the library installation directory it but I think
> |GNUInstallDirs.cmake| should pick |lib64| by default.

It does not work automatically, neither with -DCMAKE_INSTALL_LIBDIR:

cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=lib64
-DBUILD_LIBZ3_SHARED=true -DUSE_LIB_GMP=true
-DBUILD_PYTHON_BINDINGS=true -DINSTALL_PYTHON_BINDINGS=true
...
CMake Warning:
  Manually-specified variables were not used by the project:

     CMAKE_INSTALL_LIBDIR

I see:
/dev/shm/xxx//home/abuild/rpmbuild/BUILDROOT/z3-20160309-0.x86_64/usr/lib/libz3.so.4.4.2.0

thanks,
-- 
js

delcypher:
@jirislaby Did you remember to wipe your source tree of the old ``CMakeLists.txt`` files and then repopulate with the latest ``CMakeLists.txt`` files before running CMake? i.e. run

```
contrib/cmake/bootstrap remove
contribute/cmake/boostrap create
```

I keep forgetting to do that myself which can result in very confusing behaviour. That's the only way the warning you show would make sense because ``CMAKE_INSTALL_LIBDIR`` is definitely used in this PR.

jirislaby:
Yep, I always use a clean tree.

But I didn't actually _apply_ the patch in the proper tree. Now it works automatically, indeed:
[   27s] -- CMAKE_INSTALL_LIBDIR: "lib64"
[   27s] -- CMAKE_INSTALL_BINDIR: "bin"
[   27s] -- CMAKE_INSTALL_INCLUDEDIR: "include"

Thanks.

delcypher:
@jirislaby Great. Thanks for testing.

@wintersteiger This PR is ready to go :)

PR #505, opened 2016-03-10
Improvement in lazy ackermannization.

MikolasJanota:
Lazy ackermannization now creates more than one congruence lemma at a time. An option was added to use the incremental SAT solver (defaulting to false as it's not performing well).


wintersteiger:
Looking good!

PR #506, opened 2016-03-10
Update README.md

hongjiawu:
Hello there!
I am using Z3 as a paragon to study the software architecture. When I did the presentation to my fellow students about Z3, they showed strong interest on it. However, after they forked Z3, they lost the interest on contributing it for the reason it is hard to understand the application and background in  a short time. Essentially, Z3 should be recognized as a powerful and functional software of practical use in computer science as long as they know the background with some degree.
Z3 deserves more contributors to develop it. So I think some straightforward information should be posted on readme which leads the guy who is green hand of the background information can quickly know what is Z3 and what Z3 can do in a short time. According to personal experience, I think the background information in wiki can play this role. Maybe not in the way I did, but more effective information about Z3 background should be posted on readme and link the green hand to the potential contributors quickly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 Z3 is a theorem prover from Microsoft Research. It is licensed
 under the [MIT license](LICENSE.txt).
 
+If you are not familiar with Z3, you can start [here](https://github.com/Z3Prover/z3/wiki#background).
+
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
Just using Z3 already requires quite some background.

PR #513, opened 2016-03-13
Fix #510

### LINKED ISSUE #510: Broken build
PR #498 broke the build on at least Ubuntu x86 and amd64, see [here](https://github.com/Z3Prover/bin/blob/master/nightly/z3-x86-ubuntu-failure-11-03-2016-21-21-14.log) and [there](https://github.com/Z3Prover/bin/blob/master/nightly/z3-x64-ubuntu-failure-11-03-2016-21-23-52.log). @delcypher could you take a look at this?

delcypher:
I've tested on

* Arch Linux (Python 3.5.1 and Python 2.7.11)
* Ubuntu 14.04LTS (Python 2.7.6)

I'm still waiting for the MSVC build (Python 3.4.2, Windows 8, Visual Studio 2013) to finish.

Note I'm not a heavy user of the Python APIs so someone should check I've not broken anything but a quick basic test seems to indicate they still work.

### DIFF
--- a/scripts/mk_genfile_common.py
+++ b/scripts/mk_genfile_common.py
@@ -105,9 +105,9 @@ def mk_z3consts_py_internal(api_files, output_dir):
                 if m:
                     name = words[1]
                     z3consts.write('# enum %s\n' % name)
-                    for k in decls:
-                        i = decls[k]
-                        z3consts.write('%s = %s\n' % (k, i))
+                    # Iterate over key-value pairs ordered by value
+                    for k, v in sorted(decls.items(), key=lambda pair: pair[1]):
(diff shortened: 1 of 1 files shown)


delcypher:
MSVC build succeeds.

delcypher:
I think this is good to go. I've retested on Ubuntu 14.04LTS, Arch Linux and Windows. The generated ``.cpp`` files are identical (apart from the line endings) in all the builds and everything builds.

The generated ``.cpp`` files are obviously not identical to the files currently generated in the ``master`` branch because I am now enforcing a different order. However they appear to be semantically equivalent.

wintersteiger:
This seems to have done the trick, thanks for the fix!

PR #517, opened 2016-03-15
Add methods for obtaining numeral values in C++ API

yaqwsx:
The C++ API is missing a lot of funcionality compared to C API. I don't like mixing C and C++ API in my code, so I added a small piece of missing functionality.


NikolajBjorner:
thanks

Issue #520, opened 2016-03-18
Feature variability 


hongjiawu:
Dear Microsoft team,

I study Z3 from software architecture view. However, I have divergence with my tutor in feature variability of Z3. Personally, I think the theory Z3 can support, for example bit-vector, arrays, arithmetic etc, is the feature variability of Z3. Is that correct?

Thank you very much!

wintersteiger:
Z3 supports all [SMT theories ](http://smtlib.cs.uiowa.edu/theories.shtml) and a few others or new ones like datatypes, sequences, strings. It also features some techniques that are not sound and complete (where possible) decision procedures, for instance, optimization techniques and incomplete procedures like local search (for some theories), and a tactic framework that allows you to construct your own procedures and integrate them. 

But, I'm not sure that's what feature variability means to you as Z3 always comes with all those features, some users might just chose not to use some of them.

hongjiawu:
Thank you very much! I study Z3 from the software architecture view instead of functional view. But you give me clear idea from the overall figure of the project.

PR #521, opened 2016-03-20
[CMake] Add CMAKE_INSTALL_PYTHON_PKG_DIR option to control where python

delcypher:
Add CMAKE_INSTALL_PYTHON_PKG_DIR option to control where python bindings are installed.


JeevesLibrarian:
The file `z3/src/util/file_path.cpp` does not exist in the repository. Issue #520 (which this depends on) already converted `file_path.h` to C23 with inline functions, making a separate .cpp file unnecessary.

Should this issue be:
1. Closed as already complete (functionality is in the .h file)?
2. Converted to a different task?

The current `file_path.h` contains only one inline function (`get_extension`) and is already pure C23-compatible.

PR #526, opened 2016-03-22
build: allow overriding of 'ar' in mk_config

bpowers:
This will still use 'ar' if AR isn't set in the environment, but lets
us override the default archive tool at configure time.

Just like CC and CXX, this doesn't apply to a ./configure for Windows.


wintersteiger:
That looks good, thanks for the fix/feature!

PR #527, opened 2016-03-23
Fix typo in examples/*/README

tabe:
They are using the Z3 C binding, rather than C++.


wintersteiger:
That's right, thanks for the fix!

PR #528, opened 2016-03-23
Clean up README

EinNarr:
k_api_doc.py also generates API documentation for Java and C++, which are not mentioned in this README file.


wintersteiger:
That's correct, thanks for the fix!

PR #529, opened 2016-03-23
Mbp fix

agurfinkel:
A few minor fixes to mbp code


NikolajBjorner:
I use a slightly different fix for the "dec" bug.

PR #537, opened 2016-03-24
adding flag to build Z3 without direct loading of the Z3-library with Java-Bindings.

kfriedberger:
This allows the user to choose other loading mechanisms, for example with a different path for the Z3-library.
This flag is needed, if no static linking is used for the binary and also no LD_LIBRARY_PATH or other searching mechanisms.
For example, in combination with our project, we know that libz3.so and libz3java.so are in the same directory.
The flag JAVA_EXTERNAL_LIBRARY_LOADING_FLAG allows to load libz3.so directly from the same directory than other libraries that access it.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -2724,6 +2724,7 @@ def mk_bindings(api_files):
           dotnet_output_dir=dotnet_output_dir,
           java_output_dir=java_output_dir,
           java_package_name=java_package_name,
+          java_load_library_directly=not use_java_external_library_loading(),
           ml_output_dir=ml_output_dir
         )
         cp_z3py_to_build()
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -81,7 +81,7 @@ VS_PROJ = False
 TRACE = False
 DOTNET_ENABLED=False
 JAVA_ENABLED=False
-JAVA_EXTERNAL_LIBRARY_LOADING=False
+JAVA_LOAD_LIBRARY=True
 ML_ENABLED=False
 PYTHON_INSTALL_ENABLED=False
 STATIC_LIB=False
@@ -620,7 +620,7 @@ def display_help(exit_code):
         print("  --optimize                    generate optimized code during linking.")
     print("  --dotnet                      generate .NET bindings.")
     print("  --java                        generate Java bindings.")
-    print("  --javaExternalLibraryLoading  the user has to load the Z3-library himself.")
+    print("  --disable-java-load           the user has to load the Z3-library himself.")
     print("  --ml                          generate OCaml bindings.")
     print("  --python                      generate Python bindings.")
     print("  --staticlib                   build Z3 static library.")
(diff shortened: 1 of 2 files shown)


scripts/update_api.py
```diff
@@ -489,10 +489,11 @@ def mk_java(java_dir, package_name):
...
 
-    java_native.write('  static {\n')
-    java_native.write('    try { System.loadLibrary("z3java"); }\n')
-    java_native.write('    catch (UnsatisfiedLinkError ex) { System.loadLibrary("libz3java"); }\n')
-    java_native.write('  }\n')
+    if not mk_util.use_java_external_library_loading():
+        java_native.write('  static {\n')
+        java_native.write('    try { System.loadLibrary("z3java"); }\n')
+        java_native.write('    catch (UnsatisfiedLinkError ex) { System.loadLibrary("libz3java"); }\n')
+        java_native.write('  }\n')
```
delcypher: @kfriedberger This is going to cause problems for me when I implement support for building the Java bindings for the CMake build system. Please try to avoid having ``update_api.py`` depending on code in ``mk_util.py``. You should make ``use_java_external_library`` a parameter to ``mk_java()`` and ``generate_files()`` so that when ``generate_files()`` is called from ``mk_util.py`` you can set the option. You will also need to update the ``main()`` function code in ``update_api.py`` so this option can be controlled from the command line.

kfriedberger:
@delcypher ok, second try :-)

delcypher:
@kfriedberger Thanks. I don't use the Java API so I can't really comment much on the rest but I'll have a quick look at the rest.

scripts/mk_util.py
```diff
@@ -2358,6 +2365,8 @@ def mk_config():
             CPPFLAGS     = '%s -DNDEBUG -D_EXTERNAL_RELEASE' % CPPFLAGS
         if TRACE or DEBUG_MODE:
             CPPFLAGS     = '%s -D_TRACE' % CPPFLAGS
+        # '$$' is needed to escape '$' in the makefile, the result should be '$ORIGIN'
+        JAVA_EXTERNAL_LIBRARY_LOADING_FLAG = '-Wl,-rpath,\'$$ORIGIN\' ' if use_java_external_library_loading() else ''
```
delcypher: @kfriedberger I seem to remember that using ``rpath``  is considered insecure if the library is installed (https://bugs.archlinux.org/task/14049). Is this necessary? 

Note for others looking at this: for Linux it looks like ``'$ORIGIN'`` has a special meaning which is the directory of the library. So the system linker will look for ``libz3.so`` in the same directory as the ``libz3java.so``

kfriedberger:
In this case rpath is needed to find `libz3.so` from `libz3java.so`.
We assume that both files are in the same directory,
but we can only load the `libz3java.so` with our loading mechanism (point of view is a Java program).
For loading `libz3.so`, the system falls back to the default behaviour and would not find it, if we would not use `rpath`.  Thus it is necessary in this case.

wintersteiger:
Generally a valid observation for all APIs, but is it necessary that we add an option to mk_maky.py for this? I think it would be nicer if we just gave the Java API (and the others) the option to load the dll/so either in "automatic" or "manual" mode. I'm not a java programmer though, so I have no clue how much effort it would be to teach the static block in Native.java to do that. But, it would at least work on all platforms, e.g., the current fix doesn't implement this for Windows. 

Not important, but I really don't like the name of the option `--javaExternalLibraryLoading` feels very different from all the other option names. Maybe `--no-java-load`?

cheshire:
@wintersteiger @kfriedberger  Doing it in Java is trivial, I would just lazy load it from some (context object?) constructor, if and only if a certain global variable is not set and depending on a certain option passed.

kfriedberger:
I do not see how the suggested implementation in Java should work.
We have 3 different components:
1) `libz3.so` in some directory
2) `libz3java.so` in the same directory, it needs access to `libz3.so`
3) the Java-library `com.microsoft.z3` which needs access to `libz3java.so`.

We (from our software) can only influence the paths, where Java looks for libraries.
I do not know how to tell `libz3java.so` where to look for `libz3.so`, 
except at the point of compilation or linking, or via system-library-paths.
The currently working solution (at least on Linux) is to use `rpath $ORIGIN`.
I have no plan how to do this on Windows.

wintersteiger:
@kfriedberger: The idea would be to tell `libz3java.so` where it should look for `libz3.so`. At the moment it leaves all of this to the system (Java and OS), but it would be easy to add a function for adding paths to look into (but not easy, at least for me, to do that inside of static blocks). 

@cheshire: Sure, doing that in the Context constructor would be easy, but there are some functions that don't use contexts, e.g., global parameter setting, version information, etc. So, those would have to be done in a different way. Do you have any idea whether this would be possible static way, e.g., such that the current static block is executed only after some other static block (which adds the extra paths)?

In Java, is there any way to modify the environment, e.g., to modify LD_LIBRARY_PATH? That would probably be the easiest way to achieve that, and it might also work in static blocks (but we'd still need a compile-time variable I think).

cheshire:
@wintersteiger I don't know how to do what you are proposing.
Generally, in Java nothing is done inside static blocks and a separate factory class is used to do any "real" work which needs to be done.

Regardless of that, [easily] setting environment variables from Java is not really supported.

What steps should we take from our side to get this pull request merged?

wintersteiger:
@cheshire: I'm not convinced we need/want a compile-time flag for that. Regardless, if we do introduce one, then we'll have to do that for all APIs and OSs (the problem is exactly the same, if it is indeed a problem). Even if we decide to make an exception for Java, we have to make it work on Windows and OSX. If I have to implement those items, it has to wait until I find the time to do that. 

Could you elaborate on why the Java system library path (`java -Djava.library.path=/path/to/libs ...`) doesn't do this job? (Apparently there's even a [hack](http://blog.cedarsoft.com/2010/11/setting-java-library-path-programmatically/) that makes that mutable at runtime.)

cheshire:
@wintersteiger Setting Java system library path doesn't do the job because we ship the library, and we don't have control over how users will run their Java process which uses that library.

However, your hack seems to work, so I think we'll use that instead [though of course it's also hacky, as setting java.library.path is not platform-independent, etc, etc]. Thank you very much for the link! Somehow it eluded us.
I think this pull request can be closed.

wintersteiger:
That's great, thanks for reporting back! I'll close this PR, but I'll remember that sometimes we need static libraries (this problem also would go away if libz3java.so contained a static copy of libz3.so). We may decide to add options for that in the future.

PR #539, opened 2016-03-27
[CMake] Teach CMake to build .NET bindings

delcypher:
This implements support for the CMake build system to build the .NET bindings. If the .NET toolchain is Mono support for installation of the .NET bindings is also implemented. Like the Python bindings this support is off by default but can be enabled by setting ``BUILD_DOTNET_BINDINGS`` to ``ON``.

This required some minor refactoring of the python code. It is almost identical to the refactoring I did to support the building the Python bindings.

I'm currently testing this PR. So far the following builds work

* CMake 3.5.0, Make on Arch Linux with Mono 4.2.2
* Python build system on Arch Linux with Python 3.5.1
* CMake 3.5.1, Ninja on Windows 8 with the C# toolchain that ships with Visual Studio 2013

@jirislaby @arrowd I know you maintain packages for Z3 so feel free to give this PR a test. If you build the .NET bindings you will need to add Mono as a build dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,5 @@ src/CMakeLists.txt
 src/*/CMakeLists.txt
 src/*/*/CMakeLists.txt
 src/*/*/*/CMakeLists.txt
+src/api/dotnet/cmake_install_gac.cmake.in
+src/api/dotnet/cmake_uninstall_gac.cmake.in
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,6 +328,25 @@ message(STATUS "CMAKE_INSTALL_BINDIR: \"${CMAKE_INSTALL_BINDIR}\"")
 message(STATUS "CMAKE_INSTALL_INCLUDEDIR: \"${CMAKE_INSTALL_INCLUDEDIR}\"")
 
 ################################################################################
+# Uninstall rule
+################################################################################
+configure_file(
+  "${CMAKE_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
+  "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
+  @ONLY
+)
+
+# Target needs to be declared before the components so that they can add
+# dependencies to this target so they can run their own custom uninstall rules.
+add_custom_target(uninstall
+  COMMAND
+  "${CMAKE_COMMAND}" -P "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
+  COMMENT "Uninstalling..."
+  ${ADD_CUSTOM_COMMAND_USES_TERMINAL_ARG}
+  VERBATIM
+)
+
+################################################################################
 # CMake build file locations
 ################################################################################
 # To mimic the python build system output these into the root of the build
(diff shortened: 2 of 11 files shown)


delcypher:
Building on 

* CMake 2.8.12.2, Make on Ubuntu 14.04LTS with Mono 3.2.8
* Python build system on Ubuntu 14.04LTS with Mono 3.2.8 and Python 2.7.6
* CMake 3.5.1, Visual Studio 2013 with Windows 8 (With this configuration ``DOTNET_CSC_EXECUTABLE`` and ``DOTNET_GACUTIL_EXECUTABLE`` may have to be set manually)

succeeds.

wintersteiger:
Thanks for adding that. Did you also check whether the non-cmake build still works as intended?

delcypher:
@wintersteiger I've just tested the non-cmake build with the dotnet bindings on Linux and on Windows and it seems to work fine. My check for this is

1. Compile Z3 and the dotnet bindings
2. Build the .NET example
3. Run the .NET example

Something I probably should mention for anyone looking at this PR is that for Windows there is an important difference between how the .NET bindings are built under the non-cmake (Python build system) build system and the CMake build system. Under the CMake build the ``/linkresource:`` flag is **not** passed to ``csc.exe`` (unlike the non-cmake build system) which means ``libz3.dll`` must be copied over to directory of the ``.NET`` application that is using the Z3 ``.NET`` bindings manually. There are technical reasons for this which are explained in the comments and in the commit message of commit 20d3bf4.

wintersteiger:
The /linkresource fine for now, but I can't imagine this not being fixable in the cmake build. Might take some additional effort of course.

delcypher:
@wintersteiger Thanks for merging. The cause of the problem is a combination of two issues

* ``csc.exe`` behaves strangely (from my perspective). It seems that the argument given to ``/linkresource:`` needs to refer to a file in the same directory that will contain ``Microsoft.Z3.dll``. If I give ``/linkresource:`` an absolute path to ``libz3.dll`` (in a different directory than ``Microsoft.Z3.dll`` will reside) then the build of ``Microsoft.Z3.dll`` succeeds but when I try to build the example .NET application  and add a reference to ``Microsoft.Z3.dll`` then Visual Studio complains that it can't find ``libz3.dll`` in the directory containing ``Microsoft.Z3.dll`` (i.e. despite giving an absolute path to the ``/linkresource`` flag that path is ignored and it only seems to use the file name).
* Multi-configuration CMake generators (i.e. Visual Studio and XCode). With these generators it currently doesn't seem possible to have custom rules (``add_custom_command()``) output files into the build directory of another target without very horrible hacks. For single configuration generators (e.g. Makefiles, NMake files, Ninja...) it is trivial to do this.

When experimenting I found for single configuration CMake generators I could use the ``/linkresource`` flag and have everything work correctly but the build would fail for multi-configuration CMake generators (i.e. Visual Studio). I could have just have ``/linkresource`` passed only for single configuration generators but I thought that would be annoyingly inconsistent. So I decided to not pass the ``/linkresource`` flag in any type of build.

We might be able to avoid these issues if the Z3 .NET bindings were simply installed on Windows (rather than trying to use them from the build directory) as they are under all other distributions. Is there a reason we don't install the .NET bindings or is this not best practice under Windows?

wintersteiger:
Right, that's all good reasons. It's quite possible that /linkresource behaves differently in/via mono, I never looked into that. 

We don't publish installers for Windows anymore, but only .zip files for each release which contain everything in the default folders and it's just generally likely that those files will always be in the same directory anyways (at least on Windows). Actual "installation" in the case of .NET libraries may also mean that the libraries are added to the GAC, which, in the past, very few people needed, but for those few it was a bit of a headache.

delcypher:
@wintersteiger 
> Right, that's all good reasons. It's quite possible that /linkresource behaves differently in/via mono, I never looked into that.

Under mono the both build systems do not pass the ``/linkresource`` flag because Mono will look for ``libz3.so`` in the standard search paths when using the .NET bindings. If the ``libz3.so`` is installed the library is found and if a user needs to use the library in the build directory ``LD_LIBRARY_PATH`` can be set appropriately.

> We don't publish installers for Windows anymore, but only .zip files for each release which contain everything in the default folders and it's just generally likely that those files will always be in the same directory anyways (at least on Windows). Actual "installation" in the case of .NET libraries may also mean that the libraries are added to the GAC, which, in the past, very few people needed, but for those few it was a bit of a headache.

It looks like under Windows the ``gacutil`` doesn't have the ``-root`` flag like mono does so there isn't a way to install outside of the system's GAC which would make distributing the tarball of ``make install DESTDIR=/temp/path/to/tarball`` containing a partial GAC impossible. Under Windows it might just make more sense to copy ``Microsoft.Z3.dll`` into ``<prefix>/lib`` instead.

This isn't really an issue yet as you're not using CMake to create release tarballs so I'm inclined to leave this for now until it actually becomes a problem.

PR #540, opened 2016-03-28
Include *.cmx files during installation of OCaml bindings.

martin-neuhaeusser:
The *.cmx files are now installed using ocamlfind. They contain information from the
compiler that can be used during optimization (the upcoming OCaml 4.03.0 issues
warning 58 if those files are missing from a package).


wintersteiger:
Sounds good, thanks for the fix!

PR #545, opened 2016-03-29
Lackr

MikolasJanota:
Silently bailing out on quantifiers in lackr plus some minor improvements in the code.


wintersteiger:
Awesome, thanks for the quick fix!

PR #553, opened 2016-04-05
Trailing

MikolasJanota:
Adds trailing 0 analysis for things bv equalities, e.g. 2x = 6y. mk_extract_proc had to be refactored from bv_rewriter into a separate file. rewriter.bv_trailing currently set to false by default.


delcypher:
@MikolasJanota If you're adding new ``.cpp`` files could you please add them to the relevant ``CMakeLists.txt`` files so that you don't break the CMake build.

For this PR you need to add ``bv_trailing.cpp`` and ``mk_extract_proc.cpp`` to ``contrib/cmake/src/ast/rewriter/CMakeLists.txt``

PR #568, opened 2016-04-17
Teach CMake to build and install the Java bindings

delcypher:
Please note I have not thoroughly tested this PR yet so it probably shouldn't be merged until I've finished doing so but I'd like those interested to give it a try early on so we can weed out any issues. So far it seems to be working for me.

@kfriedberger I think you'll be interested in these changes. Note that in the build directory with the CMake build ``libz3java.so`` has the ``rpath`` set automatically so if ``libz3java.so`` is in your library search path ``libz3.so`` will be found automatically so there is no need for #537 if you just use the CMake build system once this PR is merged.

@jirislaby @arrowd I know you maintain packages for Z3 so feel free to give this PR a test. I'm not sure if I've exposed enough options to give the control packagers need.

### DIFF
--- a/README-CMake.md
+++ b/README-CMake.md
@@ -279,6 +279,10 @@ The following useful options can be passed to CMake whilst configuring.
 * ``INSTALL_DOTNET_BINDINGS`` - BOOL. If set to ``TRUE`` and ``BUILD_DOTNET_BINDINGS`` is ``TRUE`` then running the ``install`` target will install Z3's .NET bindings.
 * ``DOTNET_CSC_EXECUTABLE`` - STRING. The path to the C# compiler to use. Only relevant if ``BUILD_DOTNET_BINDINGS`` is set to ``TRUE``.
 * ``DOTNET_GACUTIL_EXECUTABLE`` - STRING. The path to the gacutil program to use. Only relevant if ``BUILD_DOTNET_BINDINGS`` is set to ``TRUE``.
+* ``BUILD_JAVA_BINDINGS`` - BOOL. If set to ``TRUE`` then Z3's Java bindings will be built.
+* ``INSTALL_JAVA_BINDINGS`` - BOOL. If set to ``TRUE`` and ``BUILD_JAVA_BINDINGS`` is ``TRUE`` then running the ``install`` target will install Z3's Java bindings.
+* ``Z3_JAVA_JAR_INSTALLDIR`` - STRING. The path to directory to install the Z3 Java ``.jar`` file. This path should be relative to ``CMAKE_INSTALL_PREFIX``.
+* ``Z3_JAVA_JNI_LIB_INSTALLDIRR`` - STRING. The path to directory to install the Z3 Java JNI bridge library. This path should be relative to ``CMAKE_INSTALL_PREFIX``.
 
 On the command line these can be passed to ``cmake`` using the ``-D`` option. In ``ccmake`` and ``cmake-gui`` these can be set in the user interface.
 
--- a/README-CMake.md
+++ b/README-CMake.md
@@ -308,6 +308,10 @@ correct location. For example
 ```
 JAVA_HOME=/usr/lib/jvm/default cmake -DBUILD_JAVA_BINDINGS=ON ../
 ```
+Note that the built ``.jar`` file is named ``com.microsoft.z3-VERSION.jar``
+where ``VERSION`` is the Z3 version. Under non Windows systems a
+symbolic link named ``com.microsoft.z3.jar`` is provided. This symbolic
+link is not created when building under Windows.
 
 ## Developer/packager notes
 
--- a/README-CMake.md
+++ b/README-CMake.md
@@ -290,6 +290,23 @@ Example
 
 ```
 cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_TRACING=FALSE ../
+
+```
+
+## Z3 API Bindings
+
+Z3 exposes various language bindings for its API. Below are some notes on building
+and/or installing these bindings when building Z3 with CMake.
+
+### Java bindings
+
+The CMake build uses the ``FindJava`` and ``FindJNI`` cmake modules to detect the
+installation of Java. If CMake fails to find your installation of Java set the
+``JAVA_HOME`` environment variable when invoking CMake so that it points at the
+correct location. For example
+
+```
+JAVA_HOME=/usr/lib/jvm/default cmake -DBUILD_JAVA_BINDINGS=ON ../
 ```
 
 ## Developer/packager notes
--- a/contrib/cmake/src/CMakeLists.txt
+++ b/contrib/cmake/src/CMakeLists.txt
@@ -234,4 +234,16 @@ if (BUILD_DOTNET_BINDINGS)
   add_subdirectory(api/dotnet)
 endif()
 
+################################################################################
+# Java bindings
+################################################################################
+option(BUILD_JAVA_BINDINGS "Build Java bindings for Z3" OFF)
+if (BUILD_JAVA_BINDINGS)
+  if (NOT BUILD_LIBZ3_SHARED)
+    message(FATAL_ERROR "The Java bindings will not work with a static libz3. "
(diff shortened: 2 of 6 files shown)


delcypher:
Pinging @cmur2 @cheshire @seanmcl because it looks like you were interested in Z3's Java API at some point in the past. Feel free to give this a try.

delcypher:
Okay testing so far...

Checking that Z3 and Java bindings build and that Z3's Java example can be built against installed Java bindings and against an installed copy.

* CMake 3.5.1 on Arch Linux with openjdk-7
* CMake 2.8.12.2 on Ubuntu 14.04LTS with openjdk-7

Note that ``CMAKE_INSTALL_PREFIX`` needs to be changed to ``/usr`` for this to work correctly.

cheshire:
Hi @delcypher , I have tried to use this with JavaSMT, but currently the paths seem to be set up in a somewhat funny way. E.g. the absolute path to `libz3.so` is hardcoded inside `libz3java.so`.
This means the generated `libz3java.so` can not be distributed or moved around, which is suboptimal.

Also instead of searching for `libz3.so`, `libz3java.so` searches for `libz3.so.4.4.2.0`. Maybe that's a feature to ensure the right version is always picked up, but if the filename has to change the distribution through Apache Ivy / Maven Central becomes problematic as well, as they rely on the artifact name staying constant.

I'm no expert with build systems though.

delcypher:
Okay the following also work

* Python build system using Python 2.7.11 and Python 3.5.1 on Arch Linux with openjdk-7
* Python build system using Python 2.7.6 on Ubuntu 14.04LTS with openjdk-7

delcypher:
@cheshire 

Thanks for taking a look

> Hi @delcypher , I have tried to use this with JavaSMT, but currently the paths seem to be set up in a somewhat funny way. E.g. the absolute path to libz3.so is hardcoded inside libz3java.so.
This means the generated libz3java.so can not be distributed or moved around, which is suboptimal.

It is meant to be like that. Build directories **are not supposed to be relocatable**. If you intend to use the built Z3 from the build directory then you shouldn't move it.

It sounds like you should be using an installed copied of Z3 which is usually relocatable.

Note the when you install Z3 i.e.

```
make install
```

The rpath on ``libz3java.so`` is automatically removed so the installed copy doesn't reference files in the build directory.

This seems okay to me though because if you've installed Z3 properly then then ``libz3java.so`` and ``libz3.so`` will be in the search path of the system linker so the Java code in ``com.microsoft.z3.jar`` will find the libraries.

Or am I missing something here?

cheshire:
@delcypher I want to get "moveable" shared object, which I can copy from machine to machine with same architecture without any installation, while preserving the filename.
Currently I have that with vanilla Z3 with `libz3.so`.
To get the same effect with both `libz3.so` and `libz3java.so` I have to use our patch (which is in the pull request you have mentioned).

Can I get the same effect with your system?

delcypher:
@cheshire 
>  I want to get "moveable" shared object, which I can copy from machine to machine with same architecture without any installation, while preserving the filename.
Currently I have that with vanilla Z3 with libz3.so.
To get the same effect with both libz3.so and libz3java.so I have to use our patch (which is in the pull request you have mentioned).

When I say "install" that doesn't necessary have to be in a normal system location. You can set ``CMAKE_INSTALL_PREFIX`` to be anything you like. If you install outside of a normal install prefix (like ``/usr``) you will need to tell your system linker where to find the libraries though.

> Can I get the same effect with your system?

It might be possible to teach the build system to modify the rpath of the installed libraries however I think I need to take a step back here. I don't understand why you are doing this

In order for your java application to work you must have ``libz3java.so`` in the system linker's search path (e.g. by adding it to ``LD_LIBRARY_PATH``). The ``libz3.so`` library is placed in the same directory as ``libz3java.so`` so if the system linker can find ``libz3java.so`` then it can also find ``libz3.so`` as well so modifying ``libz3java.so``'s rpath shouldn't be necessary.

So under what scenario is it possible for your application to load ``libz3java.so`` but then fail to load ``libz3.so``? That sounds like the scenario you want to handle but this scenario doesn't sound like it should ever happen (at least from my perspective anyway).

delcypher:
Okay the following also works:

* Windows 8 with MSVC using CMake with the Ninja generator and Oracle Java SE 7u79

delcypher:
Okay the following also works:

* Windows 8 with MSVC using the Python build system with Oracle Java SE7u79

delcypher:
@wintersteiger @NikolajBjorner Okay I think this PR is ready to go. There are additional issues that still need resolving (like the problems @cheshire is having) but we can fix those in future PRs. I'm pretty confident I haven't broken the build of the Java bindings under the Python build system and under CMake the build seems to work okay under Linux and Windows.

kfriedberger:
@delcypher 
Is it possible to have a "static" linked library of the z3java-bindings?
The expected result would be one (bigger) binary `libz3java.so` without the need of a separate `libz3.so`, because the second file would be "statically" included in `libz3java.so`. 
This would make at least half of my pull request irrelevant, because then the whole story with `rpath`/`$ORIGIN` would be solved.

delcypher:
@kfriedberger 
I think you're abusing the word "static" here as ``libz3java.so`` is a dynamic (shared) library not a static library but I know what you are asking for. You're asking for a single monolithic shared library that effectively contains everything in ``libz3.so`` and ``libz3java.so``.

This would technically be possible to implement in the CMake build system. I'm not sure if it's a good idea though. This assumes you never have symbol conflicts with in ``Native.cpp`` with what's in ``libz3.so`` which I'm not sure if that's guaranteed.

On a hight level I'm also a little uncertain as to why we really need ``Native.cpp`` in the first place. For the ``.NET`` bindings for example this problem doesn't exist because they call directly into ``libz3.so`` without going through an intermediate library.

Anyway I would like to get this PR merged first before tackling the problem you're experiencing because I want to get support building the Java bindings using CMake on-par with the Python based build system first before adding new features.

delcypher:
@wintersteiger @NikolajBjorner 

Is there anything I can do to get this PR merged? I've done all the testing I can and it seems to work fine.

NikolajBjorner:
Very cool thanks. The scripts look fine to me (but I haven't bw to test). CW would merge when there are cycles to follow through on any build issues.

NikolajBjorner:
I very much like the cmake system. As I see, there are a few loose ends, such as this PR. Others would be some of the examples (.NET and as far as I can see in the code also the Java examples).

wintersteiger:
The main issue for me is to check that the old build system isn't broken and that the changes in mk_util.py won't break existing applications because different API code is generated. Has any of this been tested?

delcypher:
@wintersteiger 
> The main issue for me is to check that the old build system isn't broken 

I've tested that on Linux and Windows with both CMake and the old build system.

> won't break existing applications because different API code is generated

I tested that the Java example still built correctly and it does for all scenarios that I tested.

The generated API code is only different for the enums. Their declaration order is now sorted but the actual values have not changed.

delcypher:
> @NikolajBjorner 
> I very much like the cmake system. As I see, there are a few loose ends, such as this PR. Others would be some of the examples (.NET and as far as I can see in the code also the Java examples).

Thanks for your support. For the examples I think it would be better if we provided standalone projects for the examples so users can see how to build against Z3. For example for ``.NET`` we should provide a ``.sln`` to that uses MSBuild which is the standard way to build ``.NET`` projects.

NikolajBjorner:
> For example for  .NET  we should provide a  .sln  
maybe, depends how much midnight oil there is to burn, because there are tons and tons of projects in the wild that integrate mixed mode libraries. So this isn't specific to Z3. 
Before that, there is the objective of checking that the examples build and run as expected as part of regressions, and also let users just build the examples with the (same) simple setup.

wintersteiger:
@delcyper great, thanks for running the tests and confirming, this should be safe to merge then.

I agree, it would be nice to have separate makefiles for the examples instead of them being integrated into the main build system(s), so that users can essentially copy the makefile as a starting point for their own projects (as opposed to decrypting the auto-generated makefiles and cross-checking compiler flags with config.mk and such). Like @NikolajBjorner says though, not a high priority.

PR #576, opened 2016-04-26
[CMake] Fix install of libz3.dll on Windows

delcypher:
[CMake] On Windows fix the ``install`` target so that it installs ``libz3.dll``.

I've left a comment about the installation of ``libz3.lib``. I'm not
sure if we want that installed or not.


wintersteiger:
Yes, on Windows both are needed. Roughly, Visual Studio will (statically) link with libz3.lib which contains the (dynamic) interface to libz3.dll.

PR #583, opened 2016-05-03
Lots of OCaml API bugfixes/improvements courtesy of @martin-neuhaeusser

wintersteiger:


wintersteiger:
@martin-neuhaeusser: Just one detail: I had to change the "inline" annotation on compare_pointers and try_to_delete_context, see here: https://github.com/Z3Prover/z3/pull/583/commits/140f0bb7945ffefb6c01e7139e0fb02f666319e7
LLVM/clang on OSX complained that those functions weren't there anymore at link time. Probably fixable, but for now I just removed the annotations to get it working.

PR #601, opened 2016-05-13
Bugfix: character not in ascii range if output contains german characters

sashs:
Hi,

the mk_util.py script crashs in function is_clang_in_gpp_form because of a character which is not in ascii range.

`gcc (GCC) 6.1.1 20160501
Copyright (C) 2016 Free Software Foundation, Inc.
Dies ist freie Software; die Kopierbedingungen stehen in den Quellen. Es
gibt KEINE Garantie; auch nicht für MARKTGÄNGIGKEIT oder FÜR SPEZIELLE ZWECKE.`

regards
Sascha

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -852,7 +852,7 @@ def is_CXX_gpp():
 
 def is_clang_in_gpp_form(cc):
     version_string = check_output([cc, '--version'])
-    return str(version_string).find('clang') != -1
+    return str(version_string.encode('utf_8')).find('clang') != -1
 
 def is_CXX_clangpp():
     if is_compiler(CXX, 'g++'):


wintersteiger:
`check_output` already tries to decode the string (which I added for exactly the same reason), but it decodes to the stdout encoding (see [here](https://github.com/Z3Prover/z3/blob/master/scripts/mk_util.py#L112)). Could you find out why that doesn't work for you?

sashs:
Hey, 

I don't know if the decode call is needed in other cases, but for me it works without the decode call or with decode and my push request.

This works for me
`x = 'äöü'
str(x)`

and that works for me
`x = 'äöü'.decode('utf-8')
str(x.encode('utf-8))`

If you use python2 and you call decode you get an unicode string and if you have non ascii characters in that string you have to call encode again if you want to have a normal str. 

Regards
Sascha

wintersteiger:
What do you call a "normal" string? Python 3 is unicode-only as far as I know, and Python is  2 sometimes is and sometimes isn't. The complication in this case is that the gcc output contains unicode characters (which only very recent versions of gcc do). The fix for my problem cases was the decoding in `check_output` (and any new fixes should also go there so that it works for all tools). 

On Windows 10, Python 3.5.1 this:

    x = 'äöü'.decode('utf-8'); str(x.encode('utf-8')) 

tells me:

      File "<stdin>", line 1, in <module>
    AttributeError: 'str' object has no attribute 'decode'

And using Python 2.7.11 I get:

      File "<stdin>", line 1, in <module>
      File "c:\Python27\lib\encodings\utf_8.py", line 16, in decode
        return codecs.utf_8_decode(input, errors, True)
    UnicodeDecodeError: 'utf8' codec can't decode byte 0x84 in position 0: invalid start byte

I had to switch keyboard and type the umlauts manually though as my Windows refuses to paste them into the console.

sashs:
Ok, what I wrote only works with python 2. In python 3 you cannot write äöü in a bytes object. 
With "normal" string I meant str. In python 3 str is unicode und bytes is a python2 str. 

In your out object of the Popen call you have a bytes object, therefore you can call decode. If you call decode you get a str/unicode object. Because of that there is a problem with the str call. 
Can you tell me why you need the decode call. Then I will look for a solution which works for your and for my case. 

Look at my github profile for my email address. :)

Regards
Sascha

wintersteiger:
I added the decode call when I had the same problem as you did, i.e., gcc 6.x having non-ascii characters in its version string (a copyright &copy; character if I remember correctly). It worked for me, but possibly only because I had been using a non-unicode shell, perhaps we should just `.decode('ascii')` everything there to make sure it's at least predictable.

sashs:
I think that won't work if you have non ascii character in the string. I will try to reproduce your problem. If I can reproduce it, I will look for a fix for both problem and create another pull request. 

Regards
Sascha

delcypher:
@sashs Just to note if this blocking you, you could also just use the Z3's [CMake build system](https://github.com/Z3Prover/z3/blob/master/README-CMake.md) instead which likely doesn't have this problem. Of course we should try to fix the problem in the Python build system.

wintersteiger:
Closing this issue in anticipation of the new PR.

PR #618, opened 2016-05-24
speeding up mk_array_eq() in model_evaluator

agurfinkel:
Speeding up the common case


NikolajBjorner:
I wonder, is this code ever reached:


  287   +                    if (m().are_distinct (s1, s2)) {  
  288   +                        result = m().mk_false();  
  289   +                        return BR_DONE;  
  290   +                    }  

Since s1, s2 are select expressions, they are not "values" so m().are_distinct() should always evaluate to false. We rather want to access the entries for which the key came from.

agurfinkel:
yes. I wondered that too :)

NikolajBjorner:
This is maybe your subtle passive-aggressive way to have me take care of it regardless.

PR #622, opened 2016-05-26
Export default tactic for use via the SMT-LIB 2 interface.

dstaple:
This suggestion is in response to [this Stack Overflow discussion](https://stackoverflow.com/questions/37443308/check-sat-using-default-or-similar).  The purpose of this change is to permit combining the Z3 default tactic with other tactics in a custom strategy.  For example `(check-sat-using (or-else default qfnra-nlsat))`.


NikolajBjorner:
Thanks

PR #627, opened 2016-05-30
Consistent naming of methods in Java binding with lowercase "mk" prefix

cmur2:
Could you consider using consistent method names in the Java binding class Context? Not necessarily by merging this PR with my specific changes but in general?


wintersteiger:
Yes, the lower-case form was always the intention, but apparently some upper-case ones sneaked in. Thanks for fixing that right away!

PR #630, opened 2016-06-02
Add init file to Python bindings

hph86:
I wondered why the Python bindings come without an `__init__.py` file, so I just created one. I think this would help package maintainers to provide z3 Python packages which work out of the box. For example, in Fedora 23 the lack of an init file (or the presence of an empty init file created by the maintainer) renders the module unusable (unless your CWD is `/usr/lib64/python2.7/site-packages/z3`):

```
$ python -c 'import z3; print(z3.get_version_string())'
  File "<string>", line 1, in <module>
ImportError: No module named z3
```

Merging this PR should provide the expected functionality (once it arrives in the distros):

```
$ python -c 'import z3; print(z3.get_version_string())'
4.4.1
```

What do you think?

### DIFF
--- a/contrib/cmake/src/api/python/CMakeLists.txt
+++ b/contrib/cmake/src/api/python/CMakeLists.txt
@@ -4,6 +4,7 @@ message(STATUS "Emitting rules to build Z3 python bindings")
 ###############################################################################
 # This allows the python bindings to be used directly from the build directory
 set(z3py_files
+  __init__.py
   z3.py
   z3num.py
   z3poly.py
--- a/src/api/python/__init__.py
+++ b/src/api/python/__init__.py
@@ -1,0 +1,1 @@
+from z3 import *


delcypher:
@hph86 Unfortunately I don't think this correct because currently the Python files don't get put into a ``z3`` directory containing the ``*.py`` files. I don't know why Z3's python bindings are copied into the build directory or installed this way. When implementing the CMake build system I copied the behaviour of the other build system.

wintersteiger:
The reason they are in the *.py files are build directory is simply that everything else also went into the build directory and it makes it easier to find the .dll/.so/.dylib etc. 

I don't understand the reasoning of the request: This here:

    python -c "import z3; print(z3.get_version_string())"

works just fine for me, as long as `PYTHONPATH` is set up correctly, and libz3.dll is in the `PATH` (on Windows).

tyll:
> as long as PYTHONPATH is set up correctly, and libz3.dll is in the PATH (on Windows).

I do not know how the PYTHONPATH is managed on Windows, but in Linux packages are installed into directories below paths like ```/usr/lib64/python2.7/site-packages```. PYTHONPATH then contains this directory. To make it possible to import all the packages inside this directory, they need to provide a ```__init__.py``` file as @hph86 provided (compare https://docs.python.org/3/tutorial/modules.html#packages ). Setting the PYTHONPATH to each individual python package is then not reqired. However even with the __init__.py file, setting the PYTHONPATH to the directory containing the binary module would still work.

wintersteiger:
As I understand it, that file should be added to packages, e.g., at `make install` time or later. However, the recommendation is to keep it empty if no initialization code is required (see [tyll's link](https://docs.python.org/3/tutorial/modules.html#packages)). Adding `from z3 import *` would always, unconditionally import the module, even if it's not required/imported by the user.

tyll:
It is normal for the file to exist in python packages/modules before they are installed. Adding the ```from z3 import *``` is required here to make ```import z3``` work on regular Linux setups. If it is empty. users would need to type ```from z3.z3 import *``` because the libz3.so file is installed in ```/usr/lib64/python2.7/site-packages/z3```.

wintersteiger:
What exactly is the difference between `import z3` and `from z3 import *`? So far I always assumed those are exactly the same command.

cheshire:
@wintersteiger `from z3 import *` puts everything into the global namespace, while `import z3` puts it under `z3` namespace.

wintersteiger:
Ooh, I see. 

So, on OSX, for me all the following options behave exactly the same for me after running `make install`: no __init__.py, empty __init__.py, `import z3`, `from z3 import *`. Which is the right right thing to do and why?

agurfinkel:
Python has two concepts of packages: file-as-a-package (FAAP) and directory-as-a-package (DAAP). FAAP is used for one-file throw-away things, DAAP is preferred for packages that contain many internal files (like z3 bindings).

You know how FAAP works: put everything you need in `z3.py` and use `import z3` in your scripts.

For DAAP, place all python code is a directory `z3`, and create a file `__init__.py` in that directory. Then, `import z3` will import `__init__.py` from `z3` directory. 

When  converting from FAAP to DAAP and want to preserve existing behavior, you have two choices:

1. move the content of `z3.py` into `z3/__init__.py`
2. copy z3.py into directory `z3`. In `__init__.py` add `from z3 import *`. In this case, the statement `import z3` in your scripts imports `__init__.py` and the statement `from z3...` in `__init__.py` imports `z3.py` at the top level.

Whether you want to convert the bindings library into DAAP is your choice. It would be cleaner, but might be disruptive to current users. I've done it in the past for personal use, but have never tested it heavily.

Hope this helps,
arie

wintersteiger:
OK, so I think I figured it out: By default, the Z3 Python bindings, when running `make install` in the build directory, will install everything into (site-/dist-packages)/ while apparently some other packages are installed into (site-/dist-packages)/z3/ and then we want those things pulled out into the global namespace. Is this correct?

wintersteiger:
Thanks for the explanation @agurfinkel, that makes a lot of sense! So, if I'm not mistaken, whatever comes with Z3 at this time is FAAP, but some users want DAAP. @hph86: Is that correct and do we know of any packages that are being distributed as DAAP at this time?

hph86:
@wintersteiger: Sorry for my absence, but yes, I think you summed it up correctly. Currently, I know of [Debian (teseting/stretch)](https://packages.debian.org/stretch/amd64/python-z3/filelist) and [Fedora 23](http://koji.fedoraproject.org/koji/rpminfo?rpmID=7221522) distribute DAAP packages. While the Fedora maintainer at least provided an empty provided an empty `__init.py__` file (`import z3.z3` within a script should work), the Debian package lacks that file completely (I guess the `z3` folder is not recognized as an importable module at all).

wintersteiger:
Right, so I suppose the first action to take is to let those package maintainers know that they should update their packages. I feel we shouldn't prescribe them where/how to publish them, but they should of course work correctly. I know the maintainer of the Debian package, but does anybody know who's responsible for the Fedora package?

hph86:
@wintersteiger: I have just created a [bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1351580) for Fedora to inform the maintainer.

wintersteiger:
That's great, thanks for that @hph86. I've written to the Debian maintainer about a week ago, but haven't received a reply yet.

jamesjer:
I am the Fedora package maintainer.  I have made the suggested change and started new builds going.  They will have to go through testing first, so it will be a little over a week before they are shipped as updates.

tyll:
Since it is agreed to move forward with this patch, can you please merge this PR? In Fedora we like to avoid carrying patches that are not merged upstream. And after you merged this PR you can also add a notice to the next release announcement to make all oher distro maintainers aware of the change.

wintersteiger:
@tyll I don't think this patch really belongs here, because it's ultimately up to the package providers to decide in which way they want to provide their package. So, once these issues are resolved I'm rather planning to close this issue instead of merging it. Are there strong opinions against this from anybody?

cheshire:
@wintersteiger I would say it's more idiomatic to have the `__init__` file in the python package consisting of multiple files.
There is a choice to be made on how the python package should be structured, and to me the proposed solution looks more optimal and more idiomatic than the current one.

hph86:
I second the opinion of @cheshire. @wintersteiger: while your motivation to provide freedom to the maintainers is indeed noble, I do not think they have a real choice because currently the package consists of multiple files (DAAP-like). So in order for them to provide the code as FAAP, they would need to merge everything into a `z3.py` file themselves (as pointed out by @agurfinkel). Even if maintainers were doing this already, I cannot see why supplying a meaningful (with respect to the DAAP-like structure) `__init__.py` file would limit them in their course of providing the code as FAAP.

wintersteiger:
@cheshire, @hph86: Apart from the changes in the currently proposed PR, will we also have to move the default installation path for the Python API into a new `z3` directory?

wintersteiger:
Superseded by recent changes to the Python bindings (#735 etc), so I'll close this PR here.

PR #645, opened 2016-06-10
Extend build scripts to support MinGW64 cross-compilation on Windows.

martin-neuhaeusser:
This patch makes the configuration scripts detect builds of Z3 with the MinGW64 cross-compiler toolchain under Cygwin and adds the (few) additional parameters that are required to build the DLL and the OCaml bindings for native Windows 32 and 64 bits.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,7 +304,10 @@ endif()
 # library. If not building a shared library ``-fPIC`` isn't needed and would add
 # unnecessary overhead.
 if (BUILD_LIBZ3_SHARED)
-  if (NOT MSVC)
+  # Avoid adding -fPIC compiler switch if we compile with MSVC (which does not
+  # support the flag) or if we target Windows, which generally does not use
+  # position independent code for native code shared libraries (DLLs).
+  if (NOT (MSVC OR MINGW OR WIN32))
     z3_add_cxx_flag("-fPIC" REQUIRED)
   endif()
 endif()
@@ -387,4 +390,3 @@ option(ENABLE_EXAMPLE_TARGETS "Build Z3 api examples" ON)
 if (ENABLE_EXAMPLE_TARGETS)
   add_subdirectory(examples)
 endif()
-
--- a/README.md
+++ b/README.md
@@ -56,6 +56,16 @@ CXX=clang++ CC=clang python scripts/mk_make.py
 
 Note that Clang < 3.7 does not support OpenMP.
 
+You can also build Z3 for Windows using Cygwin and the Mingw-w64 cross-compiler.
+To configure that case correctly, make sure to use Cygwin's own python and not
+some Windows installation of Python.
+
+For a 64 bit build (from Cygwin64), configure Z3's sources with
+```bash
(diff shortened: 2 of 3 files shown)


NikolajBjorner:
Do you know if the cmake based build works with mingw64 under Cygwin already? 
To put it in a different way, would you object if I some day propose to sunset the original build system in favor of the cmake system?

delcypher:
@NikolajBjorner I have not tried mingw64 under Cygwin but in principle CMake supports it. Our CMake scripts may need to be tweaked a little to work correctly though.

The OCaml bindings aren't supported under the CMake build system yet though (I haven't had time to add support).

martin-neuhaeusser:
I tried to build Z3 under mingw64 yesterday, and it almost worked ;-)
Some minor issues are a warning about `-fPIC` being specified although mingw64 generates position independent code, anyways. Another issue that makes the build currently fail is that it does not compile the "collect_statistics_tactic". 

I will try to fix this today in another PR.

That being said, some projects here rely heavily on Z3's OCaml bindings. I can offer to integrate them into the CMake build system; however, this will take some time, as I am quite busy at the moment.

delcypher:
@martin-neuhaeusser The ``collect_statistics_tactic`` issue was already fixed in master (16ad33bf399d1b9aa1e333bbb89b560ffc868632) you should rebase and drop the commit you made that adds it.

wintersteiger:
That would be fantastic if you could integrate those too. 

Regarding collect_statistics: initially I had used C++11-only code in there and some compilers still don't support that by default so I removed it. Could you check whether it still has any problems with that?

CMakeLists.txt
```diff
@@ -304,7 +304,7 @@ endif()
 # library. If not building a shared library ``-fPIC`` isn't needed and would add
 # unnecessary overhead.
 if (BUILD_LIBZ3_SHARED)
-  if (NOT MSVC)
+  if (NOT (MSVC OR MINGW))
```
delcypher: @martin-neuhaeusser This LGTM. However please add a small comment here that says something similar to ``mingw64`` always generates position independent code so passing ``-fPIC`` is not necessary. I know this message currently is the commit message but this will get lost if someone touches the code here and it may not be obvious to a reader why MinGW doesn't need ``-fPIC``.

delcypher:
@wintersteiger 
> Regarding collect_statistics: initially I had used C++11-only code in there and some compilers still don't support that by default so I removed it. Could you check whether it still has any problems with that?

What exactly are asking to be be checked? You said you removed the "C++11 only code" so what is there to check.

Side note I think C++11 should be enforced I don't think you should let old compilers hold back Z3 development. It's 2016 it's time for people to upgrade.

wintersteiger:
@delcypher: Does it compile under mingw64?

We support C++11 if it's needed and the oldest compiler we support is VS2012, which supports almost everything. So far I have not seen a use-case or any request for those features to be enabled, either by default or for an exceptional piece of code. Obviously I was the first one to ever try that (and it works everywhere with all compilers by default, without any flags, except for gcc). So while nobody needs it we might just as well not enforce it to be enabled...

NikolajBjorner:
time to upgrade, https://www.youtube.com/watch?v=HeJnA8ItA6s

wintersteiger:
Thanks for the patch, I'm assuming all compilation issues are resolved at this point.

PR #646, opened 2016-06-10
Make C-layer of OCaml bindings C89 compatible.

martin-neuhaeusser:
This patch ensures that the C code generated for the OCaml stubs complies with C89. It is needed to compile Z3 with OCaml support with Visual Studio versions older than VS2013.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1324,36 +1324,80 @@ def mk_z3native_stubs_c(ml_dir): # C interface
         if len(ap) > 0:
             ml_wrapper.write('  unsigned _i;\n')
 
-        # declare locals, preprocess arrays, strings, in/out arguments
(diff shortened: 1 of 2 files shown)


wintersteiger:
I thought you had already signed a CLA, but it seems our CLA bot doesn't know about that. In case you really didn't sign it yet, please take a look at the CLA section in [Contribution Guidlines](https://github.com/Z3Prover/z3/wiki/Contribution%20Guidelines). Basically, we need to make sure that your employer is OK with you spending time on Z3 contributions, to be on the safe side.

wintersteiger:
Great, thanks for getting the CLA out of the way!

martin-neuhaeusser:
Yes, actually I was surprised that legal did not see any major obstacles ;-)

PR #675, opened 2016-07-09
Fix spelling errors

FabianWolff:


wintersteiger:
Thanks for those fixes! Although our CLA bot thinks we need a CLA for that, I think this is fine without, so I'm merging it right away.

PR #710, opened 2016-08-10
Fix off-by-one bug in computing the maximal and minimal exponents in …

### LINKED ISSUE #699: Wrong bias returned for denormal floating point numbers
The attached [example](https://github.com/Z3Prover/z3/files/388462/fp_bias.zip) constructs the largest denormal 32 bit floating point number and checks the returned exponent with Z3's API functions, e.g. with Z3_fpa_get_numeral_exponent_string. 

As a 32bit FP has a bias of 127, we'd expect to receive -127; however, Z3 returns -126. If we use an exponent of 2 (as a counter-check), we correctly receive -125.

As far as we have looked into the code, there seems to be a special case in [api_fpa.cpp](https://github.com/Z3Prover/z3/blob/b7de813c6386a5625c5570c99d8daf0b00e245f8/src/api/api_fpa.cpp#L1025), where mk_min_exp is used to compute the exponent for denormal numbers. However, [that computation](https://github.com/Z3Prover/z3/blob/b7de813c6386a5625c5570c99d8daf0b00e245f8/src/util/mpf.cpp#L1742-L1746) seems to correctly compute the result first, and then adds another 1; at least, we cannot see the reason for doing that last addition.

martin-neuhaeusser:
…FPA.

The range of admissible exponent values in FPA was too large by one both in the mpf utils and in the fpa2bv converter. This patch fixes #699 and makes mk_max_exp return the correct bound of 2^(ebits-1)-1 and adds some comments to mk_min_exp to explain why it returns -2^(ebits-1)+1.

### DIFF
--- a/src/ast/fpa/fpa2bv_converter.cpp
+++ b/src/ast/fpa/fpa2bv_converter.cpp
@@ -264,7 +264,7 @@ void fpa2bv_converter::mk_function(func_decl * f, unsigned num, expr * const * a
     rng = f->get_range();
     fapp = m.mk_app(f, num, args);
     if (m_util.is_float(rng)) {
-        sort_ref bv_rng(m);        
+        sort_ref bv_rng(m);
         expr_ref new_eq(m);
         unsigned ebits = m_util.get_ebits(rng);
         unsigned sbits = m_util.get_sbits(rng);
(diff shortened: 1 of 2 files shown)


wintersteiger:
Thanks for the PR, but this will take some time to analyze and test. The implementations of those functions and conversions have been tested very thoroughly (automated, random) and they are re-used in many places. I doubt that the off-by-one actually stems from there, but that the problem is elsewhere, e.g., in the API functions when converting internal results back to external ones (as in #699).

wintersteiger:
(Just for context: mpf's don't use biassed exponents, but the fpa2bv conversion does, and "top exp" is not the same as "max exp").

martin-neuhaeusser:
Hi Christoph,

with your comments in #699 and another look into the implementation of power2, I agree. The problem does not stem from those functions and there is no off-by-one bug. Sorry!

wintersteiger:
No need to be sorry! Where floats are concerned it's pretty easy to get confused :)

Issue #712, opened 2016-08-13
Buffer overflow in sat/sat_clause.cpp


izbyshev:
Debug build of Z3 (e8141aaa84bb47a927282ce0318686fd12c7b2f9) reports the following assertion failure when given [query.txt](https://github.com/Z3Prover/z3/files/416725/query.txt) as input.
```
ASSERTION VIOLATION
File: ../src/sat/sat_clause.cpp
Line: 149
m_num_segments <= c_max_segments
```
The stack trace follows.
```
#0  0x00007f0299cccada in waitpid () from /lib64/libc.so.6
#1  0x00007f0299c548db in do_system () from /lib64/libc.so.6
#2  0x0000000001c9b60a in invoke_gdb () at ../src/util/debug.cpp:100
#3  0x0000000001b41698 in sat::clause_allocator::get_segment (this=0x7fff9528e1c0, ptr=106583908417536)
    at ../src/sat/sat_clause.cpp:149
#4  0x0000000001b417ca in sat::clause_allocator::get_offset (this=0x7fff9528e1c0, ptr=0x60f00000b268)
    at ../src/sat/sat_clause.cpp:159
#5  0x0000000001b02525 in sat::solver::attach_nary_clause (this=0x7fff9528e0a8, c=..., reinit=@0x7fff9528d520: false)
    at ../src/sat/sat_solver.cpp:300
#6  0x0000000001b0223b in sat::solver::mk_nary_clause (this=0x7fff9528e0a8, num_lits=34, lits=0x61e000002490, learned=false)
    at ../src/sat/sat_solver.cpp:286
#7  0x0000000001b01423 in sat::solver::mk_clause_core (this=0x7fff9528e0a8, num_lits=34, lits=0x61e000002490, learned=false)
    at ../src/sat/sat_solver.cpp:206
#8  0x0000000001b00b9e in sat::solver::mk_clause (this=0x7fff9528e0a8, num_lits=34, lits=0x61e000002490)
    at ../src/sat/sat_solver.cpp:156
#9  0x00000000015ee9d3 in goal2sat::imp::mk_clause (this=0x7fff9528db80, num=34, lits=0x61e000002490)
    at ../src/sat/tactic/goal2sat.cpp:102
#10 0x00000000015eff93 in goal2sat::imp::convert_or (this=0x7fff9528db80, t=0x61300002da08, root=false, sign=true)
    at ../src/sat/tactic/goal2sat.cpp:229
[...]

PR #715, opened 2016-08-19
Make clause offset conversion safe for debug mode AMD64.

NikolajBjorner:
The following is a proposed change to address problems running with 64bit systems whose allocators use random high-order bits of the virtual address space. The compression of 64 bit address spaces into 32 bits (available in clause_offset type), relies on an assumption that addresses are allocated within a limited range of high-order bits. Address sanitizer systems will not follow this pattern as reported in #436. Furthermore, issue #712, reports the same problem, also under debug mode. 
On non-debug mode, the small object allocator uses chunks, so more clauses are required to fall back to an allocator that potentially uses different high-order virtual address ranges. In non-debug mode, the code now throws instead of falling back to hash-tables.

### DIFF
--- a/src/sat/sat_clause.cpp
+++ b/src/sat/sat_clause.cpp
@@ -131,10 +131,12 @@ namespace sat {
 
     clause * clause_allocator::get_clause(clause_offset cls_off) const {
 #if defined(_AMD64_) 
+#if defined (Z3DEBUG)
         clause const* result;
         if (m_overflow_valid && m_cls_offset2ptr.find(cls_off, result)) {
             return const_cast<clause*>(result);
         }
+#endif
         return reinterpret_cast<clause *>(m_segments[cls_off & c_aligment_mask] + (static_cast<size_t>(cls_off) & ~c_aligment_mask));
 #else
         return reinterpret_cast<clause *>(cls_off);
@@ -152,7 +154,8 @@ namespace sat {
             if (m_segments[i] == ptr)
                 return i;
         i = m_num_segments;
-        SASSERT(m_num_segments < c_max_segments);
+#if defined(Z3DEBUG)
+        SASSERT(i < c_max_segments);
         if (i + 1 == c_max_segments) {
             m_overflow_valid = true;
             i += c_max_segments * m_cls_offset2ptr.size();
@@ -160,9 +163,17 @@ namespace sat {
             m_cls_offset2ptr.insert(i, cls);
         }
         else {
-            m_num_segments++;
+            ++m_num_segments;
             m_segments[i] = ptr;
         }
+#else
+        SASSERT(i <= c_max_segments);
+        if (i == c_max_segments) {
+            throw default_exception("segment out of range");
+        }
+        m_segments[i] = ptr;
+        ++m_num_segments;
+#endif
         return i;
     }
 #endif
@@ -170,12 +181,12 @@ namespace sat {
     clause_offset clause_allocator::get_offset(clause const * ptr) const {
 #if defined(_AMD64_) 
         unsigned segment = const_cast<clause_allocator*>(this)->get_segment(ptr);
+#if defined(Z3DEBUG)
         if (segment >= c_max_segments) {
             return m_ptr2cls_offset.find(reinterpret_cast<size_t>(ptr));
         }
-        else {
-            return static_cast<unsigned>(reinterpret_cast<size_t>(ptr)) + segment;
-        }
+#endif
+        return static_cast<unsigned>(reinterpret_cast<size_t>(ptr)) + segment;
 #else
         return reinterpret_cast<size_t>(ptr);
 #endif
--- a/src/sat/sat_clause.cpp
+++ b/src/sat/sat_clause.cpp
@@ -123,21 +123,28 @@ namespace sat {
 
     clause_allocator::clause_allocator():
         m_allocator("clause-allocator") {
-#ifdef _AMD64_
+#if defined(_AMD64_) 
         m_num_segments = 0;
+        m_overflow_valid = false;
 #endif
     }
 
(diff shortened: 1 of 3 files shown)


src/sat/sat_clause.cpp
```diff
@@ -123,48 +123,80 @@ namespace sat {
...
     }
 #endif
 
     clause_offset clause_allocator::get_offset(clause const * ptr) const {
-#ifdef _AMD64_
-        return static_cast<unsigned>(reinterpret_cast<size_t>(ptr)) + const_cast<clause_allocator*>(this)->get_segment(reinterpret_cast<size_t>(ptr));
+#if defined(_AMD64_) 
+        unsigned segment = const_cast<clause_allocator*>(this)->get_segment(ptr);
+#if defined(Z3DEBUG)
+        if (segment >= c_max_segments) {
```
izbyshev: The first map-based "segment" is `c_max_segments - 1`, so shouldn't that be `segment >= c_max_segments - 1`?

src/sat/sat_clause.cpp
```diff
@@ -123,48 +123,80 @@ namespace sat {
...
+            return m_ptr2cls_offset.find(reinterpret_cast<size_t>(ptr));
```
izbyshev: IIUC, this lookup returns `segment`, so why do it?

src/sat/sat_clause.cpp
```diff
@@ -123,48 +123,80 @@ namespace sat {
...
         i = m_num_segments;
-        m_num_segments++;
-        SASSERT(m_num_segments <= c_max_segments);
-        if (i >= c_max_segments)
+#if defined(Z3DEBUG)
+        SASSERT(i < c_max_segments);
+        if (i + 1 == c_max_segments) {
+            m_overflow_valid = true;
+            i += c_max_segments * m_cls_offset2ptr.size();
+            m_ptr2cls_offset.insert(ptr, i);
```
izbyshev: `ptr` is the higher part of `cls` at this point, but we lookup by full pointers, so you probably meant `m_ptr2cls_offset.insert(reinterpret_cast<size_t>(cls), i);`.

izbyshev: Also, what is the reason of having `m_ptr2cls_offset` at all if, after we run out of normal segments, each time `clause_allocator::get_segment()` is called a new index is created at line 163 and the only lookup in `clause_allocator::get_offset()` seems to be unnecessary?

src/sat/sat_clause.cpp
```diff
@@ -123,48 +123,80 @@ namespace sat {
...
 #else
         return reinterpret_cast<clause *>(cls_off);
 #endif
     }
 
-#ifdef _AMD64_
-    unsigned clause_allocator::get_segment(size_t ptr) {
+#if defined(_AMD64_) 
+    unsigned clause_allocator::get_segment(clause const* cls) {
+        size_t ptr = reinterpret_cast<size_t>(cls);
```
veikkoeeva: This looks like taking the size of the pointer. If so, it would feel more correct to have `auto ptr = reinterpret_cast<std::uintptr_t>(cls);`, I believe. Some more information at http://stackoverflow.com/questions/153065/converting-a-pointer-into-an-integer.

src/sat/sat_clause.cpp
```diff
@@ -123,48 +123,80 @@ namespace sat {
...
-#ifdef _AMD64_
-        return static_cast<unsigned>(reinterpret_cast<size_t>(ptr)) + const_cast<clause_allocator*>(this)->get_segment(reinterpret_cast<size_t>(ptr));
+#if defined(_AMD64_) 
+        unsigned segment = const_cast<clause_allocator*>(this)->get_segment(ptr);
+#if defined(Z3DEBUG)
+        if (segment >= c_max_segments) {
+            return m_ptr2cls_offset.find(reinterpret_cast<size_t>(ptr));
+        }
+#endif
+        return static_cast<unsigned>(reinterpret_cast<size_t>(ptr)) + segment;
```
veikkoeeva: I wonder if I interpret this code correctly as adding a segment to a pointer. In that case the information at http://stackoverflow.com/questions/15934111/portable-and-safe-way-to-add-byte-offset-to-any-pointer would be helpful (might apply to other cases too).

izbyshev: @veikkoeeva No, this code is not doing pointer arithmetic. The purpose is to store additional data (`segment`, which is in [0,7]) in the lower 3 bits of the lower half of a 64-bit pointer. The lower 3 bits are normally zero because of `malloc()` alignment requirements on relevant platforms.

src/sat/sat_clause.cpp
izbyshev: `std::uintptr_t` is available starting from C++11 which AFAIK is not required by Z3. It might be possible to use `uintptr_t` from `stdint.h`,  but that wouldn't be standard-complying, and I don't know whether MSVC supports it.

veikkoeeva: Hm, I was over-eager. I'm not a theorem-prover kind of person, but I could imagine coming the right kind of a project, I could use this library. Or just to keep my C++ skills up. I started already https://github.com/Z3Prover/z3/issues/716, and I think I need to amend paying attention to your notes.

izbyshev:
I've checked that if this patchset is changed according to my comments, the debug+asan build passes all my tests. Given NikolajBjorner#1, it is not clear to me how I should proceed. Any feedback on how I can help would be appreciated.

wintersteiger:
I'm closing this pull request as I think I have a better solution in PR #722. In my tests this new solution didn't throw any assertion failures and the address sanitizer seems to be happy too (pending tests with updated clang, my version is a bit out of date).

PR #728, opened 2016-09-05
Fix VisualStudio 2010 compiler warning C4100

AlexVonB:
When compiling with Visual Studio 2010 the buildlog warns of the following:

- `z3++.h: warning C4100: 'e' : unreferenced formal parameter` and
- `z3++.h: warning C4100: 'c' : unreferenced formal parameter`.

This merge request removes the warnings.


wintersteiger:
That's great, thanks for the fix!

PR #730, opened 2016-09-07
Fix omission of Z3_model_has_interp in z3++.h

cttghc:
I ran into a desire for this while using the z3++ API for a project.


NikolajBjorner:
Thanks!

Issue #14, opened 2015-03-29
FP to real and back conversions don't seem to be operational


LeventErkok:
This is on the unstable branch.

I'm experimenting with floating-point to real (`fp.to_real`) and back-conversions (`(_ to_fp eb sb)`). It appears Z3 is always producing `unknown` or going away forever for queries that involve these conversions. Is something wrong with the way I've constructed the benchmarks (logic-selection comes to mind), or is this just not quite working well yet?

Here's an example benchmark that causes Z3 to run forever. (Well, I just killed it after a while)

```
; Automatically generated by SBV. Do not edit.
(set-option :produce-models true)
(set-logic QF_FP)
; --- uninterpreted sorts ---
; --- literal constants ---
(define-fun s_2 () Bool false)
(define-fun s_1 () Bool true)
; --- skolem constants ---
(declare-fun s0 () Real)
(declare-fun s1 () RoundingMode)
; --- constant tables ---
; --- skolemized tables ---
; --- arrays ---
; --- uninterpreted constants ---
; --- user given axioms ---
; --- formula ---
(assert ; no quantifiers
   (let ((s2 ((_ to_fp 8 24) s1 s0)))
   (let ((s3 (fp.to_real s2)))
   (let ((s4 (distinct s0 s3)))
   s4))))
(check-sat)
```

And this one causes Z3 to issue `unknown`:

```
; Automatically generated by SBV. Do not edit.
(set-option :produce-models true)
(set-logic QF_FP)
; --- uninterpreted sorts ---
; --- literal constants ---
(define-fun s_2 () Bool false)
(define-fun s_1 () Bool true)
; --- skolem constants ---
(declare-fun s0 () Real)
(declare-fun s1 () RoundingMode)
(declare-fun s2 () RoundingMode)
; --- constant tables ---
; --- skolemized tables ---
; --- arrays ---
; --- uninterpreted constants ---
; --- user given axioms ---
; --- formula ---
(assert ; no quantifiers
   (let ((s3 ((_ to_fp 8 24) s1 s0)))
   (let ((s4 ((_ to_fp 8 24) s2 s0)))
   (let ((s5 (not (fp.eq s3 s4))))
   s5))))
(check-sat)
```

wintersteiger:
[...]

Issue #13, opened 2015-03-29
fp.isNegative does not handle NaN properly


LeventErkok:
This is on the unstable branch.

For this benchmark:

```
(set-option :produce-models true)
(set-logic QF_FP)
(declare-fun s0 () (_ FloatingPoint 11 53))
(assert
   (let ((s1 (and (fp.isNaN s0) (fp.isNegative s0))))
   s1))
(check-sat)
(get-value (s0))
```

Z3 produces:

```
sat
((s0 (_ NaN 11 53)))
```

Which suggests `NaN` is accepted by the predicate `fp.isNegative`. According to the formalization, `fp.isNegative` holds for `-0.0` and when the number `lt` compares to `0` successfully, neither of which is true for `NaN`.

wintersteiger:
Fixed in unstable, thanks again!

PR #735, opened 2016-09-14
New packaging for and ability to distribute python bindings

rhelmot:
This PR tweaks many, many things about the way that z3's python bindings work. I started with the goal of making it possible for there to be an official distribution of z3 on the python package index - `pip install z3` should let the user obtain a completely functional z3 installation.

These goals required making the following changes:

- The z3 python bindings are now a single python package instead of several interdependent modules. This means that you can no longer say `import z3core` or whatever, it has to be `from z3 import z3core`. Anything that touches only the `z3` module, including `from z3 import *`, should still work exactly as before.
- Because of this, there would have needed to be a folder named `z3` in the `build` directory to be able to use the python bindings directly out of the build directory. This couldn't happen because there already needs to be a file named `z3` in the build directory, the z3 executable! To fix this, the python bindings and examples in the build directory have been moved into a `python` subdirectory.
- A symbolic link of libz3 is made pointing from the `build/python` subdir to `build/libz3.{so,dll,dylib}`. This is important so that python interpreters sitting in `build/python` can use the bindings. This involved adding a feature to support this to the build infrastructure. On platforms without symbolic linking, a copy is made.

The crux of this PR is the file `src/api/python/setup.py`, which can be used to perform installations and distributions of z3. `python setup.py install`, `python setup.py sdist`, `python setup.py bdist_egg`, and `python setup.py bdist_wheel` all work in my testing. You would use `python setup.py register` and `python setup.py upload` to upload any of the distributions you'd built like this.

Some notes about the current state of the branch that will be addressed if you decide that this is a PR that you want to accept:

- The metadata in the setup.py file is for the `angr-only-z3-custom` package that the angr project (which I am a part of) distributes in order to distribute the unstable branch of z3 over pip. This should be changed to the real z3 metadata. :)
- The documentation needs to be updated in order to reflect the new file layout and the new setup option
- The cmake build needs to be fixed for the new file layout
- We probably want to squash those commits

### DIFF
--- a/contrib/cmake/src/api/python/CMakeLists.txt
+++ b/contrib/cmake/src/api/python/CMakeLists.txt
@@ -64,9 +64,15 @@ add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3/z3consts.py"
 )
 list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/z3/z3consts.py")
 
+if (UNIX)
+  set(LINK_COMMAND "create_symlink")
+else()
+  set(LINK_COMMAND "copy")
+endif()
+
 # Link libz3 into the python directory so bindings work out of the box
 add_custom_command(OUTPUT "${z3py_bindings_build_dest}/libz3${CMAKE_SHARED_MODULE_SUFFIX}"
-  COMMAND "${CMAKE_COMMAND}" "-E" "create_symlink"
+  COMMAND "${CMAKE_COMMAND}" "-E" "${LINK_COMMAND}"
     "${CMAKE_BINARY_DIR}/libz3${CMAKE_SHARED_MODULE_SUFFIX}"
     "${z3py_bindings_build_dest}/libz3${CMAKE_SHARED_MODULE_SUFFIX}"
   DEPENDS libz3
--- a/contrib/cmake/src/api/python/CMakeLists.txt
+++ b/contrib/cmake/src/api/python/CMakeLists.txt
@@ -9,7 +9,6 @@ set(z3py_files
   z3poly.py
   z3printer.py
   z3rcf.py
-  z3test.py
   z3types.py
   z3util.py
 )
@@ -18,12 +17,12 @@ set(z3py_bindings_build_dest "${CMAKE_BINARY_DIR}")
 
 set(build_z3_python_bindings_target_depends "")
 foreach (z3py_file ${z3py_files})
-  add_custom_command(OUTPUT "${z3py_bindings_build_dest}/${z3py_file}"
+  add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3/${z3py_file}"
     COMMAND "${CMAKE_COMMAND}" "-E" "copy"
-      "${CMAKE_CURRENT_SOURCE_DIR}/${z3py_file}"
-      "${z3py_bindings_build_dest}"
-    DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/${z3py_file}"
-    COMMENT "Copying \"${z3py_file}\" to ${z3py_bindings_build_dest}"
+      "${CMAKE_CURRENT_SOURCE_DIR}/z3/${z3py_file}"
+      "${z3py_bindings_build_dest}/z3.py"
+    DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/z3/${z3py_file}"
+    COMMENT "Copying \"${z3py_file}\" to ${z3py_bindings_build_dest}/z3.py/${z3py_file}"
   )
   list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/${z3py_file}")
 endforeach()
--- a/contrib/cmake/src/api/python/CMakeLists.txt
+++ b/contrib/cmake/src/api/python/CMakeLists.txt
@@ -4,31 +4,34 @@ message(STATUS "Emitting rules to build Z3 python bindings")
 ###############################################################################
 # This allows the python bindings to be used directly from the build directory
 set(z3py_files
-  z3.py
-  z3num.py
-  z3poly.py
-  z3printer.py
-  z3rcf.py
-  z3types.py
-  z3util.py
+  z3/__init__.py
+  z3/z3.py
+  z3/z3num.py
+  z3/z3poly.py
+  z3/z3printer.py
+  z3/z3rcf.py
+  z3/z3types.py
+  z3/z3util.py
 )
 
-set(z3py_bindings_build_dest "${CMAKE_BINARY_DIR}")
+set(z3py_bindings_build_dest "${CMAKE_BINARY_DIR}/python")
+file(MAKE_DIRECTORY "${z3py_bindings_build_dest}")
+file(MAKE_DIRECTORY "${z3py_bindings_build_dest}/z3")
 
 set(build_z3_python_bindings_target_depends "")
 foreach (z3py_file ${z3py_files})
-  add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3/${z3py_file}"
+  add_custom_command(OUTPUT "${z3py_bindings_build_dest}/${z3py_file}"
     COMMAND "${CMAKE_COMMAND}" "-E" "copy"
-      "${CMAKE_CURRENT_SOURCE_DIR}/z3/${z3py_file}"
-      "${z3py_bindings_build_dest}/z3.py"
-    DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/z3/${z3py_file}"
-    COMMENT "Copying \"${z3py_file}\" to ${z3py_bindings_build_dest}/z3.py/${z3py_file}"
+      "${CMAKE_CURRENT_SOURCE_DIR}/${z3py_file}"
+      "${z3py_bindings_build_dest}/${z3py_file}"
+    DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/${z3py_file}"
+    COMMENT "Copying \"${z3py_file}\" to ${z3py_bindings_build_dest}/${z3py_file}"
   )
   list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/${z3py_file}")
 endforeach()
 
 # Generate z3core.py
-add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3core.py"
+add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3/z3core.py"
   COMMAND "${PYTHON_EXECUTABLE}"
     "${CMAKE_SOURCE_DIR}/scripts/update_api.py"
     ${Z3_FULL_PATH_API_HEADER_FILES_TO_SCAN}
@@ -43,10 +46,10 @@ add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3core.py"
   COMMENT "Generating z3core.py"
   ${ADD_CUSTOM_COMMAND_USES_TERMINAL_ARG}
 )
-list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/z3core.py")
+list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/z3/z3core.py")
 
 # Generate z3consts.py
-add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3consts.py"
+add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3/z3consts.py"
   COMMAND "${PYTHON_EXECUTABLE}"
     "${CMAKE_SOURCE_DIR}/scripts/mk_consts_files.py"
     ${Z3_FULL_PATH_API_HEADER_FILES_TO_SCAN}
@@ -59,13 +62,23 @@ add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3consts.py"
   COMMENT "Generating z3consts.py"
   ${ADD_CUSTOM_COMMAND_USES_TERMINAL_ARG}
 )
-list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/z3consts.py")
+list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/z3/z3consts.py")
+
+# Link libz3 into the python directory so bindings work out of the box
+add_custom_command(OUTPUT "${z3py_bindings_build_dest}/libz3${CMAKE_SHARED_MODULE_SUFFIX}"
+  COMMAND "${CMAKE_COMMAND}" "-E" "create_symlink"
+    "${CMAKE_BINARY_DIR}/libz3${CMAKE_SHARED_MODULE_SUFFIX}"
+    "${z3py_bindings_build_dest}/libz3${CMAKE_SHARED_MODULE_SUFFIX}"
+  DEPENDS libz3
+  COMMENT "Linking libz3 into python directory"
+)
 
 # Convenient top-level target
 add_custom_target(build_z3_python_bindings
   ALL
   DEPENDS
     ${build_z3_python_bindings_target_depends}
+    "${z3py_bindings_build_dest}/libz3${CMAKE_SHARED_MODULE_SUFFIX}"
 )
 
 ###############################################################################
@@ -116,7 +129,7 @@ if (INSTALL_PYTHON_BINDINGS)
   # Using DESTDIR still seems to work even if we use an absolute path
   message(STATUS "Python bindings will be installed to \"${CMAKE_INSTALL_PYTHON_PKG_DIR}\"")
   install(FILES ${build_z3_python_bindings_target_depends}
-    DESTINATION "${CMAKE_INSTALL_PYTHON_PKG_DIR}"
+    DESTINATION "${CMAKE_INSTALL_PYTHON_PKG_DIR}/z3"
   )
 else()
   message(STATUS "Not emitting rules to install Z3 python bindings")
--- a/scripts/mk_genfile_common.py
+++ b/scripts/mk_genfile_common.py
@@ -98,7 +98,7 @@ def mk_z3consts_py_internal(api_files, output_dir):
     openbrace_pat  = re.compile("{ *")
     closebrace_pat = re.compile("}.*;")
 
-    z3consts  = open(os.path.join(output_dir, 'z3consts.py'), 'w')
+    z3consts  = open(os.path.join(output_dir, 'z3/z3consts.py'), 'w')
     z3consts_output_path = z3consts.name
     z3consts.write('# Automatically generated file\n\n')
     for api_file in api_files:
--- a/scripts/mk_genfile_common.py
+++ b/scripts/mk_genfile_common.py
@@ -98,7 +98,7 @@ def mk_z3consts_py_internal(api_files, output_dir):
     openbrace_pat  = re.compile("{ *")
     closebrace_pat = re.compile("}.*;")
 
-    z3consts  = open(os.path.join(output_dir, 'z3/z3consts.py'), 'w')
+    z3consts  = open(os.path.join(output_dir, 'z3', 'z3consts.py'), 'w')
     z3consts_output_path = z3consts.name
     z3consts.write('# Automatically generated file\n\n')
     for api_file in api_files:
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1600,8 +1600,8 @@ def write_core_py_preamble(core_py):
   core_py.write('# Automatically generated file\n')
   core_py.write('import sys, os\n')
   core_py.write('import ctypes\n')
-  core_py.write('from z3types import *\n')
-  core_py.write('from z3consts import *\n')
+  core_py.write('from .z3types import *\n')
+  core_py.write('from .z3consts import *\n')
   core_py.write(
 """
 _lib = None
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1708,7 +1708,7 @@ def generate_files(api_files,
   with mk_file_or_temp(api_output_dir, 'api_log_macros.h') as log_h:
     with mk_file_or_temp(api_output_dir, 'api_log_macros.cpp') as log_c:
       with mk_file_or_temp(api_output_dir, 'api_commands.cpp') as exe_c:
-        with mk_file_or_temp(z3py_output_dir, 'z3/z3core.py') as core_py:
+        with mk_file_or_temp(z3py_output_dir, os.path.join('z3', 'z3core.py')) as core_py:
           # Write preambles
           write_log_h_preamble(log_h)
           write_log_c_preamble(log_c)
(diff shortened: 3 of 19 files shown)


rhelmot:
blah blah blah you can have my soul or whatever, signed

NikolajBjorner:
Your firstborn will also do ;-)

delcypher:
@rhelmot 
I've not taken a proper look at this yet but
> Because of this, there would have needed to be a folder named z3 in the build directory to be able to use the python bindings directly out of the build directory. This couldn't happen because there already needs to be a file named z3 in the build directory, the z3 executable! To fix this, the python bindings and examples in the build directory have been moved into a python subdirectory.

The problem you describe is a rather artificial one for the CMake build. For the CMake build system it would be trivially easy to make sure the `z3` binary (and other binaries) end up in their own directory. I guess we need to keep the old build system working too though. I'm not sure if having the `z3` binary in the the root of the build directory is really a hard requirement.

scripts/mk_genfile_common.py
```diff
@@ -98,7 +98,7 @@ def mk_z3consts_py_internal(api_files, output_dir):
     openbrace_pat  = re.compile("{ *")
     closebrace_pat = re.compile("}.*;")
 
-    z3consts  = open(os.path.join(output_dir, 'z3consts.py'), 'w')
+    z3consts  = open(os.path.join(output_dir, 'z3/z3consts.py'), 'w')
```
delcypher: This is not portable. Do `os.path.join('z3', 'z3const.py')`

scripts/update_api.py
```diff
@@ -1706,7 +1706,7 @@ def mk_file_or_temp(output_dir, file_name, mode='w'):
   with mk_file_or_temp(api_output_dir, 'api_log_macros.h') as log_h:
     with mk_file_or_temp(api_output_dir, 'api_log_macros.cpp') as log_c:
       with mk_file_or_temp(api_output_dir, 'api_commands.cpp') as exe_c:
-        with mk_file_or_temp(z3py_output_dir, 'z3core.py') as core_py:
+        with mk_file_or_temp(z3py_output_dir, 'z3/z3core.py') as core_py:
```
delcypher: This is not portable. Do `os.path.join('z3', 'z3core.py')`.

src/api/python/setup.py
```diff
@@ -0,0 +1,88 @@
...
+build_dir = os.path.abspath(os.path.dirname(__file__))
+
+if sys.platform == 'darwin':
+    library_file = "libz3.dylib"
+elif sys.platform == 'win32':
+    library_file = "libz3.dll"
+else:
+    library_file = "libz3.so"
+
+def _configure_z3():
```
delcypher: If you're going to build Z3 I'd highly recommend using the CMake build system instead of the old build system because IMHO it is the future of Z3 development and is what packagers should be using. Of course I would say this though because I wrote the CMake build system :)

rhelmot:
I will fix all of these concerns today! My thoughts:

- Given that the z3 executable and the z3 python bindings cannot exist in the same folder, I think it would make sense that the z3 executable is in the root and the python bindings are in a separate dir. If you like, I can put the z3 executable (and possibly the z3 shared library?) in a `bin/` directory, too.
- I'm hesitant to use the cmake build by default since cmake would be another hard dependency which can't be controlled by pip, but I guess this is why we have binary distributions :)
- I'll definitely fix the non-portable slashes. I was under the impression that windows would just shrug and accept either slash as a path delimiter, but consistency is probably important.

rhelmot:
Pushed the cmake fixes. The `make install` method of installations installs python bindings that do not work without specifying `init('path/to/libz3.so')`, since cmake's library installation destination /usr/local/lib is not included in the library search path, at least on my machine but probably others too. I'd like to solve this by symlinking the library into $PYTHONPATH/z3/lib/libz3.so but apparently performing symbolic links in the install phase of a cmake build is a known hard problem? I've never touched cmake prior to deciding for some reason that it was a good idea to mess with z3's build, so I'm Not Going To Fix This.

I don't have any further changes to make until people comment on:

- Whether the z3 executable/shared library should be moved into a bin directory in the build
- Whether the cmake build should be the one used by the python build (I'm leaning toward yes)

rhelmot:
Are there any updates on this? The code is in a working state and has been tested to the extent of my ability to test it; the only required changes would be to update the metadata, the documentation, and to squash my dev commits.

wintersteiger:
Thanks for the sizeable contribution! I had a browse through this a few days ago and it looks pretty good to me. I'm currently travelling and won't be able to toroughly test this for about another two weeks though. Did anybody try this on Windows yet?

NikolajBjorner:
trying now...

We will have to check if the directory reshuffling and something that looks like a name change to ".z3" (I really hope "import * from z3" remains workable)  is going to break anything.

rhelmot:
`from z3 import *` should work identically! Anything that only ever accessed the bindings through the `z3` module should work absolutely identically, I made sure that would be the case.

contrib/cmake/src/api/python/CMakeLists.txt
```diff
@@ -60,13 +62,23 @@ add_custom_command(OUTPUT "${z3py_bindings_build_dest}/z3consts.py"
   COMMENT "Generating z3consts.py"
   ${ADD_CUSTOM_COMMAND_USES_TERMINAL_ARG}
 )
-list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/z3consts.py")
+list(APPEND build_z3_python_bindings_target_depends "${z3py_bindings_build_dest}/z3/z3consts.py")
+
+# Link libz3 into the python directory so bindings work out of the box
+add_custom_command(OUTPUT "${z3py_bindings_build_dest}/libz3${CMAKE_SHARED_MODULE_SUFFIX}"
+  COMMAND "${CMAKE_COMMAND}" "-E" "create_symlink"
```
NikolajBjorner: Seems not working on windows:

C:\py\z3\build>ninja
[1/1] Linking libz3 into python directory
FAILED: cmd.exe /C "cd /D C:\py\z3\build\src\api\python && "C:\Program Files (x86)\CMake\bin\cmake.exe" -E create_symlink C:/py/z3/build/libz3.dll C:/py/z3/build/python/libz3.dll"
failed to create symbolic link 'C:/py/z3/build/python/libz3.dll': No error

NikolajBjorner:
The part that adds a symbolic link doesn't seem to be working for me. I added a review comment at the place that uses the symlink to libz3.dll

rhelmot:
Does build/libz3.dll exist? Does cmake -E create_symlink not fall back to copying on windows?

NikolajBjorner:
libz3.dll is there:

C:\py\z3\build>dir
 Volume in drive C is OSDisk
 Volume Serial Number is E29D-C88D

 Directory of C:\py\z3\build

09/21/2016  02:37 PM    <DIR>          .
09/21/2016  02:37 PM    <DIR>          ..
09/21/2016  02:27 PM           260,868 .ninja_deps
09/21/2016  02:39 PM            57,771 .ninja_log
09/21/2016  02:37 PM         1,451,179 build.ninja
09/21/2016  02:37 PM            17,502 CMakeCache.txt
09/21/2016  02:37 PM    <DIR>          CMakeFiles
09/21/2016  02:25 PM             1,466 cmake_install.cmake
09/21/2016  02:24 PM               957 cmake_uninstall.cmake
09/21/2016  02:37 PM    <DIR>          examples
09/21/2016  02:27 PM        29,538,304 libz3.dll
09/21/2016  02:27 PM            85,284 libz3.exp
09/21/2016  02:27 PM        68,207,228 libz3.ilk
09/21/2016  02:27 PM           140,674 libz3.lib
09/21/2016  02:27 PM        85,159,936 libz3.pdb
09/21/2016  02:37 PM    <DIR>          python
09/21/2016  02:37 PM            27,675 rules.ninja
09/21/2016  02:37 PM    <DIR>          src
09/21/2016  02:34 PM                32 test.py
09/21/2016  02:27 PM        23,578,112 z3.exe
09/21/2016  02:27 PM               381 z3.exe.manifest
09/21/2016  02:27 PM               924 z3.exp
09/21/2016  02:27 PM             2,274 z3.lib
09/21/2016  02:27 PM        83,398,656 z3.pdb
              18 File(s)    291,929,223 bytes
               6 Dir(s)  16,802,131,968 bytes free

rhelmot:
lmao it looks like it doesn't fall back to copying. I will push a fix...

NikolajBjorner:
Skimming online forums (back to 2009), it seems creation of symlinks are not supported for Windows. While my version of cmake 3.5 the behavior seems to be the same as the forums point out. So from what I can tell, for windows, better to copy.

NikolajBjorner:
Thanks a lot for the contribution. I have validated it on my environment. It is quite possibly going to break z3test, which tests the python bindings. It takes a while to run this as it by default uses a fresh build from the master branch.

rhelmot:
Alright! I have a few questions then

- What would be the process by which z3 could be distributed on pip? Is there a member of the z3 team that typically handles distributions, or should I be allowed to do that?
- What would be the process of backporting these changes to the unstable branch? Can unstable be easily rebased on top of master?

NikolajBjorner:
Well, hit the first issue:

+        if self.staging_link:
+            if IS_WINDOWS or IS_OSX:
+                out.write('\n\tcp %s %s' % (self.dll_file(), self.staging_link))
+            else:
+                out.write('\n\tln -s %s %s' % (os.path.join(reverse_path(self.staging_link), self.dll_file()), self.staging_link))

The command "cp" isn't exposed in the standard windows shell. It is typically called "copy" or "xcopy".

rhelmot:
How do you want fixes submitted, just another pull request?

NikolajBjorner:
Trying a fix. A test is to ensure that win32test.py works in the z3test project. Not sure what other code paths use this feature. Maybe the python build.

NikolajBjorner:
•What would be the process by which z3 could be distributed on pip? Is there a member of the z3 team that typically handles distributions, or should I be allowed to do that?
  
You can redistribute Z3. Others redistribute it in various ways. If the redistribution mechanism touches many users and they start asking questions, then it falls on you to follow up.

•What would be the process of backporting these changes to the unstable branch? Can unstable be easily rebased on top of master?

Unstable isn't used any longer. The main branch is "master". It may very well be unstable at times.

NikolajBjorner:
Here is another: 
<code>

def testz3py(branch="master", debug=True, clang=False):
    z3dir = find_z3depot()
    bdir = get_builddir(branch, debug, clang)
    p    = os.path.join(z3dir, bdir)
    with cd(p):
        if subprocess.call([config.PYTHON, 'z3test.py']) != 0:
            raise Exception("Failed to execute Z3 python regression tests 'z3test.py' at '%s'" % p)
        if subprocess.call([config.PYTHON, 'z3num.py']) != 0:
            raise Exception("Failed to execute Z3 python regression tests 'z3num.py' at '%s'" % p)

</code>

previous behavior expect z3test.py and z3num.py to be in the build directory. z3test.py seems MIA while z3num is under the python\z3 directory.

contrib/cmake/src/api/python/CMakeLists.txt
```diff
@@ -4,32 +4,34 @@ message(STATUS "Emitting rules to build Z3 python bindings")
 ###############################################################################
 # This allows the python bindings to be used directly from the build directory
 set(z3py_files
-  z3.py
-  z3num.py
-  z3poly.py
-  z3printer.py
-  z3rcf.py
-  z3test.py
```
NikolajBjorner: why is z3test.py gone?

NikolajBjorner:
What is the suggested way to test z3num.py?

C:\z3\build\master\debug\python\z3>z3num.py
  File "C:\z3\build\master\debug\python\z3\z3num.py", line 8, in <module>
    from .z3 import *
ValueError: Attempted relative import in non-package

C:\z3\build\master\debug\python\z3>cd ..

C:\z3\build\master\debug\python>z3\z3num.py
  File "C:\z3\build\master\debug\python\z3\z3num.py", line 8, in <module>
    from .z3 import *
ValueError: Attempted relative import in non-package

NikolajBjorner:
Since z3test.py is still under src/api/python and the python build scripts are updated to copy python files from the z3 sub-directory it isn't being copied in the python build environment either. Based on your design intent, can you fix this?

rhelmot:
z3test got removed because in a previous version of me attempting to write this code, there was no clean way for me to describe where z3test.py would get put in the build dir without adding a new build target, and then I forgot about it when I switched to the current system.

z3num is tricker - I'm about to submit a pull request that fixes this by making z3test.py able to test both the main z3 module and z3num, controlled by a command line argument.

NikolajBjorner:
Ok, will test later tonight. Thanks

rhelmot:
implemented in #742

NikolajBjorner:
Maybe you can help with the regression tests. They currently fail on all platforms.
The environment setup for the new paths has changed, but I was not able to figure out how to adjust it.

This is the relevant piece in z3test/scripts/util.py:

    def test_pyscripts(z3libdir, scriptdir, ext="py", timeout_duration=60.0):
    pydir = os.path.join(z3libdir, "python")
    print("Testing scripts at %s using %s" % (scriptdir, z3libdir))
    with setenv('LD_LIBRARY_PATH', z3libdir):
        with setenv('PYTHONPATH', pydir):
            with setenv('DYLD_LIBRARY_PATH', z3libdir):
                print("Testing python scripts at %s using %s" % (scriptdir, z3libdir))
                error = False
                for script in filter(lambda f: f.endswith(ext), os.listdir(scriptdir)):
                    script = os.path.join(scriptdir, script)
                    print("Testing %s" % script)
                    try:
                        if timeout(exec_script,
                                   args=[script, timeout_duration],
                                   timeout_duration=timeout_duration,
                                   default=False) == False:
                            raise Exception("Timeout executing script '%s' at '%s' using '%s'" % (script, scriptdir, z3libdir)) 
                    except Exception as ex:
                        print("Failed")
                        print(ex)
                        error = True
                if error:
                    raise Exception("Found errors testing scripts at '%s' using '%s'" % (scriptdir, z3libdir))


The previous version did not have pypath, and just used z3libdir for everything.
You should be able to build tests by checking out the z3test repository and 
running "scripts/smalltest.py"

NikolajBjorner:
Several issues are coming up.
The Cygwin port seems now broken. 
The regression tests from z3test repository are also failing on all platforms.
I may have to revert these changes if we are not able to fix the regressions.

Below is a report by Bryan Parno on the Cygwin port:

Hi Nikolaj,

I deleted the entire build directory, ran "python scripts/mk_make.py" and then "cd build; make", then "cd python", but I'm still hitting segfaults with python 2.7:

~/git/z3/build/python$ which python
/usr/bin/python
~/git/z3/build/python$ python --version
Python 2.7.10
~/git/z3/build/python$ python example.py Segmentation fault (core dumped)

FWIW, I tried commenting out portions of example.py, and running:

from z3 import *

x = Real('x')
y = Real('y')
s = Solver()
#s.add(x + y > 5, x > 1, y > 1)
#print(s.check())
#print(s.model())

works, but uncommenting either of the first two lines segfaults.  Uncommenting the last one on its own just throws an exception saying that a model isn't available (unsurprisingly).  

I tried looking at the stack trace in gdb, but I'm afraid it's not very informative (see below).  

I also tried Python 3.4 but it appears to be missing a module that's not available via pip3.4.  If Python 3 is the preferred one though, I can try to figure out where to find the missing module.

~/git/z3/build/python$ python3.4 example.py Traceback (most recent call last):
  File "example.py", line 3, in <module>
    from z3 import *
  File "/cygdrive/c/home/git/z3/build/python/z3/__init__.py", line 1, in <module>
    from .z3 import *
  File "/cygdrive/c/home/git/z3/build/python/z3/z3.py", line 44, in <module>
    from . import z3core
  File "/cygdrive/c/home/git/z3/build/python/z3/z3core.py", line 4, in <module>
    import pkg_resources
ImportError: No module named 'pkg_resources'


-Bryan


~/git/z3/build/python$ gdb python
GNU gdb (GDB) (Cygwin 7.10.1-1) 7.10.1
Copyright (C) 2015 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "i686-pc-cygwin".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from python...(no debugging symbols found)...done.
(gdb) r min.py
Starting program: /usr/bin/python min.py [New Thread 11112.0x28a0] [New Thread 11112.0x4bc0] [New Thread 11112.0x2df4] [New Thread 11112.0x4c5c] [New Thread 11112.0x3780]
gdb: unknown target exception 0x406d1388 at 0x770396c2

Program received signal ?, Unknown signal.
0x770396c2 in RaiseException () from /cygdrive/c/WINDOWS/System32/KERNELBASE.dll
(gdb) bt
#0  0x770396c2 in RaiseException () from /cygdrive/c/WINDOWS/System32/KERNELBASE.dll
#1  0x61081614 in SetThreadName(unsigned long, char const*) () from /usr/bin/cygwin1.dll
#2  0x61004a6b in cygthread::create() () from /usr/bin/cygwin1.dll
#3  0x61004b5f in cygthread::async_create(unsigned long) () from /usr/bin/cygwin1.dll
#4  0x7722655b in ntdll!WinSqmEventWrite () from /cygdrive/c/WINDOWS/SYSTEM32/ntdll.dll
#5  0x611c2900 in cygthread::main_thread_id () from /usr/bin/cygwin1.dll
#6  0xed85d3a5 in ?? ()
#7  0x0060f9e0 in ?? ()
#8  0x7720fe1d in ntdll!KiUserApcDispatcher () from /cygdrive/c/WINDOWS/SYSTEM32/ntdll.dll
#9  0x61004b50 in cygthread::create() () from /usr/bin/cygwin1.dll
#10 0x771ca3cd in ntdll!LdrShutdownProcess () from /cygdrive/c/WINDOWS/SYSTEM32/ntdll.dll
#11 0xed85d031 in ?? ()
#12 0x00008e40 in ?? ()
#13 0x0060fca4 in ?? ()
#14 0x77206bc0 in ntdll!LdrInitializeThunk () from /cygdrive/c/WINDOWS/SYSTEM32/ntdll.dll
#15 0x00000000 in ?? ()
(gdb)

rhelmot:
I'm unable to reproduce the cygwin segfaults - I just did a fresh cygwin install, ran configure, ran build, and the example scripts and the tests run perfectly.

With regards to the python 3 issue, pkg_resources is provided by setuptools - I had to explicitly mark the `python3-setuptools` package for installation before this would work.

I will look into the regressions tests now.

NikolajBjorner:
Awesome,

I was able to get my cygwin 2.7 working after downloading pkg_resources using 

       wget https://bootstrap.pypa.io/ez_setup.py -O - | python

pkg_resources apparently doesn't come out of the box with cygwin.
This is a small issue and can be mitigated with documentation for the few cygwin users.

Looking into the regression tests would be great. I am not sure how to address what seems like an environment variable setting issue. The errors look pretty much the same on all platforms:

Z3 was successfully built.
Z3Py scripts can already be executed in the 'build/master/debug/python' directory.
Z3Py scripts stored in arbitrary directories can be executed if the 'build/master/debug/python' directory is added to the PYTHONPATH environment variable and the 'build/master/debug' directory is added to the LD_LIBRARY_PATH environment variable.
Use the following command to install Z3 at prefix /usr/local.
sudo make install
Exception AttributeError: "Context instance has no attribute 'lib'" in <bound method Context.__del__ of <z3.z3.Context instance at 0x80628af38>> ignored
Exception AttributeError: "Context instance has no attribute 'lib'" in <bound method Context.__del__ of <z3.z3.Context instance at 0x8063364d0>> ignored
**********************************************************************
File "regressions/python/ac1.py", line 6, in __main__
Failed example:
    x, y = Reals('x y')
Exception raised:
      File "/usr/local/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest __main__[1]>", line 1, in <module>
        x, y = Reals('x y')
      File "/home/nightly/z3/build/master/debug/python/z3/z3.py", line 2864, in Reals
        ctx = _get_ctx(ctx)
      File "/home/nightly/z3/build/master/debug/python/z3/z3.py", line 219, in _get_ctx
        return main_ctx()
      File "/home/nightly/z3/build/master/debug/python/z3/z3.py", line 214, in main_ctx
        _main_ctx = Context()
      File "/home/nightly/z3/build/master/debug/python/z3/z3.py", line 159, in __init__
        conf = Z3_mk_config()
      File "/home/nightly/z3/build/master/debug/python/z3/z3core.py", line 1219, in Z3_mk_config
        r = lib().Z3_mk_config()
      File "/home/nightly/z3/build/master/debug/python/z3/z3core.py", line 22, in lib
        raise Z3Exception("init(Z3_LIBRARY_PATH) must be invoked before using Z3-python")
    Z3Exception: init(Z3_LIBRARY_PATH) must be invoked before using Z3-python
**********************************************************************
File "regressions/python/ac1.py", line 7, in __main__
Failed example:
    2 - (x - y)
Exception raised:
      File "/usr/local/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest __main__[2]>", line 1, in <module>
        2 - (x - y)
    NameError: name 'x' is not defined
**********************************************************************
File "regressions/python/ac1.py", line 9, in __main__
Failed example:
    2 + (x - y)
....

rhelmot:
You probably don't have to run the bootstrap script, just mark the `python-setuptools` package for installation. I must have checked it without thinking during my installation, or maybe one of the packages I installed had a dependency on it. :)

What do I run in the z3test repo to run these tests? None of the scripts are running for me without errors. The closest I came was win64test.py, which failed because I don't have java. Do I need java to run the regression tests?

NikolajBjorner:
Try running "smalltest", you can comment out all other tests from smalltest.py
(also comment out buildz3 after the first time)

rhelmot:
ah, shit, I found what's probably the issue. Some changes I made accidentally neutered the ability for the python bindings to load the dynamic library using the dynamic linker's default search mechanism. I'm running tests with the fix right now, and if they pass I will push them!

nunoplopes:
This is also breaking my setup here. Please revert it.

I also had a 'from z3 import *', I copied libz3.dll and build/z3/*.py to my proj dir and my project stopped working.
I now get a lot of "NameError: global name 'xx' is not defined".  I'm not a python expert, so I can't really debug these packaging issues.  All I know is that before this commit everything was working fine.

rhelmot:
@nunoplopes: please copy `libz3.dll` and the folder `build/python/z3` to your project dir. The folder `z3` acts as a single entry point for python to pick up the library, so as long as your current directory contains this folder, `import z3` and `from z3 import *` should both work, the former giving you a single object named `z3` with all z3's components available as properties, and the latter importing every single z3 object into the current namespace.

PR #736, opened 2016-09-14
Create socrates.py

MathieuRoger:
I hope this is a correct example of classical syllogism in Z3.
Many samples talks about integer, reals; and there are few sample available on non integer things.

I hope it may helps.

Regards

### DIFF
--- a/examples/python/socrates.py
+++ b/examples/python/socrates.py
@@ -1,0 +1,33 @@
+############################################
+# Copyright (c) Microsoft Corporation. All Rights Reserved.
+# 
+# all humans are mortal
+# Socrates is a human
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
Nice! 
Two comments: Human(socrates) == True can be replaced by Human(Socrates), and Mortal(Socrates) == False can be replaced by Not(Mortal(Socrates)).
Also, you don't query for proofs in this example, so set_param(proof=True) is not required. If you add print(s.proof()) you will see the proof object, which is rather unreadable and essentially shows that these are not for human consumption.

MathieuRoger:
Good to know.
I will apply your recommendations and send another pull request.

PR #739, opened 2016-09-14
Set soname version correctly in cmake build

rhelmot:
Resolves a FIXME in the cmake build. See diff.

Previously, the DT_SONAME of libz3.so was libz3.so.4.4.2.1, which was kind of horrifying to me? This means that if any package maintainer distributes a binary that depends on the z3 shared library, they are forced to redo their distribution any time z3 publishes a _tweak_ update.

Note that this is only if the z3 distributor uses the cmake build, and it looks like the libz3.so in the ubuntu xenial archives was not built this way. This... confuses me, since the python build system produces a libz3.so that doesn't have a DT_SONAME and it's not clear if there's supposed to be a flag to enable this behavior, so I have no idea how the ubuntu distribution was built? The ubuntu distribution currently has a soname of libz3.so.4, which is what this pull request changes the soname to.

All this, of course, operates under the assumption that z3 only breaks its APIs on major version increments. If this is not the case we can change the soname to major.minor.

### DIFF
--- a/contrib/cmake/src/CMakeLists.txt
+++ b/contrib/cmake/src/CMakeLists.txt
@@ -121,9 +121,12 @@ set_target_properties(libz3 PROPERTIES
   # VERSION determines the version in the filename of the shared library.
   # SOVERSION determines the value of the DT_SONAME field on ELF platforms.
   # On ELF platforms the final compiled filename will be libz3.so.W.X.Y.Z
-  # but symlinks will be made to this file from libz3.so and also from libz3.so.W
+  # but symlinks will be made to this file from libz3.so and also from
+  # libz3.so.W.X.
+  # This indicates that no breaking API changes will be made within a single
+  # minor version.
   VERSION ${Z3_VERSION}
-  SOVERSION ${Z3_VERSION_MAJOR})
+  SOVERSION ${Z3_VERSION_MAJOR}.${Z3_VERSION_MINOR})
 
 if (NOT MSVC)
   # On UNIX like platforms if we don't change the OUTPUT_NAME
--- a/contrib/cmake/src/CMakeLists.txt
+++ b/contrib/cmake/src/CMakeLists.txt
@@ -117,14 +117,13 @@ else()
   set(lib_type "STATIC")
 endif()
 add_library(libz3 ${lib_type} ${object_files})
-# FIXME: Set "VERSION" and "SOVERSION" properly
 set_target_properties(libz3 PROPERTIES
-  # FIXME: Should we be using ${Z3_VERSION} here?
-  # VERSION: Sets up symlinks, does it do anything else?
+  # VERSION determines the version in the filename of the shared library.
+  # SOVERSION determines the value of the DT_SONAME field on ELF platforms.
+  # On ELF platforms the final compiled filename will be libz3.so.W.X.Y.Z
+  # but symlinks will be made to this file from libz3.so and also from libz3.so.W
   VERSION ${Z3_VERSION}
-  # SOVERSION: On platforms that use ELF this sets the API version
-  # and should be incremented everytime the API changes
-  SOVERSION ${Z3_VERSION})
+  SOVERSION ${Z3_VERSION_MAJOR})
 
 if (NOT MSVC)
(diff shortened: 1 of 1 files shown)


delcypher:
@rhelmot Thanks for taking a look at this. Setting the SONAME correctly indeed is desirable.

> All this, of course, operates under the assumption that z3 only breaks its APIs on major version increments. If this is not the case we can change the soname to major.minor.

I don't know what the API stability policy is but if I take a look at the current version reported in `master` of `4.4.2.1` as the next version number then your assumption of breaking the API major version increments is wrong. I know because I [had to work around this in KLEE](https://github.com/klee/klee/commit/5960bc6387df8ec89e3e98ae3867b0a3c025033b). Specifically the interface of `Z3_get_error_msg()` changed between 4.4.1 and what is in the `master` branch.

If `4.4.2.1` is the next version of Z3 then `SONAME` will need to be `${Z3_VERSION_MAJOR}.${Z3_VERSION_MINOR}.${Z3_VERSION_PATCH}`

@wintersteiger @NikolajBjorner What will be the version number of the next release of Z3? I think it would be useful if it was `4.5.0.0` rather than `4.4.2.1` because then we could make it the policy that Z3's API may change if the the major or minor version gets bumped (i.e. there will be no API changes if the patch or tweak version get bumped). This would allow us to set the SONAME to `${Z3_VERSION_MAJOR}.${Z3_VERSION_MINOR}`

Is there a policy on API changes? If we were to follow [semantic versioning](http://semver.org/) then we would actually bump the major version every time we made an incompatible API change. I'm not sure if that's what we want but there needs to be some sort of policy so clients of Z3 can expect API breakages.

wintersteiger:
So far we never had an explicit policy for API changes, but I suspect that in the past they aligned fairly often with changes in build number (as in major.minor.build.revision). So, by that history, I'd say the next release will get 4.4.3.0. But, since we added new features (e.g., get-implications and optimization features), we have reason enough to call it 4.5.0.0 too.

I agree that we should have some sort of policy for this, so that things like the breaking change to get_error_msg(_ex) won't make it into a release at all.

delcypher:
As a personal preference I'd like to see breaking API changes require that either major and/or minor  build numbers get incremented.

It is up to you though what policy you'd like to use. Once a decision is made the CMake code can be fixed appropriately.

delcypher:
Damn. Looks like `4.5.0` shipped with this broken behaviour which is a shame.

@wintersteiger @NikolajBjorner did you decide on policy for API changes (i.e. how Z3's version should be incremented when API changes are made?).

wintersteiger:
We didn't decide on a formal policy. In the past we incremented the 2nd/minor version number when there were new features, which usually also implied API additions or changes, and that's also how we decided on 4.5.0.

wintersteiger:
Aside: I think the case of `Z3_get_error_msg`/`_ex` may have been the first and only hard "breaking" API change in the 3.x/4.x era (and it really wasn't necessary). Apart from that the API usually grows with new features/functions. Some functions are (already) marked as deprecated, those may go away anytime now of course.

wintersteiger:
Is this PR intended to be merged and, if so, is it complete? Or will this be replaced with something else and we should close this unmerged?

rhelmot:
I wrote it with the intention of it being complete, but ultimately the
merging of it or any replacement PR needs to be indicative of some sort of
commitment from the core devs about how they're going to version future
releases with respect to API compatibility.

delcypher:
My guess is given the API break between (4.4 and 4.5) the SONAME and VERSION should be `${Z3_VERSION_MAJOR}.${Z3_VERSION_MINOR}`. I'm not an expert here though.

wintersteiger:
I agree with @delcypher, that's the intended way of handling this in the future. Consider this a commitment from a core dev; if we ever decide to change this, this will be just one of the places that we have to change something.

rhelmot:
I have pushed changes to include the minor version number, along with documenting the semantics thereof! I also rebased everything on the current master. :)

wintersteiger:
Looks good to me, merging! @delcypher may want to take another look.

PR #750, opened 2016-09-29
allow python to load libz3 using loader's default search

rhelmot:
This should fix the regression tests. I think at some point I added the `realpath` call to fix some other issue, and this broke the last search case, which was meant to search globally. This adds an explicit path for that case.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1611,7 +1611,7 @@ _lib = None
 def lib():
   global _lib
   if _lib is None:
-    _dirs = ['.', pkg_resources.resource_filename('z3', 'lib'), os.path.join(sys.prefix, 'lib'), '']
+    _dirs = ['.', pkg_resources.resource_filename('z3', 'lib'), os.path.join(sys.prefix, 'lib'), None]
     for _dir in _dirs:
       try:
         init(_dir)
@@ -1641,9 +1641,12 @@ else:
         return ""
 
 def init(PATH):
-  PATH = os.path.realpath(PATH)
-  if os.path.isdir(PATH):
-    PATH = os.path.join(PATH, 'libz3.%s' % _ext)
+  if PATH:
+    PATH = os.path.realpath(PATH)
+    if os.path.isdir(PATH):
+      PATH = os.path.join(PATH, 'libz3.%s' % _ext)
+  else:
+    PATH = 'libz3.%s' % _ext
 
   global _lib
   _lib = ctypes.CDLL(PATH)


NikolajBjorner:
I tried this, but doesn't seem to deal with the situation (on my Win10 machine).

rhelmot:
I suspect that there'll have to be some other hack for windows, where
LD_LIBRARY_PATH and the os x equivalent don't apply. Let me take a look...

NikolajBjorner:
I suspect it is about fixing the following:

    with setenv('LD_LIBRARY_PATH', z3libdir):
        with setenv('PYTHONPATH', pydir):
            with setenv('DYLD_LIBRARY_PATH', z3libdir):
                print("Testing python scripts at %s using %s" % (scriptdir, z3libdir))
                error = False

because the paths for the various references changed.

NikolajBjorner:
I instrumented my small test script and z3core.py to log the path that gets used.
I can observe that it uses the case where PATH = None, so ends up with a reference to libz3.dll without absolute path prefix. So maybe this reference gets mixed up as the unit tests are run from some other directory.

rhelmot:
I can't think of any way that could mess things up... I am running into some issues trying to run the tests on native windows - the pure smt tests are failing, claiming that there's unexpected output but the output is the same as what's expected. I attached a log of the output after I disabled the build step.

[log.txt](https://github.com/Z3Prover/z3/files/499807/log.txt)

NikolajBjorner:
Ignore the smt tests. We know how to deal with this (using git in a mode where it checks out with unix file ending etc.)

I instrumented the code to print the paths it is trying to initialize with:
 ['.', 'C:\\z3\\build\\master\\debug\\python\\z3\\lib', 'C:\\Python27\\lib', None]

The second option looks fishy. It comes from

   pkg_resources.resource_filename('z3', 'lib')

rhelmot:
That is the correct result, it is supposed to work with the new python-based distribution I designed. That folder doesn't exist when you do a build with either of the typical build systems, but it will with setup.py.

I commented out everything but the python tests and from what I can tell from my instrumentation, the only instance where libz3.dll is being found correctly is through z3test.py. In the rest it isn't able to load libz3.dll, even when PATH=None.

NikolajBjorner:
I get a little further by changing the code in util.py (back) to:

    with setenv('LD_LIBRARY_PATH', z3libdir):
        with setenv('PYTHONPATH', z3libdir):
            with setenv('DYLD_LIBRARY_PATH', z3libdir):
                print("Testing python scripts at %s using %s" % (scriptdir, z3libdir))
                error = False
                for script in filter(lambda f: f.endsw

It now finds _lib, but cant find z3 inside the scripts

rhelmot:
Got it to work successfully by appending libdir to PATH - windows uses this to search for libraries, apparently :|

```
    with setenv('LD_LIBRARY_PATH', z3libdir):
        with setenv('PYTHONPATH', pydir):
            with setenv('DYLD_LIBRARY_PATH', z3libdir):
				newpath = os.environ['PATH']
				if is_windows():
					newpath += ';' + z3libdir
				with setenv('PATH', newpath):
					print("Testing python scripts at %s using %s" % (scriptdir, z3libdir))
					error = False
					for script in filter(lambda f: f.endswith(ext), os.listdir(scriptdir)):
						script = os.path.join(scriptdir, script)
						print("Testing %s" % script)
						try:
							if timeout(exec_script,
									   args=[script, timeout_duration],
									   timeout_duration=timeout_duration,
									   default=False) == False:
								raise Exception("Timeout executing script '%s' at '%s' using '%s'" % (script, scriptdir, z3libdir)) 
						except Exception as ex:
							print("Failed")
							print(ex)
							error = True
					if error:
						raise Exception("Found errors testing scripts at '%s' using '%s'" % (scriptdir, z3libdir))
```

NikolajBjorner:
Didn't work for me, but this did:

    pydir = os.path.join(z3libdir,"python")
    with setenv('LD_LIBRARY_PATH', z3libdir):
        with setenv('PYTHONPATH', z3libdir + ";" + pydir):

NikolajBjorner:
Commit added to z3test (does not cross reference this issue).

rhelmot:
Does this resolve the rest of the test failures you were seeing?

NikolajBjorner:
There is a very good chance it does. Will see when the builds kick in. Good night, thanks.

NikolajBjorner:
A partial status is in. It is somewhat mixed: the tests go through on some, but not all platforms. Debian, Ubuntu 64, OSX fail for sure. For some reason Ubuntu 32 build succeeds, but my own Ubuntu shell (under Win10 redstone) fails to find the z3 library even when supplied with the additional environment variables.

NikolajBjorner:
Update: I have tried to use different include path configurations for the various operating systems, yet our builds are still failing on most unix derivatives. The nightly builds under z3prover/bin/nightly indicate the status of which targets build.

wintersteiger:
Just to avoid redundant work: I'm working on this now.

wintersteiger:
It's LD_LIBRARY_PATH on Linux, DYLD_LIBRARY_PATH on OSX and PATH on Windows. Fix committed, tests still running.

rhelmot:
Oh, we eventually fixed this I think. There are several discussions spanning this repository and z3test. Were you still seeing failures when you posted the first message?

wintersteiger:
No, it's not fixed yet. Turns out that, under some (but not all) circumstances, Python exports PYTHONPATH when added to os.environ, but not DYLD_LIBRARY_PATH. Still working on it.

wintersteiger:
(Yes, still seeing different types of failures on all OSs.)

wintersteiger:
Turns out that on OSX, the system integrity protection will remove changes to DYLD_LIBRARY_PATH when new processes are spawned. Fix committed, builds are running.

Issue #570, opened 2016-04-19
Failing assertion (probably related to the use of QF_BV)


martin-neuhaeusser:
In the attached log, the debug version of Z3 fails with an assertion violation in the SAT solver's clause handling. The release version of Z3 works (i.e. hides the error without a seg fault), but valgrind reports an out-of-memory error and aborts the execution.

This issue arises only if the Z3 solver is configured for QF_BV. If we feed it the same problem and configure it for QF_UFBV, the problem does not occur.

[z3_assertion.zip](https://github.com/Z3Prover/z3/files/225884/z3_assertion.zip)

wintersteiger:
I'm afraid I can't reproduce this problem. Could you confirm that it still exists? Also, which platform do you get this segfault on?

martin-neuhaeusser:
I tried on the latest master (commit-id 67e49b4ad), and the problem persists. 
My platform is Debian 8.4 amd64.

With a _debug_ build of Z3, I get the following:
```
marneu@obelix:~/verifier$ z3 --log z3_assertion.log 
ASSERTION VIOLATION
File: ../src/util/vector.h
Line: 238
idx < size()
(C)ontinue, (A)bort, (S)top, (T)hrow exception, Invoke (G)DB
c
Segmentation fault
marneu@obelix:~/verifier$ 
```

NikolajBjorner:
Note: It doesn't repro on windows.

martin-neuhaeusser:
Actually, we have another segfault in our benchmarks, which I guess is related to the same problem.
The corresponding log file is larger, but replaying it with Z3 from the current https://github.com/wintersteiger/z3/tree/new-ml-api branch (commit 140f0bb79) produces a segfault on my machine.
Here is the log: [segfault.zip](https://github.com/Z3Prover/z3/files/247047/segfault.zip)

martin-neuhaeusser:
@NikolajBjorner: I cannot reproduce both issues on Windows, either.

NikolajBjorner:
If you have a stack trace, it may be possible to identify the problem without a repro.

martin-neuhaeusser:
The stack trace of gdb (for the segfault.log example) looks as follows:

```
Program received signal SIGSEGV, Segmentation fault.
0x000000000112206c in sat::clause_set::erase() ()
(gdb) where
#0  0x000000000112206c in sat::clause_set::erase() ()
#1  0x000000000110f754 in sat::simplifier::subsume() ()
[...]

PR #760, opened 2016-10-15
Added FPA numeral accessor API functions.

wintersteiger:
Added `Z3_fpa_get_numeral_*_bv(...)`


johndoe888:
Christoph,

When I test the commit eedf76f with the following smt2 code:
`(declare-fun x () (_ FloatingPoint 8 24))
(assert (fp.eq x (fp #b0 #x81 #b00000000000000000000000)))
(check-sat)
(get-model)`

I receive the following model:
`sat
(model
  (define-fun x () (_ FloatingPoint 8 24)
    (_ +zero 8 24))
)
`
Can you confirm this?
Maybe it is better not to merge this code yet ;)

- John

wintersteiger:
@johndoe888: That's been fixed in the master branch in the meanwhile, I'll update this PR.

johndoe888:
@wintersteiger: Great! That worked.

Something else popped up instead. Using this OCaml code:

---------------
let _ =
  
  let ctx = Z3.mk_context [] in
  let f1 = Z3.FloatingPoint.mk_numeral_s ctx "4.0" (Z3.FloatingPoint.mk_sort_32 ctx) in
  Printf.printf "FP numeral = %s\n" (Z3.Expr.to_string f1);

  (* only used with "wintersteiger:fpa_numeral_accessors"
  begin
    try
      let _ = Z3.FloatingPoint.get_numeral_sign_bv ctx f1 in ()
    with
    | Z3.Error(x)-> Printf.printf "Z3.Error(%s) on numeral_sign_bv\n" x;
  end;
  
  begin
    try
      let _ = Z3.FloatingPoint.get_numeral_exponent_bv ctx f1 in ()
    with
    | Z3.Error(x)-> Printf.printf "Z3.Error(%s) on numeral_exponent_bv\n" x;
  end;

 begin
    try
      let _ = Z3.FloatingPoint.get_numeral_significand_bv ctx f1 in ()
    with
    | Z3.Error(x)-> Printf.printf "Z3.Error(%s) on numeral_significand_bv\n" x;
  end;
  *)
  
  begin
    try
      let _ = Z3.FloatingPoint.get_numeral_exponent_string ctx f1 in ()
    with
    | Z3.Error(x)-> Printf.printf "Z3.Error(%s) on numeral_exponent_string\n" x;
  end;

----------------------------

Using commit bada7e5e638 on branch wintersteiger:fpa_numeral_accessors yields (erroneous):
FP numeral = (fp #b0 #x81 #b00000000000000000000000)
Z3.Error(invalid argument) on numeral_sign_bv
Z3.Error(invalid argument) on numeral_exponent_bv
Z3.Error(invalid argument) on numeral_significand_bv
Z3.Error(invalid argument) on numeral_exponent_string

Using commit f97ffce479e328 of branch Z3Prover:master yields (correctly):
FP numeral = (fp #b0 #x81 #b00000000000000000000000)


Is my code wrong or is there an error somewhere in the patch?
- John

wintersteiger:
None of the Z3.FloatingPoint.get_numeral_*_bv function have been tested yet, so yes, most likely a real bug. 

The master branch does not contain Z3.FloatingPoint.get_numeral_*_bv yet, so I'm not sure what your "correct" version does.

johndoe888:
The code for Z3.FloatingPoint.get_numeral_*_bv is commented out for branch wintersteiger:fpa_numeral_accessors.

So it just shows, that get_numeral_exponent_string (old function) is working in the branch Z3Prover:master and no more in wintersteiger:fpa_numeral_accessors.

wintersteiger:
Oh yes, that's quite possible; the implementation of the get_numeral_*_string functions also changed in this branch.

johndoe888:
@wintersteiger The get_numeral_*_string functions seem to work now :+1: 

The Z3.FloatingPoint.get_numeral_*_bv functions still throw an invalid argument exception using the code above.

wintersteiger:
Yes, last week's github outage meant I couldn't push everything. It's all there now and I'm testing it on all platforms.

johndoe888:
Great :+1:  Only some minor things to report right now:
- The bias parameter of get_numeral_exponent_string is not changing the behavior of the function as in get_numeral_exponent_bv.
- There seems to be some debug output of Z3. I often get something like: "val=1.0 2"

johndoe888:
One more thing I noticed is that ML bindings for is_numeral_* seem to be missing.

wintersteiger:
Thanks for the further tests, will be fixed presently. 

The fpa-numeral functions are definitely there. I fixed them yesterday. Are you talking about the non-fpa is_numeral? I may have adjusted the function names so they are consistent everywhere.

johndoe888:
@wintersteiger We are getting near to the optimum ;)

Could you please point me to the OCaml call for Z3.FloatingPoint.is_nan or similar that bind Z3_fpa_is_numeral_* from z3_fpa.h? I cannot find those.

johndoe888:
Further tests pop up further questions ;)

Using this fp (normal):
xxxxxxxxxxxxxxxxxxxxxx
(define-fun x () (_ FloatingPoint 8 24)
  (fp #b0 #x01 #b00000000000000000000000))
xxxxxxxxxxxxxxxxxxxxxx

I get the following results with the named functions (biased and unbiased):
xxxxxxxxxxxxxxxxxxxxxx
get_numeral_exponent_string_unbiased -126   (fine)
get_numeral_exponent_string___biased 1   (fine)
get_numeral_exponent_bv_____unbiased #x82   (fine)
get_numeral_exponent_bv_______biased #x01   (fine)
xxxxxxxxxxxxxxxxxxxxxx




Using this fp (subnormal or denormal):
xxxxxxxxxxxxxxxxxxxxxx
(define-fun x () (_ FloatingPoint 8 24)
  (fp #b0 #x00 #b10000000000000000000000))
xxxxxxxxxxxxxxxxxxxxxx

I get the following results with the named functions (biased and unbiased):
xxxxxxxxxxxxxxxxxxxxxx
get_numeral_exponent_string_unbiased -253   (should'nt this be -127?)
get_numeral_exponent_string___biased -126   (should'nt this be 0?)
get_numeral_exponent_bv_____unbiased #x03   (should'nt this be x81?)
get_numeral_exponent_bv_______biased #x82   (should'nt this be x00?)
xxxxxxxxxxxxxxxxxxxxxx

wintersteiger:
What does Z3.FloatingPoint.is_nan have to do with Z3_fpa_is_numeral_* ? Quite possible that I haven't added is_* to the ML API yet.

Regarding unbiased exponents: unbiased "special" exponents (#00, #11) don't have a well-defined unbiased version. What is your preferred value in those cases?

johndoe888:
I was looking for is_subnormal to identify subnormal numbers. At the moment we are using Z3.FloatingPoint.get_numeral_significand_string to decide for that, which is a bad idea (imo).

As for the unbiased versions of special exponents: no need to change right now. I was just using them in wrong way. Knowing that it is not well-defined is enough for now.

wintersteiger:
I see! Sure, the new is_* functions are the better choice for that task. I think I still need to add those to all other (non-C) APIs. I'm hoping to find time for that later today.

I'll now adopt the behavior that special exponents are the same whether biased or unbiased.

This:
    get_numeral_exponent_string___biased -126 (seems wrong to me)

Is fine, the exponent of denormal singles is indeed -126, but the significand is of the form 0.xxx.

#x03 and #x82 are a bit of a mystery to me too :)

wintersteiger:
Ah! No, #x82 is correct, just looks unusual, but it's the signed BV for -126.

src/api/api_fpa.cpp
```diff
@@ -1152,7 +1154,11 @@ extern "C" {
                         mpfm.is_denormal(val) ? mpfm.mk_min_exp(ebits) :
                         mpfm.is_inf(val) ? mpfm.mk_top_exp(ebits) :
                         mpfm.bias_exp(ebits, mpfm.exp(val));
-        if (!biased) exp = mpfm.unbias_exp(ebits, exp);
+        if (mpfm.is_normal(val) && !biased)  {
+            std::cout << "unbiassing" << std::endl;
```
johndoe888: This can be removed, I guess ;)

johndoe888:
The usage of OCamls is_subnormal is killing z3 :)

wintersteiger:
That was not restricted to OCaml and/or Z3.FloatingPoint.is_subnormal :-) Ran out of time yesterday, just used the repository as a buffer between PCs/VMs. Added some more, but didn't have time to test any of this yet.

src/api/api_fpa.cpp
```diff
@@ -1155,10 +1155,10 @@ extern "C" {
                         mpfm.is_inf(val) ? mpfm.mk_top_exp(ebits) :
                         mpfm.bias_exp(ebits, mpfm.exp(val));
         if (mpfm.is_normal(val) && !biased)  {
-            std::cout << "unbiassing" << std::endl;            
+            std::cout << "unbiassing" << std::endl;
```
johndoe888: Can be removed again ;)

wintersteiger:
Sweet, thanks for the tests!

martin-neuhaeusser:
@johndoe888 Thanks for doing all the testing!

Looking at the code, is there a reason why the Z3_fpa_get_numeral_exponent_bv function raises an exception if the numeral is NaN? I know that the sign and significand bits are undefined in that case and raising exceptions in the corresponding functions is clearly fine; but the exponent bits are well-defined in case of a NaN, afaik. Could we instead return the "1....1"-bitvector as the exponent in case of NaN?

Another thing: If one extracts the "mathematical" exponent, specifying whether the return-value should be biased or not is plausible. But when the bitvector-representation is concerned, I would have expected to always receive a biased bit-vector according to the IEEE754 encoding? I doubt a little that the Boolean flag for the biased value is necessary. But that is a matter of taste, of course.

wintersteiger:
In SMT there is no prescribed bias and in Z3 numerals are internally represented by unbiased exponents, so we might as well provide both (and unbiased is much easier to read during debugging).

Is the NaN-exponent well-defined in the SMT standard? (IEEE754 is not the goal, those tradeoffs were discussed during lots of SMT discussions.)

martin-neuhaeusser:
@wintersteiger: You are right; according to the SMTLIB paper, NaN indeed stands for itself and has no bit-vector representation (see Sec. IV, A). Sorry for bothering!

wintersteiger:
Great, thanks for checking! I also took another look and came to the same conclusion again. 

Are those functions alright or should we add something else before we merge this?

martin-neuhaeusser:
I have two more issues: First, the is_numeral_negative function is missing in z3.ml{i}
Second, something is still wrong in accessing subnormal numerals. If one constructs a subnormal from three bit-vectors, deconstructs the resulting floating point numeral using the fpa_numeral_accessor functions into three bit-vectors, and uses them to construct a floating point expression again, the results are not equal. This is independent of the biased flag.

An example that reproduces the bug is the following ML code (I think this problem is not ML specific):

```Ocaml
let () =
  Printf.printf "Denormal example with %s\n" (Z3.Version.full_version);
  let ctx = Z3.mk_context [] in

  let denormal_float =
    let sign_bv = Z3.BitVector.mk_numeral ctx "0" 1 in
    let exp_bv = Z3.BitVector.mk_numeral ctx "0" 8 in
    let significand_int = 1 lsl 22 in
    let significand_bv = Z3.BitVector.mk_numeral ctx (string_of_int significand_int) 23 in
    Printf.printf "Creating denormal from bit vectors (fp %s %s %s):\n"
      (Z3.Expr.to_string sign_bv)
      (Z3.Expr.to_string exp_bv)
      (Z3.Expr.to_string significand_bv);
    let tmp = Z3.FloatingPoint.mk_fp ctx sign_bv exp_bv significand_bv in
    Z3.Expr.simplify tmp None
  in
  Printf.printf "Z3 reports: %s\n" (Z3.Expr.to_string denormal_float);


  Printf.printf "Reconstructing bit-vectors using fpa_numeral_accessors with biased exponent:\n";
  let sign_bv = Z3.FloatingPoint.get_numeral_sign_bv ctx denormal_float in
  let exp_bv = Z3.FloatingPoint.get_numeral_exponent_bv ctx denormal_float true in
  let significand_bv = Z3.FloatingPoint.get_numeral_significand_bv ctx denormal_float in
  Printf.printf "Obtained bit-vectors (%s %s %s)\n"
    (Z3.Expr.to_string sign_bv)
    (Z3.Expr.to_string exp_bv)
    (Z3.Expr.to_string significand_bv);
  Printf.printf "Constructing float-value from those bit-vectors:\n";
  let result =
      let tmp = Z3.FloatingPoint.mk_fp ctx sign_bv exp_bv significand_bv in
      Z3.Expr.simplify tmp None
  in
  Printf.printf "Z3 reports: %s\n" (Z3.Expr.to_string result);


  Printf.printf "Reconstructing bit-vectors using fpa_numeral_accessors with unbiased exponent:\n";
  let sign_bv = Z3.FloatingPoint.get_numeral_sign_bv ctx denormal_float in
  let exp_bv = Z3.FloatingPoint.get_numeral_exponent_bv ctx denormal_float false in
  let significand_bv = Z3.FloatingPoint.get_numeral_significand_bv ctx denormal_float in
  Printf.printf "Obtained bit-vectors (%s %s %s)\n"
    (Z3.Expr.to_string sign_bv)
    (Z3.Expr.to_string exp_bv)
    (Z3.Expr.to_string significand_bv);
  Printf.printf "Constructing float-value from those bit-vectors:\n";
  let result =
      let tmp = Z3.FloatingPoint.mk_fp ctx sign_bv exp_bv significand_bv in
      Z3.Expr.simplify tmp None
  in
  Printf.printf "Z3 reports: %s\n" (Z3.Expr.to_string result)
```

wintersteiger:
That's a great test, thanks for sharing that! There were indeed a couple issues with denormal exponents, and one additional issue with significands as well (wrong bit-vector size), but those are all fixed now.

martin-neuhaeusser:
Perfect, thanks!
@johndoe888: Could you please adapt our Z3 layer and run the tests again?

johndoe888:
I have problems compiling the code:

../src/api/api_fpa.cpp: In function ‘const char* Z3_fpa_get_numeral_exponent_string(Z3_context, Z3_ast, Z3_bool)’:
../src/api/api_fpa.cpp:1091:9: error: expected ‘;’ before ‘}’ token
         }
         ^

wintersteiger:
@johndoe888 Indeed, missing semicolon that some compilers didn't like. Fixed now.

johndoe888:
My tests look fine. :+1:

martin-neuhaeusser:
Mine look fine, as well. There are one or two instances where Z3 segfaults (on QF_FP - see also #570), but the numeral accessor functions seem work, as expected.

wintersteiger:
That's great, thanks for all the tests. This is now merged into master, so please file new issues if anything should come up.

PR #779, opened 2016-11-07
Standardize on __uint64 instead of unsigned __int64

ddcc:


wintersteiger:
That looks good, thanks for the fix!

delcypher:
@wintersteiger @ddcc 
Is there any reason not to use `uint64_t` (from `stdint.h`) instead of `__uint64`?

wintersteiger:
@declyper: Historical platform compatibility. If I remember correctly, Visual Studio didn't support C99 until around VS 2010 or so? For earlier ones there was no stdint.h at all.

ddcc:
One problem that I've run into is that `__uint64` and `uint64_t` are incompatible on some platforms; the former is declared as a `unsigned long long`, whereas on most x86_64 Linux platforms, the latter is declared as a `unsigned long`. Even though they're both 8 bytes, the two aren't implicitly convertible without an explicit cast.

delcypher:
@wintersteiger I wasn't aware that MSVC ever supported C99. I thought it was stuck at C89. I think C++ expects the header to exist though (as `#include <cstdint>`).

@ddcc That's interesting and also very unfortunate.

PR #780, opened 2016-11-08
[CMake] Bump the version number in the CMake build.

delcypher:
@wintersteiger It's great to see another Z3 release was made but in the future it would be nice if you could either bump the version in the `CMakeLists.txt` file or give me a heads up so I can make the change before making the release. The tagged `4.5.0` release has the wrong version in the CMake build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,9 +32,9 @@ project(Z3 C CXX)
 # Project version
 ################################################################################
 set(Z3_VERSION_MAJOR 4)
-set(Z3_VERSION_MINOR 4)
-set(Z3_VERSION_PATCH 2)
-set(Z3_VERSION_TWEAK 1)
+set(Z3_VERSION_MINOR 5)
+set(Z3_VERSION_PATCH 1)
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
It's mostly icing on the cake: the release builds are binaries so they will have the right version number.
The release sources are not going to be perfect, and furthermore I believe we still have a couple of pending items to complete the cmake system for features (some examples, ocaml) which could allow us to also retire the legacy approach.

delcypher:
@NikolajBjorner My worry is if anyone builds Z3 from that release tag with CMake bult binary will report the wrong version number which is unfortunate.

wintersteiger:
Sorry, this one slipped past me. I grep'd for hard-coded version numbers and fixed one in the Python API, but I didn't find yours. Obviously, the version string should only be hard-coded in one place (until now that was mk_project.py). However, very few people will download the source tarball and the binaries report the right number. I've never heard of anyone using the tags, but I'm sure there is at least one person that will use it, but as I discussed with @NikolajBjorner, we need to increase our release frequency anyways, so there'll be a new tag to replace it soon.

PR #796, opened 2016-11-20
Fix GCC/Clang compilation.

rickyz:
The calls to negate use a non-dependent name, so GCC and Clang do not
examine dependent base classes when looking up the name. Adds a using
declaration as suggested at
https://isocpp.org/wiki/faq/templates#nondependent-name-lookup-members.


NikolajBjorner:
Thanks!

PR #800, opened 2016-11-22
Add convenience functions for x87 80-bit extended floating-point type

ddcc:


delcypher:
Cool!

At a glance this looks good.

It occurs to me that using `Z3_mk_fpa_to_ieee_bv()` on expressions of this sort though might not work properly as [extended precision](https://en.wikipedia.org/wiki/Extended_precision#x86_extended_precision_format) is not IEEE (e.g. the integer part of the significant is explicit rather than implicit and you also have this weird "unnormal" number thing)

wintersteiger:
This looks technically ok, but I'm not sure it's a good design idea to add those. Like @delcypher said, in IEEE754 extended and "extendable" formats are only a suggestion. The SMT FP standard doesn't include a name for these sorts either. The danger is that users will expect all FP operations in 80-bit will behave like Intel x87 - 80-bit numbers, which is not necessarily the same as using 80-bit IEEE754 numbers or 80-bit SMT numbers. I don't remember the exact details of those differences though; they could be trivial.

ddcc:
Yeah, the semantics don't quite line up with IEEE 754, so it'd be tricky to verify that mixed constraints involving both sorts would be correct. And especially from a user perspective as well, since the differences aren't immediately obvious.

wintersteiger:
Exactly, I'd really like to avoid additional confusion for users; "normal" floats are tricky enough :-) Still, thanks for the contribution, even if it didn't go in this time!

PR #805, opened 2016-11-23
Do not request time stamp if not needed

MartinNowack:
If no timeout is needed for solving queries, time stamps do not
need to be acquired.

My personal experiments show that measuring the time in the context of symbolic execution (KLEE) can take up to 25% of the execution time.


NikolajBjorner:
thanks

Issue #836, opened 2016-12-09
Many -Woverloaded-virtual warnings


delcypher:
Clang 3.9 emits many `-Woverloaded-virtual` warnings

Here's an example:

```
[659/664] Building CXX object src/shell/CMakeFiles/shell.dir/datalog_frontend.cpp.o
In file included from /home/dsl11/dev/klee-afr/newer_z3/src/src/shell/datalog_frontend.cpp:30:
In file included from /home/dsl11/dev/klee-afr/newer_z3/src/src/muz/rel/dl_compiler.h:33:
In file included from /home/dsl11/dev/klee-afr/newer_z3/src/src/muz/rel/dl_base.h:31:
/home/dsl11/dev/klee-afr/newer_z3/src/src/muz/base/dl_context.h:122:22: warning: 'datalog::rel_context_base::display_profile' hides overloaded virtual function [-Woverloaded-virtual]
        virtual void display_profile(std::ostream& out) = 0;
                     ^
/home/dsl11/dev/klee-afr/newer_z3/src/src/muz/base/dl_engine_base.h:69:22: note: hidden overloaded virtual function 'datalog::engine_base::display_profile' declared here: different qualifiers (const vs none)
        virtual void display_profile(std::ostream& out) const {}
                     ^
In file included from /home/dsl11/dev/klee-afr/newer_z3/src/src/shell/datalog_frontend.cpp:32:
In file included from /home/dsl11/dev/klee-afr/newer_z3/src/src/muz/rel/dl_finite_product_relation.h:25:
/home/dsl11/dev/klee-afr/newer_z3/src/src/muz/rel/dl_table_relation.h:52:33: warning: 'datalog::table_relation_plugin::mk_full' hides overloaded virtual functions [-Woverloaded-virtual]
        virtual relation_base * mk_full(const relation_signature & s, func_decl* p, family_id kind);
                                ^
/home/dsl11/dev/klee-afr/newer_z3/src/src/muz/rel/dl_base.h:314:35: note: hidden overloaded virtual function 'datalog::tr_infrastructure<datalog::relation_traits>::plugin_object::mk_full' declared here: different number of parameters (2 vs 3)
            virtual base_object * mk_full(func_decl* p, const signature & s) {
                                  ^
[...]

PR #838, opened 2016-12-11
Potential fixes for #836

delcypher:
Here are some potential fixes for #836 .

@NikolajBjorner 
Please think carefully about whether these fixes actually make sense because I don't know this code.

I think the the change to `src/smt/proto_model/numeral_factory.h` is potentially dangerous because now there are two very similar methods

```
bv_factory::mk_value(rational const&, unsigned)
bv_factory::mk_value(rational const&, sort*)
```

Which could easily be confused. Especially if you call `mk_value(some_rational, NULL)` which will actually call `bv_factory::mk_value(rational const&, unsigned)` not `bv_factory::mk_value(rational const&, sort*)`.

A few alternative ideas

* Rename `bv_factory::mk_value` to avoid hiding the method from the parent class.
* Suppress the Clang warning here using `#pragma` (I'm not a big fan of this idea).

There are a probably other ways of fixing this too. I just haven't thought of them.

### DIFF
--- a/src/muz/rel/dl_sieve_relation.h
+++ b/src/muz/rel/dl_sieve_relation.h
@@ -104,6 +104,9 @@ namespace datalog {
         sieve_relation * mk_empty(const relation_signature & s, relation_plugin & inner_plugin);
 
         virtual relation_base * mk_full(func_decl* p, const relation_signature & s);
+        // Explicitly use `mk_full` from parent class so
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
Thanks! 
I am going to address these by using different naming conventions. 
I also already use get_num_value for other numeric plugins.

delcypher:
@NikolajBjorner Great. I'll close this PR then. Please ping me once you've made the changes so I can check the build.

PR #841, opened 2016-12-11
Make `scoped_timer` robust to spurious wakes under Linux.

### LINKED ISSUE #839: scoped_timer not firing?
I'm trying to debug instances where it appears that Z3's solver timeout is not firing on Linux in my application. I'm struggling to debug this as it doesn't seem to always happen and I'm also using very large timeouts (~1600 seconds) which makes repeating this problem rather slow. When I've tried to observe this using gdb because every time I've attached to the process the timer appears to have fired when expected.

I believe that the `scoped_timer` is used for enforcing the timeout. I was looking over the code and saw something that might be the culprit.

```c++
#elif defined(_LINUX_) || defined(_FREEBSD_)
    static void* thread_func(void *arg) {
        scoped_timer::imp *st = static_cast<scoped_timer::imp*>(arg);

        struct timespec end_time;
        clock_gettime(CLOCK_REALTIME, &end_time);
        end_time.tv_sec  += st->m_ms / 1000u;
        end_time.tv_nsec += (st->m_ms % 1000u) * 1000000ull;
        // check for overflow
        if (end_time.tv_nsec >= 1000000000) {
            ++end_time.tv_sec;
            end_time.tv_nsec -= 1000000000;
        }

        pthread_mutex_lock(&st->m_mutex);
        st->m_initialized = true;
        int e = pthread_cond_timedwait(&st->m_cond, &st->m_mutex, &end_time);
        ENSURE(e == 0 || e == ETIMEDOUT);

        pthread_mutex_unlock(&st->m_mutex);
        if (e == ETIMEDOUT)
            st->m_eh->operator()();
        return 0;
    }
```

According to the documentation of `pthread_cond_timedwait()`

```
When using condition variables there is always a Boolean predicate involving
shared variables associated with each condition wait that is true if the thread
should proceed. Spurious wakeups  from  the pthread_cond_timedwait()  or
pthread_cond_wait()  functions may occur. Since the return from
pthread_cond_timedwait() or pthread_cond_wait() does not imply anything about
the value of this predicate, the predicate should be re-evaluated upon such
return.
```

My interpretation of this is that `pthread_cond_timedwait()` could be spuriously woken up and this **does not imply** that the condition variable was signalled. It seems there should be some other predicate that should be checked after `pthread_cond_timedwait()` to check if a signal was actually sent.

The current implementation of `scoped_timer` for Linux does not do this. It presumes that if `pthread_cond_timedwait()` returns 0 that it was signalled and the thread exits.

I'm guessing that `pthread_cond_timedwait()` should be called in a loop and if `0` is returned it should be checked (probably by reading some `bool` that is set by the code that calls `pthread_cond_signal()`) that we were signalled.

Also see http://stackoverflow.com/questions/8594591/why-does-pthread-cond-wait-have-spurious-wakeups

delcypher:
Fix implementation of `scoped_timer` under Linux where it was
incorrectly assumed that `pthread_cond_timedwait()` would exit
due to a condition variable being signaled or a timeout occuring.

According to the documentation `pthread_cond_timedwait()` might
spuriously wake so we introduce a new variable `m_signal_sent` (that is
guarded by an existing mutex) that is used as the variable to check that
the thread wake was not spurious.

This is intended to partially fix #839 .

### DIFF
--- a/src/util/scoped_timer.cpp
+++ b/src/util/scoped_timer.cpp
@@ -73,6 +73,7 @@ struct scoped_timer::imp {
     pthread_cond_t  m_cond;
     unsigned        m_ms;
     bool            m_initialized;
+    bool            m_signal_sent;
 #else
     // Other
 #endif
@@ -119,10 +120,18 @@ struct scoped_timer::imp {
 
         pthread_mutex_lock(&st->m_mutex);
         st->m_initialized = true;
-        int e = pthread_cond_timedwait(&st->m_cond, &st->m_mutex, &end_time);
-        ENSURE(e == 0 || e == ETIMEDOUT);
-
+        int e = 0;
+        // `pthread_cond_timedwait()` may spuriously wake even if the signal
+        // was not sent so we loop until a timeout occurs or the signal was
+        // **really** sent.
+        while (!(e == 0 && st->m_signal_sent)) {
+          e = pthread_cond_timedwait(&st->m_cond, &st->m_mutex, &end_time);
+          ENSURE(e == 0 || e == ETIMEDOUT);
+          if (e == ETIMEDOUT)
+            break;
+        }
         pthread_mutex_unlock(&st->m_mutex);
+
         if (e == ETIMEDOUT)
             st->m_eh->operator()();
         return 0;
@@ -170,6 +179,7 @@ struct scoped_timer::imp {
         // Linux & FreeBSD
         m_ms = ms;
         m_initialized = false;
+        m_signal_sent = false;
         ENSURE(pthread_mutex_init(&m_mutex, NULL) == 0);
         ENSURE(pthread_cond_init(&m_cond, NULL) == 0);
         ENSURE(pthread_create(&m_thread_id, NULL, &thread_func, this) == 0);
@@ -218,7 +228,10 @@ struct scoped_timer::imp {
             if (!init)
                 sched_yield();
         }
+        pthread_mutex_lock(&m_mutex);
+        m_signal_sent = true;
         pthread_cond_signal(&m_cond);
+        pthread_mutex_unlock(&m_mutex);
 
         pthread_join(m_thread_id, NULL);
         pthread_cond_destroy(&m_cond);


delcypher:
@wintersteiger This is a sketch of the intended fix. It's WIP that I haven't tested thoroughly yet so I advise not merging it until I've tested it some more.

delcypher:
@wintersteiger What would like to do with this? I think the fix is technically correct but it hasn't received a huge amount of testing yet.

NikolajBjorner:
so this is ready to merge?

delcypher:
@NikolajBjorner No I need to test this some more. It turns the problem I was observing was a bug in the tool using Z3 rather than in Z3 itself.

This fix is therefore a fix for a problem I never observed (even though I think it's a bug) so I'd like to give it some more testing before I'm happy with this.

delcypher:
@NikolajBjorner I've tested this and I haven't found any problems so far so it's probably fine to merge. Arguably a similar change ought to be made for macOS as it has similar code.

NikolajBjorner:
You are setting m_signal_set = true under a mutex, but the while loop that checks for the timeout is also within the same mutex. I am not sure this is what you intend with the new flag. What state do you intend to protect?

src/util/scoped_timer.cpp
```diff
@@ -218,7 +228,10 @@ struct scoped_timer::imp {
             if (!init)
                 sched_yield();
         }
+        pthread_mutex_lock(&m_mutex);
```
NikolajBjorner: this mutex protects both m_signal_sent as well as the condition variable. THe only code that reads m_signal_sent seems within the same mutex.

delcypher:
@NikolajBjorner 
> You are setting m_signal_set = true under a mutex, but the while loop that checks for the timeout is also within the same mutex. I am not sure this is what you intend with the new flag. What state do you intend to protect?

`m_signal_sent` is needed because we need a way of determining if a signal was **actually sent**. This is because according to `pthread_cond_timedwait()` it may spuriously be woken up.

From `man pthread_cond_timedwait`
>  When using condition variables there is always a Boolean predicate involving
shared variables associated with each condition wait that is true if the thread
should proceed. Spurious wakeups  from  the pthread_cond_timedwait()  or
pthread_cond_wait()  functions may occur. Since the return from
pthread_cond_timedwait() or pthread_cond_wait() does not imply anything about
the value of this predicate, the predicate should be re-evaluated upon such
return.

Here the "Boolean predicate" is the `m_signal_sent` variable that I have added.

src/util/scoped_timer.cpp
NikolajBjorner: Thanks, you are right.

PR #849, opened 2016-12-18
Making z3 python look in its installation directory for the z3 lib

ColdHeat:
Thought this made things a little more organized for those using the Python bindings since you can just move libz3 into the Python bindings directory.


NikolajBjorner:
thanks!

PR #854, opened 2016-12-22
Add -fpic to armv7/armv8 build

### LINKED ISSUE #585: Build fails on certain ARMv7 linux environments demanding -fPIC
On some but not all of my ARMv7 linux machines, linking fails with:

```
/usr/bin/ld: api/dll/mem_initializer.o: relocation R_ARM_THM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
api/dll/mem_initializer.o: error adding symbols: Bad value
```

I get this failure on an Nvidia Jetson-TK1 devkit running their provided Ubuntu 14.04 and g++ 4.8.4 and on a https://www.scaleway.com/ server running Ubuntu 15.10 and g++ 5.2.1. It builds properly on a Raspberry Pi running Raspbian 8.0 (jessie) and g++ 4.9.2.

I have a patch that fixes it by adding -fPIC when building on armv7 but I want to think if there is a more fine-grained solution here, since it doesn't affect all my machines.

rhelmot:
resolves #585 

I have not been able to test this end-to-end, since my only armv7 machine is qemu-system-armhf, upon which building z3 takes approximately 6 hours, but I have verified that it produces a correct-looking config.mk.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -2441,7 +2441,7 @@ def mk_config():
                 CXXFLAGS     = '%s -O3 -D _EXTERNAL_RELEASE -fomit-frame-pointer' % CXXFLAGS
         if is_CXX_clangpp():
             CXXFLAGS   = '%s -Wno-unknown-pragmas -Wno-overloaded-virtual -Wno-unused-value' % CXXFLAGS
-        sysname = os.uname()[0]
+        sysname, _, _, _, machine = os.uname()
         if sysname == 'Darwin':
             SO_EXT    = '.dylib'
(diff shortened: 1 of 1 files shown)


rhelmot:
What is needed for this to be merged? Should I find a raspberry pi somewhere and make sure everything works correctly?

wintersteiger:
Looks innocent enough, but I guess it's a good idea to let it run through at least once. I have a couple RPIs to test on, but I'm not sure when I'll get around to it.

wintersteiger:
Just tested this on an armv7 and all builds and runs fine.

PR #857, opened 2016-12-26
update guards around MSVC-specific functions to avoid errors in mingw

gyorokpeter:


wintersteiger:
@gyorokpeter This has been in here for a while now. Is the CLA a problem for you? Also, are these changes still necessary for Z3 to compile using mingw and are those all the changes that are necessary?

gyorokpeter:
At the time I tried to build it these were the necessary changes. I haven't tried since then as my focus has moved to other things.

PR #869, opened 2017-01-12
Fix .NET Core bindings

danpere:
Fix two separate issues making the Z3 .NET bindings not work on .NET Core:
1. While `DummyContracts` made the code compile, the expressions of some of the contracts threw exceptions at runtime. This adds `[Conditional]` attributes to the `void` `DummyContracts` methods to make those method calls (and therefore their arguments which throw exceptions) not get included in the compiled code.
2. Unlike Mono, .NET Core will not remove the `.dll` from the end of a native library name in attempt to find the correct filename on non-Windows platforms. This is fixed by not including any extension at all.

Note the build script still does not actually build the .NET Core bindings, but they can be built manually following the instructions already in the `README` in the `src/api/dotnet/core` directory.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -365,7 +365,7 @@ def mk_dotnet(dotnet):
     dotnet.write('        public delegate void Z3_error_handler(Z3_context c, Z3_error_code e);\n\n')
     dotnet.write('        public class LIB\n')
     dotnet.write('        {\n')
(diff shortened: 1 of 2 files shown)


delcypher:
@danpere While you are looking at this could you check the CMake based build of the dot net bindings (see https://github.com/Z3Prover/z3/blob/master/README-CMake.md )?

When I wrote this I tested using Windows's .NET implementation and mono on Linux but I did not check `.NET core`. I suspect a few changes will need to be made to make this work.

PR #881, opened 2017-01-27
Thread labels through tactic system

dwoos:
This adds support for Simplify-style labels to the tactic system. Previously, `(labels)` calls would return no labels after `(check-sat-using smt)`; now they should return the same labels as if `(check-sat)` was called instead.

Model converters now convert both models and labels; another option would be to add a separate `label_converter` class.

### DIFF
--- a/src/cmd_context/tactic_cmds.cpp
+++ b/src/cmd_context/tactic_cmds.cpp
@@ -165,7 +165,23 @@ public:
     }
 };
 
-typedef simple_check_sat_result check_sat_tactic_result;
+struct check_sat_tactic_result : public simple_check_sat_result {
+public:
+  svector<symbol> labels;
+
+  check_sat_tactic_result(ast_manager & m) : simple_check_sat_result(m) {
+  }
+
+  virtual void get_labels(svector<symbol> & r) {
+    r.append(labels);
+  }
+
+  virtual void add_labels(svector<symbol> & r) {
+    labels.append(r);
+  }
+};
+
+typedef svector<symbol> & labels_ref;
 
 class check_sat_using_tactict_cmd : public exec_given_tactic_cmd {
 public:
@@ -189,6 +205,7 @@ public:
         ast_manager & m = ctx.m();
         unsigned timeout   = p.get_uint("timeout", ctx.params().m_timeout);
         unsigned rlimit  =   p.get_uint("rlimit", ctx.params().m_rlimit);
+        svector<symbol> labels;
         goal_ref g = alloc(goal, m, ctx.produce_proofs(), ctx.produce_models(), ctx.produce_unsat_cores());
         assert_exprs_from(ctx, *g);
         TRACE("check_sat_using", g->display(tout););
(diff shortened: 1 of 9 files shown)


src/tactic/model_converter.cpp
```diff
@@ -102,8 +126,12 @@ model_converter * concat(model_converter * mc1, unsigned num, model_converter *
 
 class model2mc : public model_converter {
     model_ref m_model;
+    buffer<symbol> m_labels;
 public:
     model2mc(model * m):m_model(m) {}
+
+    model2mc(model * m, buffer<symbol> r):m_model(m), m_labels(r) {}
```
NikolajBjorner: buffer<symbol> const&

src/tactic/model_converter.cpp
```diff
@@ -76,6 +82,24 @@ class concat_star_model_converter : public concat_star_converter<model_converter
         }
         UNREACHABLE();
     }
+
+    virtual void operator()(svector<symbol> & r, unsigned goal_idx) {
+        unsigned num = this->m_c2s.size();
+        for (unsigned i = 0; i < num; i++) {
+            if (goal_idx < this->m_szs[i]) {
```
NikolajBjorner: how are m_szs and m_c2s.size related? Hopefully num == m_szs.size(), but if so, then why not use num = m_szs.size()?

src/cmd_context/tactic_cmds.cpp
```diff
@@ -165,7 +165,23 @@ class exec_given_tactic_cmd : public parametric_cmd {
...
+  virtual void get_labels(svector<symbol> & r) {
+    r.append(labels);
+  }
+
+  virtual void add_labels(svector<symbol> & r) {
+    labels.append(r);
+  }
+};
+
+typedef svector<symbol> & labels_ref;
```
NikolajBjorner: Using labels_ref seems nicer than svector<symbol> throughout. Having class labels_ref : public svector<symbol> {}; furthermore allows to maintain that the vector of symbols is related to labels.

msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

src/tactic/model_converter.cpp
dwoos: This code is copy-pasted from `virtual void operator()(model_ref & m, unsigned goal_idx)`; to be honest, I'm not exactly sure how the interaction between the goal index and the model converters works. Two lines later, though, `i` is used as an index into `m_c2s`. Should I still change it?

NikolajBjorner: Sorry, my comment wasn't informed. No need to change. The seem to be aligned.

NikolajBjorner: Sorry, my comment wasn't informed. No need to change. They seem to be aligned.

dwoos:
Nikolai, I think I fixed everything you mentioned. Is this ready to be merged in now?

NikolajBjorner:
I think so. @wintersteiger is it ready as far as you are concerned?

wintersteiger:
Took me a while to figure out what it's for, but yes, that totally makes sense.

PR #882, opened 2017-01-27
Add basic Sine Qua Non filtering

dwoos:
Filter assertions, choosing only those which could affect the goal.

Performance is really poor right now; I suspect I'm doing something wrong, but I'm not sure what. I'm going to keep looking into it, but if something pops out at you guys please let me know!

### DIFF
--- a/src/tactic/sine_filter.cpp
+++ b/src/tactic/sine_filter.cpp
@@ -26,6 +26,7 @@ Revision History:
 #include "var_subst.h"
 #include "ast_util.h"
 #include "obj_pair_hashtable.h"
+#include "ast_pp.h"
 
 class sine_tactic : public tactic {
 
@@ -121,7 +122,7 @@ private:
                     app *a = to_app(curr);
                     func_decl *f = a->get_decl();
                     if (!consts.contains(f)) {
-                        TRACE("sine", tout << f; tout << "\n";);
+                        TRACE("sine", tout << mk_pp(f, m) << "\n";);
                         p_matched = false;
                         next_consts.push_back(f);
                         break;
--- a/src/tactic/sine_filter.cpp
+++ b/src/tactic/sine_filter.cpp
@@ -31,15 +31,11 @@ class sine_tactic : public tactic {
 
     ast_manager&  m;
     params_ref    m_params;
-    datatype_util m_dt;
-    ref<extension_model_converter> m_ext;
-    ref<filter_model_converter>    m_filter;
-    unsigned                       m_num_transformed;
 
 public:
 
     sine_tactic(ast_manager& m, params_ref const& p): 
-        m(m), m_params(p), m_dt(m) {}
+      m(m), m_params(p) {}
     
     virtual tactic * translate(ast_manager & m) {
         return alloc(sine_tactic, m, m_params);
--- a/src/tactic/sine_filter.cpp
+++ b/src/tactic/sine_filter.cpp
@@ -77,11 +77,6 @@ public:
 
 private:
 
-    // is this a user-defined symbol name?
-    bool is_name(func_decl * f) {
-        return f->get_family_id() < 0;
-    }
-
     typedef std::pair<expr*,expr*> t_work_item;
 
     t_work_item work_item(expr *e, expr *root) {
@@ -97,8 +92,8 @@ private:
             stack.pop_back();
             if (is_app(curr)) {
                 app *a = to_app(curr);
-                func_decl *f = a->get_decl();
-                if (is_name(f)) {
+                if (is_uninterp(a)) {
+                    func_decl *f = a->get_decl();
                     consts.insert_if_not_there(f);
                 }
             }
@@ -159,8 +154,8 @@ private:
             stack.pop_back();
             if (is_app(curr.first)) {
                 app *a = to_app(curr.first);
-                func_decl *f = a->get_decl();
-                if (is_name(f)) {
+                if (is_uninterp(a)) {
+                    func_decl *f = a->get_decl();
                     if (!consts.contains(f)) {
                         consts.insert(f);
                         if (const2quantifier.contains(f)) {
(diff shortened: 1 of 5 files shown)


src/tactic/sine_filter.cpp
```diff
@@ -0,0 +1,252 @@
+/*++
+Copyright (c) 2016 Microsoft Corporation
+
```
NikolajBjorner: Add your name somewhere

src/tactic/sine_filter.cpp
```diff
@@ -0,0 +1,252 @@
...
+#include "rewriter_def.h"
+#include "filter_model_converter.h"
+#include "extension_model_converter.h"
+#include "var_subst.h"
+#include "ast_util.h"
+#include "obj_pair_hashtable.h"
+
+#if 0
+TODO documentation here
+#endif
```
NikolajBjorner: delete

src/tactic/sine_filter.cpp
```diff
@@ -0,0 +1,252 @@
...
+    params_ref    m_params;
+    datatype_util m_dt;
+    ref<extension_model_converter> m_ext;
+    ref<filter_model_converter>    m_filter;
+    unsigned                       m_num_transformed;
+
+public:
+
+    sine_tactic(ast_manager& m, params_ref const& p): 
+        m(m), m_params(p), m_dt(m) {}
```
NikolajBjorner: Initialize m_num_transformed(0) for good order

src/tactic/sine_filter.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+    
+    virtual void cleanup() {
+    }
+
+private:
+
+  // is this a user-defined symbol name?
+  bool is_name(func_decl * f) {
+    return f->get_family_id() < 0;
```
NikolajBjorner: There must be some better method than this. What is the specification of this function? We have methods called "is_uninterp" or uninterp_const or something like this for user-defined functions.

src/tactic/sine_filter.cpp
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  typedef std::pair<expr*,expr*> t_work_item;
+
+  t_work_item work_item(expr *e, expr *root) {
+    return std::pair<expr*, expr*>(e, root);
+  }
+
+  void find_constants(expr *e, obj_hashtable<func_decl> &consts) {
```
NikolajBjorner: to be uniform with the rest: 2 => 4 spaces

src/tactic/sine_filter.cpp
```diff
@@ -0,0 +1,252 @@
...
+    ptr_vector<expr> stack;
+    stack.push_back(e);
+    expr *curr;
+    while (!stack.empty()) {
+      curr = stack.back();
+      stack.pop_back();
+      if (is_app(curr)) {
+        app *a = to_app(curr);
+        func_decl *f = a->get_decl();
+        if (is_name(f) && !consts.contains(f)) {
```
NikolajBjorner: const.contains(f) seems redundant. Insertion should do this check already.

src/tactic/sine_filter.cpp
```diff
@@ -0,0 +1,252 @@
...
+    for (obj_hashtable<func_decl>::iterator constit = consts.begin(), constend = consts.end(); constit != constend; constit++) {
+      TRACE("sine", tout << *constit; tout << "\n";);
+    }
+    bool matched = false;
+    for (int i = 0; i < q->get_num_patterns(); i++) {
+      bool p_matched = true;
+      vector<expr *> stack;
+      expr *curr;
+      // patterns are wrapped with "pattern"
+      stack.push_back(to_app(q->get_pattern(i))->get_arg(0));
```
NikolajBjorner: we should add a function that wraps this to_app(q->get_pattern(i))->get_arg(0) and checks that indeed there is exactly one argument and that the function is "pattern". Since patterns are in the ast.h, the function would be a matcher on patterns. ast_manager has a method called "bool is_pattern(expr* e)". Patterns can be multi-patterns, so take more than 1 argument. The relevant matcher could then be bool is_pattern(expr* e, ptr_vector<expr>& args);.
You currently assume single patterns. Multi-patterns require matches on all argument terms (the set of patterns associated with a quantifier, on the other hand, are disjunctive: a match succeeds if one of the alternative patterns is matched).

src/tactic/sine_filter.cpp
```diff
@@ -0,0 +1,252 @@
...
+      while (!stack.empty()) {
+        curr = stack.back();
+        stack.pop_back();
+        if (is_app(curr)) {
+          app *a = to_app(curr);
+          func_decl *f = a->get_decl();
+          if (!consts.contains(f)) {
+            TRACE("sine", tout << f; tout << "\n";);
```
NikolajBjorner: tout << mk_pp(f, m) << "\n"
where m is an ast_manager.

msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

NikolajBjorner:
looks fine to me

dwoos:
Cool! Performance is still bad though, so I'm looking at that now. If you have time at some point, would you mind taking another look and seeing if anything jumps out at you as a performance issue? Thanks!

wintersteiger:
This looks great! It shouldn't interfere with anything else as it's all kept local to it's own tactic. I'm happy to merge this, even if you're still working on improving the performance. 

Regard performance, I just noticed that you're using your own stacks to run over expressions, which, if there are many shared subterms, can be slow. We have various strategies for doing that kind of task, for instance `for_each_expr` + `expr_mark` (see e.g. [here](https://github.com/Z3Prover/z3/blob/master/src/smt/smt_solver.cpp#L253)).

dwoos:
@wintersteiger I looked into doing that for the main body but I couldn't see a way of recursing conditionally, e.g., I only want to process the body of a quantifier if its pattern could actually match (and I want to go back and process it if I find a constant that causes it to match). Apologies if I'm missing something obvious, but is there a way to do this?

It does look like it should be possible to do `quantifier_matches` this way, though.

PR #885, opened 2017-01-30
Fix off-by-one bug in array indexing in the OCaml bindings

martin-neuhaeusser:
This patch fixes an off-by-one bug that occurred in the construction of output arrays
in the OCaml bindings.


wintersteiger:
That's great, thanks for the quick fix!

PR #907, opened 2017-02-22
theory_str for Z3 based on the Z3str2 framework

mtrberzi:
This PR adds a new theory solver, `theory_str`, for quantifier-free strings, integer length, and regular expression membership. The solver algorithm is based on Z3str2, which previously only existed as a wrapper script around Z3 without full integration with the Z3 core.

The input language is based on the string SMT draft standard and should be compatible with what is currently accepted by `theory_seq`.

In addition, this adds bindings for the Z3 C API to `theory_str`. If I need to do any additional work on adding bindings for different languages, I can look at that as well (I can't remember how much of the Python bindings are automatically generated, for instance).

I've tested this code on Windows 8 (compiled with Visual Studio 2015) and Debian Linux (compiled with g++ 4.9.2). I've evaluated the performance and stability of the solver on the Kaluza, AppScan, and PISA benchmarks, along with some small handcrafted regression cases (which can be provided).

Here are some points we may want to address before this is merged:
- How will this interact with the existing `theory_seq`? The approach I took for testing is to replace invocations of `theory_seq` with `theory_str` almost everywhere
- There are some changes to `smt_context` which might need more attention as they modify the core solver directly. The features which use these modifications aren't enabled by default but can provide better performance. 

I'm available for code review and feedback over the next few weeks, so if you have any questions or suggestions I'm happy to take a look right away.


src/util/obj_pair_set.h
```diff
@@ -46,6 +46,11 @@ class obj_pair_set {
     bool contains(obj_pair const & p) const { return m_set.contains(p); }
     void reset() { m_set.reset(); }
     bool empty() const { return m_set.empty(); }
+
```
NikolajBjorner: this would be easy to add as stand-alone PR

src/smt/smt_context.h
```diff
@@ -227,6 +227,17 @@ namespace smt {
 
         // -----------------------------------
         //
+        // Theory case split
+        //
+        // -----------------------------------
+        typedef int_hashtable<int_hash, default_eq<int> > int_set;
```
NikolajBjorner: we mainly use uint_set for sets over dense domains, such as literals. Hash-tables may have less memory overhead, but tend to not perform as well. So hash-tables make more sense if we have to maintain a large number of them simultaneously.

src/smt/smt_context.cpp
```diff
@@ -2392,6 +2403,9 @@ namespace smt {
     */
     unsigned context::pop_scope_core(unsigned num_scopes) {
 
+        TRACE("t_str_refcount_hack", tout << "begin pop_scope_core in smt_context" << std::endl;);
```
NikolajBjorner: These trace statements can be deprecated. They clutter the code going forward.

src/smt/smt_case_split_queue.cpp
```diff
@@ -1086,33 +1111,275 @@ namespace smt {
...
             p.m_case_split_strategy = CS_ACTIVITY;
         }
         if (p.m_auto_config && (p.m_case_split_strategy == CS_RELEVANCY || p.m_case_split_strategy == CS_RELEVANCY_ACTIVITY || 
-                                p.m_case_split_strategy == CS_RELEVANCY_GOAL)) {
+                p.m_case_split_strategy == CS_RELEVANCY_GOAL)) {
             warning_msg("auto configuration (option AUTO_CONFIG) must be disabled to use option CASE_SPLIT=3, 4 or 5");
             p.m_case_split_strategy = CS_ACTIVITY;
         }
+
+        case_split_queue * baseQueue;
```
NikolajBjorner: what is the purpose of this variable?

src/smt/smt_case_split_queue.cpp
```diff
@@ -1086,33 +1111,275 @@ namespace smt {
                 m_params.m_qi_eager_threshold += start_gen;
         }
     };
-   
+
+  /*
```
NikolajBjorner: delete commented out code

src/smt/smt_case_split_queue.cpp
```diff
@@ -35,6 +39,27 @@ namespace smt {
...
+    struct theory_aware_act_lt {
+        svector<double> const & m_activity;
+        theory_var_priority_map const & m_theory_var_priority;
+        theory_aware_act_lt(svector<double> const & act, theory_var_priority_map const & a):m_activity(act),m_theory_var_priority(a) {}
+        bool operator()(bool_var v1, bool_var v2) const {
+            double p_v1, p_v2;
+            if (!m_theory_var_priority.find(v1, p_v1)) {
+  	        p_v1 = 0.0;
+	    }
+            if (!m_theory_var_priority.find(v2, p_v2)) {
```
NikolajBjorner: indent

src/smt/params/theory_str_params.cpp
```diff
@@ -0,0 +1,34 @@
...
+Revision History:
+
+--*/
+
+#include"theory_str_params.h"
+#include"smt_params_helper.hpp"
+
+void theory_str_params::updt_params(params_ref const & _p) {
+    smt_params_helper p(_p);
+    m_AssertStrongerArrangements = p.str_strong_arrangements();
```
NikolajBjorner: aligning external with internal names makes it easier to track things

src/parsers/smt2/smt2parser.cpp
```diff
@@ -1077,10 +1086,29 @@ namespace smt2 {
...
+      
+        void parse_string() {
+            SASSERT(curr() == scanner::STRING_TOKEN);
+            char const *original_token = m_scanner.get_string();
+            size_t bufsize = strlen(original_token);
+            char * buf = alloc_svect(char, bufsize + 1);
+            strncpy(buf, original_token, bufsize);
+            buf[bufsize] = '\0';
+            TRACE("parse_string", tout << "new string constant: " << buf << " length=" << bufsize << "\n";);
+            expr_stack().push_back(strutil().mk_string_with_escape_characters(buf));
```
NikolajBjorner: This ownership model is prone to leaks and cluttering the call site.

src/ast/seq_decl_plugin.cpp
```diff
@@ -259,7 +259,7 @@ std::ostream& zstring::operator<<(std::ostream& out) const {
 
 
 seq_decl_plugin::seq_decl_plugin(): m_init(false),
-                                    m_stringc_sym("String"),
+                                    m_stringc_sym("StringSequence"),
```
NikolajBjorner: we will have to address this

src/smt/smt_case_split_queue.cpp
mtrberzi: This was a holdover from when I had a slightly different approach to maintaining a custom case split queue. I'll refactor it out, as it's no longer needed.

src/ast/str_decl_plugin.cpp
```diff
@@ -0,0 +1,496 @@
...
+    op_names.push_back(builtin_name("str.suffixof", OP_STR_SUFFIXOF));
+    op_names.push_back(builtin_name("str.to-int", OP_STR_STR2INT));
+    op_names.push_back(builtin_name("str.from-int", OP_STR_INT2STR));
+    op_names.push_back(builtin_name("str.to.re", OP_RE_STR2REGEX));
+    op_names.push_back(builtin_name("str.in.re", OP_RE_REGEXIN));
+    op_names.push_back(builtin_name("re.++", OP_RE_REGEXCONCAT));
+    op_names.push_back(builtin_name("re.*", OP_RE_REGEXSTAR));
+    op_names.push_back(builtin_name("re.union", OP_RE_REGEXUNION));
+    op_names.push_back(builtin_name("re.+", OP_RE_REGEXPLUS));
+    op_names.push_back(builtin_name("Unroll", OP_RE_UNROLL));
```
NikolajBjorner: if this is internal, then don't need to expose it here

src/ast/ast.cpp
```diff
@@ -58,6 +59,7 @@ parameter& parameter::operator=(parameter const& other) {
     case PARAM_SYMBOL: new (m_symbol) symbol(other.get_symbol()); break;
     case PARAM_RATIONAL: new (m_rational) rational(other.get_rational()); break;
     case PARAM_DOUBLE: m_dval = other.m_dval; break;
+    case PARAM_STRING: m_string = other.m_string; break;
```
NikolajBjorner: why a separate category of PARAM_STRING? Symbols should be able to encode strings.

mtrberzi:
synchronizing with the octal escape merge, nothing new in this commit

mtrberzi:
New commits have been added; this updates to the `seq_decl_plugin` frontend, removes `str_decl_plugin` and `str_rewriter`, and cleans up the code in a few other places.

src/api/api_context.h
```diff
@@ -82,6 +82,7 @@ namespace api {
         family_id                  m_pb_fid;
         family_id                  m_fpa_fid;
         family_id                  m_seq_fid;
+        family_id                  m_str_fid;
```
NikolajBjorner: do you still need this?

src/api/api_context.h
```diff
@@ -135,6 +136,7 @@ namespace api {
         family_id get_pb_fid() const { return m_pb_fid; }
         family_id get_fpa_fid() const { return m_fpa_fid; }
         family_id get_seq_fid() const { return m_seq_fid; }
+        family_id get_str_fid() const { return m_str_fid; }
```
NikolajBjorner: do you still need this?

src/api/z3_api.h
```diff
@@ -1169,6 +1169,10 @@ typedef enum {
     Z3_OP_RE_FULL_SET,
     Z3_OP_RE_COMPLEMENT,
 
+	// theory_str
```
NikolajBjorner: Z3_OP_SEQ_CONCAT and Z3_OP_SEQ_LENGTH should work?

src/ast/ast.cpp
```diff
@@ -93,6 +95,7 @@ bool parameter::operator==(parameter const & p) const {
     case PARAM_SYMBOL: return get_symbol() == p.get_symbol();
     case PARAM_RATIONAL: return get_rational() == p.get_rational();
     case PARAM_DOUBLE: return m_dval == p.m_dval;
+    case PARAM_STRING: return (m_string == NULL && p.m_string == NULL) || strcmp(m_string, p.m_string)==0;
```
NikolajBjorner: can use PARAM_SYMBOL

NikolajBjorner:
not sure if git is playing tricks with me, but I see some references to str in the family/theory ids.

mtrberzi:
I've resolved your notes on `PARAM_SYMBOL` and the old API methods. Currently testing and I should have the new commits up here tomorrow. After that I will take a look at the overlap detection in unsat core idea.

NikolajBjorner:
OK, looking forward.

mtrberzi:
I added some new commits that implement an interface for unsat core assumptions and validation. Take a look and let me know your thoughts.

mtrberzi:
This merges the unsat core check from earlier, and adds static feature check for str vs. seq. It also unifies tracing in `theory_str.cpp` (the only trace tag is "str") now.

Testing this on some benchmarks right now, but I think we are almost ready to go.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -329,7 +329,13 @@ br_status seq_rewriter::mk_app_core(func_decl * f, unsigned num_args, expr * con
     switch(f->get_decl_kind()) {
 
     case OP_SEQ_UNIT:
-        return BR_FAILED;
+        // TODO configuration param
+        if (true) {
+            SASSERT(num_args == 1);
+            return mk_seq_unit(args[0], result);
```
NikolajBjorner: This does not need to be a configuration parameter. Just return mk_seq_unit(args[0], result)

src/ast/seq_decl_plugin.h
```diff
@@ -273,6 +273,21 @@ class seq_util {
         bool is_in_re(expr const* n)    const { return is_app_of(n, m_fid, OP_SEQ_IN_RE); }
         bool is_unit(expr const* n)     const { return is_app_of(n, m_fid, OP_SEQ_UNIT); }
 
+        bool is_string_term(expr const * n) const {
```
NikolajBjorner: not sure how _STRING_SORT gets into play. The seq_decl_plugin converts this to a sequence over bit-vectors.

NikolajBjorner:
will be somewhat easier and faster to merge if you create smaller pull requests based on units.
1. For seq_rewriter + seq_decl_plugin, there is a clear independent pull request.
2. Some references to str_family_id. I believe these are the same as seq_family_id.
3. static features, cmd_context and perhaps a few more files can be a separate standalone pull too.
4. smt_case_split queue seems stand-alone, or should be.
5. theory_arith logging
6. smt_setup
7. and finally theory_str

src/ast/seq_decl_plugin.h
mtrberzi: Oh, oops, this is a no-op. Surprised it worked without this, but this should be something like `is_seq(s) && is_string(s)`.

mtrberzi:
Point 1 on the list opened #988

mtrberzi:
Should be no more references to str_fid now. I'll start working out the static features / cmd_context parts next.

mtrberzi:
Point 3 opened #992

mtrberzi:
Point 4 part 1 opened as #995 (the `smt_case_split_queue` is actually another heuristic which will be a separate PR)

mtrberzi:
Point 4 part 2 opened as #999 (this is the `smt_case_split_queue` patch)

mtrberzi:
Added some commits that remove the logging in theory_arith. Also updated #999

mtrberzi:
Opened point 6 as #1000

src/util/obj_pair_set.h
mtrberzi: Opened as #1002

mtrberzi:
Cleaned up the PR a bit and merged with master.

src/smt/smt_context.cpp
```diff
@@ -37,6 +37,7 @@ Revision History:
 #include"model_pp.h"
 #include"ast_smt2_pp.h"
 #include"ast_translation.h"
+#include"theory_seq.h"
```
NikolajBjorner: should be redundant

src/smt/params/smt_params.cpp
```diff
@@ -32,6 +32,7 @@ void smt_params::updt_local_params(params_ref const & _p) {
     m_restart_factor = p.restart_factor();
     m_case_split_strategy = static_cast<case_split_strategy>(p.case_split());
     m_theory_case_split = p.theory_case_split();
+    m_theory_aware_branching = p.theory_aware_branching();
```
NikolajBjorner: Isn't this is moved to case_split_strategy

NikolajBjorner:
there are rather few differences now. Seems essentially ready to merge.

mtrberzi:
Sounds great. I'll address these small changes later today. There is one other thing I need to do: in a commit earlier today, rewriting of `re.range` was disabled, which breaks theory_str (we don't handle `re.range` terms directly, instead expecting that they will always be rewritten to `re.union`). Unless we can enable these rewrites only when theory_str is being used, I will need to add support for handling these terms directly.

src/smt/params/smt_params.cpp
mtrberzi: This should have been in #999, not sure how it was missed.

src/smt/smt_context.cpp
mtrberzi: It is. Removed

NikolajBjorner:
the expansion of range constraints is not going to work well. A regression reported in issue #998 uses many range constraints and expanding these comes with significant overhead.

mtrberzi:
I think I can handle it natively in the solver.

mtrberzi:
Works fine with native support. Testing now and should be ready to go shortly.

mtrberzi:
Everything looks good on my end, OK to merge when ready

PR #923, opened 2017-02-28
Fixed bug in `mk_make.py --build=`...

mlr-msft:
Fixed bug where `mk_make.py --build=...` would fail to handle absolute paths correctly.


wintersteiger:
This looks good, thanks for the fixes!

PR #924, opened 2017-03-01
Free allocated char arrays in JNI API

### LINKED ISSUE #886: Java api memory leak?
I was looking at the JNI code for the java api (in Native.cpp) and noticed a number of calls to `GetStringUTFChars` without corresponding calls to `ReleaseStringUTFChars`. At best that will result in a memory leak. I haven't tracked down the use of the resulting allocated array, but there is no guarantee that the JVM returns a piece of memory obtained from `malloc`, thus you cannot safely free the memory. Furthermore, if the JVM is running with native memory tracking turned on, then the memory allocated by `GetStringUTFChars` will not be safe to pass to `free`.

cheshire:
Fixes #886 

Tested using JavaSMT test suite.


wintersteiger:
That's great, thanks for the fix!

PR #925, opened 2017-03-02
[CMake] Validate CMAKE_BUILD_TYPE

delcypher:
For single configuration generators only allow `CMAKE_BUILD_TYPE` to be one of the pre-defined build configurations that we support.


wintersteiger:
Thanks!

Issue #915, opened 2017-02-25
cmake Z3Config not installed


jirislaby:
@delcypher z3 cannot be found by find_package in cmake currently. This is at least used by https://github.com/hoangmle/metaSMT/blob/qf_abv/CMakeLists.txt#L96 at least. You have to do this hack in klee due to that for example:
https://github.com/klee/klee/blob/master/.travis/metaSMT.sh#L24

Could you prepare Z3Config.cmake based on something like this:
https://github.com/agra-uni-bremen/dependencies/blob/development/Z3-2.19/Z3Config.cmake
and let it install?

delcypher:
@jirislaby This is something I intend to do when I have free time to work on Z3's CMake build system.

There is a `FIXME` about it in the existing code.

https://github.com/Z3Prover/z3/blob/7d37203653cffaab97d5cc75ca99cf0a6d662727/contrib/cmake/examples/c/CMakeLists.txt#L1

PR #926, opened 2017-03-04
[CMake] Z3 Package config files

delcypher:
This implements a Z3 CMake config package so that users of CMake can easily use `libz3` in their own projects. This was requested by #915.

The `Z3Config.cmake` and `Z3Targets.cmake` files are generated in the root of the build tree and are also installed during installation. This allows projects to use a build of Z3 directly from the build tree or from an install of Z3.

To demonstrate this new feature the `c_example`, `cpp_example` and `z3_tptp5` examples have been given their own standalone build systems in which they use `libz3` via this new mechanism.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,11 +294,11 @@ if (OPENMP_FOUND)
     # flag by MSVC and breaks the build
     if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang") OR
         ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU"))
-      list(APPEND Z3_DEPENDENT_EXTRA_CXX_LINK_FLAGS ${OpenMP_C_FLAGS})
+      list(APPEND Z3_DEPENDENT_EXTRA_CXX_LINK_FLAGS ${OpenMP_CXX_FLAGS})
     endif()
     if (("${CMAKE_C_COMPILER_ID}" MATCHES "Clang") OR
         ("${CMAKE_C_COMPILER_ID}" MATCHES "GNU"))
-      list(APPEND Z3_DEPENDENT_EXTRA_C_LINK_FLAGS ${OpenMP_CXX_FLAGS})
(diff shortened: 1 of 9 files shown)


NikolajBjorner:
Thanks
If I understand more changes are underway so we shouldn't merge it just yet.

delcypher:
@jirislaby Please take a look and let me know if this meets your needs.

There are a currently a few deficiencies that I am not happy with.

* The `Z3Config.cmake` file does not state where the `z3` binary is. This is slightly complicated due to the fact that in a multi configuration generator (e.g. Visual Studio) there may be multiple builds of `z3`. One idea I have for handling this to export the `shell` target (this is Z3's internal name for the `z3` executable) as `z3::shell` and then require the client of `Z3Config.cmake` to inspect the exported target's properties to determine the location of the binary. I haven't done this so far because I really didn't want to leak the `shell` target to the outside world unless I have to.

* The example projects (and any other project) will fail to build if `libz3` is built as a static library and we use OpenMP. This is because we don't export the requirement that `libz3.a` needs OpenMP. I need to have a think about how to fix this. In the past I found the problem was that doing `target_link_libraries(libz3 INTERFACE ${OPENMP_C_FLAGS})` broke the MSVC build because it interpreted `/openmp` as a file to link against... I need to do some testing on Windows and see if I can come up with something that works.

* I've not created a `Z3Version.cmake` file. This file would be used if the caller of `find_package(Z3 CONFIG)` specified a version. I'm not sure if it's worth going to the effort of doing this.

delcypher:
@NikolajBjorner Yes more changes are needed to the pull request. I need feedback from @jirislaby, I need to fix the static library build and also do some testing with MSVC.

delcypher:
@NikolajBjorner I've done more testing (on Windows) and I've made a few minor improvements so the example projects work out of the box. I've not heard from @jirislaby so I think this should be merged.

There are some missing features (as noted above) but those can be added in the future if it turns out someone actually needs them.

jirislaby:
I am only slow, it works as expected. Thanks.

delcypher:
@jirislaby Thanks. Could you please close #915 ?

Issue #939, opened 2017-03-14
Double free when solving QF_BV formulas


rainoftime:
We are using z3 4.5.1 and met an issue the same as https://github.com/Z3Prover/z3/issues/798, with the following stacktrace:

```
*** Error in a.out: double free or corruption (out): 0x00007f4d3618baf0 ***

#0 0xbc3eb8 llvm::sys::PrintStackTrace(_IO_FILE*) (/home/qingkaishi/pinpoint/Release+Asserts/bin/pp-check+0xbc3eb8)
#1 0xbc53ab SignalHandler(int) (/home/qingkaishi/pinpoint/Release+Asserts/bin/pp-check+0xbc53ab)
#2 0x7f52fd5cd3e0 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x113e0)
#3 0x7f52fb927428 gsignal /build/glibc-Qz8a69/glibc-2.23/signal/../sysdeps/unix/sysv/linux/raise.c:54:0
#4 0x7f52fb92902a abort /build/glibc-Qz8a69/glibc-2.23/stdlib/abort.c:91:0
#5 0x7f52fb9697ea /build/glibc-Qz8a69/glibc-2.23/libio/../sysdeps/posix/libc_fatal.c:175:0
#6 0x7f52fb971e0a _int_free /build/glibc-Qz8a69/glibc-2.23/malloc/malloc.c:4103:0
#7 0x7f52fb97598c cfree /build/glibc-Qz8a69/glibc-2.23/malloc/malloc.c:2967:0
#8 0x7f52fcb917ce memory::deallocate(void*) (/usr/lib/libz3.so+0xed67ce)
#9 0x7f52fcad7092 sat::clause_allocator::del_clause(sat::clause*) (/usr/lib/libz3.so+0xe1c092)
#10 0x7f52fcabe533 sat::solver::~solver() (/usr/lib/libz3.so+0xe03533)
#11 0x7f52fc752a3c sat_tactic::imp::~imp() (/usr/lib/libz3.so+0xa97a3c)
#12 0x7f52fc752d55 sat_tactic::operator()(ref<goal> const&, sref_buffer<goal, 16u>&, ref<model_converter>&, ref<proof_converter>&, obj_ref<dependency_manager<ast_manager::expr_dependency_config>::dependency, ast_manager>&) (/usr/lib/libz3.so+0xa97d55)
[...]

PR #940, opened 2017-03-14
z3 double free primary fix

rainoftime:


NikolajBjorner:
They can be tricky to fix. The root causes are that cancellation in the middle of simplification or another process may leave pointers in a state that cause a double free. Often I can address this by ensuring that the code is exception safe. It is useful to know which place in the code the cancellation was thrown.

rainoftime:
We run z3 in multithread context(each thread has its own context of z3).  

I thought the cancellation was thrown in `sat_simplifier.cpp`, and with my primary fix `iss939fix`, z3 has not crashed after several experiments. 

However, when I tried to delete all `checkpoint()` in `sat_simplifier.cpp`, z3 crashed again. Maybe functions in `sat_simplifier.cpp` call `propatate_core()` or other functions that have their own `checkpoint()`?

NikolajBjorner:
Correct, line 620 in sat_simplifier.cpp, for instance.

rainoftime:
It is hard to give a concrete trace that trigers the double free. Our system crashed only after several hours or running.  
We cannot just delete `checkpoint()` in `propagate_core()`(assume it's the source of the issue),  since it is called much more often during SAT solving.

NikolajBjorner:
Some inspection may reveal places where cancellation causes double frees.
Inside cleanup_clauses, for instance we have something like this:

            if (sz == 0) {
                s.set_conflict(justification());
                return;
            }

It should more likely be:

            if (sz == 0) {
                s.set_conflict(justification());
                for (; it != end; ++it) {
                    *it2 = *it;
                    ++it2;
                }
                break;
            }

in this way de-allocated clauses are removed from the lists, while non-deallocated clauses are retained.

NikolajBjorner:
Also, asym_branch::process calls propagate_core, which can throw and this could easily lead to double frees. This is the more likely cause of your crash.

rainoftime:
Thanks! I will try your new commit https://github.com/Z3Prover/z3/commit/72651e2e98b38a8eec25acffa41f1b5470a76fa8

rainoftime:
@NikolajBjorner We have tried https://github.com/Z3Prover/z3/commit/72651e2e98b38a8eec25acffa41f1b5470a76fa8
but z3 still crashed in all three experiments. with the same information put in https://github.com/Z3Prover/z3/issues/939

rainoftime:
Perhaps I could add some extra exception handling code at some suspicious sources? E.g., the `probing::operator()` in `sat_probing.cpp` calls `process()` that may also call `propagate_core()` (with `try_lit()` function)

rainoftime:
Perhaps I could add some extra exception handling code at some suspicious sources? E.g., the `probing::operator()` in `sat_probing.cpp` calls `process()` that may also call `propagate_core()` (with `try_lit()` function)

NikolajBjorner:
Yes, sorry, it may be easier to just create a version of the code with heavy instrumentation to pinpoint the places it throws. Something as crude as the following could help localize the troublemaker.
        
        try {
             < block of code >
        }
        catch (solver_exception & ex) {
            std::cout << __FILE__ << " " << __LINE << "\n";
            throw ex;
        }

There are some more sophisticated ways: logging the exception stacks in the exception handlers, but 
I last I looked into this it was highly platform dependent for C++ and also really obscure.

rainoftime:
I have added extra `try..catch` to `sat_asym_branch.cpp`
~~~~
    bool asymm_branch::process(clause & c) {
            ......
      
        // try asymmetric branching
        // clause must not be used for propagation
        s.dettach_clause(c);
        s.push();
  try {
        for (i = 0; i < sz - 1; i++) {
            literal l = c[i];
            SASSERT(!s.inconsistent());
            TRACE("asymm_branch_detail", tout << "assigning: " << ~l << "\n";);
            s.assign(~l, justification());
            s.propagate_core(false); // must not use propagate(), since check_missed_propagation may fail for c
            if (s.inconsistent())
                break;
        }
     } catch (solver_exception & ex) {
             std::cout << "Timeout out at " << __FILE__ << __LINE << "\n";
             throw ex;
     }
        s.pop(1);
         
            ......
        new_sz = j;
        m_elim_literals += sz - new_sz;
  try {
        switch(new_sz) {
        case 0:
            s.set_conflict(justification());
            return false;
        case 1:
            TRACE("asymm_branch", tout << "produced unit clause: " << c[0] << "\n";);
            s.assign(c[0], justification());
            s.del_clause(c);
            s.propagate_core(false); 
            SASSERT(s.inconsistent() || s.m_qhead == s.m_trail.size());
            return false; // check_missed_propagation() may fail, since m_clauses is not in a consistent state.
                      ......
        }
      } catch (solver_exception & ex) {
             std::cout << "Timeout at " << __FILE__ << __LINE__ << "\n";
             throw ex;
      }
    }
~~~~

It seems that the double free happens when z3 times out at the first loop (about line 142)
~~~~
        for (i = 0; i < sz - 1; i++) {
            literal l = c[i];
            SASSERT(!s.inconsistent());
            TRACE("asymm_branch_detail", tout << "assigning: " << ~l << "\n";);
            s.assign(~l, justification());
            s.propagate_core(false); // must not use propagate(), since check_missed_propagation may fail for c
            if (s.inconsistent())
                break;
        }
~~~~
But I guess there is no clause that will be deleted during the loop?

Issue #861, opened 2017-01-05
cplusplus no exception support


humeafo:
in some projects like llvm,  no exception is required, even if I just want to link dynamicly, so is it possible for z3 to support no exception enabled usage case?
of course I can use c interface, but it's a bit hard to use than c++ interface, so at least no throws in z3++.h?

NikolajBjorner:
reasonable. Then user is required to check the error.

NikolajBjorner:
Will this do?

NikolajBjorner:
pull request?

humeafo:
it's my fault, I just mean this:

int get_numeral_int() const {             
        int result = 0;
        if (!is_numeral_i(result) && ctx().enable_exceptions()) {
            throw exception("numeral does not fit in machine int");
        }
        return result;
    }

maybe shoud be changed to:

int get_numeral_int() const {             
        int result = 0;
        if (!is_numeral_i(result) && ctx().enable_exceptions()) {
#ifndef CXX_NOTHROWS
            throw exception("numeral does not fit in machine int");
#endif
        }
        return result;
    }

or:
#ifndef CXX_NOTHROWS
#define THROW_EXCEPTION(msg) throw exception("numeral does not fit in machine int");
#else 
#define THROW_EXCEPTION(msg)
#endif

int get_numeral_int() const {             
        int result = 0;
        if (!is_numeral_i(result) && ctx().enable_exceptions()) {
           THROW_EXCEPTION("numeral does not fit in machine int");
        }
        return result;
    }

and this should be applied to entire z3++.h except check_error()...

wintersteiger:
Could you give us an example of a program that LLVM rejects? I can't imagine that it has a mode that allows C++ classes but not exceptions.

humeafo:
to wintersteiger :

please see llvm documents for why llvm not using exceptions:
http://llvm.org/docs/CodingStandards.html#do-not-use-rtti-or-exceptions

or this question:
http://stackoverflow.com/questions/4080601/not-using-c-exceptions-by-design-in-llvm-clang

you can also see llvm compilation scripts:
http://llvm.org/svn/llvm-project/llvm/trunk/cmake/modules/AddLLVM.cmake line 23:

list(APPEND LLVM_COMPILE_FLAGS "-fno-exceptions")

with -fno-exceptions on commandline:

c++ codes that use throw, try, catch won't compile under gcc clang

[...]

PR #941, opened 2017-03-14
added no exception support to z3++.h

humeafo:
added no exception support to z3++.h in case where no exception is expected or required(such as in llvm sub-projects), this had been discussed in https://github.com/Z3Prover/z3/issues/861, but not implemented yet.


NikolajBjorner:
hmm, the default behavior I get is that exceptions are disabled.
This is a bit unfortunate because it seems we are now getting reports based on users not getting exceptions.

wintersteiger:
Yes, exceptions must definitely be enabled by default! Are `__cpp_exceptions` and `_CPPUNWIND` defined by all compilers?

PR #942, opened 2017-03-14
alternate str.extract semantics in seq_rewriter

mtrberzi:
I've added new edge case semantics again in `seq_rewriter`, this time in `str.extract`. This patch allows the following rewrites to be made:

- constant negative starting position or length: rewrite to ""
- constant base string and starting position, and `pos` is beyond the end of the string: rewrite to ""
- `pos+len` is beyond the end of the string: take as many characters as possible up to the end of the string

This is necessary to make `str.extract` total in the rewriter, for model validation. This is consistent with the behaviour of the C++ `substr` method on `std::string`.

### DIFF
--- a/src/ast/rewriter/seq_rewriter.cpp
+++ b/src/ast/rewriter/seq_rewriter.cpp
@@ -517,13 +517,13 @@ br_status seq_rewriter::mk_seq_extract(expr* a, expr* b, expr* c, expr_ref& resu
     // case 1: pos<0 or len<0
     // rewrite to ""
     if ( (constantPos && pos.is_neg()) || (constantLen && len.is_neg()) ) {
-        result = m_util.str.mk_string(symbol(""));
+        result = m_util.str.mk_empty(m().get_sort(a));
         return BR_DONE;
     }
     // case 1.1: pos >= length(base)
     // rewrite to ""
     if (constantBase && constantPos && pos.get_unsigned() >= s.length()) {
-        result = m_util.str.mk_string(symbol(""));
+        result = m_util.str.mk_empty(m().get_sort(a));
         return BR_DONE;
     }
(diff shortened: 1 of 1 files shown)


src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -509,15 +509,40 @@ br_status seq_rewriter::mk_seq_length(expr* a, expr_ref& result) {
...
-        result = m_util.str.mk_string(s.extract(_pos, _len));
+
+    bool constantBase = m_util.str.is_string(a, s);
+    bool constantPos = m_autil.is_numeral(b, pos);
+    bool constantLen = m_autil.is_numeral(c, len);
+
+    // case 1: pos<0 or len<0
+    // rewrite to ""
+    if ( (constantPos && pos.is_neg()) || (constantLen && len.is_neg()) ) {
+        result = m_util.str.mk_string(symbol(""));
```
NikolajBjorner: This should be m_util.str.mk_empty(m.get_sort(a))

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -509,15 +509,40 @@ br_status seq_rewriter::mk_seq_length(expr* a, expr_ref& result) {
...
+        return BR_DONE;
+    }
+    // case 1.1: pos >= length(base)
+    // rewrite to ""
+    if (constantBase && constantPos && pos.get_unsigned() >= s.length()) {
+        result = m_util.str.mk_string(symbol(""));
```
NikolajBjorner: same here. The type of 'a' need not be a string.

mtrberzi:
Good point -- last commit should make this properly type-generic.

PR #908, opened 2017-02-22
Add support for MSYS2

Uni-:
With this patch, Z3 is compiled via MSYS shell and toolchain of MSYS2. (MINGW32 shell and MINGW64 shell is not yet cared.)

PR #969, opened 2017-04-06
Fixed platform detection on MSYS (was #908)

wintersteiger:
This is a replacement for the broken PR #908.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -613,6 +613,8 @@ elif os.name == 'posix':
         IS_CYGWIN=True
         if (CC != None and "mingw" in CC):
             IS_CYGWIN_MINGW=True
+    elif os.uname()[0][:7] == 'MSYS_NT':
+        IS_CYGWIN_MINGW=True
 
 def display_help(exit_code):
     print("mk_make.py: Z3 Makefile generator\n")


wintersteiger:
@Uni- Could you try this change to see whether Z3 compiles on your MSYS system with this?

Uni-:
@wintersteiger, still same error as seen in #908. 

````
  File "scripts/mk_make.py", line 21, in <module>
    mk_makefile()
  File "/home/Eon/Development/z3/scripts/mk_util.py", line 2576, in mk_makefile
    mk_config()
  File "/home/Eon/Development/z3/scripts/mk_util.py", line 2475, in mk_config
    raise MKException('Unsupported platform: %s' % sysname)
mk_exception.MKException: 'Unsupported platform: MSYS_NT-10.0'
````

wintersteiger:
What kind of Python do you use? Did you compile your own or is that in the MSYS package management repository somewhere?

wintersteiger:
Bumping. Is this still a problem or has it been resolved?

Uni-:
@wintersteiger, sorry for coming late. I use python2, python3 packages as-is delivered from MSYS Pacman. /usr/bin/python.exe is a copy of /usr/bin/python3.exe, as the packages default.

And still the same exception.

Thank you for your efforts, but I think you should see what's happening on the point build error occurs, [scripts/mk_util.py:2475 (at your commit 514f08f)](https://github.com/Z3Prover/z3/blob/514f08f/scripts/mk_util.py#L2475). The if-else around this line does not use is_cygwin_mingw(), but the raw uname. That's why at first I modified exactly that line in commit 387970f6 (initially suggested via #908).

Moreover, I think you commit 514f08f is going little weird. Please see [scripts/mk_util.py:614](https://github.com/Z3Prover/z3/blob/514f08f/scripts/mk_util.py#L614), [scripts/mk_util.py:2492](https://github.com/Z3Prover/z3/blob/514f08f/scripts/mk_util.py#L2492), [scripts/mk_util.py:2526](https://github.com/Z3Prover/z3/blob/514f08f/scripts/mk_util.py#L2526). The condition setting IS_CYGWIN_MINGW and usages of is_cygwin_mingw() supposes that is_cygwin_mingw() for the MinGW cross-compiling toolchain running on Cygwin, not for the Cygwin native compiling toolchain. Definitely native compiling on MSYS2 is more likely to compiling on Cygwin for itself, not cross-compiling.

I also looked around the whole appearance of '_CYGWIN' macro. As you said in #908 it changes the behavior of Z3, but many of them are unsignificant (like in [src/util/trace.h:23](https://github.com/Z3Prover/z3/blob/514f08f/src/util/trace.h#L23), and the rests are used with '_WINDOWS' which means the macro-defined sections should be used instead of other sections for OS X or Linux (because of the limits of Windows environment). Note that IS_CYGWIN always gets True before IS_CYGWIN_MINGW gets True ([scripts/mk_util.py:614](https://github.com/Z3Prover/z3/blob/514f08f/scripts/mk_util.py#L614) again), so the cross-compiled artifact of Z3 for Windows (cross-compiled in Cygwin) obviously has been defined '_CYGWIN' (because the condition making IS_CYGWIN True is same as the condition defining '_CYGWIN', see [scripts/mk_util.py:2469](https://github.com/Z3Prover/z3/blob/514f08f/scripts/mk_util.py#L2469)). In short, defining _CYGWIN in MSYS2 does not change the behavior of Z3 on Windows.

wintersteiger:
This is now fixed (https://github.com/Z3Prover/z3/commit/e50470f2c4d561e42989d74b1a49e606c8e84415). 

Whether the _CYGWIN macro actually, in it's current version, changes the behavior of Z3 is ultimately irrelevant. If -D _CYGWIN is on, we will rely on the platform actually being cygwin and not mingw. Anyways, MSYS2 is now recognized as it's own platform.

Uni-:
@wintersteiger Why did you decide shared object artifact to have `.so` extension? In MSYS2 it should be `.dll`.

wintersteiger:
I assumed MSYS uses .so and binaries with .exe, if that's not so we should of course change that. The install path can be set manually and I'd really like to avoid have a huge list of default install paths for every option on flavors of unpopular OSs...

wintersteiger:
Ah, I see, the .so/.dll gets put into bin instead of lib? Strange. May be related to the file extensions not matching up.

wintersteiger:
(Should work as expected now.)

PR #981, opened 2017-04-22
pre-init assumptions and unsat core validation for smt theories

mtrberzi:
This adds two hooks to `smt::theory` allowing the context to ask it for assumptions it wants to add to the search and allowing it to "validate" an unsat core in the theory solver. The theory solver can change the result of an unsat to unknown if it finds a problem with the unsat core (e.g. an assumption was used that makes the search actually incomplete).


mtrberzi:
one minute, I forgot something

mtrberzi:
okay, should be good to go. Forgot to add the core that actually uses the result from `mk_unsat_core`

PR #982, opened 2017-04-23
Prettifying license.

JulienPalard:
For readability, wrapping around 80th column is nice, but also I fixed the:

""Software"" to "Software".

I tried not to modify the license more than this, so I kept your
header almost unchanged:

```
Z3
Copyright (c) Microsoft Corporation
All rights reserved.

MIT License
```

I just removed a trailing whitespace at the end of `All rights reserved``.

Note that your header still not match the
https://opensource.org/licenses/MIT template, you're missing the year
in the Copyright notice. I personally don't care, but you should check
this with your legal team.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,7 +1,23 @@
 Z3
 Copyright (c) Microsoft Corporation
-All rights reserved. 
+All rights reserved.
+
 MIT License
-Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the ""Software""), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
-The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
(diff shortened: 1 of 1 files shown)


wintersteiger:
Looks fine to me, but I'm not sure we should merge this change. So far I assumed that the formatting should be the same as on [opensource.org](https://opensource.org/licenses/MIT), but I guess it doesn't make a big difference.

NikolajBjorner:
As far as I can see Julien's updates does get the text aligned with that on http://opensource.org. Maybe his crawler started barking when it hit this.
I don't know why we have double quotes around "Software". It reminds me of something like http://knowyourmeme.com/photos/594192-dr-evil-air-quotes.

JulienPalard:
What my parser disliked the most was the double-double-quote, looks like the only important thing to fix.

Second important thing, if you want to keep your license unwrapped like in [opensource.org](https://opensource.org/licenses/MIT) you should at least also keep blank lines between paragraphs, it helps with readability.

For the wrapping around ~80 column, that's just for readability, for humans: I just though "as I'm doing a pull requests for this, let's do something as clean as possible" (so I even removed ugly-invisible trailing whitespaces…).

NikolajBjorner:
I removed double quotes and added new lines per. Kept the wrap-around for now.

JulienPalard:
Works for me, take care!

PR #983, opened 2017-04-24
Return check result in fixedpoint object

Baltoli:
This method was missing a return statement - the value returned from `check_result` was discarded. The only change in this PR is to add the return statement, correcting the error.


NikolajBjorner:
Thanks!

PR #984, opened 2017-04-25
[CMake][Doxygen] Support building/installing API documentation and fix lots of bugs

delcypher:
This PR starts with fixing the `mk_api_doc.py` to be in a good enough and flexible enough that it can be used from CMake and then adds CMake support (81a7a55). The CMake support also installs documentation by default.

I have only added new behaviour to `mk_api_doc.py` so it should function as before when none of the new command line arguments are used.

I ended up finding quite a few bugs a long the way so I fixed those as I went (e.g. 5508bfee2ba1f6129a55974a06418d58e5dcfcc2, 896c82924d51c1dabc3ac2bee969509558dfd7ff , fd6602727f1691d7ec56899003b6b92b5b512786, c9a462d9bf32dd26ca0b523ab64f2ff0bae75003)


delcypher:
Just in case someone was reviewing this I've just force pushed a few fixes.

PR #988, opened 2017-04-27
Frontend changes for theory_str

mtrberzi:
This patch makes a small number of frontend changes (`seq_decl_plugin`, `seq_rewriter`) in preparation for the `theory_str` merge.

### DIFF
--- a/src/ast/seq_decl_plugin.cpp
+++ b/src/ast/seq_decl_plugin.cpp
@@ -573,7 +573,7 @@ void seq_decl_plugin::set_manager(ast_manager* m, family_id id) {
     m_char = bv.mk_sort(8);
     m->inc_ref(m_char);
     parameter param(m_char);
-    m_string = m->mk_sort(symbol("String"), sort_info(m_family_id, SEQ_SORT, 1, &param));
+    m_string = m->mk_sort(symbol("StringSequence"), sort_info(m_family_id, SEQ_SORT, 1, &param));
     m->inc_ref(m_string);
     parameter paramS(m_string);
     m_re = m->mk_sort(m_family_id, RE_SORT, 1, &paramS);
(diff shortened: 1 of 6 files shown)


src/ast/seq_decl_plugin.cpp
```diff
@@ -573,7 +573,7 @@ void seq_decl_plugin::set_manager(ast_manager* m, family_id id) {
     m_char = bv.mk_sort(8);
     m->inc_ref(m_char);
     parameter param(m_char);
-    m_string = m->mk_sort(symbol("String"), sort_info(m_family_id, SEQ_SORT, 1, &param));
+    m_string = m->mk_sort(symbol("StringSequence"), sort_info(m_family_id, SEQ_SORT, 1, &param));
```
NikolajBjorner: Seems it should be "String". We will end up printing these types. _STRING_SORT gets used by the parser front-end.

src/ast/seq_decl_plugin.h
```diff
@@ -273,6 +273,15 @@ class seq_util {
         bool is_in_re(expr const* n)    const { return is_app_of(n, m_fid, OP_SEQ_IN_RE); }
         bool is_unit(expr const* n)     const { return is_app_of(n, m_fid, OP_SEQ_UNIT); }
 
+        bool is_string_term(expr const * n) const {
+            sort * s = get_sort(n);
+            return (u.is_seq(s) && u.is_string(s));
```
NikolajBjorner: u.is_seq should be redundant.

NikolajBjorner:
looks mostly good, main question is about "StringSequence"

mtrberzi:
I had some reason for needing to change that name. Let me look into that tomorrow.

src/ast/seq_decl_plugin.h
mtrberzi: Right, `is_string` does make that check.

mtrberzi:
Actually it seems to work with "String" instead of "StringSequence", I really can't remember why I thought I needed to do that but if it comes up again I'll try to work around it.

PR #993, opened 2017-04-29
[CMake] Make MSVC flags much more similar to python/Makefile build system.

### LINKED ISSUE #990: [CMake] Broken debug info in RelWithDebInfo build mode
Taken from #986 

Stated by @NikolajBjorner 
For starters, I get these kind of warnings (RelWithDebInfo) on my various machines.

api_log_macros.cpp.obj : warning LNK4099: PDB 'vc140.pdb' was not found with 'api_log_macros.cpp.obj' or at 'C:\z3\build\vc140.pdb'; linking object as if no debug info
api_commands.cpp.obj : warning LNK4099: PDB 'vc140.pdb' was not found with 'api_commands.cpp.obj' or at 'C:\z3\build\vc140.pdb'; linking object as if no debug info
z3_replayer.cpp.obj : warning LNK4099: PDB 'vc140.pdb' was not found with 'z3_replayer.cpp.obj' or at 'C:\z3\build\vc140.pdb'; linking object as if no debug info
api_tactic.cpp.obj : warning LNK4099: PDB 'vc140.pdb' was not found with 'api_tactic.cpp.obj' or at 'C:\z3\build\vc140.pdb'; linking object as if no debug info
api_stats.cpp.obj : warning LNK4099: PDB 'vc140.pdb' was not found with 'api_stats.cpp.obj' or at 'C:\z3\build\vc140.pdb'; linking object as if no debug info
api_solver.cpp.obj : warning LNK4099: PDB 'vc140.pdb' was not found with 'api_solver.cpp.obj' or at 'C:\z3\build\vc140.pdb'; linking object as if no debug info
api_seq.cpp.obj : warning LNK4099: PDB 'vc140.pdb' was not found with 'api_seq.cpp.obj' or at 'C:\z3\build\vc140.pdb'; linking object as if no debug info

I am not sure where "vc140.pdb" gets pinned by the build system and where I would fetch it from.
The post http://stackoverflow.com/questions/33292241/generating-pdb-files-correctly-for-a-native-dll-using-visual-studio-2013 indicates it is entirely handled by the build system.

About repros
C:\z3>cmake --version
cmake version 3.5.2
C:\z3>ninja --version
1.6.0
C:\z3>cl --version
Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24215.1 for x86
Repro scenario: take some smt2 file that takes more than 10 seconds to run. Select sampling based profiling from the VS profiling wizard, configure the profiler to run z3.exe on the appropriate long running input. The stack trace information in the profiler is going to be inaccurate (compared to what gets generated by python/nmake).

delcypher:
This PR attempts to make the compiler defines, flags and linker flags for MSVC much more similar to what the python/Makefile build system. This is motivated by problems reported in #990 . This might not fix #990 but it makes the build systems flags much similar which will help with debugging.

The most notable things are

* Removal of `compiler_flags_override.cmake`. It wasn't working correctly.
* Addition of the `msvc_legacy_quirks.cmake` file.
* Support for link time optimization. The new `LINK_TIME_OPTIMIZATION` option is equivalent
to the python/Makefile build system's `--optimize` flag for MSVC. It's also supported for Clang and GCC too but this very experimental.

This `msvc_legacy_quirks.cmake` file contains flags that I couldn't find a proper home for. I haven't duplicated some of the flags from the old build system because some of them do not make sense.

* `/D _DEBUG` is not set. [MSVC sets this itself when the runtime is set](https://msdn.microsoft.com/en-us/library/b0084kay.aspx)  so this should not be set
* `/D _CONSOLE` is not set. I can't find any proper documentation for this but apparently this is [a define set in older environments](https://codeyarns.com/2010/12/02/visual-c-windows-and-console-subsystems/).
* `/D _LIB` is not set. [Apparently this macro](http://stackoverflow.com/questions/35034683/how-to-tell-if-current-project-is-dll-or-static-lib) is used indicate if building a static library. This doesn't seem to be documented though.
* `/D ASYNC_COMMANDS` is not set. I have no idea why this was being set. It is not documented and is not being used.

Other flags from the old build system are used. I've changed when some of them are used (e.g. `/D UNICODE /D _UNICODE` is set always but the old build system only set this is some configuration) and some are set but comments note that setting some don't really have a purpose.


NikolajBjorner:
Ready to merge?

delcypher:
@NikolajBjorner Should be. There are things that need to be fixed (see all the `FIXME`s) but I am not the right person to fix this. There are many flags whose purpose I don't understand.

However given that these flags are actively used with the old build system I probably won't be making things worse.

NikolajBjorner:
Indeed, your FIXME comments are very useful. Thanks!

PR #995, opened 2017-05-01
Theory case split heuristic (for theory_str)

mtrberzi:
This patch adds a heuristic to smt_context called "theory case split", which enforces cardinality constraints on sets of boolean literals. The interface to this heuristic is the method `mk_th_case_split()`, which takes a set of literals which are to be treated as mutually exclusive. The context then performs relevant checks both to propagate truth values over this set of literals and to detect conflicts when more than one literal in a case split set is set to true simultaneously.

The option controlling this heuristic is `smt.theory_case_split`. By default it is turned off, which means the context will just assert axioms that constrain any case split sets that would be generated to be mutually exclusive.


src/smt/smt_context.cpp
```diff
@@ -2960,6 +2964,125 @@ namespace smt {
...
+                    TRACE("theory_case_split", tout << "found case split set { ";
+                    for(literal_vector::iterator set_it = case_split_set.begin(); set_it != case_split_set.end(); ++set_it) {
+                        tout << set_it->index() << " ";
+                    }
+                    tout << "}" << std::endl;);
+                    for(literal_vector::iterator set_it = case_split_set.begin(); set_it != case_split_set.end(); ++set_it) {
+                        literal l2 = *set_it;
+                        if (l2 != l) {
+                            b_justification js(l);
+                            switch (get_assignment(l2)) {
```
NikolajBjorner: the code for assign should be handling these cases already, so I am not sure you want to unfold the case splits here. You can log one place by displaying the literal assignment.

src/smt/smt_context.cpp
```diff
@@ -2960,6 +2964,125 @@ namespace smt {
...
+        unsigned assigned_literal_end = m_assigned_literals.size();
+        while(assigned_literal_idx < assigned_literal_end) {
+            literal l = m_assigned_literals[assigned_literal_idx];
+            TRACE("theory_case_split", tout << "check literal " << l.index() << std::endl; display_literal_verbose(tout, l); tout << std::endl;);
+            ++assigned_literal_idx;
+            // check if this literal participates in any theory case split
+            if (m_all_th_case_split_literals.contains(l.index())) {
+                TRACE("theory_case_split", tout << "assigned literal " << l.index() << " is a theory case split literal" << std::endl;);
+                // now find the sets of literals which contain l
+                vector<literal_vector> case_split_sets = m_literal2casesplitsets.get(l.index(), vector<literal_vector>());
```
NikolajBjorner: if m_literal2casesplitsets isn't modified you could use a reference, but dangerous when the memory below changes.

mtrberzi: Case splits shouldn't be removed in this code so I think it is okay -- have a look at the latest commit.

src/smt/smt_context.cpp
mtrberzi: Makes sense. Is it correct in the new version?

NikolajBjorner: your code looks correct to me, just I wonder if it has some redundancies

PR #999, opened 2017-05-02
Theory-aware branching heuristic

mtrberzi:
This patch adds the ability for theory solvers to give a small additive "bump" to the activity of boolean literals during the search. This is added to the normal activity found during conflict analysis. A specialization of the case split queue is used to track this information, and can be enabled with the parameter `smt.theory_aware_branching`.


src/smt/smt_case_split_queue.cpp
```diff
@@ -1086,32 +1111,151 @@ namespace smt {
...
+
+        virtual void push_scope() {}
+
+        virtual void pop_scope(unsigned num_scopes) {}
+
+        virtual void next_case_split(bool_var & next, lbool & phase) {
+            phase = l_undef;
+
+            if (m_context.get_random_value() < static_cast<int>(m_params.m_random_var_freq * random_gen::max_value())) {
+                next = m_context.get_random_value() % m_context.get_num_b_internalized();
```
NikolajBjorner: I would add some asserts here: multiplication could overflow, get_num_b_internalized() could be 0.

src/smt/smt_case_split_queue.cpp
```diff
@@ -1086,32 +1111,151 @@ namespace smt {
...
-        case CS_RELEVANCY:
-            return alloc(rel_case_split_queue, ctx, p);
-        case CS_RELEVANCY_ACTIVITY:
-            return alloc(rel_act_case_split_queue, ctx, p);
-        case CS_RELEVANCY_GOAL:
-            return alloc(rel_goal_case_split_queue, ctx, p);
-        default:
-            return alloc(act_case_split_queue, ctx, p);
+
+        if (p.m_theory_aware_branching) {
```
NikolajBjorner: why not reuse the case-split strategy configuration instead of the Boolean?

src/smt/smt_case_split_queue.cpp
mtrberzi: This is copied from the `act_case_split_queue` above, should those asserts go there as well?

src/smt/smt_case_split_queue.cpp
mtrberzi: I wasn't sure if it was okay to do that. I'll refactor.

mtrberzi:
New commits are up.

src/smt/params/smt_params.h
```diff
@@ -240,6 +241,7 @@ struct smt_params : public preprocessor_params,
         m_case_split_strategy(CS_ACTIVITY_DELAY_NEW),
         m_rel_case_split_order(0),
         m_lookahead_diseq(false),
+        m_theory_case_split(false),
```
NikolajBjorner: delete?
Does this compile at this point?

mtrberzi: Yes. Compiles fine without that

PR #1000, opened 2017-05-02
smt_setup for strings/sequences

mtrberzi:
This adds a small hook for selecting between multiple string solvers, for theory_str. At present whenever theory_str would be used, theory_seq is instantiated instead.


src/smt/smt_setup.cpp
```diff
@@ -201,7 +205,7 @@ namespace smt {
     void setup::setup_QF_BVRE() {
         setup_QF_BV();
         setup_QF_LIA();
-        setup_seq();
+        m_context.register_plugin(alloc(theory_seq, m_manager));
```
NikolajBjorner: is this setup_seq without static features?

mtrberzi: Yes

PR #1013, opened 2017-05-08
Integrate new LRA solver capabilities by Lev Nachmanson. 

NikolajBjorner:


CMakeLists.txt
```diff
@@ -311,7 +314,7 @@ endif()
 # FIXME: Support ARM "-mfpu=vfp -mfloat-abi=hard"
 if (("${TARGET_ARCHITECTURE}" STREQUAL "x86_64") OR ("${TARGET_ARCHITECTURE}" STREQUAL "i686"))
   if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang"))
-    set(SSE_FLAGS "-mfpmath=sse" "-msse" "-msse2")
+    set(SSE_FLAGS "-mfpmath=sse" "-msse" "-msse2" "-std=c++0x")
```
delcypher: @NikolajBjorner The `-std=c++0x` flag doesn't really belong here. It isn't related to SSE flags.

Is Z3 moving to C++11 now?

NikolajBjorner: The LRA solver uses C++11, so yes.
We will have to fix this misplacement.

delcypher: Okay so something like this in the appropriate place in the top-level `CMakeLists.txt` file would do the job.

```cmake
# C++ language version
# FIXME: Use CMake's own mechanism for selecting language version
if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang"))
  z3_add_cxx_flag("-std=c++11" REQUIRED)
else()
  message(AUTHOR_WARNING "Not setting C++ language version for compiler")
endif()
```

delcypher: @NikolajBjorner Thanks. Is there also a corresponding changes to the python/Makefile build system. I did a quick skim of this very large PR and I couldn't see it. We should keep the build systems in sync.

Requiring a C++11 capable compiler is a breaking change. AFAIK this hasn't been required up until now. I have no problem with this requirement because most modern C++ compilers support C++11.

I do worry if such a change would break existing code. For example will the introduction of C++11's implicit move constructors silently break existing code?

NikolajBjorner: Yes it is a breaking change. C++11 was tested in Lev's branch for some time on a reasonable slice of Z3.
The C++11 directive to the master branch was added Saturday by Wintersteiger. One observation was that the gcc build time increased and our builds started hitting a timeout.

delcypher: @NikolajBjorner Oh I missed that change. Sorry.

delcypher:
@NikolajBjorner @levnach 
I can't really provide much more feedback because I'm not familiar with this code. For the benefit of others who wish to review your code you may wish to clean up this PR by rebasing it on the current master and removing all the merge commits (Seeing changes already present in master in the PR makes it harder to review). You could also squash some commits together if you think its appropriate.

There also commits fixing mistakes (e.g. e53fa5a) in here and it would be cleaner to fix the commits that made the mistakes to avoid this.

Of course rebasing is a lot of work so this is just a suggestion.

NikolajBjorner:
Is this better?

delcypher:
@NikolajBjorner There certainly are less commits :). Again I can't really review the code but hopefully this makes it a little easier for others. I didn't envisage squashing everything into a single commit because some of the history might have been useful, but it's up to you. Now that you've done I guess we can use this.

Two minor nits:

* @levnach is not the author of your squashed commits
* The commit message is very terse considering how much code the commit introduces. Maybe something like this?
```
Add new LRA (<spell out acronym>) solver.

This solver is aimed at <something> and is available via the <something> tactic.
See <reference> for more information.
```

PR #1040, opened 2017-05-23
sys/timeb.h not available on OpenBSD

zshipko:
I ran the tests against my changes on OpenBSD and everything passed -- please let me know if there are any other issues with my update.


NikolajBjorner:
I have removed the code that caused the incompatibility. Hope this helps.

zshipko:
Even better, thank you!

Issue #1035, opened 2017-05-22
Unexpected huge performance difference between solver using "default" tactic, Z3_mk_solver() and Z3_mk_simple_solver()


delcypher:
I'm currently investigating queries that Z3 is taking a very long time to solve in my tool and I ran into one that Z3 is handling in a surprising way.

This is the query

```
(declare-fun f_ackermann!0 () (_ BitVec 64))
(assert
 (not (fp.isNaN ((_ to_fp 11 53) f_ackermann!0))))
(assert
 (let ((?x8 ((_ to_fp 11 53) f_ackermann!0)))
(let ((?x18 (fp.sqrt roundNearestTiesToEven (fp.mul roundNearestTiesToEven ?x8 ?x8))))
(let ((?x24 (fp.mul roundNearestTiesToEven ?x18 ((_ to_fp 11 53) (_ bv4607092346807469998 64)))))
(let ((?x11 (fp.abs ?x8)))
(let (($x20 (or (fp.isNaN (fp.mul roundNearestTiesToEven ?x11 ((_ to_fp 11 53) (_ bv4607092346807469998 64)))) (fp.isNaN ?x18))))
(let (($x22 (or $x20 (fp.gt (fp.mul roundNearestTiesToEven ?x11 ((_ to_fp 11 53) (_ bv4607092346807469998 64))) ?x18))))
(not (or $x22 (or (or (fp.isNaN ?x11) (fp.isNaN ?x24)) (fp.lt ?x11 ?x24)))))))))))
(check-sat)
```

In my tool libz3 is taking ~1679 seconds to solve this. When I give this query to the Z3 binary it is solving the query in ~2 seconds.

I started digging into this and by making my tool increase libz3's verbosity it looks like the slow `smt` tactic is being used when my tool uses libz3.

```
(smt.searching)
(smt.simplifying-clause-set :num-deleted-clauses 4)
(smt.searching)
(smt.simplifying-clause-set :num-deleted-clauses 30)
...
```

but when Z3 binary solves the query a different tactic is being used.

```
(sat.sat-solver)
(sat.simplify)
(sat-restart :conflicts 202 :decisions 677 :restarts 1 :clauses 185638 :learned 159 :gc-clause 0 :memory 46.75 :time 1.35)
(sat-restart :conflicts 303 :decisions 1100 :restarts 2 :clauses 185638 :learned 246 :gc-clause 0 :memory 46.75 :time 1.36)
...
```

[...]

PR #1069, opened 2017-06-11
[Doxygen] Rewrite documentation of `Z3_mk_solver()` and `Z3_mk_simple_solver()`

delcypher:
Rewrite documentation of `Z3_mk_solver()` and `Z3_mk_simple_solver()` `Z3_mk_simple_solver()` to try to make it clearer what the differences
are between these APIs.

This an attempt to address issues noted in #1035.

@wintersteiger @NikolajBjorner Could you please review you please review this? You both have much better knowledge than me so I want the documentation to be accurate.

@ddcc I noticed that you made a similar change to your fork of the clang-static analyzer as I did for KLEE which is related to these functions that I'm documenting. Could you take a look at the documentation and check you are happy with it?

### DIFF
--- a/src/api/z3_api.h
+++ b/src/api/z3_api.h
@@ -5795,9 +5795,35 @@ extern "C" {
     /** @name Solvers*/
     /*@{*/
     /**
-       \brief Create a new (incremental) solver. This solver also uses a
-       set of builtin tactics for handling the first check-sat command, and
-       check-sat commands that take more than a given number of milliseconds to be solved.
+       \brief Create a new solver. This solver is a "combined solver" (see
+       combined_solver module) that internally uses a non-incremental (solver1) and an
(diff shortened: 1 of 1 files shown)


delcypher:
@wintersteiger Side note: Once this is merged the online [Z3 documentation](https://z3prover.github.io/api/html/group__capi.html) should probably be re-generated. It doesn't look like this has been done since Mon Oct 5 2015.

ddcc:
Thanks for the patch! Minor grammar nit: `Does not changes its behaviour` => `Does not change its behaviour`. The only remaining case you may want to address is the fallback behavior from solver2 to solver1 when solver2 returns unknown.

As an aside,  I did test changing the simple solver to be default, but runtime on the CSA test suite went up significantly, so I dropped the patch.

PR #1070, opened 2017-06-12
[CMake] Move CMake files into their intended location

delcypher:
This PR revisits the compromise proposed by @agurfinkel in #461.

While this has served us well (allowing progress
to be made) over time limitations of this approach have appeared.

The main problem is that doing many git operations (e.g. pull, rebase)
means the CMake files don't get updated unless the user remembers to
run the script.

This PR

* moves the CMake files into their intended location.
* Changes `contrib/cmake/boostrap.py` to do nothing except emit a warning
  that it is now deprecated. This is so that we don't break out-of-tree scripts
  that build Z3 using CMake without any warning.
* Updates the CMake build system and documentation to be aware of this change.

This PR is a stepping stone to complete #986 .


delcypher:
@NikolajBjorner Thanks for merging.

PR #1073, opened 2017-06-13
Add translate method for FuncDecl in java api

chaserhkj:
When implementing a multithread application using z3, I find that I need to translate `FuncDecl` objects to another `Context`, but noticed that `FuncDecl` does not have a `translate(Context ctx)` method as `Expr` does. I think it indeed needs one so I implemented one for it.

### DIFF
--- a/src/api/java/FuncDecl.java
+++ b/src/api/java/FuncDecl.java
@@ -1,6 +1,6 @@
 /**
 Copyright (c) 2012-2014 Microsoft Corporation
-   
+
 Module Name:
 
     FuncDecl.java
@@ -12,8 +12,8 @@ Author:
     @author Christoph Wintersteiger (cwinter) 2012-03-15
 
 Notes:
-    
-**/ 
+
+**/
 
 package com.microsoft.z3;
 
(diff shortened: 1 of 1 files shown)


chaserhkj:
I forgot to get the docstrings done right, just fixed it.

wintersteiger:
This should really not be necessary as `FuncDecl` inherits `translate` from `AST`. Is there some other problem that prevents the use of that?

chaserhkj:
`FuncDecl` could not use `translate` from `AST`, cause `AST.translate()` is creating an `AST` java object, which cannot be cast to its heir class object `FuncDecl`. Doing so will yield a runtime exception.

In order to fix this, we need to implement either a `FuncDecl.translate()` like I did in this branch, or a dispatcher in `AST` to dispatch its object creation in `AST.translate()` to heir class constructors. (like `Expr.create()` in `Expr.translate()`, but maybe this could be done easier with java reflection)

wintersteiger:
Oh yes, `AST.translate` needs to call `create`, that's a bug (which probably also exists in some of the other APIs).

PR #1094, opened 2017-06-21
[WIP][CMake] Fix broken regeneration of some .cpp files

### LINKED ISSUE #1030: [CMake] Regeneration of some `.cpp` files are broken
Currently the CMake build system fails to re-generate the `install_tactic.cpp`, `mem_initializer.cpp`, and `gparams_register_modules.cpp` files when the headers that are used to generate these files change.

The reason for this is because CMake is currently unaware that the generation of these files depends on certain header files.

We cannot use shell globing in CMake to fix this because this will fail in the case that new header files are added. The right way to fix this is to change the interface of `z3_add_component()` so that header files that need to be scanned to generate header files are explicitly listed.

When I have the time I'll fix this. This blocks #986.

delcypher:
This PR fixes #1030.

This adds new optional arguments to `z3_add_component()` CMake macro

* `TACTIC_HEADERS` - Allows header files that declare tactics to be explicitly listed
* `EXTRA_REGISTER_MODULE_HEADERS` - Allows additional headers that declare module parameters/description to be explicitly listed. Note that all header files that are generated from
`.pyg` files are implicitly scanned too.
* `MEMORY_INIT_FINALIZER_HEADERS` - Allows header files that declare memory
initializers/finalizers to be explicitly listed.

The basic approach is to make the header files that need to be scanned explicit (in most cases)
so that CMake knows that when those headers change or the list of headers to scan changes that the 
`install_tactic.cpp`, `gparams_register_modules.cpp`, or `mem_initializer.cpp` files get automatically rebuilt.

Although this adds a little more maintenance work we get the benefit that incremental builds work correctly (i.e. when I do `$ git pull ; ninja` everything rebuilds correctly).

This PR is labelled WIP because I need to do more testing. So far I have verified on Arch Linux that the
generated `install_tactic.cpp`, `gparams_register_modules.cpp`, or `mem_initializer.cpp` files are identical between what is on Z3's `master` branch (both build systems) and what is in this PR (both build systems).

I want to check everything works fine on Windows and macOS before saying I think this is ready to review/merge.

### DIFF
--- a/cmake/z3_add_component.cmake
+++ b/cmake/z3_add_component.cmake
@@ -55,6 +55,7 @@ endfunction()
 #   SOURCES source1 [source2...]
 #   [COMPONENT_DEPENDENCIES component1 [component2...]]
 #   [PYG_FILES pygfile1 [pygfile2...]]
+#   [TACTIC_HEADERS header_file1 [header_file2...]]
 # )
 #
 # Declares a Z3 component (as a CMake "object library") with target name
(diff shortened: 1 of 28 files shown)


delcypher:
@NikolajBjorner This merged a little earlier than I intended.

However I've just finished testing on macOS and on Windows and this PR seems to work correctly.

PR #1095, opened 2017-06-22
model_evaluator fix

agurfinkel:
The major change is a bug fix to handling of expansion of array equalities.


NikolajBjorner:
There should be a few regressions under z3test/regressions/smt2 that check properties of the evaluator.

agurfinkel:
added a regression that shows 3 bugs in model_evaluator

Issue #1098, opened 2017-06-22
undefined reference to `lean::dense_matrix<rational, rational>::dense_matrix(unsigned int, unsigned int)'


delcypher:
I just tried compiling `master` (@ 77ffa9f32fa5a202b95a7b2b745e608969ec9309) with CMake (Ninja generator with Debug build type) and with Clang (4.0) and ASan enabled (set environment variable `CXXFLAGS="-fsanitize=address -fno-omit-frame-pointer"` in environment before configuring) and I get a linking error I haven't seen before.

```
src/util/lp/CMakeFiles/lp.dir/lu_instances.cpp.o: In function `lean::dense_matrix<rational, rational> lean::get_B<rational, rational>(lean::lu<rational, rational>&, vector<unsigned int, true, unsigned int> const&)':
/home/dan/dev/jfs/z3/src/src/util/lp/lu.hpp:935: undefined reference to `lean::dense_matrix<rational, rational>::dense_matrix(unsigned int, unsigned int)'
clang-4.0: error: linker command failed with exit code 1 (use -v to see invocation)            
ninja: build stopped: subcommand failed.       
```

@levnach I think this is your code. Any idea what's going on here?

I took a look and I think it's complaining about this template instantiation.

```c++
template lean::dense_matrix<lean::mpq, lean::mpq> lean::get_B<lean::mpq, lean::mpq>(lean::lu<lean::mpq, lean::mpq>&, vector<unsigned int> const&);
```

It looks like `lean::mpq` is a typedef for `rational` (see `src/util/lp/numeric_pair.h`). I think the constructor it's trying to call is

```c++
template <typename T, typename X> dense_matrix<T, X>::dense_matrix(unsigned m, unsigned n) : m_m(m), m_n(n), m_values(m * n, numeric_traits<T>::zero()) {
}
```

in `src/util/lp/dense_matrix.hpp`.

**EDIT:** ~~If I build without ASan I don't get the linker error which is weird.~~ I made a mistake. The a non-ASan Clang debug fails too.

delcypher:
@levnach Thanks for the quick fix. The build is now fixed for me. Closing.

levnach:
I just created a pull request that should fix it. Nikolaj, can you please merge this pull request?

PR #1099, opened 2017-06-22
add a template instantination

levnach:


NikolajBjorner:
This is to address #1098

PR #1114, opened 2017-06-26
Fix Python API examples

delcypher:
This PR does two things

* Teaches CMake to copy over some of the Python API examples it wasn't aware of previously
* Fixes the Python API examples so they are compatible with Python 3.


wintersteiger:
Looks good to me, thanks!

PR #1129, opened 2017-07-04
clean up warnings in theory_str

mtrberzi:
This fixes some warnings exposed by the new build process. `theory_str.cpp` now compiles under g++-5.4.0 with no warnings.

PR #1131, opened 2017-07-05
fix theory_str warnings: rename get_value() to get_arith_value()

mtrberzi:
Fixes a warning exposed while cleaning up #1129. Should build `theory_str` cleanly in debug/release without warnings, if anything else gets exposed by clang during this build I'll address it before merge.


NikolajBjorner:
thanks!

PR #1136, opened 2017-07-07
Add flag to mk_make.py to optionally enable Control Flow Guard

danpere:
Adds a flag (disabled by default) `--guardcf` to `mk_make.py` to enable [Control Flow Guard checks](https://docs.microsoft.com/en-us/cpp/build/reference/guard-enable-guard-checks), which is a security feature that checks that indirect calls go to known good addresses. `libz3.dll` compiled with this flag satisfies [`binskim`](https://github.com/Microsoft/binskim)'s security checks. This PR leaves it disabled by default as it may have a performance impact.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -104,6 +104,8 @@ GIT_DESCRIBE=False
 SLOW_OPTIMIZE=False
 USE_OMP=True
 LOG_SYNC=False
+GUARD_CF=False
+ALWAYS_DYNAMIC_BASE=False
 
 FPMATH="Default"
 FPMATH_FLAGS="-mfpmath=sse -msse -msse2"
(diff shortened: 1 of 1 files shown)


delcypher:
@danpere I don't suppose you'd be interested in adding this to the CMake build system too?

wintersteiger:
That's great, thanks a lot! 

As @delcypher says, it would be great if you could add that to the cmake system as well.

danpere:
I had seen the comments about wanting to get rid of the Python build system in favor of CMake, but I hadn't found the MSVC path in CMake when making the PR. I think I see how to do it now, but won't have time to do the PR today, so I'll probably do it Monday. Also, doing research for this, I notice that `llvm` has a similar feature called [Control Flow Integrity](https://clang.llvm.org/docs/ControlFlowIntegrity.html); I'll figure out the right flags so when compiling `z3` with `clang`, that feature will be used.

NikolajBjorner:
super, thanks!

Issue #1130, opened 2017-07-04
[TravisCI] Treating some warnings as errors


delcypher:
Now that we have TravisCI in place we could consider treating certain categories of warnings found by Clang/GCC as errors. This would hopefully prevent certain issues from staying in the master branch for long as they would break the build.

Is there any interest in doing this? If so what warnings should be treated as errors?

This was inspired by  #1129.

E.g.

## GCC

- [ ] `-Wodr` 

## Clang

- [ ] `-Wdelete-non-virtual-dtor`
- [ ] `-Woverloaded-virtual`

NikolajBjorner:
I am fairly anal about getting Z3 to compile without warnings even if they may be benign. They help flag issues even when it comes to code maintainability. 
The main issue now becomes that it is a mind-numbing task to nit over some warnings that have no effect especially in the unit tests (I took care of several of these, but many to go). 

The warnings you list seem not in the benign category. Are these flagged as warnings at this point? I don't recall seeing them for a while. 
Deleting a non-virtual destructor is asking for trouble: we tend to treat these as bugs because they can mask memory leaks. I am not sure what the ramifications of the other warnings are. I was unable to compile a couple of days ago when overloading a virtual, so this was fixed long before pushing.

delcypher:
@NikolajBjorner 
> The warnings you list seem not in the benign category. Are these flagged as warnings at this point? I don't recall seeing them for a while.

Yes most of the warnings I gave as examples are not emitted by Z3's current code. These are warnings are remember being in Z3's code at some point which I would be keen to not have re-introduced. It's certainly not exhaustive list, they are just things I thought of just now.

NikolajBjorner:
OK, sounds good. What should be changed to treat them as errors? The virtual destructor issue is definitely something I would treat as an error unless learning about any compelling evidence to the contrary. The only kind of warnings I feel compelled to punt on are the ones that appear in compiled macros (there are some on potentially uninitialized or unused variables, but these are going to be used properly)

delcypher:
> OK, sounds good. What should be changed to treat them as errors?

There are two places this could be done either as custom flags set in the TravisCI build scripts or as part of the CMake build system itself.

I'd prefer to do it in CMake itself because we can test the compiler at runtime and actually check whether or not it supports the warning we'd like to treat as an error.

[...]

PR #1138, opened 2017-07-09
Support serious warnings in CMake and TravisCI

delcypher:
This aims to implement the ideas in #1130.

~~DO NOT MERGE until we are sure that this works correctly in TravisCI.~~ Seems to work okay in TravisCI.


delcypher:
@NikolajBjorner @wintersteiger I think this is ready to go. Can this be merged?

PR #1142, opened 2017-07-11
Add get_num_scopes to python solver api.

jfeser:
Add a function to call Z3_solver_get_num_scopes from the python solver api.


wintersteiger:
Looks good to me, thanks!

PR #1146, opened 2017-07-12
fix get_model in lar_solver

levnach:

### DIFF
--- a/src/util/lp/column_namer.h
+++ b/src/util/lp/column_namer.h
@@ -15,7 +15,7 @@ public:
         T a;
         unsigned i;
         while (it->next(a, i)) {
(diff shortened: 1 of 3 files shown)


contrib/cmake/src/test/lp/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+add_executable(lp_tst lp_main.cpp lp.cpp $<TARGET_OBJECTS:util> $<TARGET_OBJECTS:polynomial> $<TARGET_OBJECTS:nlsat>  $<TARGET_OBJECTS:lp> )
```
delcypher: This file is in the wrong location. I suggest you rebase on to the latest master.

PR #1147, opened 2017-07-14
Improved theory_arith integration in theory_str::get_arith_value()

mtrberzi:
This is a simple fix for a common case in Z3str3 that performs badly when the equivalence class of arithmetic terms is very large (e.g. a lot of terms having length 0). Instead of iterating over the equivalence class to find an arithmetic constant, we first ask the arithmetic solver directly whether it knows a value for the term; if that fails, we fall back to the eqc approach.

### DIFF
--- a/src/smt/theory_str.cpp
+++ b/src/smt/theory_str.cpp
@@ -4653,30 +4653,23 @@ namespace smt {
     }
 
     bool theory_str::get_arith_value(expr* e, rational& val) const {
-        if (opt_DisableIntegerTheoryIntegration) {
-            TRACE("str", tout << "WARNING: integer theory integration disabled" << std::endl;);
-            return false;
-        }
-
         context& ctx = get_context();
         ast_manager & m = get_manager();
-        theory_mi_arith* tha = get_th_arith(ctx, m_autil.get_family_id(), e);
-        if (!tha) {
+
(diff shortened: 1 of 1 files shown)


delcypher:
Looks like this is breaking some regression tests E.g. 

https://travis-ci.org/Z3Prover/z3/jobs/253640858#L2582

These are also some timeouts, which I didn't even know we were enforcing.

mtrberzi:
Hmm, the "invalid model" response is troublesome. I hope that's not an issue in the arithmetic solver. I'll investigate it.

delcypher:
If it helps you can use https://github.com/Z3Prover/z3/blob/master/contrib/ci/scripts/test_z3_system_tests.sh to run the regression tests. The script is parameterized by a set of environment variables (check the top of the script to see which are needed).

mtrberzi:
I changed the way the check is done as suggested by @NikolajBjorner -- now we get the root of the EQC directly. Can confirm that this fixes the "invalid model" case on my end, want to see if tests come out more cleanly before I investigate further.

mtrberzi:
Don't merge this yet, it is segfaulting on some cases.

mtrberzi:
Okay, this is doing much better after adding the obvious check. I want to try doing this for string constants as well but I think this can potentially be reviewed as-is and I'll open a new PR for the string constant version.

PR #1158, opened 2017-07-23
pretty printer: fix typo with ReSort sort name

facanferff:
The pretty printer prints "Re" instead of "RegEx" (as http://rise4fun.com/Z3/tutorial/sequences).


NikolajBjorner:
thanks

PR #1188, opened 2017-08-01
moved obj_equiv_class to ast

agurfinkel:


NikolajBjorner:
right, "git add factor_equivs.cpp, commit and push"

agurfinkel:
:blush:, should be good now.

PR #1205, opened 2017-08-12
[CMake] Change how the default value of `USE_OPENMP` is set.

delcypher:
This change means if the user explicitly passes `-DUSE_OPENMP=ON` to
CMake during the first configure and the compiler does not support
OpenMP the configure will fail but if the user doesn't specify it the
build system will automatically enable/disable OpenMP support depending
on whether it is supported by the compiler.

This is an improvement on the previous behaviour because previously we
would just emit a warning if `-DUSE_OPENMP=ON` was passed and the
compiler didn't support OpenMP.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,33 +281,37 @@ endif()
 ################################################################################
 # OpenMP support
 ################################################################################
-option(USE_OPENMP "Use OpenMP" ON)
-set(OPENMP_FOUND FALSE)
-if (USE_OPENMP)
-  # Because this is on by default we make the configure succeed with a warning
-  # if OpenMP support is not detected.
-  find_package(OpenMP)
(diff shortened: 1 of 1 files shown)


delcypher:
@wintersteiger When did Appveyor get enabled for this repo?

delcypher:
This is currently failing for reasons unrelated to this PR as far as I know. See #1206

NikolajBjorner:
Regarding Appvoyer: https://github.com/Z3Prover/z3test/pull/10

wintersteiger:
@delcypher: some time before last week. Nothing is set up yet, it shouldn't run anything or report anything.  @NikolajBjorner: Your link points to an old PR for the new performance testing infrastructure. This is now at https://github.com/Z3Prover/PerformanceTest . Our developers set up AppVeyor just for their code, not for Z3.

Issue #1206, opened 2017-08-12
regressions/smt2/string-ops.smt2 failing


delcypher:
It looks like a recent commit (probably 7b47b0380e5d2fb403c9da0b89af498064b5e286 ) broke the continuous integration build the `regressions/smt2/string-ops.smt2` test seems to be failing.

https://travis-ci.org/Z3Prover/z3/builds/263272789

You can see that it's the configurations using Clang that are failing

delcypher:
Looks like this was fixed by a change in z3test ( https://github.com/Z3Prover/z3test/commit/34bd4a2a04b3c36adf0e2b69660cac2bc246d23a ). Closing.

PR #1207, opened 2017-08-12
[TravisCI] Enable macOS build

delcypher:
Only one configuration is enabled due to macOS builds taking a long time to start. This is based on previous experience. We can easily add more configurations later if we really need them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,11 +59,16 @@ env:
     # 64-bit GCC 4.8 Debug
     - LINUX_BASE=ubuntu_14.04 C_COMPILER=/usr/bin/gcc-4.8 CXX_COMPILER=/usr/bin/g++-4.8 TARGET_ARCH=x86_64 Z3_BUILD_TYPE=Debug
 
-# TODO: OSX support
-#matrix:
-#  include:
-#    - os: osx
-#      osx_image: xcode 8.2
+# macOS (a.k.a OSX) support
+matrix:
+  include:
+    # For now just test a single configuration. macOS builds on TravisCI are
+    # very slow so we should keep the number of configurations we test on this
+    # OS to a minimum.
+    - os: osx
+      osx_image: xcode8.3
+      # Note: Apple Clang does not support OpenMP
+      env: Z3_BUILD_TYPE=RelWithDebInfo USE_OPENMP=0
 script:
   # Use `travis_wait` when doing LTO builds because this configuration will
   # have long link times during which it will not show any output which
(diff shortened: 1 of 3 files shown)


delcypher:
This is currently failing for reasons unrelated to this PR as far as I know. See #1206

delcypher:
@wintersteiger The OSX builds run on TravisCI's own build machines just like the Linux builds. I am not aware of any additional conditions when running on OSX. There's certainly no mention of "apple"/"macos"/"osx" in [their terms of service](https://billing.travis-ci.com/pages/terms). I am not a lawyer though.

If you want to know about OSX environment itself [see here](https://docs.travis-ci.com/user/reference/osx/).

We already have working builds of this configuration. Here's an example
https://travis-ci.org/Z3Prover/z3/jobs/264172108

wintersteiger:
This isn't so much about Travis' T&C, but about Apple's. When last I checked it was illegal to run OSX on a VM, i.e. you had to run on Apple hardware, and this is also the reason for me having a Mac Mini on my office desk (and for Azure and AWS not offering OSX VMs). The Travis OSX environment overview starts with "Travis CI runs builds in virtual machines...", and I wouldn't want to get into trouble over that. Can you please disable these builds until we have figured out the legal situation?

delcypher:
I'm not sure if it's still relevant but at least in the past TravisCI used a third party service (called [MacStadium](https://www.macstadium.com/) for their Mac machines. MacStadium appear to provide bare-metal macs.

There's an article on the site specifically about TravisCi

https://www.macstadium.com/customers/travis-ci/

If TravisCI is not using mac machines for their builds then that would be illegal for all projects that use them (not just Z3) which I highly doubt is the case.

However I will create a PR to disable the macOS configuration for now. I suggest you reach out to TravisCI team (they've been pretty responsive to my e-mails in the past) to assuage your legal concerns.

It would be nice to solve this reasonably quickly though because having macOS tested as well as Linux is useful.

delcypher:
@wintersteiger PR created ( #1208 )

NikolajBjorner:
Seems that the more likely situation is that Travis is a business that ensures its infrastructure is properly licensed, e.g., uses Mac based services where appropriate.

wintersteiger:
@NikolajBjorner Sure, one would expect so many things... They claim to have found an "amazing sponsor" (see https://github.com/travis-ci/travis-ci/issues/216), but I haven't found any clear statement on this issue yet. I'm fine with having Travis builds on OSX, but for now we should use our own hardware (if Travis allows us to do that).

delcypher:
@wintersteiger @NikolajBjorner Moving discussion to new issue (#1211) as this PR was already merged.

PR #1216, opened 2017-08-17
Simpler include paths (fixes #534)

### LINKED ISSUE #534: Simpler header includes
This is a proposal for a non-essential code clean up.

Currently header includes look like this

``` c++
#include "fpa2bv_converter.h"
```

and a component has to transitively obtain the correct include directories from its dependencies. This seems unnecessarily complicated and potentially error prone (i.e. what happens we someone creates two or more header files with the same name?).

A much simpler solution is write **all** includes (even if the header file being included is in the same directory as the `#include ""` declaration) relative to the `src/` directory i.e.

``` c++
#include "ast/fpa/fpa2bv_converter.h"
```

then the build systems only need to specify one extra include path (i.e. the `src` directory) rather than loads of different include directories. This is much simpler and also makes it clearer in the code where a header file resides.

In order to do this both build systems (i.e. the Python and CMake build systems) would need to be changed at the same time. This change might be quite disruptive for anyone who has in progress work or out of tree patches.

@wintersteiger @NikolajBjorner Any thoughts on this? I'm happy to give this a go if you think making this change would be beneficial but I would want to do it at a time where it would cause minimal disruption.

delcypher:
# DO NOT MERGE UNTIL TESTS PASS

This PR simplifies the remaining includes and then performs a clean up to the CMake build system that we can gain from doing this.

A nice bonus from doing this is compiler invocations are a lot easier to read because now we only pass two include paths to the compiler rather than the directories of the transitive dependencies.

### DIFF
--- a/cmake/z3_add_component.cmake
+++ b/cmake/z3_add_component.cmake
@@ -36,15 +36,8 @@ function(z3_add_component_dependencies_to_target target_name)
   # Remaing args should be component names
   set(_expanded_deps ${ARGN})
   foreach (dependency ${_expanded_deps})
-    # FIXME: Adding these include paths wouldn't be necessary if the sources
-    # used include paths rooted in the ``src`` directory.
-    get_property(_dep_include_dirs GLOBAL PROPERTY Z3_${dependency}_INCLUDES)
-    foreach (inc_dir ${_dep_include_dirs})
-      target_include_directories(${target_name} PRIVATE "${inc_dir}")
-    endforeach()
-    unset(_dep_include_dirs)
     # Ensure this component's dependencies are built before this component.
-    # This important because we might need the generated header files in
+    # This is important because we might need the generated header files in
     # other components.
     add_dependencies(${target_name} ${dependency})
   endforeach()
@@ -214,18 +207,14 @@ macro(z3_add_component component_name)
     target_compile_options(${component_name} PRIVATE ${flag})
   endforeach()
 
-  # It's unfortunate that we have to manage the include directories and dependencies ourselves.
+  # It's unfortunate that we have to manage dependencies ourselves.
   #
   # If we weren't building "object" libraries we could use
   # ```
-  # target_include_directories(${component_name} INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_CURRENT_BINARY_DIR}")
   # target_link_libraries(${component_name} INTERFACE ${Z3_MOD_COMPONENT_DEPENDENCIES})
   # ```
   # but we can't do that with "object" libraries.
 
-  # Record this component's include directories
-  set_property(GLOBAL PROPERTY Z3_${component_name}_INCLUDES "${CMAKE_CURRENT_SOURCE_DIR}")
-  set_property(GLOBAL APPEND PROPERTY Z3_${component_name}_INCLUDES "${CMAKE_CURRENT_BINARY_DIR}")
   set_property(GLOBAL PROPERTY Z3_${component_name}_DEPS "")
   # Record this component's dependencies
   foreach (dependency ${Z3_MOD_COMPONENT_DEPENDENCIES})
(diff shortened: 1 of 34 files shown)


NikolajBjorner:
tests pass...

delcypher:
@NikolajBjorner Thanks for merging. Looks like we should invest some effort in setting up continuous integration for Windows. It would have caught what you had to fix in ee00852151f6e0b34f244878eaa5643a42ec3335

wintersteiger:
We have 2 different CIs for Windows already. What's your experience with AppVeyor? I have an account but haven't configured anything yet. VSTS was hell to set up so I'm not looking for a repeat of that.

delcypher:
@wintersteiger 
>  What's your experience with AppVeyor? 

I'm afraid I don't have any personal experience setting it up (yet). The problem with the existing Windows CIs is they don't integrate with GitHub pull requests. The great thing about TravisCI's github integration is that for every pull request we get to see whether the configurations build (thus preventing pre-maturely merging broken PRs into master). If they fail to build the build log is easily available to anyone with a few clicks. These build logs can be very helpful to the PR author to fix things.

I think AppVeyor is very similar to TravisCI in that you have a small YAML file in the root of the repository that controls the build. It is also software-as-a-service like TravisCI so everything build's on AppVeyor's servers so there's no maintenance involved. Although there does seem to be [a private cloud option](https://www.appveyor.com/docs/build-environment/#private-build-cloud) if you want it.

So it would just be a case of setting up the scripts. If you did that I'd recommend following the design pattern I used for the TravisCI scripts. Different stages of the build are split into separate bash scripts (e.g. `https://github.com/Z3Prover/z3/blob/master/contrib/ci/scripts/build_z3_cmake.sh`). The scripts are parameterized  using environment variables to change their behaviour. This makes the scripts re-usable and most of them are not tied to TravisCI at all and so can be used to build and test many different Z3 configurations on any Linux/macOS machine. This modular design made adding the TravisCI macOS support pretty trivial.

wintersteiger:
> they don't integrate with GitHub pull requests

They do! And we get the cross/checkmark indicators on our commits! The only problem is that it's apparently impossible to make the build log files publically accessible on VSTS. So, my options are to either add every single e-mail address that wants to see them, or I have to set up my own custom web-server for it (which I started to do, but it was just taking way too much time...) Other projects also just check the log files into a git repository as the last step of the build. 

In terms of scripts we have everything in z3test/scripts and our current CIs and binary builds just run those .py scripts with various parameters (and env vars) on them (plus a couple shell scripts z3test/build).

NikolajBjorner:
> @NikolajBjorner Thanks for merging. Looks like we should invest some effort in setting up continuous integration for Windows. It would have caught what you had to fix in ee00852

The fix in ee00852 was required after optional became heap allocated.

The proper approach might be that I need to change my process now that Travis checks pull requests: I just commit to master instead of creating a pull request. I am somewhat incompetent when it comes to operating machinery: my current private versions of Z3 contain some very unstable code that should not be merged.

delcypher:
> They do! And we get the cross/checkmark indicators on our commits! The only problem is that it's apparently impossible to make the build log files publically accessible on VSTS.

Ah sorry. I meant the integration that shows up in the "checks" user interface in PRs. If you look in this PR you'll see a "details" button. You'll see it shows two checks as passed

* continuous-integration/appveyor/pr AppVeyor build succeeded
* continuous-integration/travis-ci/pr The Travis CI build passed

There isn't any mention of the VSTS builds there.

> So, my options are to either add every single e-mail address that wants to see them, or I have to set up my own custom web-server for it (which I started to do, but it was just taking way too much time...) Other projects also just check the log files into a git repository as the last step of the build.

That sounds like too much work. It's probably just simpler to set up AppVeyor. This is something I could look at in the future but right now I'm too busy to give it the time it deserves.

delcypher:
@NikolajBjorner 

> The proper approach might be that I need to change my process now that Travis checks pull requests: I just commit to master instead of creating a pull request.

That is up to you. I think pull requests make a lot of sense for external contributors. For core developers it is debatable what the best approach is given that you might not want to be delayed waiting for tests to finish running. If you commit code that breaks the build TravisCI will eventually send you an e-mail :)

> I am somewhat incompetent when it comes to operating machinery: my current private versions of Z3 contain some very unstable code that should not be merged.

There are lots of different approaches to handling this. I always work on a separate branches so that `master` on my local machine is always the same (or behind) upstream `master`.

If it helps you could also just clone a separate copy of the public Z3 repository on your machine and keep that for public stuff only. You could share code between these different clones by adding additional remotes (URLs for git to pull/push code from/to).

wintersteiger:
I'm talking about this: image

I assumed the PRs would show the same, possibly a configuration issue. I never configured any settings for PRs.

Issue #871, opened 2017-01-17
Confusing result when combining regular expressions and string results


jawline:
Hi,

The following problem

```
(declare-const st (String))
(declare-const capture1 (String))
(declare-const capture2 (String))

(define-fun regexPortion1 () (RegEx String) (str.to.re "abcdefg"))
(define-fun in_portion_1 ((a String)) Bool (str.in.re a regexPortion1))

(define-fun regexPortion2 () (RegEx String) (re.++ regexPortion1 regexPortion1))
(define-fun in_portion_2 ((a String) (b String)) Bool (and (str.in.re a regexPortion2) (str.in.re b regexPortion1) (= a (str.++ b b))))

(define-fun regexPortion3 () (RegEx String) (re.* regexPortion2))
(define-fun in_portion_3 ((a String) (b String) (c String)) Bool (and (str.in.re a regexPortion3) (or (= a "") (and (in_portion_2 b c) (str.in.re a (re.+ (str.to.re b)))))))

(assert (in_portion_3 st capture1 capture2))

(check-sat)
(get-model)
(get-info :reason-unknown)
```

Fails to solve, with the message
```

unknown
(error "line 17 column 10: model is not available")
(:reason-unknown "smt tactic failed to show goal to be sat/unsat (incomplete (theory seq))")

```
However if I remove the or

(or (= a "") (and (in_portion_2 b c) (str.in.re a (re.+ (str.to.re b))))

and just have (and (in_portion_2 b c) (str.in.re a (re.+ (str.to.re b))) in its place, I get a solution. Is there any obvious reason for this?

I'm trying to create the regular expression ((SomeStuff)SomeStuffRepeated)*, which should leave 
st as a whole string accepted by the regular expression and capture1 and capture2 as portions of it, or st capture1 and capture2 as the empty string.

Thanks

NikolajBjorner:
@mtrberzi: using z3str3 we get:
z3 test.smt2 smt.string_solver=z3str3
Failed to verify: regexNFA.is_valid()

mtrberzi:
It looks like there's a term in this instance that's constructing a regex based on other terms found while solving -- Z3str3 doesn't have support for "non-constant" regex terms. For example we don't handle terms such as `(str.to.re capture1)` in this instance because `capture1` is not a string constant.

(But the error message could be improved. I'll add that to my list of small regex fixes in an upcoming PR.)

NikolajBjorner:
@mtrberzi , thanks.
[...]

PR #1220, opened 2017-08-19
Small regex fixes in theory_str

mtrberzi:
This fixes three issues with regex support in theory_str:

* Matching against the empty string in `str.to.re` would cause a segfault.
* theory_str now supports `re.all`.
* The error message emitted when encountering an invalid term in `str.to.re` (such as a string variable) has been improved. This addresses some feedback discussed in #871.

### DIFF
--- a/src/smt/theory_str.cpp
+++ b/src/smt/theory_str.cpp
@@ -6189,27 +6189,33 @@ namespace smt {
             expr * arg_str = a->get_arg(0);
             zstring str;
             if (u.str.is_string(arg_str, str)) {
-                TRACE("str", tout << "build NFA for '" << str << "'" << "\n";);
-                /*
-                 * For an n-character string, we make (n-1) intermediate states,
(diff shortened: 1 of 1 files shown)


src/smt/theory_str.cpp
```diff
@@ -6284,6 +6295,10 @@ namespace smt {
             }
 
             TRACE("str", tout << "range NFA: start = " << start << ", end = " << end << std::endl;);
+        } else if (u.re.is_full(e)) {
+            NOT_IMPLEMENTED_YET();
```
NikolajBjorner: is this ready to be merged?

mtrberzi: Oops, my mistake. I actually just caught this as well and came here to make sure this hadn't been merged yet.

mtrberzi:
OK, this should be better. It is a temporary fix for the NFA until we can fully incorporate the automata-based reasoning.

PR #1225, opened 2017-08-22
Add injectivity tactic

nbraud:
This tactic solves performance issues with Z3 interleaving uninterpreted-sort reasoning and theory reasoning, as encountered by @nickgian in http://rise4fun.com/Z3/0Ujk

In general, we expect a significant improvement in F* verification time once Everest updates to a Z3 version where this tactic is enabled.

### DIFF
--- a/scripts/mk_project.py
+++ b/scripts/mk_project.py
@@ -23,6 +23,10 @@ def init_project_def():
     add_lib('subpaving', ['interval'], 'math/subpaving')
     add_lib('ast', ['util', 'polynomial'])
     add_lib('rewriter', ['ast', 'polynomial', 'automata'], 'ast/rewriter')
+    # Simplifier module will be deleted in the future.
+    # It has been replaced with rewriter module.
+    add_lib('simplifier', ['rewriter'], 'ast/simplifier')
+    add_lib('macros', ['simplifier'], 'ast/macros')
     add_lib('normal_forms', ['rewriter'], 'ast/normal_forms')
     add_lib('model', ['rewriter'])
     add_lib('tactic', ['ast', 'model'])
@@ -30,7 +34,7 @@ def init_project_def():
     add_lib('parser_util', ['ast'], 'parsers/util')
     add_lib('grobner', ['ast'], 'math/grobner')
     add_lib('euclid', ['util'], 'math/euclid')
-    add_lib('core_tactics', ['tactic', 'normal_forms'], 'tactic/core')
+    add_lib('core_tactics', ['tactic', 'macros', 'normal_forms', 'rewriter'], 'tactic/core')
     add_lib('sat_tactic', ['tactic', 'sat'], 'sat/tactic')
     add_lib('arith_tactics', ['core_tactics', 'sat'], 'tactic/arith')
     add_lib('nlsat_tactic', ['nlsat', 'sat_tactic', 'arith_tactics'], 'nlsat/tactic')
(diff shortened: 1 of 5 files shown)


nbraud:
@wintersteiger It seems like we still have some minor issues: `(check-sat-using injectivity)` is fast, but returns `unknown`...

Here is the input I used:
```
(declare-sort Term)
(declare-fun Box ((_ BitVec 32)) Term)
(declare-fun UnBox (Term) (_ BitVec 32))
(declare-const x (_ BitVec 32))

(assert (forall ((x (_ BitVec 32))) (= (UnBox (Box x)) x)))
(assert (not (= (Box (bvshl x #x00000002)) (Box (bvmul x #x00000004)))))
```

Using the injectivity tactic, I get:
```
(check-sat-using injectivity)
(injectivity :num-exprs 14 :num-asts 192 :time 0.00 :before-memory 2.10 :after-memory 2.10)
unknown
```

Without the tactic, the solver goes into the loop @nickgian encountered:
```
(check-sat)
(simplifier :num-exprs 15 :num-asts 203 :time 0.00 :before-memory 2.23 :after-memory 2.23)
(smt.simplifier-done)
(smt.searching)
(smt.restarting :propagations 0 :decisions 129 :conflicts 101 :restart 150 :agility 0.05)
(smt.simplifying-clause-set :num-deleted-clauses 1)
(smt.restarting :propagations 1 :decisions 304 :conflicts 252 :restart 225 :agility 0.12)
(smt.working :conflicts 3080 :agility 0.70)
(smt.working :conflicts 4216 :agility 0.77)
(smt.simplifying-clause-set :num-deleted-clauses 0)
(smt.delete-inactive-clauses  :num-deleted-clauses 4616)
(smt.working :conflicts 5919 :agility 0.81)
(smt.working :conflicts 6920 :agility 0.82)
(smt.working :conflicts 8473 :agility 0.83)
(smt.working :conflicts 9474 :agility 0.83)
(smt.delete-inactive-clauses  :num-deleted-clauses 0)
(smt.working :conflicts 10475 :agility 0.83)
(smt.working :conflicts 12303 :agility 0.83)
(smt.working :conflicts 13304 :agility 0.83)
(smt.working :conflicts 14305 :agility 0.83)
(smt.delete-inactive-clauses  :num-deleted-clauses 0)
[...]
```

wintersteiger:
Looking good! 

`(check-sat-using ...)` expects the tactic to decide the problem, which takes the shape of either an empty goal or a goal with only on constraint that is `false`, otherwise it says `unknown`. Your tactic produces something else, so we need to tell it what to use to decide the problem, for instance, by using the default tactic:

`(check-sat-using (then injectivity default))`

That's a major performance improvement on this simple example! :-) 

(Minor: set your editor to 4 spaces instead of tabs.)

delcypher:
@nbraud There are some missing changes here required to make this work with CMake.

* In `src/tactic/core/CMakeLists.txt` you need to add `injectivity_tactic.h` under the `TACTIC_HEADERS`.
* If `core_tacticts` component now directly depends on the `rewriter` component (as your change to `scripts/mk_project.py`) suggests then you need modify  `src/tactic/core/CMakeLists.txt` so that it is listed under `COMPONENT_DEPENDENCIES`.

nbraud:
@delcypher Yes, I was waiting on feedback from @wintersteiger (regarding whether it's appropriate to put the tactic in `core/`) before doing this.  Thanks for the info, though  :)

nbraud:
@wintersteiger OK, that explains it  :)

nbraud:
@delcypher OK, this builds (and the `injectivity` tactic is available) under CMake.

Issue #1227, opened 2017-08-23
CMake-based build fails in git worktrees


nbraud:
The CMake-based build system assumes that if `.git` exists, it is a directory, which isn't always the case:

```
 ~/everest/z3-injectivity/build $ cmake ..        
-- The CXX compiler identification is GNU 6.3.0                                            
-- Check for working CXX compiler: /usr/bin/c++                                            
-- Check for working CXX compiler: /usr/bin/c++ -- works                                   
-- Detecting CXX compiler ABI info                                                         
-- Detecting CXX compiler ABI info - done                                                  
-- Detecting CXX compile features                                                          
-- Detecting CXX compile features - done                                                   
-- Z3 version 4.5.1.0                                                                      
CMake Error at cmake/git_utils.cmake:20 (message):                                         
  GIT_DIR ("/home/t-nibrau/everest/z3-injectivity/.git") is not a directory                
Call Stack (most recent call first):                                                       
  CMakeLists.txt:102 (add_git_dir_dependency)                                              
                                                                                           
                                                                                           
-- Configuring incomplete, errors occurred!                                                
See also "/home/t-nibrau/everest/z3-injectivity/build/CMakeFiles/CMakeOutput.log".         

 ~/everest/z3-injectivity/build $ ll ../.git      
-rw-r--r-- 1 t-nibrau t-nibrau 64 Aug 16 17:21 ../.git

 ~/everest/z3-injectivity/build $ cat ../.git     
gitdir: /home/t-nibrau/everest/z3/.git/worktrees/z3-injectivity
```

delcypher:
You can probably workaround this by passing `-DINCLUDE_GIT_DESCRIBE=OFF` and `-DINCLUDE_GIT_HASH=OFF` to CMake.

Your shell output is kind of hard to read but it looks like `.git` in your case in a plain file that contains text that points to a different tree. How did you create your git clone? If I can reproduce it I can probably fix it.

nbraud:
Sorry, I didn't realise the prompt was somewhat in the way; I just edited it out.
Yes, `.git` is indeed a plain file that only refers to another `.git` tree.

Here is how to reproduce it:
```
/tmp/z3-cmake $ git clone https://github.com/Z3Prover/z3    
Cloning into 'z3'...                                                           
remote: Counting objects: 58704, done.                                         
remote: Compressing objects: 100% (39/39), done.                               
remote: Total 58704 (delta 37), reused 44 (delta 25), pack-reused 58640        
[...]

PR #1228, opened 2017-08-23
[CMake] Teach CMake to support git worktrees

delcypher:
This should fix the bug reported by @nbraud reported in #1227.

Previously the CMake build system assumed that the `.git` file must
be a directory. This is not the case when the working directory
is a "git worktree". In this case the `.git` file is just a plain
file that points to a directory within the true `.git` directory.
This commit essentially implements the logic to traverse this extra
level of indirection and removes some assumptions that the `.git`
file is a directory.

@nbraud Could you test this to check that it works correctly for your use case?


nbraud:
@delcypher This indeed works, thanks a lot.  :)

PR #1238, opened 2017-08-30
Update injectivity tactic

nbraud:
- [x] Add extra rewriter for `(g (f x)) -> (x)`.
- [x] Add parameters controlling which rewriters are called.
- [x] Support UNSAT core generation.

This is a follow-up on #1225.

### DIFF
--- a/src/tactic/core/CMakeLists.txt
+++ b/src/tactic/core/CMakeLists.txt
@@ -25,6 +25,8 @@ z3_add_component(core_tactics
     normal_forms
     rewriter
     tactic
+  PYG_FILES
+    injectivity_params.pyg
   TACTIC_HEADERS
     blast_term_ite_tactic.h
     cofactor_term_ite_tactic.h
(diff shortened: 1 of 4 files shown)


nbraud:
@wintersteiger Interestingly, this works fine using `check-sat-using`, but adding it to the default tactic segfaults  :/
Here is the patch I used for the « hacked » build: https://paste.debian.net/hidden/747b0349/
(It turns out patches are not valid attachments on Github...)

nbraud:
This seems to be caused by the `params_ref` object going `NULL` while the `injectivity_params` object is still live.  It probably involves Z3 object wizardry beyond my ken.  :(

nbraud:
Note to self: Clang has an helpful comment about passing a temporary to a member's constructor.

wintersteiger:
@nbraud did you figure out what's going wrong there?

nbraud:
@wintersteiger I was ill the past 2 weeks and didn't have a change to poke at it, but it's likely the lifetime issue you pointed out.

NikolajBjorner:
is this ready or should it be retired?

NikolajBjorner:
no activity for 18 months

PR #1262, opened 2017-09-20
Fix warnings in C++ API

UniQP:
When assertions are disabled, the compiler warns about unused function parameters.


wintersteiger:
I hate this style of ignoring arguments, but I guess that's the latest fashion in C++ land...

PR #1271, opened 2017-09-27
Fix Python API docs done with `make api_docs`

ulidtko:
Hi,

I've noticed that `make api_docs` (I'm building with CMake) on current master produces broken html tree with 404 links right on the `index.html` page. The C and C++ parts are okay; not so much for Python. Both pydoc and doxygen/python outputs seemed to be missing from the build directory.

So I fixed this. Please consider pulling; thanks!


wintersteiger:
That's great, thank you very much!

PR #1282, opened 2017-10-05
[TravisCI] Refactor CI defaults and fix bug handling Z3_BUILD_TYPE

delcypher:
This is work in progress.


delcypher:
@wintersteiger @NikolajBjorner This PR has exposed a very silly bug (now fixed in this PR) where the requested build type was ignored. We were always doing release builds previously, even when my intention was to do a debug builds.

Now that I've fixed the bug this has lead to a new problem.

https://travis-ci.org/Z3Prover/z3/jobs/283714508#L2322

The unoptimized unit tests run very slowly. On my machine it took ~20 minutes to run them. TravisCI doesn't see any output for 10 minutes and cancels the debug builds.

How would you like to handle this?

* **EASY OPTION** For debug builds do not run unit tests.
* For debug builds teach the `run_quiet` function (which suppresses tool output) to output something periodically so that TravisCI does not cancel the build.

NikolajBjorner:
Don't run into nit tests for debug builds. We already addressed checking assertionsinside the test directory in release

delcypher:
@NikolajBjorner Thanks for the feedback. I've disabled running the unit tests for the debug builds. Note we still build them in this configuration because this is fairly fast and I'd like to catch compilation errors here.

Let's see if the tests pass now.

delcypher:
@NikolajBjorner I think this is ready to be merged now.

Issue #1288, opened 2017-10-06
[UBSan] Multiple instances of undefined behaviour when running Z3 unit tests


delcypher:
When building Z3 with UBSan,  when the unit tests are executed (`./test-z3 /a`) the following errors are reported.

```
/home/user/z3_src/src/muz/rel/tbv.cpp:79:25: runtime error: shift exponent 69 is too large for 64-bit type 'unsigned long long'
/home/user/z3_src/src/util/mpz.cpp:564:26: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
```

This is for 110d558ee4262721f3ab2c999d789eaf84ac15d6

delcypher:
@NikolajBjorner Thanks. I don't see any reference to commit that also fixes the other reported error

```
/home/user/z3_src/src/util/mpz.cpp:564
```

Did you miss that one?

NikolajBjorner:
Thanks, I didn't notice this was for something different.

Issue #1287, opened 2017-10-06
[UBSan] call to function through pointer to incorrect function type when running C++ API examples


delcypher:
When building Z3 with UBSan, when the C++ api examples are executed the following error is reported.

```
/home/user/z3_src/src/api/api_context.cpp:278:13: runtime error: call to function (unknown) through pointer to incorrect function type 'void (*)(_Z3_context *, Z3_error_code)'
(/home/user/z3_build/examples/cpp_example_build_dir/cpp_example+0x4ad480): note: (unknown) defined here
```

This is for 110d558ee4262721f3ab2c999d789eaf84ac15d6

delcypher:
Unfortunately the UBSan error is still there, although now the error message has changed.

```
/home/user/z3_src/src/api/api_context.cpp:278:13: runtime error: call to function z3::context::error_handler(_Z3_context*, Z3_error_code) through pointer to incorrect function type 'void (*)
(_Z3_context *, Z3_error_code)'
/home/user/z3_src/src/api/c++/z3++.h:143: note: z3::context::error_handler(_Z3_context*, Z3_error_code) defined here
SUMMARY: AddressSanitizer: undefined-behavior /home/user/z3_src/src/api/api_context.cpp:278:13 in 
```

I will continue to suppress this warning in TravisCI. The code looks okay to me so I suggest we just keep this issue open until we figure out if this actually a bug or a problem with UBSan.

NikolajBjorner:
suppressing issue

Issue #1286, opened 2017-10-06
[UBSan] member call on null pointer of type when running C examples


delcypher:
When building Z3 with UBSan, when the "interpolation_example" (in the C API examples) runs the following errors are reported.

```
/home/user/z3_src/src/interp/iz3mgr.h:107:13: runtime error: member call on null pointer of type 'ast_manager'
/home/user/z3_src/src/interp/iz3mgr.h:99:13: runtime error: member call on null pointer of type 'ast_manager'
```

This is for 110d558ee4262721f3ab2c999d789eaf84ac15d6

Issue #964, opened 2017-04-05
UBSan reports problems with synch_mpq_manager, mpq_manager and mpq_manager


delcypher:
The Z3 binary when instrumented with UBSan reports Z3 is doing something bad with the synch_mpq_manager, mpq_manager and mpq_manager types.

Here is a simple smt file.

```
(get-info :name)
(get-info :version)
(set-logic ALL)
(set-option :produce-models true)
(set-info :smt-lib-version 2.5)
(exit)
```

Running Z3 (at c4b26cd691ddfa303b36a6b5a37ac1948352e5be ) yields the following error messages.
```
$ ./z3 mini.smt2
/home/dan/dev/z3/src/src/util/rational.h:32:46: runtime error: reference binding to null pointer of type 'struct synch_mpq_manager'
/home/dan/dev/z3/src/src/util/rational.h:63:26: runtime error: member call on null pointer of type 'struct synch_mpq_manager'
/home/dan/dev/z3/src/src/util/mpq.h:186:12: runtime error: member call on null pointer of type 'struct mpq_manager'
/home/dan/dev/z3/src/src/util/mpq.h:183:48: runtime error: member call on null pointer of type 'struct mpz_manager'
/home/dan/dev/z3/src/src/util/mpq.h:187:12: runtime error: member call on null pointer of type 'struct mpq_manager'
/home/dan/dev/z3/src/src/util/mpq.h:183:48: runtime error: member call on null pointer of type 'struct mpz_manager'
/home/dan/dev/z3/src/src/util/rational.h:32:46: runtime error: reference binding to null pointer of type 'struct synch_mpq_manager'
/home/dan/dev/z3/src/src/util/rational.h:63:26: runtime error: member call on null pointer of type 'struct synch_mpq_manager'
/home/dan/dev/z3/src/src/util/mpq.h:186:12: runtime error: member call on null pointer of type 'struct mpq_manager'
/home/dan/dev/z3/src/src/util/mpq.h:183:48: runtime error: member call on null pointer of type 'struct mpz_manager'
/home/dan/dev/z3/src/src/util/mpq.h:187:12: runtime error: member call on null pointer of type 'struct mpq_manager'
/home/dan/dev/z3/src/src/util/mpq.h:183:48: runtime error: member call on null pointer of type 'struct mpz_manager'
/home/dan/dev/z3/src/src/util/rational.h:32:46: runtime error: reference binding to null pointer of type 'struct synch_mpq_manager'
[...]

PR #1289, opened 2017-10-06
[TravisCI] Enable UBSan and ASan configurations

delcypher:
This is work in progress to add CI configurations that build with the AddressSanitizer and UndefinedBehaviorSanitizer.

This PR includes a set of suppression files that can be used to suppress warning about non-fatal errors if we need to.

Bugs found so far

## UBSan

* #964 
* #1286 
* #1287 
* #1288

## ASan

* #1290 
* #1296
* #1303 
* #1301
* #1299 
* #1305
* #1309 
* #1308 
* #1307


NikolajBjorner:
the build seems blocked on the OMP errors. I don't see what can be done from our side. Can you suppress these errors so I can merge? #1300 seems to be getting ready and I am inclined to merge it when it is.

delcypher:
@NikolajBjorner TravisCI is now green. There are a few questionable changes I've made here. I've had to replace calls to `exit()` with `_Exit()` in order to prevent ASan's LeakSanitizer component from running. This isn't really the right way to fix the issue. The better fix is probably to disable running the LeakSanitizer component for certain tests. This would require a change to the testing infrastructure so that's out of scope for this PR.

PR #1300, opened 2017-10-10
fix vector<> to support non-POD types

nunoplopes:


delcypher:
@nunoplopes You might of noticed that in PR #1289 I'm trying to get ASan and UBSan TravisCI configurations enabled. This work is getting close to completion (thanks a lot to @NikolajBjorner  for fixing the issues found so far). I don't know how far along you are with this work but would you be willing to wait for PR #1289 to be merged into master before merging this? That way you can rebase this PR and have ASan/UBSan checks run.

ASan/UBSan can't catch everything but its certainly better than nothing.

nunoplopes:
I've run tests under valgrind and I haven't seen any failure besides the ones I was already seeing in master. My biggest worry is not introducing leaks, but rather introducing new copies and regressing perf.  The patch is fairly big, though.

@delcypher BTW, would it be possible to bump gcc version to at least 5.0 on Travis build bots? Some of them fail with this patch because they are still using gcc 4.8..

NikolajBjorner:
Besides memory leaks Asan also reports use after free (I take Valgrind would catch the same).
At this point I am unable to reproduce crashes in t171.smt2, t201.smt2, z3,53.smt2 from the latest travis run. Perhaps your valgrind or ASan tools are useful figuring out what is going on.

delcypher:
@nunoplopes 
> @delcypher BTW, would it be possible to bump gcc version to at least 5.0 on Travis build bots? Some of them fail with this patch because they are still using gcc 4.8..

That was a deliberate decision on my part. GCC 4.8 is the compiler that shipped with Ubuntu 14.04LTS which is still in active use (its end of life is April 2019). This is why I added that configuration to TravisCI so that we catch breaking changes.

Is it really not possible for these changes to work with GCC 4.8? If not the PR basically requires that we drop support for most Ubuntu 14.04LTS users (unless they are willing to install a newer compiler).

nunoplopes:
@NikolajBjorner 
> Besides memory leaks Asan also reports use after free (I take Valgrind would catch the same).
At this point I am unable to reproduce crashes in t171.smt2, t201.smt2, z3,53.smt2 from the latest travis run. Perhaps your valgrind or ASan tools are useful figuring out what is going on.

I can't reproduce any problem on these either. Valgrind is clean on my VMs for these.

NikolajBjorner:
One detail. Unit tests are run with model_validate=true

delcypher:
> I can't reproduce any problem on these either. Valgrind is clean on my VMs for these.
The TravisCI infrastructure is designed to make it easy to reproduce build failures by using Docker.

https://travis-ci.org/Z3Prover/z3/jobs/288109767

To reproduce the above failed build you can run this on a Linux machine.

```
LINUX_BASE=ubuntu_16.04 \
C_COMPILER=/usr/bin/gcc-5 \
CXX_COMPILER=/usr/bin/g++-5 \
TARGET_ARCH=x86_64 \
Z3_BUILD_TYPE=Debug \
RUN_UNIT_TESTS=BUILD_ONLY \
contrib/ci/scripts/travis_ci_linux_entry_point.sh
```

This is documented in https://github.com/Z3Prover/z3/blob/master/contrib/ci/README.md


Seeing the failure is with the regression tests you can change `RUN_UNIT_TEST` to `SKIP` to speed things up. You will need to have Docker installed for this to work.

The way I've been debugging failures is letting this script run until it hits a failure. When this happens I then create a new interactive container that starts at last successful `RUN` command in the docker file (just run `docker images` and see which one is the newest).

```
docker run --rm -ti <image_id>
```

In this environment you can use `sudo` without a password to install any debugging dependencies you might need (e.g. gdb). To use gdb inside the container you might need to pass `--privileged` so that it has permission to use ptrace.

nunoplopes:
fixed support for gcc 4.8 and rebased. rerunning tests..

PR #1307, opened 2017-10-14
[ASan] Fix reported memory leak

delcypher:
[ASan] Fix reported memory leak. When calling exit() the `s` string and `timeit` don't get free'd and so a memory leak is reported.

An alternative implementation would be to keep `tst_exit_all_tests()`
but perform leak collection (by calling the LSan API) before calling
`exit()`. This would require pre-processor macros though so just
using `return 0` seems more elegant.

### DIFF
--- a/src/test/main.cpp
+++ b/src/test/main.cpp
@@ -10,10 +10,6 @@
 #include "util/memory_manager.h"
 #include "util/gparams.h"
 
-static void tst_exit_all_tests() {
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
I added a conditional return so that normal tests could be run.
Running all tests seems to either simply time out on the build machines or they set some global parameters that cause later tests to be slow.

PR #1315, opened 2017-10-19
Add special case handling for theory_str constant backpropagation

mtrberzi:
This fixes a regression introduced by a recent commit that causes a crash when adding certain terms during the search. Specifically, it avoids asserting terms of the form "string constant = string constant", which `smt_justification` doesn't like.


mtrberzi:
Any idea why the build is failing? It seems to compile okay, but there is a message regarding running a 32-bit DLL on a 64-bit platform. I hope this isn't causing that somehow.

NikolajBjorner:
Ignore appveyor for now

delcypher:
@mtrberzi Could you add a test to https://github.com/Z3Prover/z3test ?

PR #1323, opened 2017-10-24
print proofs in graphviz

c-cube:
This adds a command `(get-proof-graph)` to the input language, and an option `dot_proof_file` to the solver.
Example: `./z3 test_dot.smt2 dot_proof_file=/tmp/proof.dot`, with the following file ('test_dot.smt2'):

```
(set-option :produce-proofs true)
(set-logic AUFLIA)
(declare-sort A$ 0)
(declare-sort B$ 0)
(declare-sort C$ 0)
(declare-sort D$ 0)
(declare-sort E$ 0)
(declare-fun f$ (A$ B$) Bool)
(declare-fun n$ () A$)
(assert (! (not (=> (forall ((?v0 A$)) (and (forall ((?v1 B$)) (f$ n$ ?v1)) (forall ((?v1 C$) (?v2 B$)) (f$ ?v0 ?v2)))) (forall ((?v0 E$) (?v1 D$) (?v2 A$)) (or (= ?v2 n$) (forall ((?v3 D$) (?v4 B$)) (f$ ?v2 ?v4)))))) :named a0))
(check-sat)
(get-proof-graph)
```

gives the attached proof (after `dot -Tpdf /tmp/proof.dot > proof.pdf`): [proof.pdf](https://github.com/Z3Prover/z3/files/1410051/proof.pdf)

Note: there seems to be an option `dot_proof_display` somewhere, but I didn't find where it was used nor how to use it.

Also, I'm not very familiar with C++, so please bear with me!

### DIFF
--- a/src/ast/ast_pp_dot.cpp
+++ b/src/ast/ast_pp_dot.cpp
@@ -11,6 +11,7 @@ Abstract: Pretty-printer for proofs in Graphviz format
 // string escaping for DOT
 std::string escape_dot(std::string const & s) {
     std::string res;
+    res.reserve(s.size()); // preallocate
     for (auto c : s) {
         if (c == '\n')
             res.append("\\l");
@@ -21,8 +22,8 @@ std::string escape_dot(std::string const & s) {
 }
 
 // map from proofs to unique IDs
-typedef map<const expr*, unsigned, obj_ptr_hash<const expr>, default_eq<const expr*> > expr2id;
-typedef map<const expr*, bool, obj_ptr_hash<const expr>, default_eq<const expr*> > set_expr;
+typedef obj_map<const expr, unsigned> expr2id;
+typedef obj_map<const expr, bool> set_expr;
 
 // temporary structure for traversing the proof and printing it
 struct ast_pp_dot_st {
@@ -33,6 +34,7 @@ struct ast_pp_dot_st {
     std::ostream &          m_out;
     unsigned                m_next_id;
     bool                    m_first;
+    ast_manager &           m_manager;
 
     ast_pp_dot_st(const ast_pp_dot * pp, std::ostream & out) :
         m_pp(pp),
@@ -41,19 +43,21 @@ struct ast_pp_dot_st {
         m_id_map(),
         m_to_print(),
         m_printed(),
-        m_first(true) {}
+        m_first(true),
+        m_manager(pp->get_manager()) {}
+
+    ~ast_pp_dot_st() = default;
     
     void push_term(const expr * a) { m_to_print.push_back(a); }        
 
     void pp_loop() {
         // DFS traversal
-        auto& m = get_manager();
         while (!m_to_print.empty()) {
             const expr * a = m_to_print.back();
             m_to_print.pop_back();
             if (!m_printed.contains(a)) {
                 m_printed.insert(a, true);
-                if (m.is_proof(a))
+                if (m().is_proof(a))
                     pp_step(to_app(a));
                 else
                     pp_atomic_step(a);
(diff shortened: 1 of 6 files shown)


src/ast/ast_pp_dot.cpp
```diff
@@ -0,0 +1,132 @@
...
+    for (auto c : s) {
+        if (c == '\n')
+            res.append("\\l");
+        else
+            res.push_back(c);
+    }
+    return res;
+}
+
+// map from proofs to unique IDs
```
NikolajBjorner: We typically use:
obj_map<expr, unsigned> and 
obj_map<expr, bool> 

It may not make a visible difference here, but generally pointers used for comparisons in maps make their behavior non-reproducible. The obj_map uses unique identifiers associated with the first arguments.

src/ast/ast_pp_dot.cpp
```diff
@@ -0,0 +1,132 @@
...
+        return escape_dot(out.str());
+    }
+
+    void pp_atomic_step(const expr * e) {
+        unsigned id = get_id(e);
+        m_out << "node_" << id << " [shape=box,color=\"yellow\",style=\"filled\",label=\"" << label_of_expr(e) << "\"] ;" << std::endl;
+    }
+    
+    void pp_step(const proof * p) {
+        auto m = get_manager();
```
NikolajBjorner: Not sure if this gets a reference or copy.
Suggestion: add an attribute 

               ast_manager& m; 

to the class and use this instead of retrieving "get_manager()".

In the long run, using get_manager() turned out to be a bit verbose.

src/ast/ast_pp_dot.cpp
c-cube: Ok, will do. I was wondering what kind of value to use for the set (I'd use `unit` in ML, but I don't know of any equivalent in C++ so I used `bool`)?

src/ast/ast_pp_dot.cpp
```diff
@@ -0,0 +1,132 @@
...
+        unsigned num_args = p->get_num_args();
+        TRACE("pp_ast_dot_step", tout << " :kind " << p->get_kind() << " :num-args " << num_args);
+        if (num_args > 0) {
+            // print result
+            expr* p_res = p->get_args()[num_args-1]; // result
```
NikolajBjorner: Is this ast_manager::get_fact?

src/ast/ast_pp_dot.cpp
```diff
@@ -0,0 +1,132 @@
...
+            unsigned id = get_id(p);
+            const char* color =
+                m_first ? (m_first=false,"color=\"red\"") : num_args==1 ? "color=\"yellow\"": "";
+            m_out << "node_" << id <<
+                " [shape=box,style=\"filled\",label=\"" << label_of_expr(p_res) << "\""
+                << color << "]" << std::endl;
+            // now print edges to parents (except last one, which is the result)
+            std::string label = p->get_decl()->get_name().str();
+            for (unsigned i = 0 ; i+1 < num_args; ++i) {
+                expr* parent = p->get_args()[i];
```
NikolajBjorner: p->get_arg(i) instead of p->get_args()[i]

src/ast/ast_pp_dot.cpp
```diff
@@ -0,0 +1,132 @@
...
+                    << "[label=\"" << label << "\"];" << std::endl;;
+            }
+        } else {
+            pp_atomic_step(p);
+        }
+    }
+
+    // find a unique ID for this proof
+    unsigned get_id(const expr * e) {
+        if (m_id_map.contains(e)) {
```
NikolajBjorner: unsigned id = 0;
           if (!m_id_map.find(e, id)) {
                id = m_next_id++;
                m_id_map.insert(e, id);
           }
           return id;

NikolajBjorner:
cool, comments added

src/ast/ast_pp_dot.cpp
NikolajBjorner: Use obj_hashtable.

src/ast/ast_pp_dot.cpp
c-cube: Ah, neat. I assume this saves one lookup if `e` is present already.

Issue #1223, opened 2017-08-22
Feature request: API to manually construct a Z3_model objects


delcypher:
I'm currently using components of Z3 to build my own solver (the parser and some of its tactics) but I use my own custom method external to Z3 to do the solving.

A problem I have is that when my custom method of solving finds a satisfying assignment I have no good way to create a `Z3_model` object. I need to have the found assignment in this form because I need to feed this model through the Z3 tactics I have used to perform model conversion so that I can give a satisfying assignment to the original problem given to my solver.

The only way I can see of creating a `Z3_model` in my case is to create a new solver and assert the satisfying assignment as a set of equalities and then ask the solver to find a satisfying assignment.
This approach seems pretty wasteful so what I really need is an API that would just let me build a `Z3_model`. In my case I'm working in the `QF_BV`, `QF_FPBV`, and `QF_FP` theories so the model I need to create is free from functions (that need an interpretation) and infinite datatypes.

Could something like this be implemented?

NikolajBjorner:
It's possible. It requires adding a fair amount of additional API functions to 
1. create a Z3_model (the reference count should be 0 to follow conventions of other functions).
2. create Z3_func_interp objects
3. populate Z3_func_interp objects with cases and defaults. 
4. populate Z3_model objects with constants and functions.
5. add corresponding functions to the Java/.Net/C++/OCaml

wintersteiger:
+1. The tactic framework should allow tactics to use external solving methods and to support that, we need a model construction API. Ultimately it's probably not significantly faster to construct models via dedicated functions than to assert a bunch of equalities into a fresh solver, but it definitely makes the user-code simpler/more readable. 

@delcypher Let me know if you want to tackle this yourself, I'm happy to help if you get stuck.

NikolajBjorner:
I am adding a first pass..

delcypher:
> @delcypher Let me know if you want to tackle this yourself, I'm happy to help if you get stuck.

Unfortunately I'm rather busy with my PhD thesis these days. Once I'm done with that I'd happily take a look at this if it isn't already implemented.

delcypher:
@NikolajBjorner 
> create a Z3_model (the reference count should be 0 to follow conventions of other functions).

Should `Z3_model`s be partially immutable? I.e. once you assign an interpretation to a function should you be able to change it?

> populate Z3_func_interp objects with cases and defaults.

[...]

PR #1324, opened 2017-10-24
Add example of using Z3's new model construction C API. 

delcypher:
This API was requested in #1223.

This example uses the new `Z3_mk_model()`, `Z3_add_const_interp()`
, `Z3_add_func_interp()`, and `Z3_mk_as_array()` API calls.

### DIFF
--- a/examples/c/CMakeLists.txt
+++ b/examples/c/CMakeLists.txt
@@ -7,6 +7,19 @@
 # the C++ standard library in resulting in a link failure.
 project(Z3_C_EXAMPLE C CXX)
 cmake_minimum_required(VERSION 2.8.12)
+
+# Set C version required to C99
+if ("${CMAKE_VERSION}" VERSION_LESS "3.1")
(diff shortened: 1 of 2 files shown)


delcypher:
@NikolajBjorner The TravisCI tests pass. The AppVeyor build seems to be broken for unrelated reasons. So this is ready to go unless you have things you'd like to change about the API example itself?

wintersteiger:
The AppVeyor builds are disabled for now. They'll need a lot more work before they become useful and they seem to be using an earlier version of Visual Studio 2012 than our other Windows build machines.

delcypher:
@wintersteiger You can set the base image to control which version of Visual Studio is used. See https://www.appveyor.com/docs/build-environment/#choosing-image-for-your-builds

Was there a reason the `appveyor.yml` file and the scripts it depends on were not placed in the Z3 repository itself? Typically these need to evolve with the project itself so it would probably be simpler (and provide more transparency into how the build is working) if they were actually in the Z3 repository.

PR #1325, opened 2017-10-25
[DO NOT MERGE][CMake] .NET signing and resource linking

delcypher:
@wintersteiger This implements

* An option  (`DOTNET_BINDINGS_SNK`) to specify the SN key file used to sign the .NET assembly
* Passing the `/linkresource` flag on Windows when using a single configuration generation (e.g. `make`, `ninja`)

Although these changes won't cause any problems for Mono (Linux/macOS) I don't know about Windows. I unfortunately no longer have access to a Windows test machine so I can't this change. Windows CI also isn't set up (AppVeyor set up looks broken and also isn't using CMake).

Given that you asked for these changes could you check that the .NET bindings and built as you would expect?


NikolajBjorner:
We currently release nuget packages that are strongly signed by Microsoft keys. 
The process is based on assembling binaries from a tagged release and creating a nuget package that contains binaries across platforms.

Issue #1350, opened 2017-11-09
API doc build


wintersteiger:
@delcypher: which command failed [here](https://travis-ci.org/Z3Prover/z3/jobs/299638876) and where should libz3.so be found? I already added the build directory to the PATH just in case, but that didn't help. It works fine on my personal Ubuntu though.

delcypher:
@wintersteiger I can reproduce this. Something is going wrong when `pydoc.writedoc()` is called. When this happen the code in `z3core.py` fails to find `libz3.so`.

The problem is the paths being searched for in `z3core.py`. I hacked `z3core.py` to print out the directories that are checked for `z3core.py`.

```
XXX: trying "/home/dan/.gem/ruby/2.4.0/bin"
XXX: trying "/home/dan/dev/llvm/bin/bin"
XXX: trying "/home/dan/dev/iscp"
XXX: trying "/home/dan/dev/unrarall"
XXX: trying "/usr/local/sbin"
XXX: trying "/usr/local/bin"
XXX: trying "/usr/bin"
XXX: trying "/usr/lib/jvm/default/bin"
XXX: trying "/usr/bin/site_perl"
XXX: trying "/usr/bin/vendor_perl"
XXX: trying "/usr/bin/core_perl"
XXX: trying "/home/dan/path"
XXX: trying "/home/dan/dev/rubber/bin"
XXX: trying "."
XXX: trying "/home/dan/dev/z3/build/python/z3"
XXX: trying "/home/dan/dev/z3/build/python/z3/lib"
XXX: trying "/usr/lib"
XXX: trying "None"
```

None of them match. There are several problems here.

* Currently `libz3.so` is next to the `z3` package directory (i.e. `/home/dan/dev/z3/build/python/`) on my system. Therefore checking `os.path.dirname(os.path.abspath(__file__))` is wrong. Instead it should be checking one directory above.
* Checking `pkg_resources.resource_filename('z3', 'lib')` doesn't make sense as we don't copy the library into `/home/dan/dev/z3/build/python/z3/lib`.
* The paths in `PATH` probably shouldn't be checked before paths like `.`. This will cause us to pick up the wrong `libz3` if the user just so happens to already have a copy of Z3 installed.

delcypher:
@wintersteiger I've created #1351 which does the bare minimum to fix TravisCI.

> It works fine on my personal Ubuntu though.
[...]

PR #1351, opened 2017-11-10
Unbreak generation of pydoc documentation for python API bindings.

delcypher:
There are lots of problems with the way the bindings search for libz3
(see #1350). This commit does the bare minimum to unbreak the TravisCI
build.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1640,7 +1640,9 @@ from .z3consts import *
 _ext = 'dll' if sys.platform in ('win32', 'cygwin') else 'dylib' if sys.platform == 'darwin' else 'so'
 _lib = None
 _default_dirs = ['.',
-                 os.path.dirname(os.path.abspath(__file__)),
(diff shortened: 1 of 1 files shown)


delcypher:
@wintersteiger TravisCI build now passes.

wintersteiger:
Sorry, but this breaks user installations of the Python bindings. If we touch this, we need to first check whether it will find the library in a different path or what kind of message we need to throw to make it easy for the users to fix it.

wintersteiger:
I guess we could add the modified path in addition to the others? That way it should be safe.

delcypher:
@wintersteiger 
> Sorry, but this breaks user installations of the Python bindings.

How? The libz3 dll does not live next to the `z3core.py` file. It may have done so in the past (when the Z3 python bindings were not a package) but the dll no longer lives there (at least in the CMake build directory and for installation).

wintersteiger:
There are plenty of people that don't compile or install via the CMake system. Plenty of users copy files around manually.

PR #1355, opened 2017-11-13
[CMake] When invoking `mk_api_doc.py` pass the build directory.

delcypher:
[CMake] When invoking `mk_api_doc.py` pass the build directory.

This change was requested by @wintersteiger as an alternative way
to unbreak the TravisCI builds.


NikolajBjorner:
thanks

PR #1363, opened 2017-11-21
[DO NOT MERGE] theory for efficient recursive functions

c-cube:
The purpose of this PR is to get some feedback and reviews, as the branch is far from ready.

State of the branch:

- [x] basic `decl_plugin` (not debugged)
- [x] basic `smt::theory` plugin (not debugged)
- [x] integration in `cmd_context`
- [x] handling fairness/termination using depth limit
- [x] basic tests (with test files)
- [x] parametrize depth limit on CLI
- [ ] parametrize depth limit with a `set-option`
- [ ] iterative deepening to better explore search space

When everything works, I will rebase and squash properly.


c-cube:
My current problem is how to integrate into `cmd_context` (I see that something custom has been done for datatypes, with `m_dt_eh` field; not sure I need the same since I don't introduce new polymorphic sorts).

In particular, since `recfun::decl::plugin` and `recfun::util` are stateful, they should probably be shared between the solver(s)' context(s) and the cmd_context?

NikolajBjorner:
The declaration plugins are owned by the ast_manager objects so the state they maintain is independent of cmd_context or external API.

NikolajBjorner:
Currently, rec-funs are entered only over the cmd_context (from the parser). 
They are compiled to quantified axioms. 
This code path could change to instead use the declaration plugin for rec funs.
There could still be a path that extracts quantified axioms from recursive function declarations, but it would preferably be owned by the utilities associated with recursive functions.

For recursive functions over the API, one could so far hack recursive functions by creating quantifiers that have the same shape as recursive functions. If supporting recursive functions over the API, it may be more suitable with a dedicated API method.

NikolajBjorner:
So a starting point is to adapt cmd_context::insert_rec_fun, which is used by the parser to populate recursive functions to the plugin. Thus, create an instance of rec_fun::util and add the declaration to the plugin there. The quantifier encoding could be delayed and controlled by the plugin.

doc/design_recfuns.md
```diff
@@ -0,0 +1,93 @@
...
+
+The 2 first clauses are only added when `A_fact_0[t]` is true
+(respectively `A_fact_1[t]` is true).
+The 2 other clauses are added as soon as `fact(t)` is internalized.
+
+## Termination
+
+To ensure termination, we define variables:
+
+- `unfold_depth: int`
```
NikolajBjorner: we can add configuration options for this along with the other smt parameters: edit smt/params/smt_params_helper.pyg, smt_params.cpp and smt_params.h to add them.
Access them from the theory plugin from get_fparams(), or using the form:
     params_ref p = smt_params_helper p(get_context().get_params());
     p.max_unfound_depth()...

src/ast/recfun_decl_plugin.h
```diff
@@ -0,0 +1,264 @@
...
+        sort_ref_vector const & get_args() const { return m_arg_sorts; }
+        unsigned get_num_args() const { return m_arg_sorts.size(); }
+    };
+
+    typedef ptr_vector<var>  vars;
+
+    class case_def {
+        friend class def;
+        case_pred           m_pred; //<! predicate used for this case
+        vector<expr_ref>    m_guards; //<! conjunction that is equivalent to this case
```
NikolajBjorner: expr_ref_vector?
I know it is actually not as convenient as vector<expr_ref> when accessing elements.

src/ast/recfun_decl_plugin.h
```diff
@@ -0,0 +1,264 @@
...
+            SASSERT(m_kind == OP_FUN_MACRO);
+            return get_cases()[0].get_rhs();
+        }
+    };
+
+    class util;
+
+    namespace decl {
+        class plugin : public decl_plugin {
+            typedef map<symbol, def*, symbol_hash_proc, symbol_eq_proc> def_map;
```
NikolajBjorner: we really should have added a symbol_map to util/map.h a while ago.

src/ast/recfun_decl_plugin.h
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    // Varus utils for recursive functions
+    class util {
+        friend class decl::plugin;
+        typedef map<symbol, func_decl*, symbol_hash_proc, symbol_eq_proc> func_decl_map;
+        
+        ast_manager &           m_manager;
+        family_id               m_family_id;
+        th_rewriter             m_th_rw;
```
NikolajBjorner: so far we haven't used rewriting inside declaration utilities

src/ast/recfun_decl_plugin.h
```diff
@@ -0,0 +1,264 @@
...
+            SASSERT(m_plugin->has_case_def(s));
+            return m_plugin->get_case_def(s);
+        }
+
+        app* mk_fun_defined(def const & d, unsigned n_args, expr * const * args) {
+            func_decl * fd = 0;
+            auto& name = d.get_name();
+            if (! m_func_decls.find(name, fd)) {
+                fd = m_plugin->get_decl(d);
+                m_func_decls[name] = fd;
```
NikolajBjorner: m_func_decls[name] = fd is probably not going to work as intended. This could perhaps be fixed in our hash-tables, but m_func_decls.insert(name, fd) works too, or insert_if_not_there does the job in a single call.

src/smt/smt_setup.cpp
```diff
@@ -28,13 +28,15 @@ Revision History:
...
 #include "smt/theory_datatype.h"
+#include "smt/theory_recfun.h"
 #include "smt/theory_dummy.h"
 #include "smt/theory_dl.h"
 #include "smt/theory_seq_empty.h"
 #include "smt/theory_seq.h"
 #include "smt/theory_pb.h"
 #include "smt/theory_fpa.h"
 #include "smt/theory_str.h"
+#include "ast/recfun_decl_plugin.h" // TODO: instead, theory_recfuns
```
NikolajBjorner: theory_recfun.h could just include ast/recfun_decl_plugin.h so you don't have to import it here too.

src/smt/theory_recfun.cpp
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    bool theory_recfun::internalize_atom(app * atom, bool gate_ctx) {
+        return internalize_term(atom);
+    }
+
+    bool theory_recfun::internalize_term(app * term) {
+        TRACE("recfun", tout << "internalizing term:\n" << mk_pp(term, m()) << "\n";);
+        context & ctx = get_context();
+        unsigned num_args = term->get_num_args();
```
NikolajBjorner: You can save a few characters:

for (expr* arg : *term) {
     ...
}

src/smt/theory_recfun.cpp
```diff
@@ -0,0 +1,283 @@
...
+    void theory_recfun::assert_case_axioms(case_expansion & e) {
+        TRACE("recfun", tout << "assert_case_axioms"<< pp_case_expansion(e)<<"\n";);
+        SASSERT(e.m_def->is_fun_defined());
+        // add case-axioms for all case-paths
+        auto & vars = e.m_def->get_vars();
+        for (recfun::case_def const & c : e.m_def->get_cases()) {
+            // applied predicate to `args`
+            expr * pred_applied = apply_pred(c.get_pred(), e.m_args);
+            // substitute arguments in `path`
+            ptr_vector<expr> path;
```
NikolajBjorner: use expr_ref_vector to avoid reference count problems

src/smt/theory_recfun.cpp
```diff
@@ -0,0 +1,283 @@
...
+                expr* g_applied = apply_args(vars, e.m_args, g);
+                path.push_back(g_applied);
+            }
+            // build formula `p(args) <=> And(guards)`
+            expr * and_path = m().mk_and(path.size(), path.c_ptr());
+            expr * form = m().mk_iff(pred_applied, and_path);
+            TRACE("recfun", tout << "assert_case_axioms " << pp_case_expansion(e)
+                << " axiom " << mk_pp(form) <<"\n";);
+            // assert formula
+            literal l(smt::theory::mk_eq(form, m().mk_true(), true));
```
NikolajBjorner: iff is the same as eq, so eq(iff(x,y), true) is the same as eq(x,y).

src/ast/recfun_decl_plugin.cpp
```diff
@@ -0,0 +1,434 @@
...
+     * compilation of functions to a list of cases
+     */
+    
+    class compute_case_op; // fwd decl of operations called during exploration of cases
+    class compute_case_alternative; // fwd decl of operations to explore alternative cases
+
+    // state for compiling a function's definition into cases
+    struct compute_case_state {
+        ast_manager &               m_manager;
+        region                      m_region;
```
NikolajBjorner: Do you really need a region inside of here?
It is mostly useful in settings where a lot of memory is allocated and then reclaimed quickly, such as during backtracking.

src/ast/recfun_decl_plugin.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+        void operator()(compute_case_state & st) override;
+        void undo(compute_case_state & st) override {
+            st.m_ssubst.pop(1);
+        }
+    };
+
+    void compute_case_explore::operator()(compute_case_state & st) {
+        if (st.m().is_ite(e)) {
+            app * a = static_cast<app*>(e);
```
NikolajBjorner: use to_app(e) instead of static_cast<app*>

src/ast/recfun_decl_plugin.cpp
```diff
@@ -0,0 +1,434 @@
...
+            compute_case_op * op = new (st.reg()) compute_case_ite(st.m(), a, true);
+            st.m_ops.push_back(op);
+        }
+        else if (is_app(e)) {
+            // explore all subterms
+            app * a = static_cast<app*>(e);
+            for (unsigned i=0; i < a->get_num_args(); ++i) {
+                compute_case_op * op = new (st.reg()) compute_case_explore(a->get_arg(i));
```
NikolajBjorner: It could get tricky with using regions since destructors, and therefore reference counting on objects, will be bypassed.

src/ast/recfun_decl_plugin.cpp
```diff
@@ -0,0 +1,434 @@
...
+    // used to know which `app` are from this theory
+    struct is_imm_pred : is_immediate_pred {
+        util & u;
+        is_imm_pred(util & u) : u(u) {}
+        bool operator()(expr * rhs) {
+            struct find : public i_expr_pred {
+                util & u;
+                find(util & u) : u(u) {}
+                bool operator()(expr * e) override {
+                    if (!is_app(e)) return false;
```
NikolajBjorner: if (is_app(e)) return u.owns_app(to_app(e));

src/ast/recfun_decl_plugin.h
c-cube: I use it for substituting terms by other terms, at preprocessing time.
For example, `f(x) = p(ite(x>0, a, b))` is compiled into two cases (guarded by `x>0` and `¬(x>0)` respectively), with RHS expression `p(a)` and `p(b)` respectively.

src/smt/theory_recfun.cpp
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    theory* theory_recfun::mk_fresh(context* new_ctx) {
+        return alloc(theory_recfun, new_ctx->get_manager());
+    }
+
+    bool theory_recfun::internalize_atom(app * atom, bool gate_ctx) {
+        context & ctx = get_context();
+        if (! ctx.b_internalized(atom)) {
+            bool_var v = ctx.mk_bool_var(atom);
```
NikolajBjorner: You probably want to attach the Boolean variable with the theory.
See for example theory_array.cpp line 254.

c-cube: It doesn't seem to be sufficient, still no calls to `assign_eh` 😕

c-cube: Ah, it seems that now, marking the boolean variable as relevant does the trick. I'll dive into debugging again, thank you!!

c-cube:
Thanks, that seems to work. The branch now works on a few very simple examples. I'm now focusing on enforcing the depth limit properly (with a special predicate, that is pushed as an assumption before search starts, plus conflict clauses created when a branch becomes too deep).

NikolajBjorner:
When initializing search the context calls add_theory_assumptions on the theory solvers. This allows the solvers to add assumption literals. If the assumption literal is used in an unsat core the result is not going to be unsat. This feature is used in theory_str and from what I understand would do the trick.

c-cube:
@NikolajBjorner indeed, that's what I'm using (see https://github.com/c-cube/z3/blob/9a5afa85190aff7fdb21228d5c0898cbf78f45bc/src/smt/theory_recfun.cpp#L370). There seems to be a problem with my conflict clauses, for some reason during conflict resolution some literal is not false when it should be. I'll be investigating by myself :-)

c-cube:
Right, so I simplified the conflicts caused by the depth limit being reached, by just adding a clause (typically a conflict clause) with `mk_theory_axiom` instead of calling `ctx.set_conflict(…)`. This seems to fix my problems.

On the following problem, I get this:

```
$ ./build-rel/z3 test_sum_sat.smt2 smt.recfun.max_depth=20
sat
(model 
  (define-fun l_1 () list
    (cons 5 (cons (- 6) (cons 5 nil))))
)
$ ./build-rel/z3 test_sum_sat.smt2 smt.recfun.max_depth=10
unsat
```

so it's making progress. Next items to handle:

- the assumption on depth should not be an assertion, but a local assumption that, if used (i.e. part of the unsat-core / proof), needs to return "unknown" instead of "unsat"
- a tactic to call the core smt with increasingly big depth limits, to explore the search space in a ± fair way.

```smtlib
(set-logic ALL)

(declare-datatypes
   ()
   ((list (cons (select_cons_0 Int) (select_cons_1 list)) 
          (nil))))
(define-funs-rec
   ((append ((x list) (y list)) list))
   ((match x (((cons n tail) (cons n (append tail y))) 
              (nil y)))))
(define-funs-rec
   ((rev ((l list)) list))
   ((match l
      (((cons x_1 xs) (append (rev xs) (cons x_1 nil))) 
       (nil nil)))))
(define-funs-rec
   ((size ((l list)) Int))
   ((match l
      (((cons x_1 xs) (+ 1 (size xs)))
       (nil 0)))))
(define-funs-rec
   ((sum ((l list)) Int))
   ((match l
      (((cons x_1 xs) (+ x_1 (sum xs))) 
       (nil 0)))))

(declare-const l_1 list)

(assert (= l_1 (rev l_1)))
(assert (= (size l_1) 3))
(assert (= (sum l_1) 4))
(check-sat)
(get-model)
```

c-cube:
It seems `add_theory_assumption` is not called when proofs are enabled, is that normal? Also, I think it'd be cleaner to add the assumption from outside the theory, to later check if it's part of the proof/unsat-core and return "unknown" instead of "unsat".

NikolajBjorner:
I am not sure what you mean by "add_theory_assumptions" not being called. The code paths are in smt::context setup_and_check and in check. They invoke the theory solvers to add assumptions. I don't see where proofs play in at this level.

c-cube:
With a tracing build, on the problem above, I get "sat" with a correct model, and there is a call to `add_theory_assumption` in the trace. If I add `(set-option :produce-proofs true)` after the `set-logic` command, I get "sat" with a wrong model, and no theory assumption in the trace.

Also, there's stuff in the main SMT code path that checks the unsat-core for assumptions, and yet I get "unsat" instead of "unknown" if I call `z3 smt.recfun.max_depth=5 foo.smt2` (when the depth limit is too small).

NikolajBjorner:
Can't see what is the culprit, but note the following:
- You can use different random seeds to force different executions. For example, smt.random_seed=3. In proof mode it sometimes produces the right result with a random seed. Other times takes a very long time. The discrepancy with proof-mode vs. non-proof mode could be in code that is unrelated to assumptions and unrelated to recfun.
- There is a leak in the depth_limit_map. You probably know about this already. The entries should be deallocated when the plugin is destructed.

NikolajBjorner:
Note also, that a reduced repro of the palindrome violation for smt.random_seed = 7 is:

(assert (= l_1 (rev l_1)))
(assert (= (size l_1) 2))
;(assert (= (sum l_1) 4))
(check-sat)

NikolajBjorner:
OK, in proof mode, theory_recfun is registered with the solver too late, only after the assumption literals are initialized, so it ends up using the plugin, but without the assumption literal in play. Perhaps this speaks for your comment about a cleaner solution, which I don't really understand yet.
Here is one way of addressing it. It is probably not very good.

Before adding theory assumptions, ensure that the context has been set up and relevant theories have been registered. Add the line setup_context(false) inside the code for add_theory_assumptions().
I will be updating the master branch to do this. 
Adding theory assumptions is a relatively fresh feature.

c-cube:
I fixed the leak, and rebasing on master does indeed fix the spurious models! The opposite problem (spurious "unsat" with certain depth limits‽) only seems to occur when proofs are disabled, I'll try and find what differs.

What would be required to make the solver return "unknown" when the depth limit assumption belongs in the unsat core?

NikolajBjorner:
The method "validate_unsat_core" is virtual in smt_theory. You should override this to check that your assumption literals are not used.

NikolajBjorner:
Should we try to finish the lose ends?

c-cube:
There still are strange bugs, and a lot of work remains to be done on assessing performance and optimizing. I really don't have the time or motivation to pursue this on my own time for now (and I'm really sorry about it) so I'll close the PR, but I will come back to it later. With @grantpassmore (at [AI](https://imandra.ai)) we plan to come back to this eventually (hopefully sooner rather than later! :crossed_fingers:). Thank you for your help anyway!

NikolajBjorner:
Could I perhaps take over this pull request and finish it in a way it can be integrated and used?
Applications of the class of Suter functions come up in applications. The latest I heard of is related to when axiomatizing and reasoning about priority queues. 

- What is a good test suite? 
- Is anything significant feature wise left from your todo list?

c-cube:
If you're interested in taking it over, sure! First I had weird answers on simple problems, which may be caused by something not marked as relevant (I'm not exactly sure).

The biggest missing item on the todo list is the outer under-approximation loop which would iteratively increase the maximum depth of unrolling of functions (typically relying on unsat-cores to check whether the limit was actually used or not).

I'm not aware of a single, clean set of benchmarks to use, sadly. SMTLIB 2.6 is expressive enough to write them, but I don't think there are many benchmarks that use many functions and no quantifiers (maybe ask Andrew Reynolds). I have a few custom benchmarks in [smbc](https://github.com/c-cube/smbc) and there's the [TIP problems](https://github.com/tip-org/benchmarks) which are mostly inductive theorems: mutated version of the TIP problems (e.g. over generalizations, or removals of hypothesis) could make for good satisfiable problems too.

NikolajBjorner:
Thanks.
The smbc set seems already a very good way of testing for basic issues.

NikolajBjorner:
oh, smbc/examples uses a dated format that isn't in SMT-LIB2.6. The "case" keyword seems to have disappeared.

c-cube:
It's in TIP (https://tip-org.github.io/), yes, sorry. The smt2 files should not be hard to convert to modern SMTLIB though.

NikolajBjorner:
FYI, updates are currently going into https:/github.com/nikolajbjorner/z3.git in the csp branch.

NikolajBjorner:
I am revisiting: 
> The biggest missing item on the todo list is the outer under-approximation loop which would iteratively increase the maximum depth of unrolling of functions (typically relying on unsat-cores to check whether the limit was actually used or not).

The use of unsat cores and assumptions seems not right: if you just avoid adding max_depth_conflict and instead in the final_check measure whether the set of inserted guards is too big, then return FC_GIVEUP.

What seems missing is a check for (un)satisfiability when the guards are saturated relative to sufficient surjectivity.

c-cube:
The rationale for using an assumption + the unsat core, is that with the outer refinement loop you can increase the maximum depth progressively. Iterative deepening is convenient for finding small counter-examples and yet benefit from backtracking "locally" (with a given depth limit). Do you mean `FC_GIVEUP` would potentially allow that?

NikolajBjorner:
FC_GIVEUP doesn't give iterative deepening.
One can expand m_guards depth in final_check to throttle expansion due to functions.

Guiding case splits of guards to follow base cases would 
mimic ADT unfolding in theory_datatypes and seems to achieve the desired objective of trying base cases before expanding depth. We would set the phase for guards that prefer base cases to true and other guards to false.

NikolajBjorner:
upon further review: the guard conflicts seemed to do something I wasn't expecting.
A reasonable expectation to me is that each function has a limit on how many unfoldings are permitted that avoid base cases. So for each function we could assert:
     depth_limit(k) => base_guard_f_1 or … or base_guard_f_k
where base_guard are the predicates associated with non-recursive unfoldings for a function f.
for a different function g, we end up with similar predicates.
It's still possible this isn't fair when recursive functions have more than one non-recursive branches and we keep unfolding on just one of them.

PR #1381, opened 2017-11-28
remove smtlib1 dependencies

NikolajBjorner:
Signed-off-by: Nikolaj Bjorner <nbjorner@microsoft.com>


delcypher:
@NikolajBjorner This is an API breaking change which means the major or minor version need to be incremented. Well according to @rhelmot 's comment anyway.

https://github.com/Z3Prover/z3/blob/d7042c234f32607c01324da7a3bbf74384d0aaab/src/CMakeLists.txt#L120-L130

How you actually do it is up to you but the next public release needs to increment the minor or major version number.

NikolajBjorner:
Thanks. depending if this makes it into the "next" release.

scripts/mk_project.py
```diff
@@ -9,7 +9,7 @@
 
 # Z3 Project definition
 def init_project_def():
-    set_version(4, 5, 1, 0)
+    set_version(4, 6, 1, 0)
```
delcypher: @NikolajBjorner 
Unfortunately you need to update the version in the root `CMakeLists.txt` too.

https://github.com/Z3Prover/z3/blob/d7042c234f32607c01324da7a3bbf74384d0aaab/CMakeLists.txt#L32-L41

I never refactored this because the intention was for the CMake build system to replace the older one.

NikolajBjorner: ok

wintersteiger:
Why 4.6.1 instead of 4.6.0? There have been multiple API-breaking changes since 4.5.0, so we should definitely bump that up to 4.6.x.

PR #1391, opened 2017-12-07
[CMake] Use C++11 is C++ API example

delcypher:
@NikolajBjorner  Hopefully this will do what you want.

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,7 +1,11 @@
 RELEASE NOTES
 
-Version 4.5.x
+Version 4.6.0
 =============
+- New requirements:
+  - C++11 capable compiler to build Z3.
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
Thanks!

NikolajBjorner:
.. of course, I only realized a bit later that this is an arbitrary choice: the use of "for" forces the C++ API to be C++11 and not earlier. This was not the case before. I could have rewritten the three for loops to using iterators and avoided the issue. So in the end the right thing may be to not use C++11 features in the header file, even though applications would be able to use the new iterator that was introduced for convenience.

delcypher:
@NikolajBjorner Well a release hasn't been made yet so you can still revert back to C++98/03 if you want.

PR #1400, opened 2017-12-13
Fix memory leak in theory_str

mtrberzi:
Add trail pushes to pre-rewrite terms in str.replace and str.substr axiom setup to avoid memory leaks involving these terms.


mtrberzi:
I have a large case from PyEx that might be too big to be a good repro. I'll try to minimize it, or possibly check further to see if there is an underlying cause.

mtrberzi:
Closing this until I have time to review it in greater detail.

PR #1401, opened 2017-12-13
fixes compilation flags for OCaml plugins

ivg:
The `-linkall` option is needed for a plugin to be standalone, otherwise it will miss those dependencies that are not used. Thus, without this option, the generated `cmxs` file is unusable (i.e., it can't be used to load the z3ml in runtime, since it misses `z3.cmx` module).


NikolajBjorner:
thanks

Issue #1408, opened 2017-12-18
Python assertions negative index and slicing


bannsec:
It would be helpful to have the ability to negatively index and slice in assertions. For example:

```python
a = solver.assertions()[1] # This works
a = solver.assertions()[-1] # This currently does not
a = solver.assertions()[10:15] # Does not
# Etc
```

NikolajBjorner:
@bannsec 
Could you add a pull request for supporting such operator overloading to AstVector in z3.py?

bannsec:
Yep. Will do.

PR #1410, opened 2017-12-18
Allowing slices and negative index in assertions

bannsec:
Ref #1408, this seems to work for me. Normal index, negative index, and slices all appear to work as expected.


NikolajBjorner:
Thanks!

PR #1435, opened 2018-01-02
raise_exception: Annotate that this doesn't return.

waywardmonkeys:
This lets the compiler do better work.

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -53,6 +53,12 @@ Revision History:
 #pragma warning(disable : 4355)
 #endif
 
+#ifdef __GNUC__
+#  define Z3_NORETURN __attribute__((noreturn))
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
I'm pretty sure this is permissible usage of ``noreturn``.

NikolajBjorner:
noreturn is part of C++ so not specific to a compiler. The right approach might be to use the language defined attributes, e.g., [[ noreturn ]] void ....

waywardmonkeys:
I wasn't sure about the required compiler support for that... if that's workable, I'd be happy to do that!

NikolajBjorner:
Currently the lowest denominator is VS2012's C++11 support, which is to be deprecated in the new year when our build machines are updated.

PR #1436, opened 2018-01-02
Use noreturn attribute and __declspec version.

waywardmonkeys:


waywardmonkeys:
This should address the issues with Windows support from #1435.

PR #1453, opened 2018-01-17
Add cffi support for pypy to z3py wrapper.

forward-wfw:
In this pr, I added a standalone pypy support for z3py, which replace ctypes with cffi as the interface to libz3 library, so pypy may use it. According to experimental results, using pypy, it's at least 10% faster to use cffi than ctypes.

I make a new directory src/api/pypy/ to hold the sources. z3types.py is no longger needed. scripts/update_api.py is modified to support using either ctypes or cffi, which is tuned with a mk_make.py option `--use-cffi'. Some necessary modifications are made to mk_util.py and mk_project.py also.

Because ctypes is replaced, the z3py functions need to be invoked in cffi manner, where the passed in parameters need to be anounced like `param1 = z3.ffi.new('int[]', 8)` instead of ctypes manner.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -44,8 +44,8 @@ INSTALL_LIB_DIR=getenv("Z3_INSTALL_LIB_DIR", "lib")
 INSTALL_INCLUDE_DIR=getenv("Z3_INSTALL_INCLUDE_DIR", "include")
 INSTALL_PKGCONFIG_DIR=getenv("Z3_INSTALL_PKGCONFIG_DIR", os.path.join(INSTALL_LIB_DIR, 'pkgconfig'))
 
-CXX_COMPILERS=['g++', 'clang++']
-C_COMPILERS=['gcc', 'clang']
+CXX_COMPILERS=['clang++', 'g++']
+C_COMPILERS=['clang', 'gcc']
 CSC_COMPILERS=['csc', 'mcs']
 JAVAC=None
 JAR=None
@@ -107,6 +107,7 @@ USE_OMP=True
 LOG_SYNC=False
 GUARD_CF=False
 ALWAYS_DYNAMIC_BASE=False
+USE_CFFI=False
 
 FPMATH="Default"
 FPMATH_FLAGS="-mfpmath=sse -msse -msse2"
@@ -649,6 +650,7 @@ def display_help(exit_code):
     print("  --java                        generate Java bindings.")
     print("  --ml                          generate OCaml bindings.")
     print("  --python                      generate Python bindings.")
+    print("  --use-cffi                    use cffi instead of ctypes for Python bindings.")
     print("  --staticlib                   build Z3 static library.")
     print("  --staticbin                   build a statically linked Z3 binary.")
     if not IS_WINDOWS:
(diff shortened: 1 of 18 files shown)


NikolajBjorner:
This is pretty cool. 

Question: is cffi supposed to be Python's main FFI method or is it a self-contained package?

One concern: having two copies of z3.py is not going to work. 
One version will bitrot and unless both are checked on all builds there will be lots of regressions.

forward-wfw:
The cffi version of z3py is introduced as a self-contained package. Considering cffi also applies for CPython, it could as well be used as the main (only) FFI for z3py; The drawback is, some z3py functions exposes bare C interface to invokers with the parameters announced originally as ctypes definations, after switching to the cffi version, these invocation codes need rewriting. That's why in this pull request, I made it as a standalone package (perhaps we can call it z3pypy).

Many hot projects, like `angr`, heavily rely on z3 python wrapper as the backend, whose operations are quite cpu-dense. PyPy is encouraged to be used for such applications to save python runtime, but PyPy+ctypes is much slower than either PyPy+cffi or CPython+ctypes. So it makes much sense to provide an alternative to entirely replace ctypes with cffi in z3py module.

If the cffi version (namely z3pypy) is maintained alongside the ctypes package, the synchronization between them will definitely be painful. This could be solved by merging 2 z3.py files into 1, adding global flag and conditional invocation between ctypes and cffi methods. Not too much work, but also ugly codes and extra branches :( Considering patches to the original z3.py should be easily applied to the cffi version, perhaps maintaining both is a good idea for now.

NikolajBjorner:
I can promise you that I will make a mess out of patching two versions of z3.py. 
Having a pypy version will also put you on the hook at times you likely would rather not, and then it would just be deleted if we can't maintain it. Getting it straight now would be useful, or rather, imperative.

Are users required to install cffi or does it come with all distributions?
I seem not to have it with Python 3.6.2 from python.org but with Anaconda (3.6.1).

In the past a really painful bottleneck in the python API was some debug code that would print an expression to a string and then throw it away. I take we are well past these kinds of bottlenecks.

forward-wfw:
cffi is built-in for PyPy. It does not come with CPython; It's required for original Python users to manually install it with `pip install cffi`.

Surely I could contribute to maintaining the individual pypy version for the builds and changes to come.

As I understand it, z3py is just a thin layer wrapping the z3 library, but the interpreter and conversion between native/python data types, such as strings, can be quite annoying and slow. This problem seems far less painful with cffi, and there isn't even any need to define the former classes of z3types which wrap the opaque structure pointers. Hopefully you could find it useful on your occasions.

NikolajBjorner:
no doubt it is useful. The tax is maintenance and user-experience: so far you can download pre-built binaries and only rely on a standard python distribution.

scripts/mk_util.py
```diff
@@ -649,6 +650,7 @@ def display_help(exit_code):
     print("  --java                        generate Java bindings.")
     print("  --ml                          generate OCaml bindings.")
     print("  --python                      generate Python bindings.")
+    print("  --use-cffi                    use cffi instead of ctypes for Python bindings.")
```
delcypher: @forward-wfw The coresponding option for the CMake build system is missing. I don't see how your changes are supposed to work in the CMake build system.

scripts/update_api.py
```diff
@@ -305,17 +354,71 @@ def display_args_to_z3(params):
     for p in params:
         if i > 0:
             core_py.write(", ")
-        if param_type(p) == STRING:
-            core_py.write("_to_ascii(a%s)" % i)
-        else:
-            core_py.write("a%s" % i)
+        #if param_type(p) == STRING:
```
delcypher: Leaving commented out code lying around isn't a good idea.

scripts/update_api.py
```diff
@@ -1850,6 +2020,10 @@ def main(args):
                       dest="ml_output_dir",
                       default=None,
                       help="Directory to emit OCaml files. If not specified no files are emitted.")
+  parser.add_argument("--use-ffi",
+                      dest="use_ffi",
+                      default="cffi",
```
delcypher: ctypes probably should be the default given that it works out of the box.

forward-wfw:
Thank you delcypher. Your advices have been accepted. The corresponding modifications are implied in the latest commit: https://github.com/Z3Prover/z3/pull/1453/commits/b38e600792950f9afd218a1ac8f8b6d39333bf77

Support for cmake build system is added now. Instead of adding a new option to `build.sh`, I judged the `PYTHON_EXECUTABLE`, and if a valid path to `pypy` is specified, cffi is auto-selected as the ffi method in generating python bindings.
Some commented-out lines are corrected, and ctypes is set as default in the python build script.

Issue #1461, opened 2018-01-25
is_clang_in_gpp_form fails in mk_util.py


Firobe:
This function fails when checking the output of g++, which contains a non-ASCII character. Doing encode('utf-8') actually tries to first *decode* the string as ASCII, which throws an UnicodeDecodeError exception.

PR #1462, opened 2018-01-25
Fix encoding error

Firobe:
The encode/decode is not needed and fails if any non-ASCII character is returned by g++ or clang++
#1461

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -889,7 +889,7 @@ def is_CXX_gpp():
     return is_compiler(CXX, 'g++')
 
 def is_clang_in_gpp_form(cc):
-    version_string = check_output([cc, '--version']).encode('utf-8').decode('utf-8')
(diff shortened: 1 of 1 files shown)


wintersteiger:
The reason those encode/decoder calls are there is that it failed on other systems with different versions of gcc. I.e., we will now fail on those platforms that have been fixed the last time around...

Firobe:
Could you explain what was the problem and how the encode/decode solved it ?

I submitted the issue because gcc outputs a copyright symbol and .encode('utf-8') assumes the string is pure ASCII, which throws an exception.

wintersteiger:
Yes, that's exactly why those decode/encode calls were in there. It's fine with Python 3.x (which assumes all strings are unicode), but not with 2.x (which assumes ASCII I think). All our scripts are littered with this as we haven't found a good, general solution yet.

PR #1485, opened 2018-02-13
Remove redundant void arg.

waywardmonkeys:
While this was needed in ANSI C, it isn't in C++ and triggers a warning
in clang-tidy when `modernize-redundant-void-arg` is enabled.


wintersteiger:
As far as I remember, this is not so on Windows/MSVC...

PR #1506, opened 2018-02-26
Make z3py datatype recognizers export correctly in sexpr()

pschanely:
Love z3. Thanks!
I noticed an issue with solver.sexpr() not properly exporting algebraic datatype recognizers from the python API. (they get exported with the python-style "is_" prefix rather than "is-", which the command-line z3 binary rejects) I was able to fix it with the following change, but perhaps you have some more principled approach you'd prefer to take :)

### DIFF
--- a/src/api/python/z3/z3.py
+++ b/src/api/python/z3/z3.py
@@ -4451,7 +4451,7 @@ class Datatype:
         if __debug__:
             _z3_assert(isinstance(name, str), "String expected")
             _z3_assert(name != "", "Constructor name cannot be empty")
-        return self.declare_core(name, "is_" + name, *args)
+        return self.declare_core(name, "is-" + name, *args)
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
The "is_" prefixing is legacy, with SMT-LIB 2.6, recognizers are parametric functions. So perhaps the right fix is to use these.

NikolajBjorner:
I have updated the recognizers internally to use the SMT-LIB 2.6 convention so the change to z3.py is different from the pull request. Hope this addresses the issue you ran into.

pschanely:
Figured you'd have a better solution. I just prefer to submit pull requests over submitting bugs :)  Cheers, and thanks!

NikolajBjorner:
The pull request was much better than a bug report. So thanks yes. I retained some or most of it, but the issue pointed to that something else had to be addressed.

PR #1518, opened 2018-03-07
Make proof_checker more const correct.

waywardmonkeys:


NikolajBjorner:
Note that most of our code uses expr* instead of expr const*. It would be pedantically more appropriate with the const qualifier, but would add clutter and I can't see how this solves an acute problem.

PR #1537, opened 2018-03-13
Add support for NetBSD

khorben:
From what I can tell, this is what we (NetBSD users) need to change to let Z3 work on our platform. Unfortunately I could only compile-test it so far, but it definitely reduces the amount of patching we have to perform there. As far as I can tell it should work, and should not impact any other platform anyway.

Originally from David Holland <dholland@NetBSD.org>.


NikolajBjorner:
cool!

PR #1546, opened 2018-03-19
theory_str determinism refactoring

mtrberzi:
(DON'T MERGE YET)

Testing this patch to see if some of the collection changes I had earlier have improved things at all.

@NikolajBjorner one of the comments in your patch mentions a "curried map", how would I go about setting that up?


mtrberzi:
Okay, that patch did significantly better, looks like ASan found a few issues on some of the string benchmarks but nothing I shouldn't be able to fix soon.

NikolajBjorner:
This is super!
I actually started nibbling at some parts of the code over the weekend, but didn't push as it ended up making local improvements without really fixing anything.

Note that there are additional occurrences of maps and sets that use pointers.

This one uses a pointer:
   std::map<std::pair<expr*, zstring>, expr*> regex_in_bool_map;

(if you want to use the same kind of map, you could use the expression ids instead of pointers, 
the drawback is that you have to switch between expressions and their ids).

There is also a map std::map<std::set<expr*>, T>, 
which would require some work to port.

Inside the theory_str.cpp file, some local variables also use such maps.

Your code also uses the properties of operator[] in std::map. We could change the corresponding function in obj_hashtable to insert a default element. This does compile.

mtrberzi:
I left some of the tricky collections for last so that I could get the simple ones working first, but I will try to port the remaining ones over tomorrow. The local variables will take a bit more work but shouldn't be that bad.

Having default element semantics would be useful; would there be any performance/correctness reason not to insert default elements in obj_map? (I assume we are talking about obj_map instead of obj_hashtable.) I would be fine with working around it a bit if changing the behaviour would introduce a problem elsewhere. On the other hand, it would be more intuitive and readable to have the corresponding behaviour in that way, since I wouldn't have to do this check each time myself. I imagine if the code didn't use this before, adding it won't break anything.

NikolajBjorner:
Sorry for the confusion. I meant obj_map, defined in obj_hashtable.
The change is:

 value & find(key * k)  {
         obj_map_entry * e = find_core(k);
+        if (!e) {
+            e = insert_if_not_there2(k, value());
+        }
         SASSERT(e);
         return e->get_data().m_value;
     }

All previous code works without change. Adding default values would mask cases where a lookup is made for something that was assumed inserted.

mtrberzi:
The change you suggest doesn't compile on Clang, with an error regarding an implicitly-deleted copy constructor. 

I'll take a look at the remaining collections in the meantime.

PR #1552, opened 2018-03-23
Fix how add_git_dir_dependency() looks for .git.

### LINKED ISSUE #1551: add_git_dir_dependency() doesn't work in subrepositories
`add_git_dir_dependency()` was updated as per issue #1227 to handle git worktrees.  Unfortunately, it fails if z3 is being used as subrepository, failing with the following:

    -- Found git worktree
    CMake Error at cmake/git_utils.cmake:37 (message):
      Found git worktree dir but could not find
      "../../.git/modules/repos/z3/commondir"

I suggest that instead of checking if .git is a file and trying to follow it via cmake, you instead use `git rev-parse --git-common-dir` and `git rev-parse --git-dir` instead.   This former invocation will get you the base .git dir, and the latter will get you the directory with the HEAD file.  This should work with all .git repositories: worktrees, subrepos, and normal.

sei-mwd:
add_git_dir_dependency() adds dependencies on files in the .git
directory.  But in certain scenarios, .git is a file instead of a
directory.  Manual cmake code was written to handle this case for git
worktrees.  Unfortunately, the solution is slightly different for git
subrepositories.  Switching to using `git rev-parse` to determine actual
directory locations fixes this problem.

Fixes #1551

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ set(GIT_DIR "${CMAKE_SOURCE_DIR}/.git")
 if (EXISTS "${GIT_DIR}")
   # Try to make CMake configure depend on the current git HEAD so that
   # a re-configure is triggered when the HEAD changes.
-  add_git_dir_dependency("${GIT_DIR}" ADD_GIT_DEP_SUCCESS)
+  add_git_dir_dependency("${CMAKE_SOURCE_DIR}" ADD_GIT_DEP_SUCCESS)
   if (ADD_GIT_DEP_SUCCESS)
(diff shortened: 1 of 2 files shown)


sei-mwd:
I'm going to have to get my employer's permission before I can sign the MS Contributor License Agreement, so it might be a bit.

sei-mwd:
I am advised by our lawyers that I cannot sign the license/cla due to the patent clause.  They will, however, let me put this patch into the public domain.  I do not really know how to proceed from here.

nunoplopes:
can't merge then, sorry.

PR #1557, opened 2018-03-30
Use stdint.h for int64_t / uint64_t in API.

waywardmonkeys:
Now that we can use stdint.h, we can use it to portably define
64 bit integer types for use in the API.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -60,7 +60,7 @@ FIRST_OBJ_ID = 100
 def is_obj(ty):
     return ty >= FIRST_OBJ_ID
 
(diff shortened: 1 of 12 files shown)


waywardmonkeys:
This handles the API. I didn't go and remove all usages of `long long` and `unsigned long long` inside the codebase. (yet?)

NikolajBjorner:
I guess we also need to update src/util/util.h

home/user/z3_src/src/api/api_datalog.cpp:216:65: error: non-const lvalue reference to type 'uint64' (aka 'unsigned long long') cannot bind to a value of unrelated type 'uint64_t' (aka 'unsigned long')

waywardmonkeys:
I will get that fixed up. Interesting that it works on macOS (where I built and tested)!

waywardmonkeys:
I'm going to get things working via this PR ... but it apparently might take some time ... and now I'm not particularly happy about how it is going (but I still think it is the right thing to do!)

PR #1560, opened 2018-04-03
improve performance of occurs check for datatypes

c-cube:
the goal of this change is to improve the performance of the acyclicity check for datatypes. It replaces the per-node occur check (which duplicates some work) by a global graph traversal with cycle detection.

Note: I don't know how to run tests on my local repo (it seems to work from a project that uses bindings, but more vetting is clearly needed)

### DIFF
--- a/src/smt/theory_datatype.cpp
+++ b/src/smt/theory_datatype.cpp
@@ -389,12 +389,14 @@ namespace smt {
     final_check_status theory_datatype::final_check_eh() {
         int num_vars = get_num_vars();
         final_check_status r = FC_DONE;
+        init_final_check();
         for (int v = 0; v < num_vars; v++) {
             if (v == static_cast<int>(m_find.find(v))) {
                 enode * node = get_enode(v);
                 if (occurs_check(node)) {
                     // conflict was detected... 
                     // return...
+                    cleanup_final_check();
                     return FC_CONTINUE;
                 }
                 if (m_params.m_dt_lazy_splits > 0) {
@@ -407,6 +409,7 @@ namespace smt {
                 }
             }
         }
+        cleanup_final_check();
         return r;
     }
 
@@ -434,6 +437,8 @@ namespace smt {
                       tout << "eq: #" << p.first->get_owner_id() << " #" << p.second->get_owner_id() << "\n";
                       tout << mk_bounded_pp(p.first->get_owner(), get_manager()) << " " << mk_bounded_pp(p.second->get_owner(), get_manager()) << "\n";
                   });
+        } else {
+            oc_mark_cycle_free(n);
         }
         return res;
     }
@@ -443,12 +448,11 @@ namespace smt {
        TODO: improve performance.
     */
     bool theory_datatype::occurs_check_core(enode * app) {
-        if (app->is_marked())
+        if (oc_cycle_free(app) || oc_explored(app))
             return false;
         
         m_stats.m_occurs_check++;
-        app->set_mark();
-        m_to_unmark.push_back(app);
+        oc_mark_explore(app);
         
         TRACE("datatype", tout << "occurs check_core: #" << app->get_owner_id() << " #" << m_main->get_owner_id() << "\n";);
 
--- a/src/smt/theory_datatype.cpp
+++ b/src/smt/theory_datatype.cpp
@@ -412,7 +412,7 @@ namespace smt {
     }
 
     // explain the cycle root -> … -> app -> root
-    void theory_datatype::occurs_check_explain(enode * app, enode * const root) {
+    void theory_datatype::occurs_check_explain(enode * app, enode * root) {
         TRACE("datatype", tout << "occurs_check_explain " << mk_bounded_pp(app->get_owner(), get_manager()) << " <-> " << mk_bounded_pp(root->get_owner(), get_manager()) << "\n";);
         enode* app_parent = nullptr;
 
@@ -424,8 +424,7 @@ namespace smt {
             SASSERT(d->m_constructor);
             if (app != d->m_constructor)
                 m_used_eqs.push_back(enode_pair(app, d->m_constructor));
-            bool found = m_parent.find(app->get_root(), app_parent);
-            SASSERT(found);
+            app_parent = m_parent[app->get_root()];
             app = app_parent;
         }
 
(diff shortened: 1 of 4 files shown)


NikolajBjorner:
nice. It will need a review, but overall is well structured.

Regarding enode equality: we normally use pointer-equality because enodes are unique given an owner. The idea is that smt_internalize checks a cache to see if an expression is already internalized, then if it isn't creates fresh enodes.

src/smt/theory_datatype.h
```diff
@@ -73,11 +72,58 @@ namespace smt {
         void propagate_recognizer(theory_var v, enode * r);
         void sign_recognizer_conflict(enode * c, enode * r);
 
-        ptr_vector<enode>    m_to_unmark;
-        enode_pair_vector    m_used_eqs;
-        enode *              m_main;
+        typedef enum { ENTER, EXIT } stack_op;
+        typedef map<enode*, enode*, obj_ptr_hash<enode>, ptr_eq<enode> > parent_tbl;
```
NikolajBjorner: I have to confess that map<enode* made me worried, but I can see it is fine given that you use obj_ptr_hash.
Just don't use ptr_hash as it leads to non-determinism across platforms and runs if there is any code that depends on iterating over the table.

src/smt/theory_datatype.h
```diff
@@ -73,11 +72,58 @@ namespace smt {
...
+            m_stack.push_back(std::make_pair(EXIT, n));
+            m_stack.push_back(std::make_pair(ENTER, n));
+        }
+
+        // class for managing state of final_check
+        class final_check_st {
+            theory_datatype * th;
+            public:
+            final_check_st(theory_datatype * th) : th(th) {
+                th->m_to_unmark.reset();
```
NikolajBjorner: I would assert that they are empty already.

src/smt/theory_datatype.cpp
```diff
@@ -410,6 +411,59 @@ namespace smt {
         return r;
     }
 
+    // explain the cycle root -> … -> app -> root
+    void theory_datatype::occurs_check_explain(enode * app, enode * const root) {
```
NikolajBjorner: Do you really intend to say that enode * const root instead of enode const* root?

src/smt/theory_datatype.cpp
```diff
@@ -410,6 +411,59 @@ namespace smt {
...
+
+        while (app->get_root() != root->get_root()) {
+            theory_var v = app->get_root()->get_th_var(get_id());
+            SASSERT(v != null_theory_var);
+            v = m_find.find(v);
+            var_data * d = m_var_data[v];
+            SASSERT(d->m_constructor);
+            if (app != d->m_constructor)
+                m_used_eqs.push_back(enode_pair(app, d->m_constructor));
+            bool found = m_parent.find(app, app_parent);
```
NikolajBjorner: most maps have operator[] and unary "find" to access the map contents, asserting the key is in the domain.
This is different from stl semantics that inserts default elements.

c-cube:
I updated the PR, but what's really missing is running the result on regression tests (and maybe also assess more thoroughly the perf difference on smt-lib?)

NikolajBjorner:
This will be possible for me on Friday.

- Nikolaj

src/smt/theory_datatype.cpp
```diff
@@ -389,10 +389,11 @@ namespace smt {
     final_check_status theory_datatype::final_check_eh() {
         int num_vars = get_num_vars();
         final_check_status r = FC_DONE;
+        final_check_st _guard(this); // RAII for managing state
         for (int v = 0; v < num_vars; v++) {
             if (v == static_cast<int>(m_find.find(v))) {
                 enode * node = get_enode(v);
-                if (occurs_check(node)) {
+                if (!oc_cycle_free(node) && occurs_check(node)) {
```
NikolajBjorner: we only need to occurs check expressions of a datatype sort. 
Both the old and new code postpone these filters until after a stack has been updated.

src/smt/theory_datatype.cpp
```diff
@@ -418,17 +493,39 @@ namespace smt {
...
+            enode * app = m_stack.back().second;
+            m_stack.pop_back();
+
+            if (oc_cycle_free(app)) continue;
+
+            TRACE("datatype", tout << "occurs check loop: #" << app->get_owner_id() << " " << mk_bounded_pp(app->get_owner(), get_manager()) << (op==ENTER?" enter":" exit")<< "\n";);
+
+            switch (op) {
+            case ENTER:
+              res = occurs_check_enter(app) || res;
```
NikolajBjorner: Why are you performing an occurs check here if res is already true?

src/smt/theory_datatype.cpp
```diff
@@ -410,6 +411,80 @@ namespace smt {
...
+            enode * app_cstor = oc_get_cstor(app);
+            if (app != app_cstor)
+                m_used_eqs.push_back(enode_pair(app, app_cstor));
+            app_parent = m_parent[app->get_root()];
+            app = app_parent;
+        }
+
+        SASSERT(app->get_root() == root->get_root());
+        if (app != root)
+          m_used_eqs.push_back(enode_pair(app, root));
```
NikolajBjorner: '  '

src/smt/theory_datatype.cpp
c-cube: If we start from variables of this theory, shouldn't they all be of a datatype sort? Or do these variables also cover recognizers?

NikolajBjorner: Correct, so my comment should be retracted

NikolajBjorner:
It passes SMT-LIB2's QF_DT benchmarks. They are all very easy anyhow.
The main build problem seems to be based on memory management of explanations.

NikolajBjorner:
I added classes to access parents and args from enodes in smt_enode.h to make it simpler to write for-loops.

c-cube:
- regarding for loop: you mean using something like `for (auto & arg : enode.get_args()) …` I suppose?
- regarding memory consumption: currently the implementation has a hashtable that eventually maps every datatype variable to a parent, but is reset after each `final_check_eh`. Is this too much still?

NikolajBjorner:
loops: yes it is basically that. The current method is enode::args(n), but I could add a method to create this class.
memory: probably a side-effect of a regression break:

ist_example

valid

valid

valid

valid

valid

valid

valid

Formula (=> ((_ is (cons (Int int_list) int_list)) u) (= u (cons (head u) (tail u))))

valid

invalid

counterexample:

u -> nil

See the output of Travis

PR #1562, opened 2018-04-06
automata-based regex engine for Z3str3

mtrberzi:
This patch adds automata-based regex solving for theory_str. Applicable parameter is `smt.str.regex_automata`: defaults to True which uses the automata solver, setting it to False uses the legacy Z3str2 regex procedure (reduction to word equations and unrolling).

Currently this is leaking a few expressions that seem to be produced somewhere in the symbolic automata methods. We can start reviewing the code now and maybe track those down as part of that process.


NikolajBjorner:
build passed. Should I merge?

NikolajBjorner:
Could you share regression tests that expose memory leaks?
Then they can be added to z3test.

mtrberzi:
Sure, I will PR the tests on that repo. I should make sure they pass before we merge.

mtrberzi:
I opened a PR on the Z3test repo for the case that causes the leak. I tried tracking down the leak with AddressSanitizer and the trace pointed me to line 10570 of theory_str.cpp where I compute an automaton product. The trace also shows that possibly `m_states1` and `m_states2` aren't getting cleaned up properly in automaton.h. Do I need to do something special to free memory after creating an automaton in this way?

mtrberzi:
Never mind, this was easy :) I forgot that `ptr_vector` doesn't call destructors on its elements when it goes away, so I do this manually.

I have a commit that fixes this, but I will push once the PR in Z3test is merged, so we can confirm that this passes the test.

NikolajBjorner:
I won't merge the new test at this point. I breaks with current bits. So better push your fixes and we can merge them.

mtrberzi:
OK, no worries, I've pushed the fix.

PR #1612, opened 2018-05-07
`lp::permutation_matrix::values` attempts an incorrect conversion

danielschemmel:
This causes compilation with GCC 8 to fail. I suspect it worked previously due to SFINAE (`permutation_matrix` is a class template).

As far as I can tell, `lp::permutation_matrix::values` is actually unused, so an alternative might be to completely remove this function.


NikolajBjorner:
@levnach, consider this fix for other branches.

Issue #1623, opened 2018-05-18
OCaml API fails to build on OCaml 4.06.1


sb98052:
When trying to build the OCaml API via mk_make.py, I get this:

ocamlc  -I api/ml -o api/ml/z3.cmi -c api/ml/z3.mli
File "api/ml/z3.mli", line 1159, characters 35-50:
Error: Unbound module Big_int
make: *** [api/ml/z3.cmo] Error 2

gwsapan-mbp:build gwsapan$ ocaml --version
The OCaml toplevel, version 4.06.1

The problem appears to be that Big_int is no longer distributed as part of the standard library, and the `num` library is not linked with.

PR #1624, opened 2018-05-18
Link in Big_int via num package

sb98052:
Big_int is no longer included in the standard library, and must be linked in via the num package: https://github.com/ocaml/num

Addresses issue #1623


NikolajBjorner:
thanks

PR #1630, opened 2018-05-19
Fix GCC Warnings

danielschemmel:
Having recently built Z3 with GCC 8.1, I spent some time hunting down all current warnings on master. While performance in release mode should theoretically not be significantly affected, this relies on the optimizer doing its job in practice.

Changes are grouped into commits by the warning that they used to cause.

At this point, only a single warning remains:
```
../src/test/hwf.cpp: In function ‘void bug_is_int()’:
../src/test/hwf.cpp:105:21: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
     double   val = *(double*)(raw_val);
                     ^~~~~~~~~~~~~~~~~~
```
While I would suggest storing the `raw_val` as an array of bytes (`std::byte`, `char` or `unsigned char`) and using `memcpy` to initialize `val`, I am unsure as of the semantics of this function.


src/util/buffer.h
```diff
@@ -39,7 +40,7 @@ class buffer {
     void expand() {
         unsigned new_capacity = m_capacity << 1;
         T * new_buffer        = reinterpret_cast<T*>(memory::allocate(sizeof(T) * new_capacity));
-        memcpy(new_buffer, m_buffer, m_pos * sizeof(T));
+        std::uninitialized_copy(std::make_move_iterator(m_buffer), std::make_move_iterator(m_buffer + m_pos), new_buffer);
```
NikolajBjorner: @nunoplopes can you check this fix?

NikolajBjorner:
Very cool. Asking @nunoplopes to check one change that uses some features I am haven't used before.

src/util/buffer.h
danielschemmel: @NikolajBjorner This change is analogue to what the LLVM people did in their [SmallVec class](http://llvm.org/doxygen/SmallVector_8h_source.html#l00194) when they encountered a similar issue.

NikolajBjorner: I have no doubts the fix is correct. Nuno changed this code last, so it is better to make sure he is aware.

nunoplopes: This change doesn't make sense. The memcpy is copying initialized memory only, so there's not undefined behavior here.

nunoplopes:
the rest of the patch looks good to me; only that uninitialized_copy change should be left out.

src/util/buffer.h
danielschemmel: There are two problems that are fixed by this change:
- Using `memcpy` to copy one initialized object into another is only legal if `T` is a [trivally copyable type](http://en.cppreference.com/w/cpp/types/is_trivially_copyable).
- `new_buffer` points to uninitialized memory, therefore new objects need to be created in the target buffer, which `memcpy` does not do.

Actually, looking at the code again, it might be more useful to either enforce [trivial destructible](http://en.cppreference.com/w/cpp/types/is_destructible) `T`s, to add destructor calls in `buffer<T>::free_memory`.

nunoplopes: I agree that memcpy is not always safe. However, not all code is Z3 is safe to change to something else, due to missing copy constructors, etc.
These kind of changes are actually high risk even though they don't look like. Without auditing all users of buffer, I can't sign off such a change. I suggest this change is dropped.

danielschemmel: fair enough, I am dropping the commit from the pull request

danielschemmel:
The commit consisting of the uninitialized_copy change has been removed.

danielschemmel:
@nunoplopes would you be interested in a pull request that has a deeper look at the buffer problematic? Otherwise I will just silence the warning in my build...

nunoplopes:
@danielschemmel I'm happy to review such a change, but we need evidence that no copies are being introduce because of some user not having the right copy constructor, etc. All uses of buffer need to be audited.
I've done similar changes to what you are proposing in the past to vector, for example, and it took a few iterations before all the bugs were fixed. Even this CallConstructors template parameter could go away at some point (by detecting if that's needed automatically).

Issue #1629, opened 2018-05-18
[Question] Z3py changing contexts


JoranHonig:
Hi, I am using z3py and would like to do some computations in parallel.
After some research, I found that this would require the use of different Context objects, due to thread safety. 
Manually changing the context of expressions results in the following exception:
``` z3.z3types.Z3Exception: b'invalid dec_ref command' ```

So now I'm wondering if setting the ctx variable for an expression to a new context is the correct approach.

NikolajBjorner:
Expressions and contexts belong together: you cannot mix expressions created from different contexts.
If you want to copy expressions from one context to another, do it in code that does not run in parallel with anything that touches the two contexts, then use the "translate" methods to perform the deep copy operations.

JoranHonig:
Thanks for the answer! Translate seems like the thing I was looking for.
The following dir contains some python examples: examples/python. Would you accept a pr from me with a parallel python example? ( that might be usefull )

NikolajBjorner:
Gladly yes

PR #1631, opened 2018-05-19
Parallel python example

JoranHonig:
I recently posted #1629, to find out how to properly change context objects.

In this pull request, I created a small parallel python example that both creates a new expression, and does some solving.

### DIFF
--- a/examples/python/parallel.py
+++ b/examples/python/parallel.py
@@ -13,7 +13,10 @@ def calculate(x, n, ctx):
     assert x.ctx == ctx
     assert x.ctx != main_ctx()
 
+    # Parallel creation of z3 object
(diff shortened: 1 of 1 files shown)


examples/python/parallel.py
```diff
@@ -0,0 +1,36 @@
...
+    solver.add(condition)
+    solver.check()
+
+
+for i in range(100):
+    # Create new context for the computation
+    # Note that we need to do this sequentially, as parallel access to the current context or its objects
+    # will result in a segfault
+    i_context = Context()
+    x_i = deepcopy(x).translate(i_context)
```
NikolajBjorner: is deepcopy(x) required?

PR #1640, opened 2018-05-23
Z3_TRUE/Z3_FALSE should be true/false, not 1/0.

waywardmonkeys:
Now that Z3_bool is a C bool, the associated constants should
be as well.


waywardmonkeys:
I noticed this while updating the Rust bindings.

PR #1646, opened 2018-05-24
Remove depedencies on interp 

NikolajBjorner:
Remove dependencies on interpolation features.

Background: small changes in preprocessing simplification and proof verbs have led to regressions in the interpolation features. They are not being maintained so with other changes going into Z3 it is a time to pull this feature allows to focus on quality elsewhere. In some cases, you can use the SPACER and PDR solvers to extract interpolants as solutions to non-recursive sets of Horn clauses. You will have to disable in-lining transformations because they just introduce quantified interpolants that are likely to be useless for applications that require quantifier-free interpolants.


PhilippWendler:
Would it be possible to provide a tutorial or even an example using the C or Python API for how to generate interpolants with one of the suggested replacements? I assume we are not the only ones who have been using Z3 interpolation, and thus I would think this could be of significant help.

Heizmann:
This is a pity. We are using the interpolation of Z3 in Ultimate Automizer and there are verification problems that we can only solve in a setting where we use the interpolants of Z3.
I have to admit that it run not always smoothly, but this was not a problem for us. In case Z3 failed, our tool just tried something else.

Shameless self-plug: As an alternative to interpolation in software verification we implemented an unsat-core-based version of an old approach from the Newton/Slam toolkit.
The [evaluation of our FSE 2017 paper](https://dl.acm.org/citation.cfm?id=3106307) shows that the unsat-core based approach can (at least for our tools) be an alternative to interpolation.

So, it would be nice if we could get the interpolant support back some time and it is vital for us that you never drop the support for unsat-cores.

NikolajBjorner:
Thanks for the feedback. At this point you can use interpolants from 4.7.1, but not in newer branches unless you can use the Horn clause solvers (that will not have the same capabilities out of the box, tutorial update to follow). The master branch is still going to churn quite a bit in the coming months and these changes are likely to break interpolation even further. So without active support it is not sustainable. This does not exclude options for what can be done in future versions. 

Unsat cores are integral to so many uses and they are not being touched. 
There is also core minimization as options for cores produced by the SMT core and SAT solver.

PR #1650, opened 2018-05-25
Fix missing SONAME in libz3.so, which breaks loading from Java

PhilippWendler:
This patch sets the `SONAME` field of `libz3.so` to `libz3.so`.

The reason is that otherwise I cannot load the binaries for Z3 from Java.

The following code just attempts to load the native libraries, without even attempting to access the API:
```
System.loadLibrary("z3");
System.loadLibrary("z3java");
```
This fails in the second line with the following error, although `libz3` could be loaded successfully in the first line, and both libraries are in the same directory, which I put in `java.library.path`:
```
Exception in thread "main" java.lang.UnsatisfiedLinkError: .../libz3java.so: libz3.so: cannot open shared object file: No such file or directory
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1941)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1857)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at Test.main(Test.java:27)
```
(Tested with OpenJDK 1.8.0_171 on Ubuntu 16.04 64bit with Z3 4.7.1, both self-compiled and the official release.)

I managed to find out that this is due to a missing `SONAME` field of `libz3.so`:
After running the following command, the same code above works:
```
patchelf --set-soname libz3.so libz3.so
```
The binaries produced by the build system of Z3 do not have `SONAME` set (`readelf -d libz3.so | grep SONAME` returns nothing).

I assume the loading fails because `libz3java` has a dependency on `libz3.so`, and it is not detected that this dependency is satisfied by the previously loaded `libz3.so` if the latter does not have a matching `SONAME`.

So I suggest to adjust the build system to set the `SONAME` field of the `libz3.so` library to `libz3.so`. This can be done by passing `-Wl,-soname,libz3.so` to the linker. As far as I understood the Z3 build system, we can simply add this to `SLINK_EXTRA_FLAGS`. I am not aware of any side effects of this.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -2460,7 +2460,7 @@ def mk_config():
         check_ar()
         CXX = find_cxx_compiler()
         CC  = find_c_compiler()
-        SLIBEXTRAFLAGS = ''
+        SLIBEXTRAFLAGS = '-Wl,-soname,libz3.so'
         EXE_EXT = ''
         LIB_EXT = '.a'
         if GPROF:


NikolajBjorner:
Our OSX machines don't like this.
Do you have a suggestion for adjusting the parameters?

a muz/dataflow/dataflow.a muz/base/muz.a duality/duality.a qe/qe.a tactic/sls/sls_tactic.a smt/tactic/smt_tactic.a tactic/bv/bv_tactics.a smt/smt.a smt/proto_model/proto_model.a smt/params/smt_params.a ast/rewriter/bit_blaster/bit_blaster.a ast/pattern/pattern.a ast/fpa/fpa.a parsers/smt2/smt2parser.a cmd_context/extra_cmds/extra_cmds.a cmd_context/cmd_context.a interp/interp.a ackermannization/ackermannization.a tactic/aig/aig_tactic.a math/subpaving/tactic/subpaving_tactic.a nlsat/tactic/nlsat_tactic.a tactic/arith/arith_tactics.a sat/tactic/sat_tactic.a solver/solver.a ast/proofs/proofs.a tactic/core/core_tactics.a math/euclid/euclid.a math/grobner/grobner.a parsers/util/parser_util.a ast/substitution/substitution.a tactic/tactic.a model/model.a ast/normal_forms/normal_forms.a ast/macros/macros.a ast/rewriter/rewriter.a ast/ast.a math/subpaving/subpaving.a math/realclosure/realclosure.a math/interval/interval.a math/automata/automata.a math/simplex/simplex.a math/hilbert/hilbert.a util/lp/lp.a nlsat/nlsat.a sat/sat.a math/polynomial/polynomial.a util/util.a -Wl,-soname,libz3.so
2018-05-25T14:47:52.5102490Z ld: unknown option: -soname
2018-05-25T14:47:52.5124460Z clang: error: linker command failed with exit code 1 (use -v to see invocation)
2018-05-25T14:47:52.5148870Z make: *** [libz3.dylib] Error 1
2018-05-25T14:47:52.5166250Z make: *** Waiting for unfinished jobs....
2018-05-25T14:47:55.9032840Z Content of build/master/release:
2018-05-25T14:47:55.9050090Z Traceback (most recent call last):
2018-05-25T14:47:55.9068040Z   File "scripts/smalltest.py", line 29, in <module>
2018-05-25T14:47:55.9084530Z     smalltest()
2018-05-25T14:47:55.9100400Z   File "scripts/smalltest.py", line 16, in smalltest
2018-05-25T14:47:55.9117880Z     util.buildz3(branch=b, everything=False, clean=True, debug=d, dotnet=False, java=False, static=False, jobs=config.NUMJOBS, clang=False)
2018-05-25T14:47:55.9135280Z   File "/Users/nightly/myagent/_work/2/s/z3test/scripts/util.py", line 174, in buildz3
2018-05-25T14:47:55.9151910Z     make(branch, debug, everything, clang, jobs)
2018-05-25T14:47:55.9169440Z   File "/Users/nightly/myagent/_work/2/s/z3test/scripts/util.py", line 151, in make
2018-05-25T14:47:55.9186430Z     raise Exception("Failed to make Z3\n%s\n" % cmd)
2018-05-25T14:47:55.9202420Z Exception: Failed to make Z3
2018-05-25T14:47:55.9220120Z ['make', '-j', '16']
2018-05-25T14:47:55.9229560Z

PhilippWendler:
I have no idea about OS X. I would suggest to avoid adding this parameter on OS X.

I am not sure what is the best way to make this distinction in the Z3 build system?

NikolajBjorner:
Could you add a pull request where this gets moved to somewhere where the flags are tested on target platform?

I guess it would be (1) reverting the update from your pull request, (2) adding something along the lines of:

       elif sysname == 'Linux':
            CXXFLAGS       = '%s -D_LINUX_' % CXXFLAGS
            OS_DEFINES     = '-D_LINUX_'
            SO_EXT         = '.so'
            LDFLAGS        = '%s -lrt' % LDFLAGS
            SLIBFLAGS      = '-shared'
            SLIBEXTRAFLAGS = '%s -lrt' % SLIBEXTRAFLAGS
            SLIBEXTRAFLAGS = '%s -Wl,-soname,libz3.so' % SLIBEXTRAFLAGS

PhilippWendler:
Ok, you beat me to it :-) Thanks!

NikolajBjorner:
sorry, was fixing some other issue and didn't know precisely when I would figure it out.

PR #1651, opened 2018-05-25
Revert "Fix missing SONAME in libz3.so, which breaks loading from Java"

NikolajBjorner:
Reverts Z3Prover/z3#1650


NikolajBjorner:
commit d66f2af45 applies possible fix.

Issue #1632, opened 2018-05-20
Modernization of `buffer<T>`


danielschemmel:
As originally discussed in #1630, the current `buffer<T>` implementation can possibly be improved upon.

I would like to submit this analysis for your consideration, so that an eventual pull request matches your expectations.

### Properties of the Current Implementation
- The current implementation restricts its capacity to 2^31, which effectively saves 8 bytes on 64 bit machines
  - The restriction 2^31 follows from how `buffer::expand` works and from `buffer::INITIAL_SIZE` always being a power of 2
  - Saving 8 bytes happens because `buffer::m_pos` and `buffer::m_capacity` should otherwise be `std::size_t`, which requires 4 bytes more on 64 bit targets. For most compilers, padding should not be introduced in either case.
- Relies on the user deciding whether destructors should be explicitly called (`buffer`'s `CallDestructors` template argument).
  - Sometimes the user may decide in a way that is inconsistent with C++ rules (the destructor may not be trivial in the C++ sense, but the use may decide to not call it anyway)
  - Sometimes the user may be wrong not just in the strict sense (cf. [`sbuffer`](https://github.com/Z3Prover/z3/blob/6938c76950f21c0e7bb958376604cb0fc4811784/src/util/buffer.h#L260-L265), [`smt::coeff_expr`](https://github.com/Z3Prover/z3/blob/6938c76950f21c0e7bb958376604cb0fc4811784/src/smt/theory_arith.h#L991), [`smt::theory_arith<Ext>::display_coeff_exprs(std::ostream&, sbuffer<coeff_expr> const&) const`](https://github.com/Z3Prover/z3/blob/6938c76950f21c0e7bb958376604cb0fc4811784/src/smt/theory_arith.h#L1129) considering [`~rational`](https://github.com/Z3Prover/z3/blob/6938c76950f21c0e7bb958376604cb0fc4811784/src/util/rational.h#L67)).
  - Maybe somewhat surprisingly, this optimization can easily be done by the compiler: If `~T()` is trivial, [the loop](https://github.com/Z3Prover/z3/blob/6938c76950f21c0e7bb958376604cb0fc4811784/src/util/buffer.h#L49-L53) becomes an empty loop over a pointer range, which can then be removed, (and so forth)
  - The same problem also exists in the opposite direction, some types which do not need to be destroyed (e.g. `bool` and `char`) exist in both specializations, which at least introduces unnecessary types
[...]

PR #1656, opened 2018-05-26
[WIP] buffer and vector modernization

danielschemmel:
Modernization of the buffer and vector classes as discussed in #1632. Please feel free to leave any comments to ensure that this is going in the right direction.

At this moment, compilation of the `z3` and `test-z3` targets succeed. Some random SMT queries have also succeeded in the past. Performance seems mostly similar to before.

### TODO:
- [ ]  Modernize interface
  - [x] `insert` => `push_back` (minimal changes)
  - [x] `data` *type* => `value_type` (minimal changes)
  - [x] mark all single-valued (non copy or move) constructors explicit (minimal changes)
  - [x] `reset` => `clear` (this causes *a lot* of changes)
  - [x] add a non-member `swap` (unifies interface w/ std::swap) (minimal changes)
  - [x] remove templated resize function hack for vector (minimal changes)
  - [x] replace `get` and `set` with `operator[]` (small changes)
  - [x] rename `reserve` to `expand` (`reserve` has a different meaning in std) (medium changes)
  - [ ] replace usage of `c_ptr` with `data` wherever possible (large amount of changes required)
  - [ ] add a `buffer_view` class and use it to replace C pointers
  - ~[ ] change `buffvec(size_type, pointer)` to `buffvec(pointer, size_type)` (deals with multiple interpretations of `vector<unsigned> vec(3, 0)` and brings the ctor more in line with its c-style heritage)~
- [x] Test VC++ for warnings/problems related to these changes
- [ ] Clean up hash implementations (currently exists in 2 different forms)
- [ ] Add additional assertions that check for overflow in variables of `size_type`
- [ ] Unify interfaces for buffer and vector
- [ ] Unify interfaces of other containers (e.g., `ref_vector`)
- [ ] Final Cleanup


NikolajBjorner:
Besides the AVs, VS 2017 runs into a compilation issue.

osrc\util\CMakeFiles\util.dir\sexpr.cpp.obj /Fdsrc\util\CMakeFiles\util.dir\ /FS -c ..\src\util\sexpr.cpp
D:\ds\z3\src\util/ref_vector.h(46): error C2146: syntax error: missing ';' before identifier 'size_type'
D:\ds\z3\src\util/ref_vector.h(185): note: see reference to class template instantiation 'ref_vector_core<T,Ref>' being compiled
D:\ds\z3\src\util/ref_vector.h(46): error C2868: 'ref_vector_core<T,Ref>::size_type': illegal syntax for using-declaration; expected qualified-name

danielschemmel:
The AV should be fixed and while I expect to have removed the construct that gave VS2017 hickups, I cannot check VS compilation right now.

Also, e.g. ` c_maxsat_example` errors out like this on master as well:
```
../src/util/rational.h:32:46: runtime error: reference binding to null pointer of type 'struct synch_mpq_manager'
../src/util/rational.h:67:26: runtime error: member call on null pointer of type 'struct synch_mpq_manager'
../src/util/mpq.h:187:9: runtime error: member call on null pointer of type 'struct mpq_manager'
../src/util/mpq.h:184:48: runtime error: member access within null pointer of type 'struct mpq_manager'
../src/util/mpq.h:184:48: runtime error: member call on null pointer of type 'struct mpz_manager'
../src/util/mpq.h:188:9: runtime error: member call on null pointer of type 'struct mpq_manager'
```

NikolajBjorner:
The compilation problem is gone. There are several warnings.
d:\ds\src\util\buffvec.h(708): note: see reference to function template instantiation 'void buffvec<generic_model_converter::entry,unsigned int,0>::reallocate<generic_model_converter::entry>(unsigned int)' being compiled
d:\ds\src\util\buffvec.h(708): note: see reference to function template instantiation 'void buffvec<generic_model_converter::entry,unsigned int,0>::reallocate<generic_model_converter::entry>(unsigned int)' being compiled
d:\ds\src\util\buffvec.h(706): note: while compiling class template member function 'void buffvec<generic_model_converter::entry,unsigned int,0>::push_back(generic_model_converter::entry &&)'
D:\ds\src\tactic/generic_model_converter.h(55): note: see reference to function template instantiation 'void buffvec<generic_model_converter::entry,unsigned int,0>::push_back(generic_model_converter::entry &&)' being compiled
D:\ds\src\tactic/generic_model_converter.h(43): note: see reference to class template instantiation 'buffvec<generic_model_converter::entry,unsigned int,0>' being compiled
[390/394] Building CXX object src\shell\CMakeFiles\shell.dir\smtlib_frontend.cpp.obj
D:\Programs\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.11.25503\include\memory(53): warning C4996: 'std::uninitialized_copy::_Unchecked_iterators::_Deprecate': Call to 'std::uninitialized_copy' with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators'
D:\Programs\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.11.25503\include\memory(53): note: see declaration of 'std::uninitialized_copy::_Unchecked_iterators::_Deprecate'
d:\ds\src\util\buffvec.h(37): note: see reference to function template instantiation '_FwdIt std::uninitialized_copy<const T*,T*>(_InIt,_InIt,_FwdIt)' being compiled
        with
        [
            _FwdIt=parameter *,
            T=parameter,
            _InIt=const parameter *
        ]

danielschemmel:
I have dealt with the C4996s by way of [unchecked_array_iterator](https://msdn.microsoft.com/de-de/library/dn607301.aspx). This should take care of the problem without any loss of performance. I also got around building the project with VS, and do not see any further warnings (except multiple unrelated LNK4044s warning about an unsupported option `/INCREMENTAL:NO`) right now.

It does however open the question if we want safer iterators for buffer and vector ourselves? After all, direct accesses via `operator[]` are already checked in debug mode.

NikolajBjorner:
There are at least two issues shown by the automatic builds. The address sanitizer complains while running unit tests and the complex.py example runs out of memory. I can reproduce the issue with "complex.py". It seems to end up expanding some vector in perpetuity.

src/util/buffvec.h
```diff
@@ -0,0 +1,893 @@
...
+
+    // Depending on initial_size, move construction may still be very expensive
+    buffvec(buffvec&& other) {
+        if(other.m_data != reinterpret_cast<pointer>(&other.m_initial_buffer)) {
+            m_data = other.m_data;
+            m_size = other.m_size;
+            m_capacity = other.m_capacity;
+            other.m_data = reinterpret_cast<pointer>(&other.m_initial_buffer);
+            other.m_size = 0;
+            other.m_size = initial_size;
```
NikolajBjorner: why assigned twice?

src/util/buffvec.h
```diff
@@ -0,0 +1,893 @@
...
+                    swap(m_size, other.m_size);
+                    swap(m_capacity, other.m_capacity);
+                } else {
+                    buffvec_detail::destroy(begin(), end());
+                    m_data = other.m_data;
+                    m_size = other.m_size;
+                    m_capacity = other.m_capacity;
+                    other.m_data = reinterpret_cast<pointer>(&other.m_initial_buffer);
+                    other.m_size = 0;
+                    other.m_size = initial_size;
```
NikolajBjorner: other.m_size is assigned twice

src/util/buffvec.h
danielschemmel: because the second assignment was supposed to be `other.m_capacity = initial_size;`

src/util/buffvec.h
danielschemmel: Copy and paste of above code, thanks for catching it!

danielschemmel:
I finally got around to tracking down the problem with running complex.py. It seems I made one (or more) mistakes while adding the replacement for the `swap` member function, which only shows itself in this case. I removed the problematic commit from the pull request for now, and am looking at what exactly I overlooked in there.

NikolajBjorner:
Thanks, very cool. Gets rid of svector as well. This is a good time for structural changes. 
A number of other structural updates are rolling in at this point.

danielschemmel:
None of the items on the list should be blocking in the sense that the changes cannot be merged before that.

There is a bit of cleanup that should possibly be done, but none of it really prevents the usage of the changes as-is.

NikolajBjorner:
Note that it isn't going to build any longer.
D:\ds\src\ast/proofs/proof_utils.h(201): error C2039: 'get': is not a member of 'buffvec<T *,unsigned int,16>'
        with
        [
            T=app
        ]
D:\ds\src\util/buffer.h(9): note: see declaration of 'buffvec<T *,unsigned int,16>'
        with
        [
            T=app
        ]
[245/693] Building CXX object src\ast\normal_forms\CMakeFiles\normal_forms.dir\pull_quant.cpp.obj

danielschemmel:
@NikolajBjorner I am rebasing it right now

danielschemmel:
However, that is a valid question: Right now, as much of the old API has been removed as possible without causing major breakage, but I can add parts of that back in (e.g. the `get` function, which I have replaced with `operator[]` as far as possible)

NikolajBjorner:
I don't see a need for "get" if you have solved the problem with const reference vectors.
reduce_invertible uses resize, which you labeled as private

agurfinkel:
`get()` and `operator[]` usually have different semantics. `get()` indicates a const access, while `operator[]` does not. For consistency with STL, it might be nice to rename `get()` into `at()`.

agurfinkel:
https://en.cppreference.com/w/cpp/container/vector/at

agurfinkel:
I tend to use `at()` for const, and `operator[]` for non const. Then a compiler flags when I am doing something that I did not expect.

NikolajBjorner:
sorry, meant reserve

NikolajBjorner:
but you can ignore this note (for whatever it intends to say) if you are taking care of reduce_invertible

agurfinkel:
btw: will it be possible to declare a function that takes a `buffer` as an argument, but does not put restriction on the initial capacity of the buffer?

danielschemmel:
@agurfinkel Sort of. The current implementation does not implement any kind of virtual interface (which could be easily added if you want one), but allows compile-time polymorphism by way of templates:

```
// already possible with the old implementations
template<typename T>
void printall(T const& iterable) {
    for(auto const& x : iterable) std::cout << x;
}
```

```
// only possible with the new implementation
template<typename SZ, SZ INITIAL_SIZE>
unsigned sum(buffer<unsigned, SZ, INITIAL_SIZE> const& buff) {
    unsigned sum = 0;
    for(auto const& x : buff) sum += x;
    return sum
}
```

agurfinkel:
It would be nice to be able to create a function that takes either `buffer` or `vector` and is not based on templates. Currently, this is done by casting everything down to `c_ptr` or by duplicating code. Unfortunately, heavy use of templates has a tendency of significantly increasing compilation time, while reducing code readability.

danielschemmel:
@agurfinkel How would you feel about a `buffer_view` similar to a `string_view`?

This would provide type-safe read-only or at least non-resizable access to any kind of buffer/vector/array.

The only other options I can think of all involve some kind of dynamic dispatch, which should probably be avoided due to its performance implications.

danielschemmel:
@NikolajBjorner Just noticed that I was talking about `reserve` as well (which in the old API already did the `resize` operation instead of just reserving additional capacity, as it does in the standard library).

Rebase is almost done btw.

agurfinkel:
`buffer_view` fits well with most current usage

NikolajBjorner:
It breaks in fp_to_fp-12.smt2 and this is an indication of a performance regression.
A sample stack is:

z3.exe!buffvec<smt::watch_list,unsigned int,0>::reallocate<smt::watch_list>(unsigned int new_capacity) Line 551 	
z3.exe!buffvec<smt::watch_list,unsigned int,0>::reserve(unsigned int new_capacity) Line 662	 	
z3.exe!buffvec<smt::watch_list,unsigned int,0>::resize(unsigned int count) Line 640	 	
z3.exe!buffvec<smt::watch_list,unsigned int,0>::expand(unsigned int new_size) Line 810
z3.exe!smt::context::mk_bool_var(expr * n) Line 866


It increments the capacity by a minimal amount. The implementation of reserve in vector.h expands a vector by some multiplicative amount such that the vectors don't have to be resized on every push.
The caller in mk_bool_var increases the vector by only one element.

NikolajBjorner:
sorry, the lra_solver branch got merged. So there are some more conflicts.

danielschemmel:
@NikolajBjorner Thanks for the heads up, I will have a look at the performance regression, as I mostly tried to copy the geometric behavior from the original implementation. There are a few changes, mostly related to how swap and copy is implemented exactly, so maybe it is due to that.

NikolajBjorner:
In this case it isn't swap or copy, but how expand gets called.
It seems inconsistent across different variants. Sometimes it uses next_size(), sometimes not.
The expected behavior is that one can call expand, push_back, from user code 
N times with at most N + O(log(N)) overhead.

danielschemmel:
@NikolajBjorner I have modified the behavior of the underlying `resize` function, which will now ensure logarithmic growth as well. This should ensure that growth when repeatedly expanding is (at least) geometric.

@agurfinkel I wrote a first sample implementation of a `buffer_view`. Could you point me towards one of the areas in the code, where you might wish to use it, so I can take it for a test drive?

NikolajBjorner:
Should we finish this or include a partial update?

nunoplopes:
Hi Daniel,

I would love to see this work going in, but right now the patch is conflicting and it just got too big. Are you still interested in working on this?  If so, cloud you please the patch in multiple patches with a single change per patch to make it easier to review?

Thanks!

danielschemmel:
I finally got around to work on this again, for a bit.

@nunoplopes I already tried to create smallish atomic changes. However, due to the nature of the problem, some semantically really small patches (e.g., those renaming `reset` to `clear`) affect a large amount of files. If you have any suggestions, I don't mind splitting up commits further.

nunoplopes:
I would suggest 1 pull request per change. As you say, the patches for renaming can be big, and right now this PR has many of these in the way and thus it's hard to review.

NikolajBjorner:
I agree with scoped pull per change in vector class. This may affect many files, but funcationality wise the change is limited. It also allows to benefit from partial progress.

NikolajBjorner:
no activity on this for a year

PR #1666, opened 2018-06-05
first draft of cmake support for ocaml

c-cube:
I'm far from knowing what I'm doing, but I tried to follow the same pattern as java and python. I will update this progressively to try and make the CI work.


cmake/modules/LICENSE-ocaml-cmake-0BSD.txt
```diff
@@ -0,0 +1,12 @@
+Copyright (C) 2018 whitequark@whitequark.org
```
NikolajBjorner: Is this MIT or BSD as the file name says? I am not keen on including other licenses even if they are fine. 
It adds to the pile of questions users may feel compelled to consider for due diligence when integrating 
external software, such as Z3.

c-cube:
Sorry, I don't know enough cmake to rewrite this from scratch, so for now I'm closing this PR.

PR #1687, opened 2018-06-17
Fix FPNumRef.significand_as_long

Vtec234:
The function doesn't work in `Python`, because it uses the C API wrong. This fixes it.


Vtec234:
XCode build failed for configuration reasons, unrelated to PR.

PR #1697, opened 2018-06-22
Refine default_tactic for better support of pure SAT instances

rainoftime:
If the constraint is an SAT instance and proof is not enabled, then use the qffd tactic.


rainoftime:
I noticed that in some cases, Z3 will apply the "smt" tactic for a pure SAT instance. 
Perhaps the "qffd" tactic can use the full power of the sat engine in `z3/src/sat`

PR #1700, opened 2018-06-25
Deep space

agurfinkel:
Fixing and improving iuc extraction from proofs

### DIFF
--- a/src/muz/spacer/spacer_unsat_core_learner.cpp
+++ b/src/muz/spacer/spacer_unsat_core_learner.cpp
@@ -54,7 +54,7 @@ void unsat_core_learner::compute_unsat_core(expr_ref_vector& unsat_core) {
         // if the node mixes A-reasoning and B-reasoning
         // and contains non-closed premises
         if (!done) {
-            if (m_pr.is_a_marked(curr) && m_pr.is_b_marked(curr)) {
+            if (is_a(curr) && is_b(curr)) {
                 compute_partial_core(curr);
             }
         }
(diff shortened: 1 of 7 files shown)


src/muz/spacer/spacer_unsat_core_learner.cpp
```diff
@@ -36,28 +36,27 @@ void unsat_core_learner::register_plugin(unsat_core_plugin* plugin) {
...
-        // if the node mixes A-reasoning and B-reasoning and contains non-closed premises
-        if (m_pr.is_a_marked(currentNode) && 
-            m_pr.is_b_marked(currentNode) && 
-            !m_closed.is_marked(currentNode)) {
-            compute_partial_core(currentNode); // then we need to compute a partial core
+        // we have now collected all necessary information,
+        // so we can visit the node
+        // if the node mixes A-reasoning and B-reasoning
+        // and contains non-closed premises
+        if (!done) {
```
NikolajBjorner: always true

src/muz/spacer/spacer_unsat_core_plugin.cpp
```diff
@@ -134,34 +133,35 @@ namespace spacer {
...
+                  for (unsigned i = 0; i < m.get_num_parents(step); ++i) {
+                      proof * prem = m.get_parent(step, i);
+                      rational coef = params[i].get_rational();
+                      bool b_pure = m_ctx.is_b_pure (prem);
+                      tout << (b_pure?"B":"A") << " " << coef << " " << mk_pp(m.get_fact(prem), m) << "\n";
+                  }
+                );
+
+            bool done = true;
+
```
NikolajBjorner: there is an iterator for that. It is called get_parents(step), so you say
for (proof * premise : get_parents(step)) {…. }

NikolajBjorner: oh, well, unless you need to access params[i]

src/muz/spacer/spacer_unsat_core_learner.cpp
agurfinkel: @NikolajBjorner why? `done` is true if all parents are `done`. This is entered only if there is an open parent that has not been interpolated over.

PR #1716, opened 2018-07-01
integrate theory_lra with master

NikolajBjorner:


src/util/lp/lar_solver.cpp
```diff
@@ -1104,6 +1104,25 @@ bool lar_solver::has_upper_bound(var_index var, constraint_index& ci, mpq& value
...
+        }
+        return true;
+    }
+    else {
+        impq const& r = get_column_value(var);
+        value = r.x;
+        return numeric_traits<mpq>::is_zero(r.y);
+    }
+}
+
```
levnach: If get_column_value(var).y is not zero, or it so any variables of the term, then the method returns false. Is it by design? I would add a comment then.

src/util/lp/lar_solver.cpp
```diff
@@ -1041,6 +1044,7 @@ bool lar_solver::explanation_is_correct(const vector<std::pair<mpq, unsigned>>&
         lp_assert(false);
         return false;
     }
+#endif
 #endif
     return true;
```
levnach: I should fix explanation_is_correct().

src/smt/theory_lra.cpp
```diff
@@ -2088,8 +2088,8 @@ class theory_lra::imp {
             vi = m_todo_vars.back();
             m_todo_vars.pop_back();
             lp::lar_term const& term = m_solver->get_term(vi);
-            for (auto i = term.m_coeffs.begin(); i != term.m_coeffs.end(); ++i) {
-                lp::var_index wi = i->first;
+            for (auto const& coeff : term.m_coeffs) {
+                lp::var_index wi = coeff.first;
```
levnach: I believe we can use this more readable syntax here 
for (auto const& p : term.m_coeffs) { 
                lp::var_index wi = p.var();  // p.coeff() is also available

src/smt/theory_lra.cpp
```diff
@@ -1177,7 +1179,12 @@ class theory_lra::imp {
...
                 for (const auto & i : term.m_coeffs) {
-                    m_todo_terms.push_back(std::make_pair(i.first, coeff * i.second));
+                    if (m_variable_values.count(i.first) > 0) {
+                        result += m_variable_values[i.first] * coeff * i.second;
+                    }
+                    else {
+                        m_todo_terms.push_back(std::make_pair(i.first, coeff * i.second));
+                    }
                 }                    
             }
```
levnach: We can replace i.first by i.var(), and i.second by i.coeff().

src/smt/theory_lra.cpp
NikolajBjorner: add these improvements directly in a pull request into z3prover/z3.git lev.

PR #1728, opened 2018-07-04
remove unsound contains/indexof heuristic in theory_str

### LINKED ISSUE #1726: Inconsistency between Z3 string solver and Z3Str3
Hi,

There's a strange inconsistency where the z3 string solver (correctly) solves an input as `sat`, but the z3str3 solver solves it as `unsat`. Here's a minimal example:
```
(set-logic QF_S)
(declare-const s String)
(assert (= -1 (str.indexof s s 5)))
(check-sat)
```
According to z3str3 documentation, `str.indexof` should simplify to `-1` and produce a sat answer.

mtrberzi:
Fixes #1726. I'll revisit the contains/indexof integration when I have time, for now this makes things behave correctly.

### DIFF
--- a/src/smt/theory_str.cpp
+++ b/src/smt/theory_str.cpp
@@ -1382,22 +1382,6 @@ namespace smt {
         expr_ref finalAxiom(m.mk_and(breakdownAssert, reduceToIndex), m);
         SASSERT(finalAxiom);
         assert_axiom(finalAxiom);
-
-        {
-            // heuristic: integrate with str.contains information
-            // (but don't introduce it if it isn't already in the instance)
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
breaks string13.smt2

mtrberzi:
Right, that's what I suspected would happen - this is the heuristic originally introduced to deal with that particular case

It seems like we need to do something to address this, since "no heuristic" breaks the test

Will revisit this week

mtrberzi:
Revisited this today. The heuristic is fine, it just needs a better precondition. Opening new PR

PR #1737, opened 2018-07-06
Equality Explanation Logging

Nils-Becker:
Whenever a quantifier is triggered and a [new-match] line appears in the log we now additionally log explanations for all relevant equalities. These appear as steps that can be taken to reach the root of an equivalence class. In general these lines look like:
[eq-expl] <id of starting node> <step type (e.g. congruence)> <potentially other information like pairs of arguments that are equal for congruence steps> ; <target node>
The full documentation will be added to the wiki.

We do not expect performance differences (apart from maybe fewer enodes fitting into a single cache line due to an additional field and resetting the log status when updating the e-graph all changes are inside if-statements) when logging is turned off. With logging enabled there is in some cases a noticeable performance overhead compared to logging with the old version.


NikolajBjorner:
looks fine. enodes are already very fat so I don't anticipate this being worthwhile compressing into other fields.

PR #1751, opened 2018-07-12
use signed_term struct to simplify the hnf interface

levnach:


levnach:
The interface with arrays was a bit faster because some of the arrays were reused by theory_lra and hnf calculations.

Issue #1791, opened 2018-08-06
Complete Javascript N-API support for Javascript


sb98052:
#1298 led to work-in-progress N-API bindings. Some work has to be done before these become usable:

- [ ] Complete packaging - calling `make; make install` after `mk_make.py --js` should install the JS bindings as an npm package
- [ ] Add unit tests mirroring other language APIs. Right now, there is a barebones sanity check.
- [ ] Integrate build and unit tests with CI
- [ ] Add an idiomatic API on top of the current thin shim

NikolajBjorner:
no progress on this for a year

PR #1792, opened 2018-08-06
Refactor JS bindings to match Z3 build system

sb98052:
- Updated initial JS support to adapt Z3 interface and value types to N-API generator
- Generate JS constants
- Check in binding generator
- Makefile, build script and basic test

This is an early-stage API and will probably need a few evolutions before it can be used reliably. See #1791.


src/api/js/test.js
```diff
@@ -0,0 +1,150 @@
+const z3 = require("./build/Release/z3");
```
NikolajBjorner: should really live in an "examples/js" directory.
The dependency on "/build/release" would also go away.

src/api/js/z3.cc.j2
```diff
@@ -0,0 +1,161 @@
+/* Automatically generated file */
```
NikolajBjorner: perhaps this file is _not_ automatically generated.

src/api/js/Makefile
```diff
@@ -0,0 +1,11 @@
+# Set to 1 to activate API logging
```
NikolajBjorner: This seems to be something to fold into the build generators (python/cmake) that produce the make files or other inputs to the compilers.

scripts/update_api.py
```diff
@@ -741,55 +741,97 @@ def mk_java(java_dir, package_name):
...
-               t = param_type(p)
-               k = t
-               ous.write("            \"name\": \"%s\",\n" % "")                        # TBD
-               ous.write("            \"c_type\": \"%s\",\n" % type2str(t))
-               ous.write("            \"napi_type\": \"%s\",\n" % type2napi(t))        
-               ous.write("            \"napi_builder\": \"%s\"\n" % type2napibuilder(t))
+
+               """ The name of the param is only useful when debugging the bindings. Otherwise,
+               it is generated on the fly. """
+               ous.write("            \"name\": \"%s\",\n" % "")
```
NikolajBjorner: the names are absent in the meta-data. If the backend needs names, this code could cook up x1, x2, x3, etc.

scripts/mk_genfile_common.py
```diff
@@ -550,6 +550,92 @@ def mk_z3consts_ml_internal(api_files, output_dir):
     # if VERBOSE:
     #     print ('Generated "%s/z3enums.mli"' % ('%s' % gendir))
 
+def mk_z3consts_js_internal(api_files, output_dir):
```
NikolajBjorner: seems like a copy of _mk_z3consts_py_internal".
A generalization could be a function that takes callbacks for the places where it produces different output.

sb98052: I started out with that in mind, but noticed that it was copied for every language - e.g. see `mk_z3consts_ml_internal`, `mk_z3consts_dotnet_internal`.

scripts/update_api.py
sb98052: The generator already does that. What I meant to say in the comment was that it is useful to link those generated names with the names of the original parameters when debugging the API. But not strictly necessary. I'll clarify.

src/api/js/test.js
sb98052: Thanks for your comments. `build/Release/z3` is the compiled add on that `test.js` is importing. Right now, it is generated inside of the `src/api/js` directory. I was thinking of the task of moving it to the right place and installing it as part of the build process as item 1 in #1791 and putting out actual examples as item 2. Can we do this incrementally in this manner? This would potentially allow other people to jump in and help out.

src/api/js/z3.cc.j2
sb98052: Fair point. It does make it seem as though it marks the `.j2` file as autogenerated, in addition to injecting the line in the actual generated file.

scripts/mk_genfile_common.py
NikolajBjorner: ouch, it is then something for me to look at.

waywardmonkeys:
Where is this PR at? What needs to be done?

I'd like to look at updating the JS stuff to work with the WebAssembly work that I've done as well ...

NikolajBjorner:
no activity on this for half a year

PR #1799, opened 2018-08-17
fix(union-find): keep values and representative in consistent order

c-cube:
NOTE: I'm not totally sure if the semantics of `merge_eh` and `after_merge_eh` implies this, as it's not commented. But it seems natural to want the 4 values' order to matter?

This patch enforces that the merge handlers should be called with r1,r2,v1,v2 or r2,r1,v2,v1
but no other order.

### DIFF
--- a/src/util/union_find.h
+++ b/src/util/union_find.h
@@ -122,8 +122,10 @@ public:
         TRACE("union_find", tout << "merging " << r1 << " " << r2 << "\n";);
         if (r1 == r2)
             return;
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
thanks, I don't see a hard dependency on this order in the existing callers, but the butterfly effect caused issue1690 repro to diverge. However, it already diverges with different random seeds so isn't prime for the regression suite anyhow.

PR #1840, opened 2018-09-21
simpler Gomory cuts

levnach:


levnach:
The experiment did not work

PR #1849, opened 2018-09-28
Adding missing call to update_max_generation

Nils-Becker:
Since `update_max_generation()` is called from `init_continue()` (l. 2203) it seems to me that it should also be called after backtracking. At least the logging code that is in `update_max_generation()` should be executed at this point.


NikolajBjorner:
will review

Nils-Becker:
Sorry, I just noticed that the call was already there.

PR #1855, opened 2018-10-01
Z3str3: refactoring, arith_value

mtrberzi:
This changes code in Z3str3 that interacts with the arithmetic solver to use the `arith_value` interface instead. There is also some minor refactoring to the regex automata subsolver that places the logic into a subroutine.


NikolajBjorner:
string3.smt2 and string4.smt2 seem to be failing.

mtrberzi:
Haven't looked into the details too closely but it seems that `arith_value::get_value` has different behaviour from what I want `theory_str::get_arith_value` to do. Maybe this is because we are calling it before we can guarantee that the arithmetic solver's `final_check` has been run. In any event, we can't make that guarantee everywhere in `theory_str` so I can't use that helper method here. It will still be useful to me later on in `final_check` where we can interact with the arithmetic solver in a better defined way (i.e. I should be able to use it to skip length testing completely).

Merging with master and updating this PR shortly.

src/smt/theory_str.cpp
```diff
@@ -4836,9 +4836,24 @@ namespace smt {
     }
 
     bool theory_str::get_arith_value(expr* e, rational& val) const {
-        arith_value v(get_context());
-        return v.get_value(e, val);
-    }
+         context& ctx = get_context();
```
NikolajBjorner: should this behavior be integrated with arith_value?

PR #1856, opened 2018-10-02
Minor fixes

waywardmonkeys:


src/smt/theory_lra.cpp
```diff
@@ -341,7 +341,6 @@ class theory_lra::imp {
         }
         app_ref cnst(a.mk_numeral(rational(c), is_int), m);
         TRACE("arith", tout << "add " << cnst << "\n";);
-        enode* e = mk_enode(cnst);
```
NikolajBjorner: I need the call to mk_enode(cnst), but the return value can be ignored.

waywardmonkeys: Okay, I will improve this fix then. Thanks for the review!

waywardmonkeys:
Pushed an update here to take into account that you fixed one of the issues.

Issue #1441, opened 2018-01-04
Clang-format file


delcypher:
It would be helpful for developers if Z3's source tree came with a `.clang-format` file so that changes can be automatically formatted.

Note that clang-format can format whole files (usually not desirable) or it can format patches (i.e. with `git clang-format`) or regions of code (e.g. there's a vim plug-in that lets you format a selected region of text).

A handy tool for visually creating a `.clang-format` file can be found at https://zed0.co.uk/clang-format-configurator/ and documentation on various options can be found at https://clang.llvm.org/docs/ClangFormatStyleOptions.html

Could a `.clang-format` file be added that matches Z3's C++ style?

NikolajBjorner:
Looks useful. Thanks for pointing this out. I wouldn't mind at all and it could help with obliterating manual reformats that happen when small but noticeable inconsistencies are introduced. Not sure how I am able to allocate bandwidth to this given the other worthwhile improvements and bug fixes that have priority.

XVilka:
Since the time of the issue opened `clang-format` is better integrated with CMake and Z3 ported (everything except OCaml bindings) to CMake, so you can check these two modules as an example:
- https://github.com/zemasoft/clangformat-cmake
- https://github.com/TheLartians/Format.cmake

PR #1862, opened 2018-10-02
[NFC] Cleanup arith_eq_solver.(cpp|h)

kbobyrev:
Use for-range loops instead of for-index loops where possible, remove
trailing whitespaces.

This patch does not affect functionality.

I was interested in working on #1441 and I wanted to start with a simple patch which introduces minor cleanup. I was basically wondering whether there is any policy regarding cleanup patches: in LLVM and Clang it is common to perform large scale changes at once (if there are any) and small-scale changes are not very common becaue `git blame` becomes less readable.

I thought that I could use Clang tools (Clang-Format and Clang-Tidy) to improve the code formatting and readability so that the project is more beginner-friendly and clean. As the part of #1441 I would propose to include `.clang-tidy` configuration file and probably run both Clang-Format style check and Clang-Tidy checks in Travis CI to improve the overall code quality.

### DIFF
--- a/src/smt/arith_eq_solver.cpp
+++ b/src/smt/arith_eq_solver.cpp
@@ -12,13 +12,10 @@ Abstract:
 Author:
 
     Nikolaj Bjorner (nbjorner) 2012-02-25
-    
+
 --*/
 #include "smt/arith_eq_solver.h"
 
-
-arith_eq_solver::~arith_eq_solver() {
-}
 
 arith_eq_solver::arith_eq_solver(ast_manager & m, params_ref const& p):
     m(m),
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
these look fine and welcome.
Larger scale is also fine, but I need to review changes so uniform batches that are easy to review are better.

kbobyrev:
@NikolajBjorner great, thank you very much! Am I correct to assume that the best thing would be to stick with large-scale changes with the same action within one PR? E.g. one PR for whitespace removal, one PR for for-range loops modernization, etc: this way it would be way easier to review such patches because each of them basically applies a single action to multiple files.

NikolajBjorner:
sure. It is also fine to pull an improvement even if it does not deal with all files.
The syntax and equivalence preserving transformations tend to be low risk. 

The higher risk changes are the ones that affect semantics. E.g, there was a change to use stdbool for the API. It turned out to break our .NET bindings for 4.7.1. It wasn't discovered by the unit tests.

PR #1867, opened 2018-10-08
Add Clang-Format configuration

### LINKED ISSUE #1441: Clang-format file

kbobyrev:
This patch introduces Clang-Format configuration and reformats a single source file to see the outcome of reformatting the sources.

Fixes #1441.

### DIFF
--- a/src/solver/mus.cpp
+++ b/src/solver/mus.cpp
@@ -39,7 +39,8 @@ struct mus::imp {
 
     imp(solver& s)
         : m_solver(s), m(s.get_manager()), m_lit2expr(m), m_assumptions(m),
-          m_soft(m) {}
+          m_soft(m) {
+    }
 
     void reset() {
         m_lit2expr.reset();
@@ -63,7 +64,9 @@ struct mus::imp {
         return idx;
     }
 
-    void add_assumption(expr* lit) { m_assumptions.push_back(lit); }
+    void add_assumption(expr* lit) {
+        m_assumptions.push_back(lit);
+    }
 
     lbool get_mus(expr_ref_vector& mus) {
         m_model.reset();
@@ -297,7 +300,9 @@ struct mus::imp {
             : m_fmls(fmls1), m_size(fmls1.size()) {
             fmls1.push_back(fml);
         }
-        ~scoped_append() { m_fmls.shrink(m_size); }
+        ~scoped_append() {
+            m_fmls.shrink(m_size);
+        }
     };
 
     template <class T> void display_vec(std::ostream& out, T const& v) const {
@@ -353,17 +358,29 @@ struct mus::imp {
     }
 };
 
-mus::mus(solver& s) { m_imp = alloc(imp, s); }
+mus::mus(solver& s) {
+    m_imp = alloc(imp, s);
+}
 
-mus::~mus() { dealloc(m_imp); }
+mus::~mus() {
+    dealloc(m_imp);
+}
 
-unsigned mus::add_soft(expr* lit) { return m_imp->add_soft(lit); }
+unsigned mus::add_soft(expr* lit) {
+    return m_imp->add_soft(lit);
+}
 
-void mus::add_assumption(expr* lit) { return m_imp->add_assumption(lit); }
+void mus::add_assumption(expr* lit) {
+    return m_imp->add_assumption(lit);
+}
 
-lbool mus::get_mus(expr_ref_vector& mus) { return m_imp->get_mus(mus); }
+lbool mus::get_mus(expr_ref_vector& mus) {
+    return m_imp->get_mus(mus);
+}
 
-void mus::reset() { m_imp->reset(); }
+void mus::reset() {
+    m_imp->reset();
+}
 
 void mus::set_soft(unsigned sz, expr* const* soft, rational const* weights) {
     m_imp->set_soft(sz, soft, weights);
--- a/src/solver/mus.cpp
+++ b/src/solver/mus.cpp
@@ -275,7 +275,7 @@ struct mus::imp {
         expr_ref_vector& m_fmls;
         unsigned m_size;
 
-        public:
+    public:
         scoped_append(imp& imp, expr_ref_vector& fmls1, expr_set const& fmls2)
             : m_fmls(fmls1), m_size(fmls1.size()) {
             expr_set::iterator it = fmls2.begin(), end = fmls2.end();
--- a/src/solver/mus.cpp
+++ b/src/solver/mus.cpp
@@ -6,7 +6,7 @@ Module Name:
     mus.cpp
 
 Abstract:
-   
+
     MUS extraction.
 
 Author:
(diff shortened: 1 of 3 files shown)


kbobyrev:
There were few inconsistencies that I noticed in the code and I'm not really sure how to break a tie in some cases:

* `*` and `&` alignment: so far I've seen `int* ptr;` and `int * ptr;` with the first one being more common - this corresponds to `PointerAlignment: Left` Clang-Format option, but feel free to correct me if this assumption is wrong.
* Single-line if statements: I've seen both `if (condition) statement;` and `if (condition) \n statement;` (both are within several lines of code in the formatted file). I didn't investigate more than a couple of files but I think `AllowShortIfStatementsOnASingleLine: false` (disallowing one-liners) would be correct here.
* I didn't set`AlignConsecutiveDeclarations: true` which caused reformatting of class/field members but if I set this option then it is applied everywhere in the code which doesn't look nice from my point of view. Again, feel free to correect me if I'm wrong.

kbobyrev:
The issue was fixed by the last commit, but the build is failing seemingly because of unrelated changes (sibling PRs fail with the same errors in the same configurations and pass in the same configurations).

src/solver/mus.cpp
```diff
@@ -18,30 +18,29 @@ Module Name:
...
-    obj_map<expr, unsigned>  m_expr2lit;
-    model_ref                m_model;
-    expr_ref_vector          m_soft;
-    vector<rational>         m_weights;
-    rational                 m_weight;
-
-    imp(solver& s): 
-        m_solver(s), m(s.get_manager()), m_lit2expr(m),  m_assumptions(m), m_soft(m)
-    {}
+    solver& m_solver;
```
NikolajBjorner: I would prefer these declarations to be mostly aligned as in the original file.

src/solver/mus.cpp
```diff
@@ -50,16 +49,18 @@ struct mus::imp {
...
-        return is_uninterp_const(lit) || (m.is_not(lit, l) && is_uninterp_const(l));
+        expr* l;
+        return is_uninterp_const(lit) ||
+               (m.is_not(lit, l) && is_uninterp_const(l));
     }
-    
+
     unsigned add_soft(expr* lit) {
-        //SASSERT(is_literal(lit));
+        // SASSERT(is_literal(lit));
```
NikolajBjorner: good

src/solver/mus.cpp
```diff
@@ -50,16 +49,18 @@ struct mus::imp {
     }
 
     bool is_literal(expr* lit) const {
-        expr* l;		
-        return is_uninterp_const(lit) || (m.is_not(lit, l) && is_uninterp_const(l));
+        expr* l;
+        return is_uninterp_const(lit) ||
```
NikolajBjorner: ok

src/solver/mus.cpp
```diff
@@ -81,9 +82,12 @@ struct mus::imp {
         ptr_vector<expr> unknown(m_lit2expr.size(), m_lit2expr.c_ptr());
         expr_ref_vector core_exprs(m);
         TRACE("mus", m_solver.display(tout););
-        while (!unknown.empty()) { 
-            IF_VERBOSE(12, verbose_stream() << "(mus reducing core: " << unknown.size() << " new core: " << mus.size() << ")\n";);
-            TRACE("mus", display_vec(tout << "core:  ", unknown); display_vec(tout << "mus:   ", mus););
+        while (!unknown.empty()) {
+            IF_VERBOSE(12, verbose_stream()
+                               << "(mus reducing core: " << unknown.size()
```
NikolajBjorner: not sure which editors support this alignment.

src/solver/mus.cpp
```diff
@@ -113,10 +117,8 @@ struct mus::imp {
                         }
                     }
                     TRACE("mus", tout << "core exprs:" << core_exprs << "\n";
-                        display_vec(tout << "core:", unknown);
-                        display_vec(tout << "mus:", mus);
-                    );
-
+                          display_vec(tout << "core:", unknown);
```
NikolajBjorner: OK

src/solver/mus.cpp
```diff
@@ -142,33 +146,38 @@ struct mus::imp {
...
         bool min_core_valid = false;
         expr* min_lit = nullptr;
         while (!unknown.empty()) {
             expr* lit = unknown.back();
             unknown.pop_back();
             model_ref mdl;
-            scoped_append assume_mss(*this,  mus, mss);      // current satisfied literals
-            scoped_append assume_nmcs(*this, mus, nmcs);     // current non-satisfied literals
-            scoped_append assume_lit(*this,  mus, lit);      // current unknown literal
+            scoped_append assume_mss(*this, mus,
```
NikolajBjorner: why does this break the line?

src/solver/mus.cpp
```diff
@@ -142,33 +146,38 @@ struct mus::imp {
...
             switch (m_solver.check_sat(mus)) {
             case l_true: {
-                TRACE("mus", tout << "literal can be satisfied: " << mk_pp(lit, m) << "\n";);
+                TRACE("mus", tout << "literal can be satisfied: "
+                                  << mk_pp(lit, m) << "\n";);
                 mss.push_back(lit);
                 m_solver.get_model(mdl);
                 model_evaluator eval(*mdl.get());
-                for (unsigned i = 0; i < unknown.size(); ) {
+                for (unsigned i = 0; i < unknown.size();) {
```
NikolajBjorner: ok

src/solver/mus.cpp
```diff
@@ -220,13 +231,13 @@ struct mus::imp {
         for (expr* e : mss) {
             mss_set.insert(e);
         }
-        for (expr * e : min_core) {
+        for (expr* e : min_core) {
```
NikolajBjorner: OK, fine to have * bind to the type. I don't like expr *e, which is very common in C/C++ code bases.
The reason is that I read * as part of the type expression. 
So expr and expr* are types, and 'e' is an expression of type expr*.

For that matter I also have developed a possibly peculiar convention of placing const qualifiers in post-fix mode.
I write T const & onstead of const T& so I don't have to recall what the const qualifier binds to.

src/solver/mus.cpp
```diff
@@ -245,56 +256,56 @@ struct mus::imp {
         if (A.size() < B.size()) {
             expr_set::iterator it = A.begin(), end = A.end();
             for (; it != end; ++it) {
-                if (B.contains(*it)) return true;
+                if (B.contains(*it))
```
NikolajBjorner: OK, fine to have this consistent

src/solver/mus.cpp
```diff
@@ -245,56 +256,56 @@ struct mus::imp {
...
+                    return true;
             }
         }
         else {
             expr_set::iterator it = B.begin(), end = B.end();
```
NikolajBjorner: Ups, I need to rewrite this to for (expr * e : B) ...

src/solver/mus.cpp
```diff
@@ -245,56 +256,56 @@ struct mus::imp {
...
             fmls1.push_back(fml);
         }
         ~scoped_append() {
             m_fmls.shrink(m_size);
         }
     };
 
-    template<class T>
-    void display_vec(std::ostream& out, T const& v) const {
+    template <class T> void display_vec(std::ostream& out, T const& v) const {
```
NikolajBjorner: This seems arbitrary. Putting the template portion on a separate line shortens the next line and separates the template portion (polymorphism quantifier) from type signature.

src/solver/mus.h
```diff
@@ -19,49 +19,52 @@ Module Name:
...
+       Assume that the solver context enforces that
        cls is equivalent to a disjunction of args.
-       Assume also that cls is a literal.           
+       Assume also that cls is a literal.
     */
     unsigned add_soft(expr* cls);
 
     void add_soft(unsigned sz, expr* const* clss) {
-        for (unsigned i = 0; i < sz; ++i) add_soft(clss[i]);
+        for (unsigned i = 0; i < sz; ++i)
```
NikolajBjorner: OK

NikolajBjorner:
items 1, 2 seem right. For 3, in the pull request there is a realignment of field declarations. This reformatting makes reading what are the field names not so easy. I would prefer (insist) that it be possible to align them in a way that fits a tabular form whenever possible.

NikolajBjorner:
closing as no reaction to fixing the realignment of field declarations. Fields are more readable when they are aligned. The other changes are fine.

PR #1873, opened 2018-10-11
Add a floating-point support to c++ api.

xlauko:


src/api/c++/z3++.h
```diff
@@ -138,11 +138,22 @@ namespace z3 {
        \brief A Context manages all other Z3 objects, global configuration options, etc.
     */
     class context {
+    public:
+        enum class rounding_mode {
```
NikolajBjorner: seems we just remove "class" from "enum class" for downlevel C++

PR #1881, opened 2018-10-15
Revert "Made Z3 compile for C++17 with MSVC"

NikolajBjorner:
Reverts Z3Prover/z3#1880


NikolajBjorner:
ups, this will require updating some build environments that live on C++11.

PR #1902, opened 2018-10-26
Mark dtors that require virtual as virtual

janisozaur:


src/muz/rel/dl_base.h
```diff
@@ -220,7 +220,7 @@ namespace datalog {
         */
         class mutator_fn : public base_fn {
         public:
-            ~mutator_fn() override {}
+            virtual ~mutator_fn() override {}
```
NikolajBjorner: because a subclass overrides this?

src/smt/theory_arith.h
```diff
@@ -345,7 +345,7 @@ namespace smt {
             friend class theory_arith;
         public:
             derived_bound(theory_var v, inf_numeral const & val, bound_kind k):bound(v, val, k, false) {}
-            ~derived_bound() override {}
+            virtual ~derived_bound() override {}
```
NikolajBjorner: ditto. 
We have many occurrences of destructors that just use override given that this is the new way to, ehm, override.

NikolajBjorner:
I don't see why this is necessary. https://en.cppreference.com/w/cpp/language/virtual says that derived methods are automatically virtual.
Is there a warning level that produces warnings for this?

PR #1918, opened 2018-11-06
feat(api/ml): release runtime lock on some long-running functions

c-cube:
When calling `Z3.Solver.check` from OCaml, the whole program can be blocked for an arbitrary amount of time, waiting for Z3 to return. Currently other OCaml threads won't even be scheduled as the whole runtime is blocked in the call. This PR changes that (for a handful of functions) by using `caml_{release,acquire}_runtime_system` for the duration of the C call (see <https://caml.inria.fr/pub/docs/manual-ocaml/intfc.html#sec476> sec 20.12.2 for more details), allowing other threads to run and giving them a chance to call `Z3.Tactic.interrupt` in case the computation takes too long.

Note that the list of functions with this wrapper is currently limited to `{check,simplify}` because I don't think this is useful for most (quickly-returning) functions.

This also adds a dependency on `threads`, but as far as I can tell, if you can run Z3 it means you're on a native machine and pthreads should be available.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1328,6 +1328,13 @@ def mk_ml(ml_src_dir, ml_output_dir):
 
     mk_z3native_stubs_c(ml_src_dir, ml_output_dir)
 
+z3_long_funs = frozenset([
+    'Z3_solver_check',
+    'Z3_solver_check_assumptions',
+    'Z3_simplify',
+    'Z3_simplify_ex',
+    ])
+
 def mk_z3native_stubs_c(ml_src_dir, ml_output_dir): # C interface
     ml_wrapperf = os.path.join(ml_output_dir, 'z3native_stubs.c')
     ml_wrapper = open(ml_wrapperf, 'w')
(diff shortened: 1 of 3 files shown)


jberdine:
Maybe adjust setting the Z3 error handler to reacquire the runtime lock before calling the registered OCaml function.

c-cube:
seems like the ml API registers an empty error handler (and creates an OCaml exception based on the return value). This should be safe, I believe (would @wintersteiger have an opinion?)

```c
 /* Internal do-nothing error handler. This is required to avoid that Z3 calls exit()
    upon errors, but the actual error handling is done by throwing exceptions in the
    n_* wrapper functions. */
```

NikolajBjorner:
Threre are basically two options:
1. have the native C code inside of Z3's dll throw exceptions
2. have the caller check an error code on each call and then throw.
Option 2 is used fairly systematically across the different APIs, and this includes the ML API.
The z3native_stubs.c is automatcally generated and contains the error handling to exception code.
It may be possible to use the OCaml Callback features together with a different automatically generated code for exception handling. Thus, instead of checking the error code on every call as in the current automatically generated code, redefine MLErrorHandler to raise the exception directly.
The catch is obviously to debug this approach and make sure it works correctly with 
the dll linked code. 
The most nasty catch is that throwing non-C++ exceptions (longjumps) from C++ code doesn't play nicely with invoking destructors on the C++ stack. This effected a very few cases in the API functions where I had to insert explicit clears before invoking error handlers. These issues are fixable, but weight it as a risk for going down the route of 1. 
I am not aware of usability/correctness issues with the current solution.

The comment that mentions "exit" is perhaps not representative of what it could do.

NikolajBjorner:
feel free to update locking calls based on experience

jberdine:
I don't know of any code that would be affected, but with this patch 
might it be possible for multiple OCaml threads to call into Z3 
concurrently? Perhaps there is some code that relied on the runtime 
system lock in order to avoid otherwise-racy parallel calls to Z3? 
Perhaps something to note in the documentation.

PR #1950, opened 2018-11-19
Improve intra-doc linking.

waywardmonkeys:


waywardmonkeys:
This is more stuff that I noticed while working on the Rust Z3 bindings.

PR #1951, opened 2018-11-19
Remove usages of Z3_TRUE / Z3_FALSE.

waywardmonkeys:
Now that this is all using stdbool.h, we can just use true/false.

For now, we leave the aliases in place in z3_api.h.

### DIFF
--- a/src/api/api_arith.cpp
+++ b/src/api/api_arith.cpp
@@ -121,7 +121,7 @@ extern "C" {
 
     Z3_bool Z3_API Z3_is_algebraic_number(Z3_context c, Z3_ast a) {
         LOG_Z3_is_algebraic_number(c, a);
-        return mk_c(c)->autil().is_irrational_algebraic_numeral(to_expr(a)) ? Z3_TRUE : Z3_FALSE;
+        return mk_c(c)->autil().is_irrational_algebraic_numeral(to_expr(a)) ? true : false;
     }
 
     Z3_ast Z3_API Z3_get_algebraic_number_lower(Z3_context c, Z3_ast a, unsigned precision) {
--- a/src/api/api_numeral.cpp
+++ b/src/api/api_numeral.cpp
@@ -146,7 +146,7 @@ extern "C" {
         Z3_TRY;
         LOG_Z3_is_numeral_ast(c, a);
         RESET_ERROR_CODE();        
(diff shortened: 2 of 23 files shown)


src/api/api_arith.cpp
```diff
@@ -121,7 +121,7 @@ extern "C" {
 
     Z3_bool Z3_API Z3_is_algebraic_number(Z3_context c, Z3_ast a) {
         LOG_Z3_is_algebraic_number(c, a);
-        return mk_c(c)->autil().is_irrational_algebraic_numeral(to_expr(a)) ? Z3_TRUE : Z3_FALSE;
+        return mk_c(c)->autil().is_irrational_algebraic_numeral(to_expr(a)) ? true : false;
```
NikolajBjorner: if we are extra courageous, we may even do

      return mk_c(c)->autil().is_irrational_algebraic_numeral(to_expr(a))

src/api/api_numeral.cpp
```diff
@@ -188,7 +188,7 @@ extern "C" {
         CHECK_IS_EXPR(a, "");
         rational r;
         Z3_bool ok = Z3_get_numeral_rational(c, a, r);
-        if (ok == Z3_TRUE) {
+        if (ok == true) {
```
NikolajBjorner: if (ok) works too

src/api/api_seq.cpp
```diff
@@ -70,35 +70,35 @@ extern "C" {
...
     }
 
     Z3_bool Z3_API Z3_is_re_sort(Z3_context c, Z3_sort s) {
         Z3_TRY;
         LOG_Z3_is_re_sort(c, s);
         RESET_ERROR_CODE();
         bool result = mk_c(c)->sutil().is_re(to_sort(s));
-        return result?Z3_TRUE:Z3_FALSE;
-        Z3_CATCH_RETURN(Z3_FALSE);
+        return result?true:false;
```
NikolajBjorner: another few levels of indirection.

waywardmonkeys:
I'll submit another PR within a day or two that cleans up those as well.

NikolajBjorner:
I went over what I could find.

waywardmonkeys:
You got all of the ones that I saw! Thanks.

PR #1952, opened 2018-11-19
chrononological backtracking, improvements to seq solver

NikolajBjorner:
Add chronological backtracking to SAT core
Some improvements to theory_seq to handle str.to.int/int.to.str and not-contains


Rosmanrms:
Thank you....on pull request

zimmski:
@NikolajBjorner I am wondering if we should test this branch with our test suite and give you the failing test cases as feedback? If yes, could you maybe write a short description about what additional functionality/improvements are in the branch besides what is already mentioned.

NikolajBjorner:
This branch implements chronological backtracking in the sat solver. It is a risky change and I had to deal with performance regressions. It also contains some xor improvements, but these are turned off by default and have no measurable effect.
Currently there is a failing unit test in the PB solver features.

NikolajBjorner:
Note that chronological backtracking features are integrated among new local search engines.
The neuro features are still very experimental so holding off on these for the master branch.

PR #1963, opened 2018-11-25
Logging Improvements for the Axiom Profiler

Nils-Becker:
This pull request contains logging support for theory constraints added after a quantifier is instantiated as well as some smaller bug fixes for the existing logging support.


NikolajBjorner:
obj_hashtable<enode> instead of unordered set would work with other features, such as memory management.

- Nikolaj

Nils-Becker:
Thanks. I've updated the pull request, now.

src/smt/smt_quantifier.cpp
```diff
@@ -108,38 +109,39 @@ namespace smt {
...
             for (enode *it = en; it != root; it = it->get_trans_justification().m_target) {
-                if (it->m_proof_logged_status == smt::logged_status::NOT_LOGGED) {
-                    it->m_proof_logged_status = smt::logged_status::BEING_LOGGED;
-                    log_single_justification(log, it);
-                    it->m_proof_logged_status = smt::logged_status::LOGGED;
-                } else if (it->m_proof_logged_status != smt::logged_status::BEING_LOGGED && it->get_trans_justification().m_justification.get_kind() == smt::eq_justification::kind::CONGRUENCE) {
-
-                    // When the justification of an argument changes m_proof_logged_status is not reset => We need to check if the proofs of all arguments are logged.
-                    it->m_proof_logged_status = smt::logged_status::BEING_LOGGED;
+                if (already_visited.find(it) == already_visited.end()) already_visited.insert(it);
```
NikolajBjorner: if (!already_visited.contains(it)) already_visited.insert(it);

src/smt/smt_quantifier.cpp
```diff
@@ -108,38 +109,39 @@ namespace smt {
...
+                else break;
+
+                if (!it->m_proof_is_logged) {
+                    log_single_justification(log, it, already_visited);
+                    it->m_proof_is_logged = true;
+                } else if (it->get_trans_justification().m_justification.get_kind() == smt::eq_justification::kind::CONGRUENCE) {
```
NikolajBjorner: is this correct if commutativity was used? There is code below in a separate case that singles out commutativity as nyi

Nils-Becker: Yes, that would be a problem. Z3's E-graph only supports commutativity for binary functions, right? In that case it would be easy to add full support in another pull request. How would one cause such a justification to be generated? Is there a way to declare functions commutative in the SMT-file or is this just supported for built in functions?

Thanks,
Nils

Issue #973, opened 2017-04-10
Swapped significand and exponent in call to Context.mkFPNumeral


acr31:
Hi,

We're writing a new analysis plugin for http://errorprone.info and it's flagged some swapped arguments in Context.java

In [Context.java on line 3377 ](https://github.com/Z3Prover/z3/blob/master/src/api/java/Context.java#L3377) there is a call to Context.mkFPNumeral

    public FPNum mkFP(boolean sgn, long exp, long sig, FPSort s)
    {
         return mkFPNumeral(sgn, sig, exp, s);
    }

However mkFPNumeral takes an `exp` and a `sig` (contrary to the javadoc)

     public FPNum mkFPNumeral(boolean sgn, long exp, long sig, FPSort s)
     {
          return new FPNum(this, Native.mkFpaNumeralInt64Uint64(nCtx(), sgn, exp, sig, s.getNativeObject()));
     }

The native code signature is also in that order:

      Z3_ast Z3_API Z3_mk_fpa_numeral_int_uint(Z3_context c, Z3_bool sgn, signed exp, unsigned sig, Z3_sort ty)

PR #1991, opened 2018-11-30
Fix java api swap params

waywardmonkeys:


waywardmonkeys:
This resolves issue #973.

PR #1993, opened 2018-11-30
cmake: Look for the .git directory correctly.

waywardmonkeys:
The function called expects to be given the path to the .git
directory, not the path to the directory that contains the .git
directory.


waywardmonkeys:
@delcypher might want to take a look at this?

waywardmonkeys:
This should be fixed better like what was in #1552.

PR #1997, opened 2018-12-01
Change how 64 bit builds are detected.

waywardmonkeys:
Instead of doing this at configure time, we look at the actual
compile time status. This also avoids hardcoding checks based on
what CPU architecture is present, which doesn't work when Z3 is
being built on non-x86_64 platforms.


waywardmonkeys:
I tested this with cmake on macOS. It should be fine on anything using gcc or clang.

The MSVC compiler should be passing `_WIN64`, I think, and if so, in that case, it should be fine.

waywardmonkeys:
It seems like only the 32 bit build on Linux failed here ... but I'm not sure why yet.

waywardmonkeys:
Okay, this passes all of the tests in CI. This should be good now and it addresses some feedback that @NikolajBjorner and I got in email.

@NikolajBjorner : Could you verify that the build is correctly getting `_WIN64` on Windows in the right scenarios, please?

NikolajBjorner:
Will check

PR #1998, opened 2018-12-01
Remove include of immintrin.h.

waywardmonkeys:
This file doesn't appear to be used and isn't available on all
platforms.


waywardmonkeys:
I'm submitting this in the hopes that it works on all platforms since I can't test Windows or Linux myself.

PR #2005, opened 2018-12-03
Remove undef max/min on macOS.

waywardmonkeys:
This is no longer needed.


waywardmonkeys:
The errors in CI weren't due to my change here.

PR #2015, opened 2018-12-04
Use nullptr, not 0 in the C++ API impl.

waywardmonkeys:


NikolajBjorner:
some example programs are compiled using downlevel compiler settings where nullptr is not available. The Travis build does not go through, I merged to quickly.

waywardmonkeys:
Sorry about all of that. :( It is all fixed and improved in PR #2017.

PR #2017, opened 2018-12-05
Use nullptr, not 0 in the C++ API impl.

waywardmonkeys:
This time, it also includes `stddef.h`, so it should work everywhere.

### DIFF
--- a/examples/tptp/CMakeLists.txt
+++ b/examples/tptp/CMakeLists.txt
@@ -13,6 +13,22 @@ find_package(Z3
   # use this option.
   NO_DEFAULT_PATH
 )
+
+################################################################################
+# Z3 C++ API bindings require C++11
+################################################################################
+if ("${CMAKE_VERSION}" VERSION_LESS "3.1")
+  # Legacy CMake support
+  if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang"))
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This is failing because we need to build the examples with C++11. I'll keep at it.

examples/tptp/CMakeLists.txt
```diff
@@ -13,6 +13,22 @@ find_package(Z3
   # use this option.
   NO_DEFAULT_PATH
 )
+
+################################################################################
```
NikolajBjorner: they didn't before. I don't know if requiring C++11 here is an advantage: what if clients forget to require C++11. They their builds fail.

waywardmonkeys: We've been building the other examples with C++11 (I copied that from the `examples` `CMakeLists.txt`) ... also, in the 4.6 release notes, it indicates that C++11 is required for both building Z3 and using the C++ API.

waywardmonkeys: We could put a check for the C++ version in `z3++.h` and issue a warning / error if it is less than C++11?

NikolajBjorner:
This would be adding extra requirements to clients and I don't see what would justify these requirements.

waywardmonkeys:
I was going to follow this up with other usages of C++11 to address some issues with copying of data ... but I don't actually use the C++ API (or Java or dotnet or Python), so I'll probably hold off then. (I sent you a direct email a while ago about my rough directions.)

PR #2020, opened 2018-12-05
Fix typos.

waywardmonkeys:


waywardmonkeys:
Typos are getting far less common! :)

PR #2021, opened 2018-12-05
Simplify boolean code.

waywardmonkeys:
Now that the C API is using bool, this can be simplified.

### DIFF
--- a/src/api/api_fpa.cpp
+++ b/src/api/api_fpa.cpp
@@ -242,8 +242,8 @@ extern "C" {
             RETURN_Z3(nullptr);
         }
         api::context * ctx = mk_c(c);
-        expr * a = negative != 0 ? ctx->fpautil().mk_ninf(to_sort(s)) :
-                                   ctx->fpautil().mk_pinf(to_sort(s));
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
This is ready to go! It passes CI!

scripts/update_api.py
```diff
@@ -73,7 +73,7 @@ def is_obj(ty):
 
 # Mapping to .NET types
 Type2Dotnet = { VOID : 'void', VOID_PTR : 'IntPtr', INT : 'int', UINT : 'uint', INT64 : 'Int64', UINT64 : 'UInt64', DOUBLE : 'double',
-                FLOAT : 'float', STRING : 'string', STRING_PTR : 'byte**', BOOL : 'byte', SYMBOL : 'IntPtr',
```
NikolajBjorner: This is very risky and could break the C# API. 
.Net is broken in release 4.7 or the one before because 
it used a non-working combination of 'bool' here and <stdbool.h>.

waywardmonkeys:
Okay, I'll remove the dotnet changes from here for now then!

waywardmonkeys:
This PR now only modifies the C and C++ code.

PR #2026, opened 2018-12-07
Remove 'register' storage class specifiers.

waywardmonkeys:
The `register` storage class is deprecated in modern C++ and
generates warnings if this code is compiled with clang for C++11
or later.

This is generated code, but is checked in without the original
files that it was generated from.


NikolajBjorner:
this is output from a compiler. I wouldn't update it in place.

waywardmonkeys:
Where is the source file that it is generated from?

NikolajBjorner:
probably from tptp.org.

PR #2033, opened 2018-12-10
Use std::hexfloat more.

waywardmonkeys:
Previously, we were only using `std::hexfloat` on Windows on VS2013
and later. This flips that to only using non-hexfloat on Windows
where we're using older than VS2013.

Since `std::hexfloat` is part of C++11 and we require C++11 to build
the Z3 library, this should be supported everywhere.

A follow up commit can probably remove the std::hex code path entirely
since we don't support older than VS2013.


waywardmonkeys:
This is broken on gcc 4.8. :(

waywardmonkeys:
I'm not going to do anything about this for now.

PR #2052, opened 2018-12-28
 Substitute "Vars" in queries and  Make sure init is included when generalize

Chen-Huanyi:
- Original test cases do not contain any primary input, and the parser did not parse them. Add code in h2t parser for parsing them.
- When doing inductive generalization, make sure generated _init_ is in the generalized cube.
- Add few more test cases.


examples/python/mini_ic3.py
```diff
@@ -97,12 +104,24 @@ def subst_vars(self, prefix, inv, fml):
...
+            return [(f, self.mk_bool(prefix)) for f in inv.children()]
+        else:
+            vars = self.get_vars(inv)
+            return [(f, self.mk_bool(prefix)) for f in vars]
 
     def mk_bool(self, prefix):
         self.index += 1
         return Bool("%s%d" % (prefix, self.index))
 
+    def get_vars(self, f, rs=[]):
```
NikolajBjorner: possibly use a table to remember which terms were visitied and avoid recursing over the same sub-terms.

PR #2065, opened 2019-01-06
Define NO_Z3_DEBUGGER for iOS builds.

waywardmonkeys:
This is defined because we can't call `system` (via `invoke_gdb`)
on iOS and related platforms.


waywardmonkeys:
I pushed the wrong version ... one moment.

waywardmonkeys:
This is the correct version now.

PR #2132, opened 2019-02-10
Don't delete the reference to the native library in the python bindings

rhelmot:
In angr, we need direct access to the reference to the native library object.

I don't think this should cause any issues, but I also don't understand the issue which prompted this change, so someone else should test this or explain to me how to test it.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1704,7 +1704,6 @@ def write_exe_c_preamble(exe_c):
 def write_core_py_post(core_py):
   core_py.write("""
 # Clean up
-del _lib
 del _default_dirs
 del _all_dirs
 del _ext


wintersteiger:
Under some conditions, Python will delete (garbage collect) `_lib` earlier than expected. This prompted the change to keeping references to the library functions (-pointers) in the argument list of all Python functions, and the addition of `del _lib` just to make sure nobody expects `_lib` to stick around. I.e., after this change, there will be random garbage collector bug reports again.

NikolajBjorner:
should we revert this or can we make non-deletion optional for angr?

wintersteiger:
I think it would be better to revert it, but I have no understanding of what angr needs access to. @rhelmot can you give us an idea of what you need access to that isn't available otherwise?

rhelmot:
The thing we need is a version of `Z3_get_symbol_string` that doesn't use `_to_pystr` at the end - we identified this as the source of a huge slowdown while moving from python 2 to python 3.

Additionally, we sometimes use the lib object as a debugging tool to be able to tell which version of libz3.so got loaded, so there should probably be a string kept around which is the full path of the library that got loaded.

NikolajBjorner:
You say we should rewrite this:

    if sys.version < '3':
      def _to_pystr(s):
         return s
    else:
      def _to_pystr(s):
         if s != None:
            enc = sys.stdout.encoding
            if enc != None: return s.decode(enc)
            else: return s.decode('ascii')
         else:
            return ""

rhelmot:
No, I'm sure there are cases where it's important to get a string back. I'm saying we should add this:

```python
def Z3_get_symbol_string_bytes(a0, a1, _elems=Elementaries(_lib.Z3_get_symbol_string)):
  r = _elems.f(a0, a1)
  _elems.Check(a0)
  return r
```

wintersteiger:
Both of  those seem like good ideas that we should add anyways! The string decoding is a bit of a mess in Python, especially when moving between 2.x and 3.x, that's why all the re-encoding is in there (Unicode being the default in 3.x, etc).

PR #2146, opened 2019-02-22
Buffer and Vector Modernization Part 1

danielschemmel:
This is the first part of #1656:
- Move the old buffer and vector implementations away, and add a shim layer.
- Does *not* change the implementation yet.
- Should be safe to use for any configuration where it compiles in the first place


NikolajBjorner:
will be adding copyright headers to beginning of new files.

PR #2147, opened 2019-02-22
Buffer and Vector Modernization Part 2

danielschemmel:
This is the second part of #1656:
- Builds upon #2146 (which has been merged)
- Replaces the old buffer and vector implementations with the new one
- Attempts to keep the interface the same as it was before
- Additional testing would be appreciated, as I personally mostly use the more "basic" features of Z3 (QFBVA checking, potentially with a bit of FP thrown in for good measure).


src/math/polynomial/upolynomial_factorization.cpp
```diff
@@ -27,8 +27,6 @@ Module Name:
 #include "math/polynomial/upolynomial_factorization_int.h"
 #include "util/prime_generator.h"
 
-using namespace std;
```
nunoplopes: this change has nothing to do with the rest of the PR. Please don't mix everything in one go.

danielschemmel: Sorry, my bad. I thought I had removed that experiment, but I must have accidentally merged that commit with this one.

NikolajBjorner:
Seems the 32 bit build has some issues.
It fails the unit tests.
When I build locally I see:

[302/706] Building CXX object src\tactic\core\CMakeFiles\core_tactics.dir\symmetry_reduce_tactic.cpp.obj
d:\ds\src\util\buffvec.h(796): warning C4018: '<': signed/unsigned mismatch
d:\ds\src\util\buffvec.h(794): note: while compiling class template member function 'app **buffvec<app *,unsigned int,0>::erase(app *const *)'
d:\ds\src\util\buffvec.h(809): note: see reference to function template instantiation 'app **buffvec<app *,unsigned int,0>::erase(app *const *)' being compiled
D:\ds\src\ast/arith_decl_plugin.h(143): note: see reference to class template instantiation 'buffvec<app *,unsigned int,0>' being compiled
[321/706] Generating "D:/ds/build/src/ast/fpa/fpa2bv_rewriter_params.hpp" from "fpa2bv_rewriter_params.pyg"

danielschemmel:
I fixed the warning during 32 bit compilation, and will have a look at 32 bit compilation in general.

danielschemmel:
I have had a look at the 32 bit build, and the bug that I run into (passing a `this == nullptr` to `ast::get_ref_count` in ´src/ast/substitution/substitution.cpp:224`) also occurs on master.

When I comment out the offending `TRACE`, it will later case an assertion violation in `src/util/obj_ref.h`.

NikolajBjorner:
what is the input for the repro?

danielschemmel:
I ran cmake on an arch linux with GCC 8.2.1 as:
```
$ env CFLAGS='-m32 -fdiagnostics-color=always' CXXFLAGS='-m32 -fdiagnostics-color=always' CC=gcc CXX=g++ cmake -GNinja -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug ..
```
Built and called the unit test binary in gdb:
```
$ ninja z3 && ninja test-z3
$ gdb --args ./test-z3 /a
```

(I am currently validating this w/o ccache, just to be sure)

NikolajBjorner:
OK, so test-z3 /a fails.

NikolajBjorner:
any idea how to proceed?

NikolajBjorner:
seems to be getting closer. Still i686 exhibits something that must be related to 32/64 bit arithmetic.

danielschemmel:
The bug that I reported the last time has since been resolved on master. I have also added a small improvement for when `max` is defined as a function-like macro (as `Windows.h` likes to do). I am trying to run down what is still going wrong and whether it is related to my proposed changes

NikolajBjorner:
It's been dormant for some time.
It is fair to drop Travis builds for 14.04 and its C++ version, but another 32 bit build fails:
Testing regressions/smt2/t169.smt2
EXPECTED
sat
======================
PRODUCED
sat
(error "line 644 column 10: max. memory exceeded")
======================
Failed

This suggests there is a 32/64 bit bug.

NikolajBjorner:
will close this soon as no activity

PR #2150, opened 2019-02-23
Logging Support for Theory Solvers

Nils-Becker:
- Support for logging the meaning of constants within a theory. E.g. the log previously only showed that there is a constant belonging to the integer theory, now we are also able to tell its value.
- Support for logging instantiations of theory axioms.
- Support for logging MBQI instantiations.
- Support for logging names (from the original SMT query) and sorts of quantified variables.


src/ast/bv_decl_plugin.h
```diff
@@ -432,7 +433,53 @@ class bv_util : public bv_recognizers {
     app * mk_bvsmul_no_udfl(expr* m, expr* n) { return m_manager.mk_app(get_fid(), OP_BSMUL_NO_UDFL, n, m); }
     app * mk_bvumul_no_ovfl(expr* m, expr* n) { return m_manager.mk_app(get_fid(), OP_BUMUL_NO_OVFL, n, m); }
 
-    app * mk_bv(unsigned n, expr* const* es) { return m_manager.mk_app(get_fid(), OP_MKBV, n, es); }
+    private:
+    void log_bv_from_exprs(app * r, unsigned n, expr* const* es) {
+        if (m_manager.has_trace_stream()) {
+            for (unsigned i = 0; i < n; ++i) {
+                if (!m_manager.is_true(es[i]) && !m_manager.is_true(es[i]))
```
NikolajBjorner: is_false?

NikolajBjorner:
look like latent bugs, independent of pull request.

NikolajBjorner:
This introduced quite a bit of code that traces theory axiom instantiations. 
It would be more maintainable if this gets logged in a single place. Currently logs are scattered wherever a theory axiom is created, e.g., in theory_lra.
What would be required to support this? 
Furthermore, I would like to see if we can get to a point where Z3 logs clausal proofs (so without information about theory propagation) that can be checked using a weaker solver (that understands theory propagation, but not case splitting). Logging to the profiler trace and logging to these DRUP proofs would use similar features.

NikolajBjorner:
Upon further review it appears that there is too much trace stream dependencies pushed to the theory solvers. This is not going to be maintainable. 
It will be necessary to push this down to the smt_context level to have any hope for this to be maintainable. The fine-grained reverse engineering of theory axioms (replacing a >= 0 or C and a <= or C by a != 0 => C) is done in an ad-hoc way and in some cases I can see it uses unnecessary preconditions (for mod/div we don't check that the divisor is non-zero when it is a non-zero constant).

Can you change this so the trace-stream dependencies live at the smt::context level?

Nils-Becker:
In the long term we would like to log the axioms for each theory, whenever possible, and then log the instantiations of these axioms as instantiations of the corresponding quantifiers (in some cases it is even possible to find something similar to a trigger). We've implemented this for the datatype theory: axioms are logged as quantifiers in datatype_decl_plugin.cpp:456-496 and when we e.g. instantiate the accessor axioms we log a trigger based instantiation in theory_datatype.cpp:183. At the smt::context level we would not have the necessary information to log such instantiations. For rewritings which I expect cannot usually be expressed as quantifiers (e.g. there is no quantifier to express rewriting 1 + 1 to 2) we're already logging them on a more abstract level outside the theory specific rewriter in th_rewriter.cpp:581-585.

Similarly `[attach-meaning]` lines are used to provide some theory specific interpretation of constants (e.g. 0 instead of just Int()), which is information that only the theory solvers have. These are only generated in 7 places, though, so maintainability may not be too much of an issue.

NikolajBjorner:
I am afraid this will break as there is nothing that prevents code modifications to forget adding these axioms and it also clutters the code. Have you considered passing an extra argument to the mk_th_axiom method that provides instructions how to reconstruct the theory axiom. This is not unlike a justification object, but you can have theory specific instructions and assemble them one place instead of scattered throughout the code. As it stands theory_lra is currently much less readable.

PR #2283, opened 2019-05-15
Change from CMAKE_*_DIR to PROJECT_*_DIR

barcharcraz:
Z3 uses CMAKE_BINARY_DIR amd CMAKE_SOURCE_DIR all over the place to refer to the root of z3's source code tree. This is great except that it prevents users from embedding z3 as a subfolder in their project. Also PROJECT_BINARY_DIR and PROJECT_SOURCE_DIR are more clear imo.

PROJECT_SOURCE_DIR and PROJECT_BINARY_DIR have existed at least since cmake 3.0. I tested the build after this change but did not run the tests because I couldn't figure out how to build them :D

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,7 @@ set(z3_polluted_tree_msg
 ################################################################################
 # Sanity check - Disallow building in source
 ################################################################################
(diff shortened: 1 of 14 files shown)


NikolajBjorner:
tests are run by the CI, so this is fine.

PR #2330, opened 2019-06-08
Generate Z3ConfigVersion.cmake.

### LINKED ISSUE #2318: cmake import targets are lacking a Z3ConfigVersion.cmake file
Following https://cmake.org/cmake/help/latest/command/find_package.html?#version-selection, the command ``find_package(Z3 4.8.0 CONFIG)`` only works, if Z3 provides a ``Z3ConfigVersion.cmake`` file that performs version checking - however, currently an installation of Z3 is lacking this file (even though it does define Z3_VERSION* in Z3Config.cmake), preventing the use of cmake's version requirements.

ekpyron:
Closes #2318.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,16 +11,11 @@ if (POLICY CMP0042)
 endif()
 
 set(CMAKE_USER_MAKE_RULES_OVERRIDE_CXX "${CMAKE_CURRENT_SOURCE_DIR}/cmake/cxx_compiler_flags_overrides.cmake")
-project(Z3 CXX)
+project(Z3 VERSION 4.8.6.0 LANGUAGES CXX)
 
 ################################################################################
 # Project version
 ################################################################################
-set(Z3_VERSION_MAJOR 4)
-set(Z3_VERSION_MINOR 8)
-set(Z3_VERSION_PATCH 6)
-set(Z3_VERSION_TWEAK 0)
-set(Z3_VERSION "${Z3_VERSION_MAJOR}.${Z3_VERSION_MINOR}.${Z3_VERSION_PATCH}.${Z3_VERSION_TWEAK}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -11,16 +11,11 @@ if (POLICY CMP0042)
 endif()
 
 set(CMAKE_USER_MAKE_RULES_OVERRIDE_CXX "${CMAKE_CURRENT_SOURCE_DIR}/cmake/cxx_compiler_flags_overrides.cmake")
-project(Z3 CXX)
+project(Z3 VERSION 4.8.6.0 LANGUAGES CXX)
 
 ################################################################################
 # Project version
 ################################################################################
-set(Z3_VERSION_MAJOR 4)
```
ekpyron: If the version is provided to the ``project`` command (possibly since cmake 3.0), then these are defined automatically.

CMakeLists.txt
```diff
@@ -557,8 +552,10 @@ unset(Z3_SECOND_PACKAGE_INCLUDE_DIR)
 unset(Z3_CXX_PACKAGE_INCLUDE_DIR)
 unset(AUTO_GEN_MSG)
 unset(CONFIG_FILE_TYPE)
-# TODO: Provide a `Z3Version.cmake` file so that clients can specify the version
-# of Z3 they want.
+
+write_basic_package_version_file("${PROJECT_BINARY_DIR}/Z3ConfigVersion.cmake"
+  COMPATIBILITY SameMajorVersion
```
ekpyron: Can be changed to ``AnyNewerVersion``, if backwards compatibility across major releases is planned.

CMakeLists.txt
NikolajBjorner: thanks for heads up. Seems nice to have, but will pass on updating right away (because I could be doing it wrong).

CMakeLists.txt
NikolajBjorner: noted.

Issue #2354, opened 2019-06-21
Regression in Z3 timeout?


andrewvaughanj:
Hello,

I have two builds of Z3:

- [`606754c09`](https://github.com/Z3Prover/z3/commit/606754c09) (May 1st)
- [`b1893f2a5`](https://github.com/Z3Prover/z3/commit/b1893f2a5) (June 20th)

Given a particular Z3 Python script, and when using `606754c09`, I get a timeout (i.e., Z3 returns `unknown`). However, when using the same script under `b1893f2a5`, Z3 never terminates (i.e., it appears that the timeout parameter is being ignored). This timeout is set to 10000 ms.

I have captured the Z3 logs from these two runs:

- [606754c09_z3.log](https://gist.github.com/andrewvaughanj/8529e1513846144b5fa3d1fc547baa61)

- [b1893f2a5_z3.log](https://gist.github.com/andrewvaughanj/3bbf7fabd8a445e06b56301f28df9e1e)

The log for `b1893f2a5` ends when I kill the process; the log for `606754c09` has been "artificially truncated" to match that of `b1893f2a5` (i.e., after the check gives `unknown`, we do further work with Z3, which is irrelevant for this issue I believe).

Please let me know if there are any more details I can provide on this issue.

Cheers,

Andrew

andrewvaughanj:
I added the following:

```
z3.set_param("timeout", 10)
```

and it seems to "get stuck" with output like this:

```
(smt.delete-inactive-clauses  :num-deleted-clauses 188)
(smt.working :conflicts 245727 :agility 0.314074)
(smt.working :conflicts 247128 :agility 0.311207)
(smt.working :conflicts 248669 :agility 0.311665)
(smt.delete-inactive-clauses  :num-deleted-clauses 8888)
(smt.working :conflicts 250364 :agility 0.269951)
(smt.working :conflicts 252228 :agility 0.334213)
(smt.working :conflicts 253229 :agility 0.295869)
(smt.working :conflicts 254278 :agility 0.325157)
(smt.delete-inactive-clauses  :num-deleted-clauses 4100)
(smt.working :conflicts 255279 :agility 0.339337)
(smt.working :conflicts 256532 :agility 0.250027)
(smt.working :conflicts 257533 :agility 0.326521)
```

andrewvaughanj:
In case it is useful, I captured the verbose output from running under these two different builds:

[...]

PR #2356, opened 2019-06-21
Restoring the checking of the timeout timer (regression from 48fc3d752)

andrewvaughanj:
Fix for #2354

Signed-off-by: Andrew V. Jones <andrew.jones@vector.com>

### DIFF
--- a/src/smt/smt_context.cpp
+++ b/src/smt/smt_context.cpp
@@ -3726,6 +3726,11 @@ namespace smt {
             if (m_last_search_failure != OK)
                 return true;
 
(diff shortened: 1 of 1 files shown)


andrewvaughanj:
Looks like this functionality was broken here:

https://github.com/Z3Prover/z3/commit/48fc3d752ef56dec5c165b2e863f89aa81caeb1e#diff-5f65287f221a60407bf05209d67c8d9bL3734

andrewvaughanj:
Seems that this is due to use `smt.timeout` (which is deprecated over `timeout`).

PR #2393, opened 2019-07-10
Fix Incorrect Logging of Newly Introduced Terms During Rewrite

Nils-Becker:
We need to log an ``attach-enode`` line for every term (and only those terms) that was newly introduced by a rewrite step. (Fixes https://bitbucket.org/viperproject/axiom-profiler/issues/13/version-486-of-z3-not-compatible-with)

### DIFF
--- a/src/ast/rewriter/th_rewriter.cpp
+++ b/src/ast/rewriter/th_rewriter.cpp
@@ -570,16 +570,46 @@ struct th_rewriter_cfg : public default_rewriter_cfg {
         }
     }
 
-    void log_enodes_for_new_terms(expr *term) {
-        map<expr *, unsigned, ptr_hash<expr>, ptr_eq<expr>> reference_map;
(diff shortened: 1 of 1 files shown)


src/ast/rewriter/th_rewriter.cpp
```diff
@@ -581,8 +606,15 @@ struct th_rewriter_cfg : public default_rewriter_cfg {
                 result_pr = m().mk_rewrite(tmp, result);
             tmp = m().mk_eq(tmp, result);
             m().trace_stream() << "[instance] " << static_cast<void *>(nullptr) << " #" << tmp->get_id() << "\n";
-            m().trace_stream() << "[attach-enode] #" << result->get_id() << " 0\n";
-            m().trace_stream() << "[attach-enode] #" << tmp->get_id() << " 0\n";
+
```
NikolajBjorner: put the has_trace_stream() dependent code in a separate function

Issue #2378, opened 2019-07-03
 (error "query failed: Sorts (Array Int Int) and Int are incompatible") on instance that should likely be SAT


markusscherer:
When trying to solve an smtlib-instance we generated from a higher level representation, we encountered the error:

~~~
(error "query failed: Sorts (Array Int Int) and Int are incompatible")
unknown
~~~

We reproduced the error in versions `4.8.3` and `4.8.4` and `4.8.5` (partially in Linux, partially on MacOS), `4.6.0` reports `sat`, current master (`df405536`) reports `unsat`.

Unfortunately we did not succeed in in generating a "minimal" example, as the smallest instance showing this behavior we found still had 35k lines. What we have is a minimally different set of files, where adding two rules and a predicate changes a `sat`-instance to an instance exhibiting the error.

[instances.zip](https://github.com/Z3Prover/z3/files/3355117/instances.zip)

NikolajBjorner:
@agurfinkel

While the crash doesn't reproduce in the latest master the bug that is exposed in previous versions remains.
The code that handles arrays expects stores whenever it is marked. https://github.com/Z3Prover/z3/blob/master/src/qe/qe_arrays.cpp#L392
However, in this case it doesn't have a store expression, but an "ite". 
The test should be at least strengthened to include && m_arr_u.is_store(a_lhs)

A good idea is to make code like this:

      idxs.append(a_lhs->get_num_args() - 2, a_lhs->get_args() + 1);

validate that a_lhs is in fact an array by using a helper function. get_array_indices(a_lhs), get_num_array_indices(a_lhs) or similar.

agurfinkel:
I will need some time to process this. 
However, the requirement for that code is that the input is a conjunction of convex literals.
`ite` terms are neither expected nor supported. They should have been simplified using 
model based implicant computation. 

Maybe I am missing something...

agurfinkel:
@NikolajBjorner digging deeper into this example, there is a problem with model evaluator. I get the following assertion failure:
```
-------- [qe] linearize ../src/qe/qe_arith.cpp:67 ---------
(= MState_0_4904!slice!5_4_n
   (store (store ((as const (Array Int Int)) 0) 64 96)
          96
          95346918756310292697882537094886449535797654316421409133056955209608488550400)) := (let ((a!1 (lambda ((x!1 Int))
             (ite (= x!1 64)
                  96
                  (ite (= x!1 96)
[...]

PR #2399, opened 2019-07-13
proper fix to issue #2378

agurfinkel:
do not evaluate arrays as lambdas

### DIFF
--- a/src/model/model_evaluator.cpp
+++ b/src/model/model_evaluator.cpp
@@ -150,7 +150,8 @@ struct evaluator_cfg : public default_rewriter_cfg {
             expr * val = m_model.get_const_interp(f);
             if (val != nullptr) {
                 result = val;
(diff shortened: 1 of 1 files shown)


src/model/model_evaluator.cpp
```diff
@@ -150,7 +150,8 @@ struct evaluator_cfg : public default_rewriter_cfg {
             expr * val = m_model.get_const_interp(f);
             if (val != nullptr) {
                 result = val;
-                return m_ar.is_as_array(val)? BR_REWRITE1 : BR_DONE;
+                return BR_DONE;
+                //return m_ar.is_as_array(val)? BR_REWRITE1 : BR_DONE;
```
NikolajBjorner: this breaks other functionality

agurfinkel: Having it breaks functionality in spacer and seahorn. Is it possible to make it optional?

PR #2454, opened 2019-07-31
Fix sudoku Java example

RichardBradley:
There is a bug in the sudoku Java example: the code that is commented "`// each column contains a digit at most once`" actually enforces that each row contains a digit.

This PR fixes that.

I have also simplified the initial conditions so that it uses an "if" in the Java layer, rather than in the Z3 constraints layer, which is a little more efficient.


NikolajBjorner:
Thank you!

Issue #2298, opened 2019-05-23
Z3 and MSAN


evmaus:
Clang provides a "Memory Sanitizer", MSAN (https://clang.llvm.org/docs/MemorySanitizer.html#how-to-build), which allows users to find pernicious and hard to diagnose issues like use-after-frees.

My team has run into a few issues with enabling MSAN and using Z3--I'm working on a repro we can share of the individual problem (as our test code is closed source), but it looks like in some cases with strings there's the possibility to de-allocate twice (which may open you up to a use after free bug, as you decrement the reference count twice).

I've also found some cases of other MSAN issues by compiling and running the built in tests.  I can share the config.mk that I used if there's interest in seeing the stack traces from that.

Is there much interest from the Z3 developers in tracking these memory issues down and wiping them out?  We'd like to take a dependency on Z3, but we'd also like to use MSAN on our own tests.

NikolajBjorner:
So far the CI builds with UBSAN and ASAN. An additional configuration seems possible from documentation. Given that MSAN takes specific OS dependencies, trying it on the build integration isn't immediate for me. So feel free to paste z3 stack traces you find in the meanwhile

evmaus:
So I'm not certain how helpful just the stack traces will be, but we're getting this stack trace:

0 0x7fc92250ed1f in ast_manager::dec_ref(ast*) third_party/z3/src/ast/ast.h:1671:13
    #1 0x7fc9241e0aac in smt::context::flush() third_party/z3/src/smt/smt_context.cpp:2911:23
    #2 0x7fc9241df772 in smt::context::~context() third_party/z3/src/smt/smt_context.cpp:233:9
    #3 0x7fc924291744 in smt::kernel::imp::~imp() third_party/z3/src/smt/smt_kernel.cpp:26:20
    #4 0x7fc924291484 in void dealloc<smt::kernel::imp>(smt::kernel::imp*) third_party/z3/src/util/memory_manager.h:96:11
    #5 0x7fc92428fb4d in smt::kernel::~kernel() third_party/z3/src/smt/smt_kernel.cpp:238:9
    #6 0x7fc924376eb2 in smt::smt_solver::~smt_solver() third_party/z3/src/smt/smt_solver.cpp:109:9
    #7 0x7fc922586189 in void dealloc<check_sat_result>(check_sat_result*) third_party/z3/src/util/memory_manager.h:96:11
[...]

PR #2477, opened 2019-08-09
MSAN Integration Build

evmaus:
This adds an MSAN build to azure-pipelines.yml per issue https://github.com/Z3Prover/z3/issues/2298

MSAN is a little weird, in that you need to link with a special copy of libc++ that's been compiled with MSAN in order to get good results (i.e. not just false positives).  The script this adds that runs at the start of the task downloads libc++ from SVN and builds it with MSAN, then uses that to link against Z3.  From testing with azure pipelines, the libc++ download and compilation seems to add a few minutes to the overall build, but certainly doesn't dominate.

Regression tests don't pass with MSAN at the moment--some of them trigger exit code 77 (which appears to be the exit code for "MSAN issue", according to a cursory search).

It seems like the example CMakes don't pick up the parameters passed for compiling/linking/etc, so I disabled those for MSAN's build.


NikolajBjorner:
I assume the Z3Build output is visible. 
Besides the odd return code of 77 it also has a set of warnings that are not flagged as errors and don't provide stack symbols.

(test fixed_bit_vector :time 0.00 :before-memory 4.59 :after-memory 4.59)
==14769==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x47f05e6  (/home/vsts/work/1/s/build/test-z3+0x47f05e6)
    #1 0x47f06c4  (/home/vsts/work/1/s/build/test-z3+0x47f06c4)
    #2 0x5ec4801  (/home/vsts/work/1/s/build/test-z3+0x5ec4801)
    #3 0x5ec38d0  (/home/vsts/work/1/s/build/test-z3+0x5ec38d0)
    #4 0x5c9b6c8  (/home/vsts/work/1/s/build/test-z3+0x5c9b6c8)
    #5 0x7f07e397782f  (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #6 0x42f278  (/home/vsts/work/1/s/build/test-z3+0x42f278)

  Uninitialized value was stored to memory at
    #0 0x47ecec2  (/home/vsts/work/1/s/build/test-z3+0x47ecec2)
    #1 0x5ec4732  (/home/vsts/work/1/s/build/test-z3+0x5ec4732)
    #2 0x5ec38d0  (/home/vsts/work/1/s/build/test-z3+0x5ec38d0)
    #3 0x5c9b6c8  (/home/vsts/work/1/s/build/test-z3+0x5c9b6c8)
    #4 0x7f07e397782f  (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

  Uninitialized value was stored to memory at
    #0 0x47ecec2  (/home/vsts/work/1/s/build/test-z3+0x47ecec2)
    #1 0x5ec4732  (/home/vsts/work/1/s/build/test-z3+0x5ec4732)
    #2 0x5ec38d0  (/home/vsts/work/1/s/build/test-z3+0x5ec38d0)
    #3 0x5c9b6c8  (/home/vsts/work/1/s/build/test-z3+0x5c9b6c8)
    #4 0x7f07e397782f  (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

evmaus:
So I dug into this a little bit, and I think I've fixed the symbolization issue--we need to have llvm-symbolizer on the path (literally named that--Ubuntu puts llvm-symbolizer-x.x on the path), or have an environment variable set to the correct location for llvm-symbolizer (which I think I have right for Ubuntu 16.04).

I've committed something I think will fix the symbolization issue.  I'll wait for the build to complete and see if it did--if it didn't, I'll figure out the best way to set the variable across 'script' blocks.  (That seems to be the most likely issue.)

As far as the odd exit code, that's caused by MSAN issues in the regression tests.  I can either disable them for this build type (keeping a green build, but not showing the issues) or leave them as is.

NikolajBjorner:
cool, now I can see it is related to the tbv test.

evmaus:
Awesome; let me know if there's anything else that you'd like changed in this build.  :)

NikolajBjorner:
the errors on the unit tests are now fixed. I disabled the regressions from the build as they fail for whatever other reason

PR #2486, opened 2019-08-13
Handle dynamic sort of Nth()'s return value in the Python API

pschanely:
When using the Python API at HEAD, I noticed sort mismatches doing things like:
```
a = Const('a', SeqSort(IntSort()))
a[3] + 1
```
No idea whether this the following fix is the right one, but it helped me. Use or ignore at will :)


NikolajBjorner:
yes the fix is correct, thanks

PR #2490, opened 2019-08-13
Allow creating select-over-lambda in python API

agurfinkel:
Python API is more restrictive than necessary. It requires
array consrtuctors to apply to ArrayRef expressions, whereas they are applicable
to lambda terms as well

### DIFF
--- a/src/api/python/z3/z3.py
+++ b/src/api/python/z3/z3.py
@@ -4288,6 +4288,11 @@ def is_array(a):
     """
     return isinstance(a, ArrayRef)
 
+def is_array_sorted_expr(a):
+    """Return `True' if `a' is a Z3 expression of sort array.
+    """
+    return a.sort_kind() == Z3_ARRAY_SORT
+
 def is_const_array(a):
     """Return `True` if `a` is a Z3 constant array.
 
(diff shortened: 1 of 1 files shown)


src/api/python/z3/z3.py
```diff
@@ -4456,9 +4461,12 @@ def Select(a, i):
     True
     """
     if z3_debug():
-        _z3_assert(is_array(a), "First argument must be a Z3 array expression")
-    return a[i]
-
+        _z3_assert(is_array_sorted_expr(a), "First argument must be a Z3 array expression")
+    if is_array(a):
+        return a[i]
```
NikolajBjorner: why are there two cases? both should becomes mk_select. The bug is the test "is_array" instead of "is_array_sorted_expr".

agurfinkel: lambda is a `QuantifierRef`. `QuantifierRef` does not implement `__getitem__`. 

The code can be moved into `__getitem__` of `QuantifierRef`, but then it should check that the quantifier is a lambda and not something else. I think it is cleaner the way it is now.

NikolajBjorner:
subsumed by other commit

PR #2496, opened 2019-08-15
Fix whitespace issues in *.pyg.

waywardmonkeys:


waywardmonkeys:
This mainly removes tabs from `*.pyg` files. In files that changed, it also removes some whitespace at the ends of lines.

PR #2611, opened 2019-10-02
Changed makefile generation for ml bindings to use OCAMLFIND variable

philzook58:
I believe the python mk_util.py file was changed recently to add a new dependency for the ocaml bindings and some variables got mixed up. It did not build on my system until I made these changes.


NikolajBjorner:
thanks!

Issue #2400, opened 2019-07-13
Getting Null FuncInterp for Arrays


schedutron:
I'm using the Z3 Java API to parse a SMT file; when I call `s.getModel()`, where `s` is a solver, I'm able to print the model correctly as follows:
```
(define-fun O_STUDENT () (Array Int STUDENT_TupleType)
  ((as const (Array Int STUDENT_TupleType))
  (STUDENT_TupleType _ID__12345 _NAME__Bourikas _DEPT_uNAME__Music 29)))
(define-fun O_DEPARTMENT () (Array Int DEPARTMENT_TupleType)
  ((as const (Array Int DEPARTMENT_TupleType))
  (DEPARTMENT_TupleType _DEPT_uNAME__Music _BUILDING__BUILDING_u5 50000.0)))
```

However, when I try to get the `FuncInterp` for `O_STUDENT` array, I get a `null`. I'm using the following to extract that value:
```
if (s.check() == Status.SATISFIABLE) {
            Model m = s.getModel();
            FuncDecl arrayDep = m.getConstDecls()[0];
            System.out.println(m.getFuncInterp(arrayDep));
        }
```

Using `m.getConstInterp` instead of `m.getFuncInterp` gives
```
com.microsoft.z3.Z3Exception: Non-zero arity functions and arrays have FunctionInterpretations as a model. Use getFuncInterp.
```

I even tried the Z3 Java example at https://github.com/Z3Prover/z3/blob/659be6940ba4108b82ed9af3f2734e9c95e9aa21/examples/java/JavaExample.java#L353, which also gives `null`s for array interpretations. Following is the output for the Java example:

```
Interpretation of MyArray:
null
Interpretation of x:
123
Interpretation of f:
[else -> 0]
Interpretation of MyArray as Term:
null
```

schedutron:
This corresponds to the [question I asked on Stackoverflow](https://stackoverflow.com/questions/56922661/z3-getting-funcinterp-as-null/56931334?noredirect=1#comment100439020_56931334).

NikolajBjorner:
I believe this is fixed now. Can you confirm?

schedutron:
Yes, I downloaded the nightly build - `getFuncInterp()` still gives `null` but `getConstInterp()` is returning the correct interpretation. Thanks!

PR #2614, opened 2019-10-04
Reconcile ocaml get_const_interp api with Java and C#

philzook58:
This commit for issue #2400 was not propagated over to the ocaml bindings.
https://github.com/Z3Prover/z3/commit/2d4e9a0f67983e8dca9a84666d0ff5f31fb5cb09


NikolajBjorner:
makes sense, thanks

PR #2645, opened 2019-10-21
Update README.md

shrin18:
Negligible but important


andrewvaughanj:
FYI, `nmake` is a Visual Studio command and it seems that, given you've changed `nmake` to `make` inside of a VS-based block, this change is incorrect.

shrin18:
Sorry my bad

PR #2655, opened 2019-10-24
Add MSVC Arm64 support

janisozaur:


janisozaur:
Thanks, that was quick

Issue #2457, opened 2019-08-02
SIGSEGV  with `--staticbin`


SaswatPadhi:
A statically linked z3 binary compiled from the master branch fails to run and terminates with a segmentation fault (SIGSEGV signal).

Steps to reproduce: (tested on Debian 10)
1. `git checkout master`
2. `python scripts/mk_make.py --staticbin`
3. `cd build && make`
4. `./z3 -v`

Without the `--staticbin` flag, the compiled z3 binary runs as expected.

NikolajBjorner:
Perhaps related to recent use of lpthread. 
https://stackoverflow.com/questions/35116327/when-g-static-link-pthread-cause-segmentation-fault-why

fpoli:
Is there any workaround to statically compile z3?

fpoli:
The `python scripts/mk_make.py --staticbin` command generates a `build/config.mk` containing `LINK_EXTRA_FLAGS= -lpthread`. If I manually modify this to `LINK_EXTRA_FLAGS= -pthread -Wl,--whole-archive -lpthread -Wl,--no-whole-archive` (as suggested in the SO question) then the compiled z3 runs fine. However, I'm not sure how to generate this modified configuration using the `mx_make.py` script.

SaswatPadhi:
I pulled the latest changes and rebuilt z3 on my Debian machine -- works as expected with both `--staticbin` and `--staticlib`.

Thanks @NikolajBjorner  and @fpoli.

SaswatPadhi:
Actually, I just noticed that although the build was successful, the `z3` binary still fails on my Debian machine with `SIGSEGV` ...

@fpoli: Does `z3` run as expected on your Ubuntu machine?

Sorry for the confusion!

fpoli:
@SaswatPadhi on various Ubuntu 16.04 computers that I tried yes, it runs as expected. Before, `z3 --version` was enough to cause a segmentation fault.

Have you checked if your `build/config.mk` actually contains `LINK_EXTRA_FLAGS= -Wl,--whole-archive -lpthread -Wl,--no-whole-archive`?

SaswatPadhi:
Hello,

Sorry for the delay in getting back -- I had a deadline.

I pulled the latest changes and checked again that `build/config.mk` contains `LINK_EXTRA_FLAGS= -Wl,--whole-archive -lpthread -Wl,--no-whole-archive`. Here is my `config.mk` file:

```make
PREFIX=/usr
[...]

PR #2683, opened 2019-11-06
Link pthread using the --whole-archive option

fpoli:
This fixes a SIGSEGV on Ubuntu 16.04 when running z3 compiled with `--staticbin` (issue #2457). It seems that without the `--whole-archive` option the linker does not statically link all pthread symbols.

The fix is described here: https://stackoverflow.com/a/45271521/2491528

Note: also `SLINK_EXTRA_FLAGS` provides a `-lpthread`, but I haven't modified it because I'm not sure how it's used.

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -2631,7 +2631,10 @@ def mk_config():
         config.write('LINK=%s\n' % CXX)
         config.write('LINK_FLAGS=\n')
         config.write('LINK_OUT_FLAG=-o \n')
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
shouldn't it depend on is_static_bin/is_static_lib?

fpoli:
> shouldn't it depend on is_static_bin/is_static_lib?

Ah, yes. Maybe also on `is_linux()`? It seems that `--whole-archive` is not available on MacOS (https://z3build.visualstudio.com/Z3Build/_build/results?buildId=1620).

PR #2732, opened 2019-11-23
Align CMake option for python executable with documentation

nmeum:
The documentation (`README-CMake.md`) mentions `Z3_PYTHON_EXECUTABLE` as
the option which should be used to specify the python executable to be
used by Z3. Unfortunately, that variable is currently not being used in
any of the CMakeLists files and therefore ignored.

It seems to me that this is a regression introduced in
53a01a07bdf245353bd0288fc50efb2739d3d3be which updated the
`README-CMake.md` but not the CMake files in regards to this variable.

---

Patch was generated by simply substituting the variable across all files in the repo.


NikolajBjorner:
thanks for the catch.
The build got broken, however. Some seems to be fixable by updating .travis.yml, but there seems to be more sources. For now, I am fixing the doc to be more accurate.

NikolajBjorner:
Seems the root is that it uses PythonInterp, that defines PYTHON_EXECUTABLE.
https://cmake.org/cmake/help/v3.0/module/FindPythonInterp.html

Thus, PYTHON_EXECUTABLE is not specific to Z3, but should have the same meaning across different projects. I believe it should not be called Z3_PYTHON_EXECUTABLE.

PR #2738, opened 2019-11-25
Make z3::exception a subclass of std::exception

ahumenberger:
I'm writing a [Julia wrapper](https://github.com/ahumenberger/Z3.jl) for Z3 by using [CxxWrap.jl](https://github.com/JuliaInterop/CxxWrap.jl). CxxWrap makes wrapping C++ code pretty easy and straightforward, and it also catches `std::exception`s and turns them into Julia exceptions. So, instead of Julia crashing because of the unhandled exception, this change would improve the experience in Julia (especially in REPL mode). Is there any reason for not subclassing `std::exception`?

### DIFF
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -80,11 +80,12 @@ namespace z3 {
     /**
        \brief Exception used to sign API usage errors.
     */
-    class exception {
+    class exception : public std::exception {
         std::string m_msg;
     public:
         exception(char const * msg):m_msg(msg) {}
         char const * msg() const { return m_msg.c_str(); }
+        char const * what() const noexcept { return m_msg.c_str(); }
         friend std::ostream & operator<<(std::ostream & out, exception const & e);
     };
     inline std::ostream & operator<<(std::ostream & out, exception const & e) { out << e.msg(); return out; }


NikolajBjorner:
Seems like a very reasonable change. At this point, the change doesn't compile.
The 'override' specifier seems to be missing from "what()".

```
/home/vsts/work/1/s/src/api/c++/z3++.h:88:29: error: expected ‘;’ at end of member declaration
         char const * what() const noexcept { return m_msg.c_str(); }
                             ^
/home/vsts/work/1/s/src/api/c++/z3++.h:88:35: error: ‘noexcept’ does not name a type
         char const * what() const noexcept { return m_msg.c_str(); }
                                   ^
/home/vsts/work/1/s/src/api/c++/z3++.h:88:35: note: C++11 ‘noexcept’ only available with -std=c++11 or -std=gnu++11
/home/vsts/work/1/s/src/api/c++/z3++.h:88:22: error: looser throw specifier for ‘virtual const char* z3::exception::what() const’
         char const * what() const noexcept { return m_msg.c_str(); }
                      ^
In file included from /usr/include/c++/5/new:40:0,
                 from /usr/include/c++/5/ext/new_allocator.h:33,
                 from /usr/include/x86_64-linux-gnu/c++/5/bits/c++allocator.h:33,
                 from /usr/include/c++/5/bits/allocator.h:46,
                 from /usr/include/c++/5/string:41,
                 from /home/vsts/work/1/s/examples/tptp/tptp5.cpp:7:
/usr/include/c++/5/exception:68:25: error:   overriding ‘virtual const char* std::exception::what() const throw ()’
     virtual const char* what() const _GLIBCXX_USE_NOEXCEPT;
                         ^
In file included from /home/vsts/work/1/s/examples/tptp/tptp5.cpp:20:0:
/home/vsts/work/1/s/src/api/c++/z3++.h:83:11: error: looser throw specifier for ‘virtual z3::exception::~exception()’
     class exception : public std::exception {
           ^
```

ahumenberger:
It seems to me that it has to do with the `noexcept` identifier and C++11.

I checked the following CMake file whether I can assume a lower bound of the C++ standard.
https://github.com/Z3Prover/z3/blob/cb6d3d245809fa0ba827f4adfc704e5cde761a26/examples/c%2B%2B/CMakeLists.txt#L17-L21

The build of the above example actually succeeds. The `tptp` example fails. It seems that `examples/tptp/CMakeLists.txt` does not set a C++ standard.

Can I actually assume C++11 as the above CMake file suggests?

NikolajBjorner:
cmake for tptp should be fixed, sorry

NikolajBjorner:
restarting build...

PR #2745, opened 2019-11-27
Fix #2736

### LINKED ISSUE #2736: [Consolidated] Soundness bugs in z3str3
Hi,
on the following unsatisfiable formula z3str3 incorrectly reports "sat" 

```
(declare-fun i () Int)
(assert (< i 1))
(assert (= (int.to.str i) "00"))
(check-sat)
```

OS: Ubuntu 18.04
Revision: ba03d99

FedericoAureliano:
Issue #2736 exploits that the translation from a string to an int in finalcheck_int2str ignores leading zeros. The fix is to enforce that int.to.str does not produce a string with a prefix of "0", unless the string is "0".

### DIFF
--- a/src/smt/theory_str.cpp
+++ b/src/smt/theory_str.cpp
@@ -1879,6 +1879,10 @@ namespace smt {
             expr_ref axiom1(ctx.mk_eq_atom(axiom1_lhs, axiom1_rhs), m);
             SASSERT(axiom1);
             assert_axiom(axiom1);
+
+            expr_ref zero(mk_string("0"), m);
+            expr_ref pref(u.str.mk_prefix(zero, ex), m);
+            assert_implication(pref, ctx.mk_eq_atom(ex, zero));
         }
     }
 
--- a/src/smt/theory_str.cpp
+++ b/src/smt/theory_str.cpp
@@ -1884,8 +1884,9 @@ namespace smt {
         // axiom 2: (str.from-int N) should not result in a string with leading zeros.
         expr_ref zero(mk_string("0"), m);
         expr_ref pref(u.str.mk_prefix(zero, ex), m);
-        // The result does not start with a "0" xor the result is "0"
-        assert_axiom(m.mk_or(m.mk_and(mk_not(m, pref), ctx.mk_eq_atom(ex, zero)), m.mk_and(pref, mk_not(m, ctx.mk_eq_atom(ex, zero)))));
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
@mtrberzi - can you review and comment/sign off?

mtrberzi:
@NikolajBjorner will do - Federico and I have been discussing this fix in an email thread. I need to double-check the axiom as it might need to change slightly.

src/smt/theory_str.cpp
```diff
@@ -1880,6 +1880,16 @@ namespace smt {
...
+
+        // axiom 2: The only (str.from-int N) that starts with a "0" is "0".
+        expr_ref zero(mk_string("0"), m);
+        // let (the result starts with a "0") be p
+        expr_ref p(u.str.mk_prefix(zero, ex), m);
+        // let (the result is "0") be q  
+        expr_ref q(ctx.mk_eq_atom(ex, zero), m);
+        // encoding: the result does NOT start with a "0" (~p) xor the result is "0" (q)
+        // ~p xor q == (~p or q) and (p or ~q)
+        assert_axiom(m.mk_and(m.mk_or(mk_not(m, p), q), m.mk_or(p, mk_not(m, q))));
```
mtrberzi: Is there a reason why the function form of `mk_not` was used and not the method on `ast_manager` (i.e. `m.mk_not(p)`)?

src/smt/theory_str.cpp
```diff
@@ -1880,6 +1880,16 @@ namespace smt {
             SASSERT(axiom1);
             assert_axiom(axiom1);
         }
+
+        // axiom 2: The only (str.from-int N) that starts with a "0" is "0".
+        expr_ref zero(mk_string("0"), m);
+        // let (the result starts with a "0") be p
+        expr_ref p(u.str.mk_prefix(zero, ex), m);
```
mtrberzi: I would suggest using more descriptive variable names, as it will greatly help the next time this code needs to be debugged or modified.

mtrberzi:
@FedericoAureliano I have some review comments for you when you are able to take a look.

src/smt/theory_str.cpp
FedericoAureliano: To be consistent with the existing code (e.g. line 1877 right above).

FedericoAureliano:
@mtrberzi fixed!

PR #2746, opened 2019-11-27
Link librt when compiling with --staticbin on Linux

fpoli:
Fix for issue #2457. The workaround is described here: https://stackoverflow.com/questions/58848694/gcc-whole-archive-recipe-for-static-linking-to-pthread-stopped-working-in-rec.

On Ubuntu 19.04 this is enough to avoid a SIGSEGV when running `z3 -v`.


NikolajBjorner:
Thank you!

PR #2876, opened 2020-01-22
Added FreshFunction to Python bindings.

ritave:
All other declarations can be done use appropriate Fresh*() call,
or FreshConst() with a desired sort, except Functions.

I've added the ability to do that in Python bindings using already existing APIs.

Other `Fresh*()` functions allow for a custom prefix, but a function signature of `def FreshFunction(*sig, prefix='f')` is not a valid Python2 code. Since we're still supporting version 2, I've decided to remove it altogether


NikolajBjorner:
Thanks!

PR #2881, opened 2020-01-23
Fix EOF detection when char is unsigned.

jamesjer:
On systems where the `char` type is unsigned, such as s390x, gcc complains:
```
../src/parsers/util/scanner.cpp: In member function ‘scanner::token scanner::scan()’:
../src/parsers/util/scanner.cpp:483:9: warning: case label value is less than minimum value for type [-Wswitch-outside-range]
  483 |         case -1:
      |         ^~~~
```
This patch creates a case label with the correct `char` value, whether `char` is signed or unsigned.


NikolajBjorner:
thanks, will let builds finish

PR #2899, opened 2020-01-29
Add explicit instantiation of update_inf_cost_for_column_tableau.

jamesjer:
We are currently building all Fedora packages with an early version of GCC 10.  The z3 package build failed on s390x due to a missing template instantiation.  Subsequent discussion is here:

https://bugzilla.redhat.com/show_bug.cgi?id=1794127

As noted in that bug, the explicit instantiation is needed to make the code correct.  See the link above for references to the standard.


NikolajBjorner:
@levnach: I believe this happened quite a bit with the way the lp solver uses templates. I wonder, is there a way we can avoid the template entanglement in a more systematic way?

Note: Pull will be merged when build finishes.

NikolajBjorner:
Note: if there is an interest we can, based on whether it is sufficiently feasible to do, add a fedora pipeline to the checkin builds to guard against some of these issues. The pipelines are defined under https://github.com/Z3Prover/z3/blob/master/azure-pipelines.yml and can use docker images (similar to the centos build for pip).

jwakely:
> I wonder, is there a way we can avoid the template entanglement in a more systematic way?

I'm not aware of any simple way to make this kind of latent bug turn into an error, so it could be caught earlier.

GCC has the `-fno-implicit-templates` option to disable implicit instantiations (and so force you to explicitly instantiate the templates you actually use), but that affects **all** templates used in the translation unit. So you'd need to explicitly instantiate `std::list<unsigned>`, `std::vector<std::pair<unsigned, unsigned>>` etc. in `lp_primal_core_solver.cpp` (or just be prepared to deal with lots of linker errors and ignore the ones that aren't instantiations of your own templates).

PR #2942, opened 2020-02-06
z3str3: bitvector-based model construction

mtrberzi:
This patch adds a new model construction algorithm to `theory_str` based on a reduction to fixed-length bitvector constraints. It performs significantly better than the previous length+value tester method, does not introduce any temporary variables into the main solver context (a subsolver is used for this), and fixes numerous issues with the previous algorithm.

There are two placeholder "not implemented" calls under a feature flag that defaults to OFF because that code is still in the process of being brought over from a different development branch and is not yet robust enough to enable.

Once this is merged I will revisit all open issues tagged `z3str3` to verify whether they have been resolved by this patch.


src/smt/theory_str_mc.cpp
```diff
@@ -43,6 +45,1068 @@ namespace smt {
...
+        if (str_chars.empty()) {
+            // check 0-length solution
+            bool zero_solution = false;
+            unsigned initial_state = aut->init();
+            if (aut->is_final_state(initial_state)) {
+                zero_solution = true;
+            } else {
+                unsigned_vector eps_states;
+                aut->get_epsilon_closure(initial_state, eps_states);
+                for (unsigned_vector::iterator it = eps_states.begin(); it != eps_states.end(); ++it) {
```
NikolajBjorner: why not just 
```
for (unsigned state : eps_states) {

}
```

src/smt/theory_str_mc.cpp
```diff
@@ -43,6 +45,1068 @@ namespace smt {
...
+            maps[0].insert(aut->init(), m.mk_true());
+            // is_accepted(a, aut) & some state in frontier is final.
+
+            for (unsigned i = 0; i < str_chars.size(); ++i) {
+                u_map<expr*>& frontier = maps[select_map];
+                u_map<expr*>& next = maps[!select_map];
+                select_map = !select_map;
+                ch = str_chars.get(i);
+                next.reset();
+                u_map<expr*>::iterator it = frontier.begin(), end = frontier.end();
```
NikolajBjorner: again, c++11 makes for loops so much nicer

src/smt/theory_str_mc.cpp
```diff
@@ -43,6 +45,1068 @@ namespace smt {
...
+            sort * bv8_sort = bv.mk_sort(8);
+            for (unsigned i = 0; i < 256; ++i) {
+                rational ch(i);
+                expr_ref chTerm(bv.mk_numeral(ch, bv8_sort), m);
+                bitvector_character_constants.push_back(chTerm);
+                fixed_length_subterm_trail.push_back(chTerm);
+            }
+        }
+
+        if (is_trace_enabled("str")) {
```
NikolajBjorner: TRACE and CTRACE

PR #2987, opened 2020-02-12
Down compatibility for C++03

### LINKED ISSUE #2984: Old C++ standards are not supported by Z3
@NikolajBjorner, it would be nice if you could avoid using C++11 in `z3++.h` header. Our tool uses old C++03 and we are not able to build with fresh z3 from now on.

https://github.com/Z3Prover/z3/blob/f161bdaf8faf92b23a07d9a3c89f94b448791f0b/src/api/c%2B%2B/z3%2B%2B.h#L2787

SweetVishnya:
Closes #2984


NikolajBjorner:
Thanks!

Issue #3020, opened 2020-02-16
z3str3 invalid model on formula with str.indexof


muchang:
Hi,
For this formula:
```
(declare-fun r () String)
(declare-fun s () String)
(declare-fun t () String)
(assert (= s t))
(assert (= (str.indexof s r 1) (str.indexof t r 0)))
(check-sat)
(get-model)
```
Z3 smt.string_solver=z3str3  gives an invalid model:
```
(model 
  (define-fun t () String
    "?")
  (define-fun r () String
    "?")
  (define-fun s () String
    "?")
)
```
if I feed this model to the formula, Z3 reports unsat.
if we swap '0' with '1' from the bug-triggering test, z3str3 now gives a correct model:
```
[525] % z3 smt.string_solver=z3str3 model_validate=true small2.smt2 
sat
(model 
  (define-fun t () String
    "?")
  (define-fun r () String
    "\x00\x00")
  (define-fun s () String
    "?")
)
[526] % cat small2.smt2 
(declare-fun r () String)
(declare-fun s () String)
(declare-fun t () String)
(assert (= s t))
(assert (= (str.indexof s r 0) (str.indexof t r 1)))
(check-sat)
(get-model)
```

OS: Ubuntu 18.04
Commit: aaf2c3b

mtrberzi:
Still open after #3032

mtrberzi:
Possible fix for this one being reviewed and tested, may fix other open issues as well

mtrberzi:
Opened #3095

Issue #3005, opened 2020-02-14
z3str3 incorrect answer on QF_S formula with str.substr str.indexof str.len


muchang:
Hi,
For this formula:
```
(declare-fun a () String)
(assert
 (let ((b (str.indexof (str.substr a 10 (- (str.len a) 0)) "*" 0))
       (c (str.substr a (+ 0 1) (- (str.len a) 0))))
  (let ((d (str.substr (str.substr a 10 (- (str.len a) 0)) 0 (- b 0)))
        (i (distinct (ite (str.contains c "-") 1 0) 0)))
   (let ((e (str.substr d 1 (- (str.len d) 0)))
         (f (str.substr d 0 (- (+ (str.indexof d "-" 0) 1) 0))))
    (let ((j (= (str.++ (str.substr d 0 0) e) "+"))
          (g (distinct (str.++ f e) "+")))
     (let ((h (distinct (ite (not j) 1 0) 0)))
      (and (not h) (distinct (ite g 1 0) 0) (not i))))))))
(check-sat)
```
Z3 smt.string_solver=z3str3 incorrectly gives sat, while this formula should be unsat.

OS: Ubuntu 18.04
Commit: 99b71a9

mtrberzi:
Still open after #3032

mtrberzi:
Times out as of the latest commit, closing this issue for now as the erroneous behaviour no longer occurs.

Issue #2996, opened 2020-02-13
z3str3 Segmentation fault on the formula with str.indexof


muchang:
Hi,
For this formula,
```
(declare-fun s () String)
(assert (distinct s "aabb"))
(assert (< (str.indexof s "aa" 10) (str.indexof s "bb" 180)))
(check-sat)
```
`z3 smt.string_solver=z3str3` throws out a segmentation fault.

OS: Ubuntu 18.04
Commit: 0146259

mtrberzi:
Fixed in #3032

Issue #2991, opened 2020-02-13
z3str3 Failed to verify: var_hasLen


muchang:
Hi,
For this formula,
```
(declare-fun b () String)
(declare-fun a () String)
(declare-fun c () String)
(assert (or (distinct (str.substr a 167 (str.len b)) "")))
(assert (let ((a!1 (str.len (str.substr (str.replace (str.replace a b "") "29M8u" "")
                                220
                                (* 92 (str.len c) 109))))
      (a!2 (= (str.at (str.replace (str.replace a b "") "29M8u" "") 141) ""))
      (a!3 (distinct (str.len (str.replace (str.replace a b "") "29M8u" ""))
                     200))
      (a!4 (to_real (str.len (str.replace (str.replace a b "") "29M8u" "")))))
  (or (distinct (ite (= a!1 121) 130 211)
                (ite a!2 168 213)
                (ite a!3 243 102)
                24)
      (< (/ (/ 29.0 a!4) 200.0) 109.0))))
(check-sat)
```
z3 smt.string_solver=z3str3 throws out a `Failed to verify: var_hasLen`

OS: Ubuntu 18.04
Commit: 0146259

mtrberzi:
Fixed in #3032

PR #3032, opened 2020-02-17
z3str3: multiple fixes

mtrberzi:
- fix an issue with `str.indexof` asserting wrong axioms for out-of-bounds cases
- fix an issue where model construction would crash when it encountered a term with no length assignment or an impossible length assignment - now we assert a length axiom instead and restart

This likely fixes multiple outstanding reported issues (may address #3020, #3005, #2996, #2991, #2982, almost certainly addresses #2951) which I'll review once this is merged.


mtrberzi:
Still unsure why this isn't building with msan - I managed to run it locally and the test passes. Any insight from the build logs?

NikolajBjorner:
msan started failing yesterday, most likely due to an update on the build VMs that causes this error, so external to Z3.

PR #3091, opened 2020-02-27
fix z3 library search order

### LINKED ISSUE #3090: libz3.dll search order
In z3core.py, the code gathers a list of search dirs (_all_dirs) for libz3.dll. However, the _default_dirs is appended to the list last. This creates a problem: there could be a different libz3.dll in other search directory, and the code will load that DLL instead of the correct one. For example, I have Triton installed on my machine and z3 loads Triton's z3 DLL. The Triron's z3 is an earlier version and ctype reports function Z3_mk_divides() is not exported by it. 

As the name suggests, _default_dirs  should be searched first. As a quick fix, add the _default_dirs to _all_dirs first before adding other search dirs: 

```
_all_dirs = []
_all_dirs.extend(_default_dirs)
```

jeffli678:
fix #3090

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1738,6 +1738,8 @@ _default_dirs = ['.',
                  os.path.join(sys.prefix, 'lib'),
                  None]
 _all_dirs = []
+# search the default dirs first
+_all_dirs.extend(_default_dirs)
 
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
Thank you

Issue #3245, opened 2020-03-12
Access defining polynomial of algebraic number via C/C++ API


ahumenberger:
Is there a way of accessing the defining components of an algebraic number via the C/C++ API?
That is, I want to get the expressions for `(+ (* 20 (^ x 2)) (* 50 x) 29)` and `1` from `(root-obj (+ (* 20 (^ x 2)) (* 50 x) 29) 1)`.

wintersteiger:
I don't really know, but doesn't `Z3_rcf_get_numerator_denominator` in `z3_rcf.h` allow you to do that?

ahumenberger:
I'm not very familiar with the types in Z3. My situation is the following: I'm using the C++ API, and there I'm getting an object of type `expr` which encapsulates an algebraic number. The `expr` is basically a `Z3_ast`, but how would I turn this into an `Z3_rcf_num`?

wintersteiger:
Good question; there doesn't seem to be an easy way. I suspect that it would work just fine if you cast the `Z3_ast` to a `Z3_rcf_num`, but if it were that easy, it probably wouldn't have been a to-do item in the Python API since 2013 [(here)](https://github.com/Z3Prover/z3/blob/d9c09dcdf94411b70bc23f817b67bf526a17724b/src/api/python/z3/z3rcf.py#L55).

NikolajBjorner:
does it work for you to have the approximations exposed (and not the root polynomials)?

NikolajBjorner:
I think it is dealt with now given the merge. Thanks

PR #3321, opened 2020-03-15
Extract defining components of algebraic number via C and C++ API

ahumenberger:
Addresses #3245


src/api/api_algebraic.cpp
```diff
@@ -418,4 +418,47 @@ extern "C" {
...
+        LOG_Z3_algebraic_get_poly_degree(c, a);
+        RESET_ERROR_CODE();
+        CHECK_IS_ALGEBRAIC(a, 0);
+        algebraic_numbers::manager & _am = am(c);
+        algebraic_numbers::anum const & av = get_irrational(c, a);
+        return _am.degree(av);
+        Z3_CATCH_RETURN(0);
+    }
+
+    int64_t Z3_API Z3_algebraic_get_poly_coeff_int64(Z3_context c, Z3_ast a, unsigned i) {
```
NikolajBjorner: what about returning an expression of type Real? 
We can create a numeric constant that contains the value, and then other parts of the API can deal with extracting either int64_t, rationals, or infinite precision rationals.
We currently don't expose a stand-alone big-num object

ahumenberger: I changed it to  `Z3_ast_vector Z3_API Z3_algebraic_get_poly(Z3_context c, Z3_ast a)` returning a vector of coefficients.

PR #3617, opened 2020-03-31
Spacer: add nonlinear counterexamples generation

leonardoalt:
Creating this PR to check first if this is wanted.

At first I wrote a separate function that generated nonlinear counterexamples, but then noticed that it is a super set, since it does generate linear counterexamples if that's the case, so I wrote it directly into `get_ground_sat_answer`.
This might makes things easier, since in this case there's no need for a separate API entry.

@agurfinkel is this wanted? Would it break things by having it directly in `get_ground_sat_answer`? I based it on the Python example you sent me once, so I think the implications and facts represent the proof correctly.
I used this code locally and it worked on my examples, but of course I'll add unit tests here if you're using that.

### DIFF
--- a/src/muz/spacer/spacer_context.cpp
+++ b/src/muz/spacer/spacer_context.cpp
@@ -55,6 +55,9 @@ Notes:
 
 #include "muz/spacer/spacer_sat_answer.h"
 
+#include <set>
+#include <stack>
+
 #define WEAKNESS_MAX 65535
 
 namespace spacer {
@@ -2950,6 +2953,14 @@ proof_ref context::get_ground_refutation() const {
     return op(*m_query);
 }
 
+/**
+    Convert a ground refutation into a linear or nonlinear counterexample.
+    The counterexample is given as a conjunction of implications of the form
+    `premises => conclusion` where `premises` is the conjunction of predicates
+    from the body of nonlinear clauses, representing the proof graph.
+    Note that if there are no nonlinear clauses, `premises` is always `true`
+    and the counterexample is linear.
+*/
 expr_ref context::get_ground_sat_answer() const {
     if (m_last_result != l_true) {
         IF_VERBOSE(0, verbose_stream()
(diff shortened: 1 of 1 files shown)


agurfinkel:
@leonardoalt , thank you for the commit. I would like to discourage the use of `ground_sat_answer`. The problem is that its result is not well defined and is easily misinterpreted. 

I prefer that proof output is used instead. It has useful additional information, and a very clear definition of what it means. The code that you are doing seems to simply unwrap the proof into a different data structure. This really belongs to the client code -- if you don't like the proof object, unwrap it at the client.

The main user of `get_ground_sat_answer` right now is `SeaHorn`. That is the only reason for not removing the method. I plan to deprecate it as soon as I have time to fix that dependency in SeaHorn. 

In short, if your commit does not change existing functionality, I'm happy to accept it. However, if you are writing code that depends on that method, note that the method is likely to be deprecated in the next few month.

leonardoalt:
Ok, understood!

The problem with unwrapping it on the client side is that the z3 data structures necessary to do that are not exported. I actually wanted to do that first, as you had already recommended, but noticed I simply couldn't, unless I'm missing something. Also, I would like to stay on the C++ API and not have to write my own Spacer wrapper, so I'm not sure what the best solution is

agurfinkel:
What API are missing? I know that I was able to construct a counterexample graph based on the proof using the Python API: 

https://spacerexamples-ariegurfinkel.notebooks.azure.com/j/notebooks/Dagstuhl2019.ipynb

leonardoalt:
The C++ API is missing `get_proof` itself for `fixedpoint`, and the definitions from `ast.h`.

agurfinkel:
Right, `get_proof` is part of the `solver` interface, which is an alternative way to get to spacer.

If you are using it through `Fixedpoint`, the proper API is `get_answer().` 
`get_answer()` returns an invariant in case fixedpoint is unsat, and a counterexample in case fixedpoint is sat. I've just changed it (commit cfe96fe92e1e6cbf265feb02580afb5f1c9d8154) to return a ground refutation in case of sat.

The refutation (i.e., proof) is just an AST. There are already API in both C and C++ layers to work with ASTs. Here is an example from my Jupyter notebook that builds a DOT graph out of a refutation. This should have everything that you need.

```python
class SpacerProof(object):
    def __init__(self, pf_ast):
        self._ast = pf_ast

    def _get_fact(self, ast):
        return ast.children()[-1]

    def _to_dot_rec(self, ast, graph, visited):
        if ast in visited:
            return

        visited.add(ast)

        dst = str(self._get_fact(ast).get_id())
        kids = ast.children()
        for k in kids[1:-1]:
            k_fact = self._get_fact(k)
            if k_fact not in visited:
                graph.node(str(k_fact.get_id()), str(k_fact))
                visited.add(k_fact)
                self._to_dot_rec(k, graph, visited)
            graph.edge(str(k_fact.get_id()), dst)

    def to_dot(self):
        import graphviz
        g = graphviz.Digraph()

        visited = set()

        fact = self._get_fact(self._ast)
        id = fact.get_id()
        g.node(str(id), str(fact))
        visited.add(id)
        self._to_dot_rec(self._ast, g, visited)
        return g

    def _repr_svg_(self):
        return self.to_dot()._repr_svg_()

    def __str__(self):
        return str(self.to_dot())
    def raw(self):
        return self._ast
```

leonardoalt:
Yep, that's exactly the one I followed when writing this code

leonardoalt:
Ok thanks for changing it to `get_ground_refutation`!

PR #3671, opened 2020-04-01
use scoped intervals

levnach:


levnach:
@NikolajBjorner , can you please review?

PR #3687, opened 2020-04-02
z3py: fix And/Or context deduction

wx-csy:
`Z3.And`, `Z3.Or` can't deduce the context when providing expressions of non-global context:
```
import z3
c = z3.Context()
a, b = z3.Bools('a b', c)
z3.And(a, b) # Z3Exception: context mismatch
```
This fixes the problem.


NikolajBjorner:
Thanks!

PR #3706, opened 2020-04-02
use scoped interval

levnach:

### DIFF
--- a/src/math/lp/nla_intervals.cpp
+++ b/src/math/lp/nla_intervals.cpp
@@ -72,11 +72,13 @@ std::ostream & intervals::print_dependencies(u_dependency* deps , std::ostream& 
 // return true iff the interval of n is does not contain 0
 bool intervals::check_nex(const nex* n, u_dependency* initial_deps) {
     m_core->lp_settings().stats().m_cross_nested_forms++;
-    auto i = interval_of_expr<e_with_deps::without_deps>(n, 1);
+    scoped_dep_interval i(get_dep_intervals());
+    interval_of_expr<e_with_deps::without_deps>(n, 1, i);
     if (!m_dep_intervals.separated_from_zero(i)) {
         return false;
     }
-    auto interv_wd = interval_of_expr<e_with_deps::with_deps>(n, 1);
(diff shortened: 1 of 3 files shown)


src/math/lp/nla_intervals.cpp
```diff
@@ -302,72 +304,68 @@ bool intervals::interval_from_term(const nex& e, interval& i) {
...
 template <e_with_deps wd>
-intervals::interval intervals::interval_of_sum_no_term(const nex_sum& e) {
-    const nex* inf_e = get_inf_interval_child(e);
-    if (inf_e) {
-        return interval();
-    }
-    interval a = interval_of_expr<wd>(e[0], 1);
+void intervals::interval_of_sum_no_term(const nex_sum& e, scoped_dep_interval & sdi) {
+    if (has_inf_interval(e))
+        return;
```
NikolajBjorner, line 309: would be good to set sdi to interval() explicitly.

src/math/lp/nla_intervals.cpp
```diff
@@ -302,72 +304,68 @@ bool intervals::interval_from_term(const nex& e, interval& i) {
...
 
-    interval a;
     m_dep_intervals.set_interval_for_scalar(a, e.coeff());
     TRACE("nla_intervals_details", tout << "a = "; display(tout, a); );
     for (const auto& ep : e) {
-        interval b = interval_of_expr<wd>(ep.e(), ep.pow());
+        scoped_dep_interval b(get_dep_intervals());
+        interval_of_expr<wd>(ep.e(), ep.pow(), b);
         TRACE("nla_intervals_details", tout << "ep = " << ep << ", "; display(tout, b); );
         interval c;
```
NikolajBjorner, line 370: make all intervals scoped, also c?

src/math/lp/nla_intervals.cpp
levnach: I can add an assert that makes sure the bounds are infinite.

src/math/lp/nla_intervals.cpp
levnach: ok

PR #3851, opened 2020-04-07
fill columns to fill in random update as in theory_arith_aux.h

levnach:

### DIFF
--- a/src/smt/theory_lra.cpp
+++ b/src/smt/theory_lra.cpp
@@ -265,8 +265,10 @@ class theory_lra::imp {
 
     svector<std::pair<theory_var, theory_var> >       m_assume_eq_candidates; 
     unsigned                                          m_assume_eq_head;
-
-    unsigned                 m_num_conflicts;
+    std::unordered_map<lp::impq, theory_var>          m_var_value_table;
+    lp::u_set                                         m_tmp_var_set;
+    
+    unsigned                                          m_num_conflicts;
 
     // non-linear arithmetic
     scoped_ptr<nra::solver>  m_nra;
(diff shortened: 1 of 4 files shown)


src/smt/theory_lra.cpp
```diff
@@ -1586,12 +1588,49 @@ class theory_lra::imp {
...
+            auto it = m_var_value_table.find(vi_val);
+            if (it == m_var_value_table.end()) {
+                m_var_value_table[vi_val] = v;
+                continue;
+            }
+            theory_var other = it->second;
+            SASSERT(other != v);
+            enode * n2 = get_enode(other);
+            if (n1->get_root() == n2->get_root())
+                continue;
```
NikolajBjorner, line 1620: this means that n1 and n2 are considered equal already globally, not only with respect to the perspective of the arithmetic solver

src/smt/theory_lra.cpp
```diff
@@ -1586,12 +1588,49 @@ class theory_lra::imp {
 
     bool assume_eqs() {        
         svector<lpvar> vars;
+        m_var_value_table.clear();
```
NikolajBjorner, line 1591: There is a table called m_model_eqs  It is intended to map impq values to variables. It leverages get_ivalue for the hash-function and value equality.

src/smt/theory_lra.cpp
```diff
@@ -1586,12 +1588,49 @@ class theory_lra::imp {
...
+            enode * n2 = get_enode(other);
+            if (n1->get_root() == n2->get_root())
+                continue;
+            if (!lp().column_is_fixed(vi)) {
                 vars.push_back(vi);
             }
+            else if (!m_tmp_var_set.contains(other)) {
+                m_tmp_var_set.insert(other);
+                lpvar other_j = lp().external_to_local(other);
+                if (lp::tv::is_term(other_j)) {
```
NikolajBjorner, line 1627: would be good with fewer round-trips with lp() in order to figure out modality a variable has.

PR #3900, opened 2020-04-10
Add ABORT and CABORT macros

agurfinkel:
Macros to gracefully abort execution on unsupported features

### DIFF
--- a/src/muz/base/rule_properties.cpp
+++ b/src/muz/base/rule_properties.cpp
@@ -68,7 +68,7 @@ void rule_properties::check_quantifier_free() {
         std::stringstream stm;
         stm << "cannot process quantifier in rule ";
         r->display(m_ctx, stm);
-        throw default_exception(stm.str());
+        ABORT(stm.str());
     }
 }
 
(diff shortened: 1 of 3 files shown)


NikolajBjorner:
not sure what problem the uses in rule_properties would address.
I thought the DEV_ASSERT macro would be something you could turn on or off depending on whether you are in dev mode or user mode. 
It is legal and should be handled for a user to feed Horn clauses outside the scope of the solver. The solver can just report that it will not handle them. In this case I use the exception mechanism.
A DEV_ASSERT would address sentinels that a developer inserts into the code to catch legal, but unexpected behavior of the code.

agurfinkel:
The intention was to catch the many uses of `throw default_exception()`. I think it would be nice to print a warning message with line numbers in that case. I can do that if you are happy with using a macro like this. The use in `rule_properties` is an example to test that the macro compiles.

NikolajBjorner:
If there is a need to trace __FILE__ __LINE__ information with default exception, then add macro to the place where default_exception is declared. Call the macro something more specific, such as 

#define throw_default_exception(_msg_) { warning_msg(std::string9__FILE__) + std::string(_msg_)); throw default_exception(_msg_); } or so

Issue #4038, opened 2020-04-21
Compile z3 static version with -fPIC


illera88:
Hi,

I'm getting the following error when using the static version of libz3 in a shared library I'm building:

```
/usr/bin/ld: /usr/local/share/vcpkg/installed/x64-linux/lib/libz3.a(memory_manager.cpp.o): relocation R_X86_64_TPOFF32 against hidden symbol `g_memory_thread_alloc_size' can not be used when making a shared object
/usr/bin/ld: /usr/local/share/vcpkg/installed/x64-linux/lib/libz3.a(api_ast.cpp.o): relocation R_X86_64_PC32 against symbol `_ZTVSt9basic_iosIcSt11char_traitsIcEE@@GLIBCXX_3.4' can not be used when making a shared object; recompile with -fPIC
```

It seems that the static version of libz3 is not being compiled using `-fPIC` producing this problem. Is there any reason why `-fPIC` is not set when building libz3 statically?

I built libz3 using `vcpkg` (it uses the default z3 build parameters).

Cheers

NikolajBjorner:
Merged.
New release: good point. Just went over a major revision so time to tone down the churn and stabilize for a release.

PR #4043, opened 2020-04-21
Use Position Independent Code flag on Static library builds

illera88:
Not compiling libz3 with `-fPIC` may lead to errors when using it in his static flavor from a shared library.

Related https://github.com/Z3Prover/z3/issues/4038


illera88:
Thanks for merging.

When are you guys thinking about releasing the next version? I'm asking because I would like to update z3 in `vcpkg`  with it.

PR #4075, opened 2020-04-24
fixes in lar_solver::maximize_term

levnach:


levnach:
The main change that maximise_term returns UNBOUNDED if the current the lar_solver cannot find a feasible solution.

PR #4092, opened 2020-04-24
z3str3: fix support for re.complement and re.intersection

mtrberzi:
Fixes a bug reported to me off-tracker where instances containing regex complement and intersection operators would crash unexpectedly when run using Z3str3. Thanks to @copumpkin for the report!

### DIFF
--- a/src/smt/theory_str_regex.cpp
+++ b/src/smt/theory_str_regex.cpp
@@ -902,7 +902,11 @@ namespace smt {
             return true;
         } else if (u.re.is_complement(re)) {
             // TODO can we do better?
-                    return false;
+            return false;
+        } else if (u.re.is_intersection(re)) {
+            return false;
+        } else if (u.re.is_complement(re)) {
+            return false;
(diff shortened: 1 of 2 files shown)


src/smt/theory_str_regex.cpp
```diff
@@ -902,7 +902,11 @@ namespace smt {
             return true;
         } else if (u.re.is_complement(re)) {
             // TODO can we do better?
-                    return false;
+            return false;
+        } else if (u.re.is_intersection(re)) {
+            return false;
+        } else if (u.re.is_complement(re)) {
```
copumpkin, line 908: I think this one is redundant given the one right above isn't it?

mtrberzi: You're right, but eventually, these will have two different heuristics for length, and I wanted to capture that by separating them out.

NikolajBjorner:
feel free to merge when it is ready on your side

Issue #4093, opened 2020-04-24
Consider using something other than __builtin_unreachable for the UNREACHABLE macro


copumpkin:
On macOS, compiling and running the following program (with clang):

```
#include <stdio.h>

int main() {
  printf("ohai!\n");
  __builtin_unreachable();
}
```

gives me this output:

```
$ ./test
ohai!
Illegal instruction: 4
```

and the same thing on Linux with `gcc` 4.4.6:

```
$ ./test
ohai!
zsh: segmentation fault  ./test
```

Since both `UNREACHABLE` and `NOT_IMPLEMENTED_YET` macros use it, it leads to less-than-ideal UX when hitting an unexpected `NOT_IMPLEMENTED_YET` case. Much more pleasant would be a call to `abort` with a message suggesting to file a bug on the tracker (maybe with macro source locations), not just because it can link to the tracker, but also because I think the user _perception_ difference between a segfault and an error message saying "sorry we haven't implemented this yet" is quite significant. I do realize that this isn't true in debug builds, but most users aren't running debug builds and will likely see what appears to be a segfault.

If folks don't object, I'd be happy to put together a PR with the change, but I wanted to solicit feedback first in case I was missing something.

NikolajBjorner:
yes, this is very reasonable. 
Currently we get undefined behavior.
It makes debugging tougher. A pull request would be great, as long as the call to abort compiles on all platforms (I guess it does as the call has been a part of the C standard library since forever)

PR #4094, opened 2020-04-25
Improve UX for unreachable/unimplemented errors

copumpkin:
This should replace several "segfaults" and "illegal instruction" errors with messages that contain a bit more context. I also put in a link to the bug tracker to make users' lives a bit easier.

For context, `__builtin_unreachable`'s behavior is undefined and is intended only as a mechanism to help the compiler see that code will not return. I do still include it in the new code because if I don't, compilation produces a lot more warnings as it can't see that `NOT_IMPLEMENTED_YET` and `UNREACHABLE` cannot return.

### DIFF
--- a/src/util/debug.cpp
+++ b/src/util/debug.cpp
@@ -23,6 +23,7 @@ Revision History:
 #include<iostream>
 #include "util/str_hashtable.h"
 #include "util/z3_exception.h"
+#include "util/z3_version.h"
 
 static volatile bool g_enable_assertions = true;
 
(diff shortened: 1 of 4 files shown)


copumpkin:
Hmm, I guess I need to check my Windows builds, but I can't click on the details link. I'll reintroduce the `_MSC_VER` check and see if that helps.

nunoplopes:
These things should never trigger. Being "nice" to users means increasing code size and reduced performance. I don't support these patches.
Z3 users are tech savvy.

copumpkin:
Hmm, help me understand please: this is adding in the order of bytes to overall code size. Context is here: https://github.com/Z3Prover/z3/issues/4093

But in short: tech savvy or not, `__builtin_unreachable` is not supposed to ever be called, and `NOT_IMPLEMENTED_YET` and `UNREACHABLE` both call it, and even tech savvy users are left wondering why pure mathematical code segfaults or causes illegal instructions. The usability (and user trust) benefits of emitting an actual error as opposed to looking like a bad memory access seem worthwhile to me, in exchange for a few extra bytes of executable size and no runtime performance impact at all.

If you still disagree, can you explain what performance aspects you’re concerned about, and why deliberately invoking undefined compiler constructs is more desirable than the alternative?

NikolajBjorner:
UNREACHABLE() is "reached" when there are bugs.
Z3 should exit gracefully then.
I have seen that the behavior when hitting __builtin_unreachable is not particularly transparent.
While we try to fix bugs when they are known, there are no claims that Z3 is bug free.

copumpkin:
Thanks @NikolajBjorner for the merge! I haven't had a chance to look into the failing checks on this PR, but will try to submit a follow-up PR later today that fixes them.

PR #4157, opened 2020-04-28
Branch perf

levnach:


levnach:
Does not repro in 97fe2c86098564d65e33821eff532ff000354dcd

PR #4205, opened 2020-05-04
Allow different parsing strategies

i-walker:
the java API only allows parsing Strings to Arrays of BoolExpr, but the ASTVector can also parse to other types like, IntExpr, RealExpr etc. 
To allow that those constructors and funtion nCtx() need to be public

### DIFF
--- a/.idea/z3.iml
+++ b/.idea/z3.iml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" inherit-compiler-output="true">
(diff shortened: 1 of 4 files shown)


i-walker:
@NikolajBjorner may you help me why the Build CI tasks all fail?

NikolajBjorner:
> @NikolajBjorner may you help me why the Build CI tasks all fail?

Your code is not type correct.
BoolExpr[] is not the same as an ASTVector. 
The previous version of the code converts the ASTVector to the type specified in the prototype.

i-walker:
Thanks @NikolajBjorner I'll add the remaining functions in the next PR :)

PR #4207, opened 2020-05-04
remaining change from #4205

i-walker:
There was no need for further functions only those modifiers

### DIFF
--- a/.idea/z3.iml
+++ b/.idea/z3.iml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" inherit-compiler-output="true">
-    <exclude-output />
-    <content url="file://$MODULE_DIR$" />
-    <orderEntry type="inheritedJdk" />
-    <orderEntry type="sourceFolder" forTests="false" />
-  </component>
-</module>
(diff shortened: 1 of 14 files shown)


i-walker:
I excuse my question, but I can't debug the CI failures, due to no access :) 
@NikolajBjorner Hope those changes are fine. 
Thanks a lot a priori.

NikolajBjorner:
The Z3Build-CI are not visible, unfortunately, because of some dependencies on signing binaries. 
The Travis builds, on the other hand are visible.
If you follow the links to the travis build you will see:

/home/user/z3_build/CMakeFiles/z3JavaJar.dir/java_compiled_z3JavaJar
/home/user/z3_src/src/api/java/Lambda.java:104: error: illegal start of expression
	public long nativeObject = Native.mkLambda(ctx.nCtx(),
	^
1 error
The command '/bin/sh -c ${Z3_SRC_DIR}/contrib/ci/scripts/build_z3_cmake.sh' returned a non-zero code: 1
/home/travis/.travis/functions: line 607:  4480 Terminated    

In other words, your code change doesn't even compile.
It helps to compile locally (in your case with the build options for Java turned on) before round-tripping to the CI.

Independently of this, I am not sure about this pull request. It exposes methods that are private
and should by the intent of @wintersteiger not be exposed. Is there a reason for exposing these methods?

PR #4215, opened 2020-05-05
z3str3: remove legacy code

mtrberzi:
This patch removes a large amount of legacy code from `theory_str`, including:

* length/value testing (replaced by fixed length model construction)
* binary search heuristic for length/value testing
* regex unroll (replaced by automata-based regex reasoning)
* fixed-length overlap testing (this was never even implemented completely)

This code has been superseded by newer modules with better performance and is prone to errors. It also won't be maintained going forward. All of the associated code has been disabled by default for a very long time and enabling it isn't the recommended way to invoke Z3str3.


mtrberzi:
As a result of this patch, the following parameters are deprecated:

* `smt.str.use_binary_search` and `smt.str.binary_search_start` (the associated methods have been removed)
* `smt.str.finite_overlap_models` (the associated methods have been removed)
* `smt.str.regex_automata` (automata-based regex reasoning is now always used)
* `smt.str.fixed_length_models` (fixed-length model construction is now always used)

PR #4247, opened 2020-05-07
fix #4054

### LINKED ISSUE #4054: Spacer hangs when unused (unconstrained) var is added
Given these two queries:

Any idea why that happens?

hgvk94:


agurfinkel:
something is wrong with your history. It includes many commits that are probably not intended to be here.

PR #4248, opened 2020-05-07
fix #4054

### LINKED ISSUE #4054: Spacer hangs when unused (unconstrained) var is added

hgvk94:


src/muz/spacer/spacer_context.cpp
```diff
@@ -3375,6 +3465,8 @@ lbool context::expand_pob(pob& n, pob_ref_buffer &out)
     lbool res = n.pt ().is_reachable (n, &cube, &model, uses_level, is_concrete, r,
                                       reach_pred_used, num_reuse_reach);
     if (model) model->set_model_completion(false);
+    if(res == l_undef) res = handle_unknown(n, r, model);
```
NikolajBjorner, line 3468: would be super good if code looks the same: space after "if" "while" "for".

hgvk94: Will make this and other changes and resubmit

PR #4277, opened 2020-05-11
fix #4054

### LINKED ISSUE #4054: Spacer hangs when unused (unconstrained) var is added

hgvk94:


src/muz/spacer/spacer_util.cpp
```diff
@@ -980,7 +980,13 @@ namespace {
         for_each_expr(cd, fml);
     }
 
-}
-
+    // set the value of a boolean function to true in model
+    void set_true_in_mdl(model *model, func_decl *f) {
```
NikolajBjorner, line 984: pass as model& model instead of model* model

src/muz/spacer/spacer_context.cpp
```diff
@@ -3321,6 +3347,96 @@ void context::predecessor_eh()
...
+  }
+  return true;
+}
+
+// Handle cases where solver returns unknown but returns a good enough model
+// model is good enough if it satisfies
+// 1. all the reachable states whose tag is set in the model
+// 2. Tr && pob
+lbool context::handle_unknown(pob &n, const datalog::rule *r, model_ref model) {
+  if (r == nullptr) {
```
NikolajBjorner, line 3426: 4 spaces
https://www.bing.com/videos/search?q=tabs+vs+spaces&&view=detail&mid=D8125B650297896DCD76D8125B650297896DCD76&&FORM=VRDGAR

hgvk94: But .... I used clang format.

NikolajBjorner: it is configurable

NikolajBjorner:
@agurfinkel - are you handling the merge status of this pull request?

agurfinkel:
@NikolajBjorner yes. waiting for spaces to normalize and builds to pass

agurfinkel:
@hgvk94 the build issue seems to be related to something outside of this PR. I can reproduce it (in release build only) on your PR, but not any longer on current `master`. 

Please rebase against master and re-submit to trigger CI one more time.

agurfinkel:
@NikolajBjorner is there some known instability in the build right now? This PR keeps failing regressions, but, as far as I can tell, no line of code from the PR is actually executed in the failing test.

Previously, failure was on a mac in seq theory initialization. I could reproduce it locally. Now a string benchmark in timing out on some linux build. I can dig deeper, but I wanted to make sure that there isn't a known anomaly.

NikolajBjorner:
string-ops.smt2 started timing out yesterday. Seems a butterfly effect of slow string reasoning

agurfinkel:
then this PR is ready to go. We can wait until the CI is more stable and re-submit. Let me know what you want to do.

PR #4278, opened 2020-05-11
Nra to nla

levnach:


levnach:
The call to nlsat from nla_solver is enabled. It helps to pass some regression tests.

Issue #4183, opened 2020-05-02
Difficulties building Z3 with Julia binding


JeffreySarnoff:
Members of the Julia Community are quite excited to work with Z3 in Julia, and develop uses that benefit from Julia and its ecosystem. We are having difficulty building Z3 with the Julia interface ([see here](https://discourse.julialang.org/t/julia-bindings-for-z3-smt-solver/) and [we know of this closed issue](https://github.com/Z3Prover/z3/issues/4035)).   It would be a great help to us if you were to add more detail (explicit step-by-step instructions known to work, and detailed expectations about environment and location of whatever files are sought).  Our large binary resource maintainers have asked me to request a new release of Z3 when the Julia API is more accessible, and directions are clearer.

Thank you for the consideration.

NikolajBjorner:
we currently build on every commit for the MacOS. The build script is here:

https://github.com/Z3Prover/z3/blob/6088da5159959c099c7b295015cef8aa4887638a/azure-pipelines.yml#L183


@ahumenberger prepared the script and also created the bindings. He uses a MacOS, so it was easier to create the script for this environment. I am very happy to include CI builds on other platforms, but similar to you, figuring out the magic dependencies is going to take some arm-twisting. If you (or others) can see what should be done for other platforms, say Ubuntu, based on the MacOS script a pull request with an update to the CI pipeline is very welcome. In this way we can build on every push. 
If there are other issues, that are orthogonal to this, that hold you back, feel welcome to add context on what will help you.

JeffreySarnoff:
That it is available for MacOS is good news.  We need to support it on vanilla Linux and Windows for Z3 to become part of Julia designer's capabilities.  I will see what others may decipher for those platforms and let you know. Thanks.

ahumenberger:
The Julia bindings sit on top of the C++ API and use [CxxWrap.jl](https://github.com/JuliaInterop/CxxWrap.jl). I'm basically waiting for a new Z3 release which then includes the Julia bindings, and then I would update [z3_jll](https://github.com/JuliaBinaryWrappers/z3_jll.jl) such that also the Julia bindings are built. The Julia package [Z3.jl](https://github.com/ahumenberger/Z3.jl) will then use z3_jll.

JeffreySarnoff:
more detailed info from @ahumenberger [partially paraphrased](https://discourse.julialang.org/t/julia-bindings-for-z3-smt-solver/38218/15?u=jeffreysarnoff)
[...]

PR #4298, opened 2020-05-12
Update README with infos about Julia bindings

ahumenberger:
#4183


XVilka:
I would have put Julia under the Python, since semantically closer to it, rather than to the C or OCaml.

NikolajBjorner:
correct, it is right after Python.

PR #4333, opened 2020-05-16
Unicode escape sequences in string literals

mtrberzi:
This introduces Unicode codepoint escape sequences of the form `\u####` and `\u{#....}` according to the SMT-LIB 2.6 standard for Unicode strings. Currently the scanner rejects codepoints greater than 255, since we expect to be working with ASCII for now, but the code paths for full support are mostly in place.

I also added some notes in comments where I observed some further changes will need to be made in the scanner and parser for future Unicode support. These are marked as `// UNICODE`.

### DIFF
--- a/src/parsers/smt2/smt2parser.cpp
+++ b/src/parsers/smt2/smt2parser.cpp
@@ -1189,7 +1189,7 @@ namespace smt2 {
 
         void parse_string_const() {
             SASSERT(curr() == scanner::STRING_TOKEN);
-            expr_stack().push_back(sutil().str.mk_string(symbol(m_scanner.get_string())));
+            expr_stack().push_back(sutil().str.mk_string(symbol(m_scanner.get_string()))); // TODO mk_string directly from a Unicode zstring
             TRACE("smt2parser", tout << "new string: " << mk_pp(expr_stack().back(), m()) << "\n";);
             next();
         }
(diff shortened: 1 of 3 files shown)


mtrberzi:
I'll update this to handle printing escape characters as well, as I realized we definitely need that.

NikolajBjorner:
I am not sure that the scanner is the place to make this change.
The utilities for zstring in seq_decl_plugin 
already have methods for converting escape sequences into characters.
These could be extended for unicode. There is a unicode mode (not yet used) for zstring.

NikolajBjorner:
The function 
https://github.com/Z3Prover/z3/blob/937afbf95b181d8ae886ca06e268abe8c56d8a61/src/ast/seq_decl_plugin.cpp#L50
converts escape characters into a character. Currently set to fit within "unsigned".
The calling code masks down the result to just 8 bits, but this would be changed.
Notice the TBD:
https://github.com/Z3Prover/z3/blob/937afbf95b181d8ae886ca06e268abe8c56d8a61/src/ast/seq_decl_plugin.cpp#L130

NikolajBjorner:
Something along the lines of

```
    if (*(s+1) == 'u' && *(s+2) == '{') {
        result = 0;
        for (unsigned i = 0; i < 5; ++i) {
            if (is_hex_digit(*(s+3+i), d1)) {
                result = 64*result + d1;
                continue;
            }
            if (*(s+3+i) == '}') {
                s += 4 + i;
                return true;
            }
            return false;
        }
    }
    if (*(s+1) == 'u' && is_hex_digit(*(s+2), d1)) {
        result = d1;
        for (unsigned i = 0; i < 3; ++i) {
            if (is_hex_digit(*(s+3+i), d1)) {
                result = 64*result + d1;
            }
            return true;
        }
    }

```

NikolajBjorner:
I have added an outline of the escape sequence handling in the function mentioned above.

PR #4361, opened 2020-05-17
z3str3: use larger unsigned integer type for length comparison in itos

### LINKED ISSUE #4346: Assertion violation at rational.h:113 (z3str3)
~~~~
(set-option :smt.string_solver z3str3)
(declare-const i14 Int)
(declare-const Str8 String)
(assert (>= (str.len Str8) 0))
(assert (>= (str.len (str.++ "" (int.to.str i14))) (* 685 (+ 0 0 908 89 970) 14 10 292)))
(check-sat)
~~~~

z3 (commit b2bfb1f)
~~~~
ASSERTION VIOLATION
../src/util/rational.h
Line: 113
is_unsigned()
(C)ontinue, (A)bort, (S)top, (T)hrow exception, Invoke (G)DB
~~~~

mtrberzi:
Fixes #4346

### DIFF
--- a/src/smt/theory_str_mc.cpp
+++ b/src/smt/theory_str_mc.cpp
@@ -728,7 +728,7 @@ namespace smt {
                 eqc_chars.reset();
                 return true;
             } else {
-                if (termLen.get_uint64() != iValue.to_string().length()) {
+                if (termLen != rational((unsigned)iValue.to_string().length())) {
                     // conflict
                     cex = expr_ref(m.mk_not(m.mk_and(get_context().mk_eq_atom(mk_strlen(term), mk_int(termLen)), get_context().mk_eq_atom(arg0, mk_int(iValue)))), m);
                     return false;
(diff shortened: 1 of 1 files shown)


src/smt/theory_str_mc.cpp
```diff
@@ -728,7 +728,7 @@ namespace smt {
                 eqc_chars.reset();
                 return true;
             } else {
-                if (termLen.get_unsigned() != iValue.to_string().length()) {
+                if (termLen.get_uint64() != iValue.to_string().length()) {
```
NikolajBjorner, line 731: I am almost sure, this still will not fix.
The value of "termLen" is decided by the arithmetic solver.
It can be any integer (>= because you add a lower bound, I hope).
So it could get beyond uint64.
The test will for sure be false even then because iValue doesn't fit in reasonable memory. So you want to augment the condition by a check that termLen is an integer (uint64, or even unsigned if you can say more about iValue).

mtrberzi: I've updated the line you suggested, comparing `rational` to `rational` should work?

PR #4382, opened 2020-05-18
SINGLE_THREAD: do not use pthread if possible

SweetVishnya:
I placed parallel stuff under `ifdef` where possible.

### DIFF
--- a/src/api/api_log.cpp
+++ b/src/api/api_log.cpp
@@ -16,18 +16,18 @@ Revision History:
 
 --*/
 #include<fstream>
-#include<mutex>
 #include "api/z3.h"
 #include "api/api_log_macros.h"
 #include "util/util.h"
 #include "util/z3_version.h"
+#include "util/mutex.h"
 
 std::ostream * g_z3_log = nullptr;
 std::atomic<bool> g_z3_log_enabled;
 
 #ifdef Z3_LOG_SYNC
-static std::mutex g_log_mux;
-#define SCOPED_LOCK() std::lock_guard<std::mutex> lock(g_log_mux)
+static mutex g_log_mux;
+#define SCOPED_LOCK() lock_guard lock(g_log_mux)
 #else
 #define SCOPED_LOCK() {}
 #endif
--- a/src/sat/sat_parallel.cpp
+++ b/src/sat/sat_parallel.cpp
@@ -129,7 +129,7 @@ namespace sat {
         if (s.get_config().m_num_threads == 1 || s.m_par_syncing_clauses) return;
         flet<bool> _disable_sync_clause(s.m_par_syncing_clauses, true);
         {
-            std::lock_guard<std::mutex> lock(m_mux);
+            lock_guard lock(m_mux);
             if (limit < m_units.size()) {
                 // this might repeat some literals.
                 out.append(m_units.size() - limit, m_units.c_ptr() + limit);
@@ -150,7 +150,7 @@ namespace sat {
         flet<bool> _disable_sync_clause(s.m_par_syncing_clauses, true);
         IF_VERBOSE(3, verbose_stream() << s.m_par_id << ": share " <<  l1 << " " << l2 << "\n";);
         {
-            std::lock_guard<std::mutex> lock(m_mux);
+            lock_guard lock(m_mux);
             m_pool.begin_add_vector(s.m_par_id, 2);
             m_pool.add_vector_elem(l1.index());
             m_pool.add_vector_elem(l2.index());            
@@ -164,7 +164,7 @@ namespace sat {
         unsigned n = c.size();
         unsigned owner = s.m_par_id;
         IF_VERBOSE(3, verbose_stream() << owner << ": share " <<  c << "\n";);
-        std::lock_guard<std::mutex> lock(m_mux);
+        lock_guard lock(m_mux);
         m_pool.begin_add_vector(owner, n);                
         for (unsigned i = 0; i < n; ++i) {
             m_pool.add_vector_elem(c[i].index());
@@ -175,7 +175,7 @@ namespace sat {
     void parallel::get_clauses(solver& s) {
         if (s.m_par_syncing_clauses) return;
         flet<bool> _disable_sync_clause(s.m_par_syncing_clauses, true);
-        std::lock_guard<std::mutex> lock(m_mux);
+        lock_guard lock(m_mux);
         _get_clauses(s);        
     }
 
@@ -227,12 +227,12 @@ namespace sat {
     }
 
     void parallel::from_solver(solver& s) {
-        std::lock_guard<std::mutex> lock(m_mux);
+        lock_guard lock(m_mux);
         _from_solver(s);        
     }
 
     bool parallel::to_solver(solver& s) {
-        std::lock_guard<std::mutex> lock(m_mux);
+        lock_guard lock(m_mux);
         return _to_solver(s);
     }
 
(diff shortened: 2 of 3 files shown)


NikolajBjorner:
thanks for the effort. One thing: I would love to avoid if-defs.
I should say I just added one temporary if-def yesterday, but it will go away after the feature is ready.
The motivation for the single threaded mode has been to avoid lock contention when there are no threads running, which is something the feature using the mode can ensure. Removing modules at compile time gives some static guarantees, but clutters a lot. Is having a runtime check workable?

cmake/z3_add_component.cmake
```diff
@@ -100,6 +100,11 @@ macro(z3_add_component component_name)
   # Note: We don't check the sources exist here because
   # they might be generated files that don't exist yet.
 
+  if(Z3_SINGLE_THREADED)
```
NikolajBjorner, line 103: I think it will be best to push this to the files themselves and outside of the build system

src/api/api_tactic.cpp
```diff
@@ -129,22 +129,32 @@ extern "C" {
     Z3_tactic Z3_API Z3_tactic_par_or(Z3_context c, unsigned num, Z3_tactic const ts[]) {
         Z3_TRY;
         LOG_Z3_tactic_par_or(c, num, ts);
+#ifdef SINGLE_THREAD
```
NikolajBjorner, line 132: par should return null or throw an exception

src/cmd_context/tactic_cmds.cpp
```diff
@@ -675,12 +679,14 @@ tactic * sexpr2tactic(cmd_context & ctx, sexpr * n) {
             return mk_and_then(ctx, n);
         else if (cmd_name == "or-else")
             return mk_or_else(ctx, n);
+#ifndef SINGLE_THREAD
```
NikolajBjorner, line 682: too many changes.
Let the tactic creation throw. You will not be using them in single threaded build anyhow.

nunoplopes:
Note that you can't completely get rid of pthread because we currently need it for the timeout handling.

NikolajBjorner:
To give an idea of what I believe is a better approach I have updated tactical.cpp to disable and not compile thread dependent functionality with SINGLE_THREAD

See d603bd7e3b08f7bdf5019017a6c2dcb025805fb5

So you will still see the options in the documentation, but get a runtime exception.
I prefer this to interleaving the ifdef in many - many places.

The remaining main modules:
- sat solver parallel, 
- smt solver parallel
- parallel-tactic

But of course, if timeouts don't work for you, then it isn't practical to complete this overhaul.
(though, I would be surprised as you should be able to use interrupts on the platforms where you want to use this).

NikolajBjorner:
Note that the parallel tactics and other parallel search features are unsafe and should not be used in SINGLE_THREAD compilation anyhow because some of the features that ensure thread safety are turned off (such as the symbol table).

nunoplopes:
@NikolajBjorner I use SINGLE_THREAD on all my builds for performance. I'm still happy to use pthreads for the timeout functionality, though.

NikolajBjorner:
right, timout does not depend on the symbol table or mpz, but the smt solver uses mpz, so these parallel tactics should never be used in single threaded builds.

src/shell/lp_frontend.cpp
```diff
@@ -16,14 +16,21 @@ Copyright (c) 2016 Microsoft Corporation
 #include "util/gparams.h"
 #include <signal.h>
 #include "smt/params/smt_params_helper.hpp"
+#ifndef SINGLE_THREAD
```
NikolajBjorner, line 19: this is the same pattern in the other frontend files. 
It will be better to coalesce the differences.

NikolajBjorner:
I have now removed thread dependencies from some of the main places.
Could you update the PR based on these?
There are many files where the #if-defs can go away since I will allow exposing these tactics/settings but they will just fail. SINGLE_THREAD mode is not an important scenario for mainstream users, so the few users that insist on compiling with SINGLE_THREAD will have to experience par tactics to just give up. I think this is very acceptable compared to sprinkling a lot of conditional compilation in many places.

nunoplopes:
Agreed, par tactics should just error.

BTW, there's this file to help reduce #ifdefs: https://github.com/Z3Prover/z3/blob/master/src/util/mutex.h

NikolajBjorner:
BTW, it should be good to establish clarity that there are two requirements:

- NO_PTHEADS - Z3 compiles without any reference to pthreads.
- SINGLE_THREAD - Z3 compiles with pthreads, but does not use threads during solving.

NO_PTHREADS implies SINGLE_THREAD, but not the other way.

I haven't looked into the reasons why NO_PTHREADS is really required, but I take @SweetVishnya knows the details well enough.

z3 should in every instance compartmentalize its dependencies on pthreads. So if supporting NO_PTHREADS requires some code refactoring it could serve into containing the dependencies.
The use of "util/mutex.h" is one of the possible tools.

NikolajBjorner:
FWIW, scoped_timer was an ugly duckling before Nuno took over

https://githistory.xyz/Z3Prover/z3/blob/master/src/util/scoped_timer.cpp

SweetVishnya:
> BTW, it should be good to establish clarity that there are two requirements:
> 
>     * NO_PTHREADS - Z3 compiles without any reference to pthreads.
> 
>     * SINGLE_THREAD - Z3 compiles with pthreads, but does not use threads during solving.

I agree that we should distinguish these requirements.

> I haven't looked into the reasons why NO_PTHREADS is really required, but I take @SweetVishnya knows the details well enough.

`NO_PTHREADS` requirement came out of [this](https://groups.google.com/d/msg/dynamorio-users/Xku7HJQd5f4/kpDZW3-7AAAJ) message from DynamoRIO users. We use Z3 inside a DynamoRIO client. But DynamoRIO does not support clients linking pthread due to its private loader and transparency. DynamoRIO uses separate libraries copies for a client. Practically, this causes a crash in pthread on old Ubuntu 10.04-16.04. Maybe @derekbruening can shed some light on it.

So, I am trying to remove all pthread dependencies from my DynamoRIO client including Z3.

> NO_PTHREADS implies SINGLE_THREAD, but not the other way.

I got some further thoughts. Consider the following. I built a Triton path predicate. Then I spawn multiple DynamoRIO client threads. In each thread I try to invert a branch:

1. Convert Triton representation to Z3 ASTs.
2. Ask Z3 for a model (non-parallel solving).

Is it safe to build independent Z3 ASTs from parallel threads and ask model for them?

nunoplopes:
> Is it safe to build independent Z3 ASTs from parallel threads and ask model for them?

You can't create ASTs in parallel, otherwise you've race conditions all over (reference counting, symbol table, etc).
I don't think it even works if you have different Z3 contexts, since it only solves part of the problem (ref counting), but the symbol table is still shared AFAIR.

SweetVishnya:
Found a [typo](https://github.com/Z3Prover/z3/commit/ff10afd2265ab87e5677287691e75711a9883eb5#r39277717)

src/cmd_context/tactic_cmds.cpp
SweetVishnya: I can create a special `util/thread.h` header with `ifndef` in it.

src/shell/lp_frontend.cpp
SweetVishnya: I can create a special `util/thread.h` header with `ifndef` in it.

SweetVishnya:
I will still try to finalize this pull request today. I leave timeouts (using pthreads) as is. At least, I will minimize pthread usage.

SweetVishnya:
Rebased and ready to merge. Still have no idea what to do with timeouts. Is there a Windows analog for [alarm](https://linux.die.net/man/2/alarm)?

NikolajBjorner:
win32 api has timers and timer queues. For example:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-settimer

SweetVishnya:
Can we merge this now? I will think about timers later.

NikolajBjorner:
The change to the 3 files are fine. Though, this doesn't remove dependency of pthreads so I doubt it is enough for you unless you tweak your build locally. The more sustainable way is to ensure the option of removing pthreads is part of the master branch (and perhaps even there is a build mode for it, though we are overflowing a bit on build modes now).

PR #4386, opened 2020-05-19
in progress

NikolajBjorner:


calebstanford-msr:
Added some commits to progress on this. str.in_re rewriter is not fully working, there appears to be either an infinite loop or a severe performance issue. I have some regression tests that we could add to z3test as well.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -657,15 +657,6 @@ br_status seq_rewriter::mk_seq_concat(expr* a, expr* b, expr_ref& result) {
         result = a;
         return BR_DONE;
     }
-    // TBD concatenation is right-associative
-    if (isc2 && m_util.str.is_concat(a, c, d) && m_util.str.is_string(d, s1)) {
-        result = m_util.str.mk_concat(c, m_util.str.mk_string(s1 + s2));
-        return BR_DONE;
-    }
-    if (isc1 && m_util.str.is_concat(b, c, d) && m_util.str.is_string(c, s2)) {
```
NikolajBjorner, line 665: this one is still a reachable case: "a" + ("b" + x) -> "ab" + x

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -1963,6 +1954,184 @@ bool seq_rewriter::is_sequence(expr* e, expr_ref_vector& seq) {
     return true;
 }
 
+br_status seq_rewriter::mk_regexp_contains_emptystr(expr* b, expr_ref& result) {
```
NikolajBjorner, line 1957: is_nullable takes care of this

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -1963,6 +1954,184 @@ bool seq_rewriter::is_sequence(expr* e, expr_ref_vector& seq) {
...
+    else if (m_util.re.is_empty(b)) {
+        result = m().mk_false();
+        return BR_DONE;
+    }
+    else {
+        UNREACHABLE();
+    }
+}
+
+br_status seq_rewriter::eval_regexp_derivative(
```
NikolajBjorner, line 2032: is this dead code now?

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2042,7 +2211,149 @@ expr_ref seq_rewriter::is_nullable(expr* r) {
     return result;
 }
 
+br_status seq_rewriter::derivative(expr* hd, expr* r, expr_ref& result) {
```
NikolajBjorner, line 2214: perhaps to avoid some clutter revert to signature

expr_ref seq_rewriter:;derivative(expr* elem, expr* r)

Have it reurn expr_ref(m()) (which is a null expression) on failure to rewrite.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2056,6 +2367,22 @@ br_status seq_rewriter::mk_str_in_regexp(expr* a, expr* b, expr_ref& result) {
         result = m().mk_eq(a, b1);
         return BR_REWRITE1;
     }
+    if (m_util.str.is_empty(a)) {
+        result = is_nullable(b);
+        return BR_REWRITE_FULL; // is_nullable doesn't rewrite
```
NikolajBjorner, line 2372: should return BR_DONE.
If is_nullable returns (str.in_re s r), it could not rewrite it any more. Returning anything but BR_DONE or BR_FAILED would reinvoke rewriting leading to a loop

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2056,6 +2367,22 @@ br_status seq_rewriter::mk_str_in_regexp(expr* a, expr* b, expr_ref& result) {
...
+    }
+
+    expr_ref hd(m());
+    expr_ref tl(m());
+    expr_ref db(m());
+    if (get_head_tail(a, hd, tl) &&
+        derivative(hd, b, db) == BR_DONE) {
+        result = m_util.re.mk_in_re(tl, db);
+        return BR_REWRITE1;
```
NikolajBjorner, line 2381: perhaps here it would be relevant to rewrite under nested terms

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2042,7 +2211,149 @@ expr_ref seq_rewriter::is_nullable(expr* r) {
...
+            result = m_util.re.mk_concat(dr1, r);
+            return BR_DONE;
+        } else {
+            return BR_FAILED;
+        }
+    }
+    else if (m_util.re.is_plus(r, r1)) {
+        return derivative(hd, m_util.re.mk_star(r1), result);
+    }
+    else if (m_util.re.is_union(r, r1, r2)) {
```
NikolajBjorner, line 2265: essentially, shortcutting on union and intersection is possible. Suppose the full set is returned by one of the branches, then the result is the full set. For intersection, if one of the branches is empty, the result can be empty. The return value of the other branch is not relevant. 

expr_ref dr1 = derivative(hd, r1);
if (dr1 && m_util.re.is_full_re(dr1))
      return dr1;
expr_ref dr1 = derivative(hd, r1);
if (dr2 && m_util.re.is_full_re(dr2))
      return dr2;
if (!dr1 || !dr2)
       return expr_ref(m());
return expr_ref(m_util.re.mk_union(dr1, dr2), m());

src/ast/rewriter/seq_rewriter.cpp
calebstanford-msr: That makes sense, I was wondering if there was a convention for returning basically 'option expr'

src/ast/rewriter/seq_rewriter.cpp
calebstanford-msr: Ah you are right, only the `is_concat(a, c, d)` case is unreachable. Fixed

src/ast/rewriter/seq_rewriter.cpp
calebstanford-msr: Yes, I kept it around in case we want it later, but now that you have created the PR I will remove it.

src/ast/rewriter/seq_rewriter.cpp
calebstanford-msr: Yes. Removing

src/ast/rewriter/seq_rewriter.cpp
calebstanford-msr: Yes, there was a loop here. But BR_DONE assumes is_nullable rewrites fully. What about the following?
    if (m_util.str.is_empty(a)) {
        result = is_nullable(b);
        // is_nullable doesn't rewrite fully. But we also want to avoid the
        // case where it didn't succeed in changing anything.
        if (m_util.str.is_in_re(result)) {
            return BR_FAILED;
        }
        else {
            return BR_REWRITE_FULL;
        }
    }

NikolajBjorner: sure

calebstanford-msr:
I have added commits which address most of the comments, but left three remaining TODOs in seq_rewriter.cpp: these should be one-line easy fixes.

A couple regression tests are failing due to the regex derivative not handling the character range case. Here's the output of `./z3 ../../z3test/regressions/smt2/re.smt2 | diff - ../../z3test/regressions/smt2/re`:

```
< (str.in_re "a" (re.comp (re.range "b" "d")))
< (not (str.in_re "c" (re.comp (re.range "b" "d"))))
---
> true
> true
```

This should be fixed by one of the one-line TODOs (implementing range check in the derivatives)

I also have a bunch of additional regression tests which can be found here:
https://github.com/calebstanford-msr/z3test/blob/b925356d7ad83c4ab040776959708cc4cc7645d1/regressions/smt2/re2.smt2
current output: https://github.com/calebstanford-msr/z3test/blob/b925356d7ad83c4ab040776959708cc4cc7645d1/regressions/smt2/re2.expected.out

Several of those involve variables denoting regular expressions, which the derivative code doesn't handle right now, and is why those are not simplifying.

src/ast/rewriter/seq_rewriter.cpp
calebstanford-msr: I didn't address this comment yet, because I think returning BR_REWRITE_FULL after each iteration of the derivative should result in simplifying such cases. This is line 2251. Should we also do the simplification in the union and intersections directly?

PR #4440, opened 2020-05-22
adding regex simplification rewriter

veanes:
Rewriter for regex constraints over the same variable.


veanes:
Regex constraint simplifier for constraints over the same variable.

PR #4501, opened 2020-06-05
BDD-based normal form for regex derivatives -- incorporate changes and boolean flags

calebstanford-msr:


src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2562,6 +2542,233 @@ expr_ref seq_rewriter::mk_derivative_rec(expr* ele, expr* r) {
...
+expr_ref seq_rewriter::combine_ites(decl_kind k, expr* a, expr* b, expr* cond) {
+    expr_ref result(m_op_cache.find(k, a, b, cond), m());
+    if (result) {
+        return result;
+    }
+    SASSERT((k == OP_ITE) == (cond != nullptr));
+    expr *acond = nullptr, *a1 = nullptr, *a2 = nullptr,
+         *bcond = nullptr, *b1 = nullptr, *b2 = nullptr;
+    expr_ref result1(m()), result2(m());
+    if (k == OP_ITE) {
```
NikolajBjorner, line 2577: Note, the BDD case for building ites has to consider comparing all three arguments, e.g,, there are cases for "cond" being an ite and having level of condition compared to level of conditions in a, b.
For example, 
https://github.com/Z3Prover/z3/blob/5f9973d8c4b1cbd86dadb58da8425d5e2acfc1a2/src/math/dd/dd_bdd.cpp#L590

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2788,6 +2995,19 @@ br_status seq_rewriter::mk_str_to_regexp(expr* a, expr_ref& result) {
     return BR_FAILED;
 }
 
+/*
```
NikolajBjorner, line 2998: yes, such comments on the functionality are really nice and missing in too many instances.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3218,6 +3470,12 @@ br_status seq_rewriter::mk_re_star(expr* a, expr_ref& result) {
         result = re().mk_star(re().mk_union(b1, c1));
         return BR_REWRITE2;
     }
+    expr *a1 = nullptr, *a2 = nullptr, *cond = nullptr;
```
NikolajBjorner, line 3473: The intent with adding the ite-lifting in mk_app_core is to factor out the special cases from each operator.

src/ast/rewriter/seq_rewriter.cpp
calebstanford-msr: Interesting, thanks. I did not know that

src/ast/rewriter/seq_rewriter.cpp
calebstanford-msr: Yes indeed, I only left these special cases in unary operations, as that avoids the term blowup which you pointed out. These were no longer necessary for the ITE lifting, I just left them in as useful rewrites more generally.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2450,40 +2484,91 @@ expr_ref seq_rewriter::mk_derivative_rec(expr* ele, expr* r) {
...
+        expr_ref is_n = is_nullable(r2);
+        expr_ref dr2 = mk_derivative(ele, r2, left);
+        result = mk_der_concat(r1, dr2);
+        if (m().is_false(is_n)) {
+            return result;
+        }
+        expr_ref dr1 = mk_derivative(ele, r1, left);
+        is_n = re_predicate(is_n, seq_sort);
+        if (lift_over_union) {
+            return mk_der_union(result, mk_der_concat(dr1, is_n));
```
NikolajBjorner, line 2504: dr2?

calebstanford-msr: This is the right-hand case so it should be dr1, but the error is actually in line 2507 (that should be dr1)

calebstanford-msr:
As discussed today I am cancelling this PR for now to investigate the different more self-contained changes.

PR #4504, opened 2020-06-06
Clean up intrinsics handling in mpz.cpp

FabianWolff:
The way conditional compilation for intrinsics support is implemented in `src/util/mpz.cpp` is somewhat confusing, because it's split across two `#if` blocks; this PR merges them into one and adds support for clang (which also [supports](https://github.com/llvm-mirror/clang/blob/aa231e4be75ac4759c236b755c57876f76e3cf05/include/clang/Basic/Builtins.def#L427) these intrinsics by now).

Since this PR mostly only reorders a few lines, it is not a copyrightable contribution, and thus I won't deem it necessary to sign your CLA.

The intrinsics implementation in `src/util/hwf.cpp` also looks fishy, because in line 34, you have
```C
#if defined(__x86_64__) || defined(_M_X64) ||    \
    defined(__i386) || defined(_M_IX86)
#define USE_INTRINSICS
#endif
```
and then, in line 51:
```C
#if defined(__clang__) || defined(_M_ARM) && defined(_M_ARM64)
#undef USE_INTRINSICS
#endif
```
But if we're on ARM, `USE_INTRINSICS` would not have been defined in the first place (lines 34-37), right? Also, I'm pretty sure clang [supports](https://clang.llvm.org/doxygen/emmintrin_8h.html) the intrinsics in question by now (but you may want to check this for yourselves by removing lines 51-53 and then compiling with clang).

### DIFF
--- a/src/util/mpz.cpp
+++ b/src/util/mpz.cpp
@@ -49,17 +49,21 @@ Revision History:
 
 
 #if defined(_WINDOWS) && !defined(_M_ARM) && !defined(_M_ARM64)
-// This is needed for _tzcnt_u32 and friends.
-#include <immintrin.h>
-#define _trailing_zeros32(X) _tzcnt_u32(X)
-#endif
-
-#if defined(__GNUC__)
-#define _trailing_zeros32(X) __builtin_ctz(X)
+ // This is needed for _tzcnt_u32 and friends.
+ #include <immintrin.h>
+ #define _trailing_zeros32(X) _tzcnt_u32(X)
+#elif defined(__GNUC__) || defined(__clang__)
+ #define _trailing_zeros32(X) __builtin_ctz(X)
+#else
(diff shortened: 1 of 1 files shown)


andrewvaughanj:
I think the licensing thing could be slightly more complicated than you not deeming it copyrightable.

There's a chance MSR/MSFT need it signed such that they can _relicense_ Z3 under terms they see fit including your contribution. Signing the CLA means you give MSFT the right to do this. If they do not have the agreement of all contributors, then this impedes their ability to relicense.

That being said, IANAL or any kind of maintainer/gatekeeper of Z3.

NikolajBjorner:
I understand it seems superfluous to sign. Though, the signing request is inserted automatically by an auditing tool and related auditing tools are used when making a call whether to release products that use this code. It will then fail such audits for no really good reason and then incur overhead in determining compliance.

PR #4514, opened 2020-06-12
Add sat option for a global simplification before the burst search and the main search loop

rainoftime:
- Control if doing a global simplification before the burst search and the main CDCL search loop.
- The option is useful (perhaps "necessary") for the `sat-preprocess` tactic in sat_tactic.cpp

The default value is `false`.


rainoftime:
The description of the `sat-preprocess` tactic in sat_tactic.h is as follows

> Apply SAT solver preprocessing procedures (bounded resolution, Boolean constant propagation, 2-SAT, subsumption, subsumption resolution

- Not sure if the tactic means using the `do_simplify` function, which contains many simplifications. The user can customize their `sat-preprocess` tactic with other options.
- Not sure if the tactic needs the burst search (default value 100)

PR #4516, opened 2020-06-12
Cheap eqs

levnach:


src/math/lp/bound_analyzer_on_row.h
```diff
@@ -13,9 +13,8 @@ Module Name:
     In the same loop trying to pin variables by pushing the partial sum up, denoting the variable related to it by _l
 
 Author:
-
-    Lev Nachmanson (levnach)
-
+    Lev Nachmanson  (levnach)
+    Nikolaj Bjorner (nbjorner)
```
NikolajBjorner, line 17: looks like a bug

src/math/lp/bound_analyzer_on_row.h
```diff
@@ -361,7 +329,20 @@ public :
...
+        case 0:
+            return;
+        case 1:
+            m_bp.cheap_eq_tree(m_row_index);
+            break;
+        case 2:
+            m_bp.cheap_eq_table(m_row_index);
+            break;
+        default:
+            UNREACHABLE();
```
NikolajBjorner, line 343: to avoid some fuzzer filing a bug by miss-configuring z3, just return here (or raise an exception)

src/math/lp/lar_solver.cpp
```diff
@@ -1709,12 +1716,55 @@ bool lar_solver::bound_is_integer_for_integer_column(unsigned j, const mpq & rig
...
     constraint_index ci = mk_var_bound(j, kind, right_side);
     activate(ci);
     return ci;
 }
 
+void lar_solver::remove_non_fixed_from_fixed_var_table() {
+    vector<value_sort_pair> to_remove;
+    for (const auto& p : m_fixed_var_table) {
+        unsigned j = p.m_value;
+        if (j >= column_count() || column_is_fixed(j) == false)
```
NikolajBjorner, line 1735: writing column_is_fixed(j) == false instead of !column_is_fixed(j) is not as bad as writing column_is_fixed(j) == true, but close.

PR #4560, opened 2020-07-03
Cheap eqs: remove the table approach

levnach:


src/smt/theory_lra.cpp
```diff
@@ -1395,10 +1395,22 @@ class theory_lra::imp {
 #endif
         }
         else {
-            literal div_ge_0   = mk_literal(a.mk_ge(div, zero));
-            literal div_le_0   = mk_literal(a.mk_le(div, zero));
-            literal p_ge_0     = mk_literal(a.mk_ge(p, zero));
-            literal p_le_0     = mk_literal(a.mk_le(p, zero));
+#if 0
```
NikolajBjorner, line 1398: seems these literals shouldn't be created?

levnach: I am not sure. These literals used in code below which is surrounded by #if 0 ... #endif

src/math/lp/hnf_cutter.h
```diff
@@ -8,7 +8,8 @@ Module Name:
 Abstract:
 
     Cuts (branches) from Hermite matrices
-
+    The implementation is based on ideas from
+    "Cutting the Mix" by Jurgen Christ and Jochen Hoenicke.
```
NikolajBjorner, line 12: and Cuts from Proofs: A Complete and Practical Technique for Solving Linear Inequalities over Integers
Isil Dillig, Thomas Dillig, Alex Aiken. Formal Methods in System Design (FMSD) CAV 2009 special issue.

src/math/lp/lp_bound_propagator.h
```diff
@@ -278,35 +233,74 @@ class lp_bound_propagator {
         }
         return false;
     }
+
+    // pol for polarity
+    int pol(const vertex* v) const { return pol(v->column()); }
+    int pol(unsigned j) const { return m_pol[j].pol(); }
```
NikolajBjorner, line 239: is polarity a sign as in util/sign.h?

levnach: Yes, it is almost the same, but cannot be zero. I am not sure using sign.h will simplify things.
For example, the code 
` if (pol(v->column()) == -1)`
would be replaced by 
`if (pol(v->column() == sign_neg)`.

PR #4596, opened 2020-07-28
Seq rewriter derivative integration

cdstanford:
Integrate the new derivative normalization functionality into the sequence rewriter.

The primary changes are:

- All conditions formed when creating a derivative now go through the `mk_der_cond` function, which enforces a desired normal form. Currently, it enforces that conditions are rewritten to the form `x <= c` where `c` is a character code.

- In the BDD form, conditions are order by character code whenever possible; otherwise expression ID only as a fallback. So, we redefine `get_id` in the context of `mk_der_op_rec` to be `c` for conditions of the form `x <= c`.

- In the BDD form, conditions are then simplified locally: unreachable branches are removed. (This uses the auxiliary function `pred_implies`)

The result is that for regexes which only involve ranges and raw characters/strings as predicates, the BDD form is linear and all leaves of the BDD are reachable. Finally:

- Additional tracing under the `seq_verbose` tag


src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2403,40 +2411,129 @@ expr_ref seq_rewriter::mk_der_concat(expr* r1, expr* r2) {
...
+    // Simplify if-then-elses whenever possible
     auto mk_ite = [&](expr* c, expr* a, expr* b) {
         return (a == b) ? a : m().mk_ite(c, a, b);
     };
+    // Use character code to order conditions
+    auto get_id = [&](expr* e) {
+        expr *ch1 = nullptr, *ch2 = nullptr;
+        unsigned ch;
+        if (u().is_char_le(e, ch1, ch2) && u().is_const_char(ch2, ch))
+            return ch;
```
NikolajBjorner, line 2495: so get_id is not going to be unique because ch could also be code of an expr

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2403,40 +2411,129 @@ expr_ref seq_rewriter::mk_der_concat(expr* r1, expr* r2) {
...
                 result = mk_ite(ca, r1, r2);
                 return result;
             }
-            else if (ca->get_id() < cb->get_id()) {
-                expr_ref r1 = mk_der_op(k, a, b1);
-                expr_ref r2 = mk_der_op(k, a, b2);
-                result = mk_ite(cb, r1, r2);
-                return result;
+            // Order with higher IDs on the outside
+            if (get_id(ca) < get_id(cb)) {
```
NikolajBjorner, line 2513: seems valid to not require unique id

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2514,6 +2613,71 @@ expr_ref seq_rewriter::mk_der_compl(expr* r) {
     return result;
 }
 
+/*
+    Make an re_predicate with an arbitrary condition cond, enforcing
+    derivative normal form on how conditions are written.
+
+    Tries to rewrites everything to (ele <= x) constraints:
```
NikolajBjorner, line 2620: Tries to rewrite

PR #4610, opened 2020-08-01
string to regex approximation used to strengthen membership constraints

veanes:
As a preprocessing step of regex membership constraints.
Converts a non-ground sequence into an additional regex and strengthens the original regex constraint into an intersection.
For example  (x ++ "a" ++ y) in b* is converted to (x ++ "a" ++ y) in intersect((.* ++ "a" ++ .*), b*).
This will guarantee termination  for some unsat cases that would otherwise not terminate.

### DIFF
--- a/src/ast/seq_decl_plugin.cpp
+++ b/src/ast/seq_decl_plugin.cpp
@@ -1454,8 +1454,8 @@ bool seq_util::re::is_loop(expr const* n, expr*& body, expr*& lo) const {
    Returns true iff e is the epsilon regex.
  */
 bool seq_util::re::is_epsilon(expr* r) const {
-    expr* r1;
-    return is_to_re(r, r1) && u.str.is_empty(r1);
+    expr* s;
+    return is_to_re(r, s) && u.str.is_empty(s);
 }
 /**
    Makes the epsilon regex for a given sequence sort.
--- a/src/ast/seq_decl_plugin.cpp
+++ b/src/ast/seq_decl_plugin.cpp
@@ -1449,3 +1449,17 @@ bool seq_util::re::is_loop(expr const* n, expr*& body, expr*& lo) const {
     }
     return false;
 }
+
+/**
+   Returns true iff e is the epsilon regex.
+ */
+bool seq_util::re::is_epsilon(expr* r) const {
+    expr* r1;
+    return is_to_re(r, r1) && u.str.is_empty(r1);
+}
+/**
+   Makes the epsilon regex for a given sequence sort.
+ */
+app* seq_util::re::mk_epsilon(sort* seq_sort) {
+    return mk_to_re(u.str.mk_empty(seq_sort));
+}
--- a/src/smt/seq_regex.cpp
+++ b/src/smt/seq_regex.cpp
@@ -90,6 +90,8 @@ namespace smt {
         expr* s = nullptr, *r = nullptr;
         expr* e = ctx.bool_var2expr(lit.var());
         VERIFY(str().is_in_re(e, s, r));
+        expr_ref _r(r, m);
+        expr_ref _s(s, m);
 
         TRACE("seq_regex", tout << "propagate in RE: " << lit.sign() << " " << mk_pp(e, m) << std::endl;);
         STRACE("seq_regex_brief", tout << "PIR(" << mk_pp(s, m) << ","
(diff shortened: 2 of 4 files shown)


src/ast/seq_decl_plugin.cpp
```diff
@@ -1449,3 +1449,17 @@ bool seq_util::re::is_loop(expr const* n, expr*& body, expr*& lo) const {
     }
     return false;
 }
+
+/**
+   Returns true iff e is the epsilon regex.
+ */
+bool seq_util::re::is_epsilon(expr* r) const {
+    expr* r1;
```
NikolajBjorner, line 1457: maybe call this s, because the type is a _sequence_ or _string_, not a regex

src/smt/seq_regex.cpp
```diff
@@ -90,6 +90,8 @@ namespace smt {
         expr* s = nullptr, *r = nullptr;
         expr* e = ctx.bool_var2expr(lit.var());
         VERIFY(str().is_in_re(e, s, r));
+        expr_ref _r(r, m);
```
NikolajBjorner, line 93: why do we need this?

src/smt/seq_regex.cpp
```diff
@@ -124,6 +152,64 @@ namespace smt {
...
+        if (m.is_value(s)) {
+            return s_to_re;
+        }
+        else {
+            expr_ref_vector es(m);
+            expr_ref s_approx = epsilon;
+            unsigned int n = 0;
+            if (str().is_concat(s)) {
+                str().get_concat(s, es);
+                n = es.size() - 1;
```
NikolajBjorner, line 173: es.empty()?

src/ast/seq_decl_plugin.cpp
veanes: you mean r1? here r is a regex

src/smt/seq_regex.cpp
```diff
@@ -124,6 +152,64 @@ namespace smt {
...
+                // make sure to simplify so that epsilons are eliminated in
+                // concatenations -- e.g. a sequence
+                // (x ++ "" ++ y ++ "") will be approximated by .*
+                for (unsigned i = n; i != (unsigned)(-1); i--) {
```
NikolajBjorner, line 177: for (unsigned i = n; i-- > 0; ) {

}

src/smt/seq_regex.cpp
```diff
@@ -124,6 +152,64 @@ namespace smt {
...
+                    expr_ref elem_i = get_overapprox_regex(es.get(i));
+                    if (i == n) {
```
NikolajBjorner, line 179: just set s_approx to null and instead of if (i == n) check if (!s_approx)

src/smt/seq_regex.cpp
```diff
@@ -124,6 +152,64 @@ namespace smt {
...
+                        s_approx = elem_i;
+                    }
+                    else if (!re().is_epsilon(elem_i)) {
```
NikolajBjorner, line 182: I thought you wanted to also avoid concatenating dotstar

src/ast/seq_decl_plugin.cpp
NikolajBjorner: expr* s = nullptr;
return is_to_re(r, s) && u.str.is_empty(s);

src/smt/seq_regex.cpp
NikolajBjorner: For example, you could have

expr_ref result(m), last(m);
for (expr* e : es) {
    s_approx = get_overapprox_regex(e);
    if (!result) 
        result = s_approx;
     else if (last != dotstar || s_approx != dotstar)
        result = str.mk_concat(result, s_approx)
     last = s_approx;
}
}

src/ast/seq_decl_plugin.cpp
cdstanford: Yes this is my mistake, r1 should be named s

src/smt/seq_regex.cpp
cdstanford: I added because I am worried about line 123, where we assign `r = re().mk_inter(r, s_approx);` but `r` is an expr* not an expr_ref so this does not save the result to make sure it is not deallocated. However I made a mistake -- `_r` needs to be set again to point to `r` right after line 123

cdstanford: And also to guard against r and s being deallocated in `get_overapprox_regex`. Basically to ensure ownership over `r` and `s` for the duration of this function. However there may be a clearer way

cdstanford: It may be best to just change r and s to `expr_ref`, then we don't have strange variables `_r` and `_s` to deal with. That may require some things to be changed to satisfy typechecking

NikolajBjorner: `e` owns r, s. So the reference is on e.

cdstanford: `e` owns `r` but we later modify `r` without saving it. When we make `re().mk_inter` no one owns it.

PR #4619, opened 2020-08-07
modular Axiom Profiler

rospoly:
This pull request mimics the excellent work done by Nils with the Axiom Profiler.
There are two main conceptual differences:

- top-down approach: Z3 logs human-readable information about quantifier instantiations while running (e.g. bindings, triggers) as opposed to the bottom-up approach used in the Axiom Profiler.

- modularity: logging is done "on-demand", Z3 logs only what is needed. This modular approach sensibly reduces the dimension of the trace and the overhead introduced by logging. In contrast, when you use the native trace of Z3 (trace=true proof=true) you need to parse the entire trace even if you are interested only in a tiny part of it. 

**Summary of the modifications:**
This approach relies on the trace mechanism already existing inside Z3 (only when Z3 is compiled in debugging mode).

We introduced 5 new tags:
- 'instance': logs the instantiation of quantifiers (in the format: fingerprint - QID);
- 'dummy': logs the instantiations that reduce to true or to sat;
- 'causality': logs the updates to the e-graph that occur as a result of quantifier instantiations;
- 'bindings': logs the bindings of instantiations;
- 'triggers': logs the triggers generating instantiations;

Finally, I added two new counters _num_instances_simplify_true_ and _num_instances_checker_sat_ to track both the instances simplied to true and to sat. In this way, the numbers returned by the native  _smt.qi.profile=true_ coincide with the numbers coming from the Axiom Profiler.

Our final goal would be to have a similar feature in release mode.

@shazqadeer @wrwg @DavidLDill


src/ast/rewriter/rewriter_def.h
```diff
@@ -654,10 +654,10 @@ void rewriter_tpl<Config>::cleanup() {
 
 template<typename Config>
 void rewriter_tpl<Config>::display_bindings(std::ostream& out) {
-    out << "bindings:\n";
+    //out << "bindings:\n";
```
NikolajBjorner, line 657: just remove line

src/smt/mam.cpp
```diff
@@ -1893,7 +1893,7 @@ namespace {
         void update_max_generation(enode * n, enode * prev) {
             m_max_generation = std::max(m_max_generation, n->get_generation());
 
-            if (m.has_trace_stream())
+            if (m.has_trace_stream() || is_trace_enabled("causality"))
```
NikolajBjorner, line 1896: There are two kinds of logs: the log associated with debug tracing and the log associated with has_trace_stream().
In this PR you use the debug trace log for quantifier information. This then takes a dependency of compiling with TRACE enabled. The release builds don't have TRACE enabled. Is this sustainable for your use? 
There could be different modalities: (1) is some logging mechanism enabled, (2) where is it logged. 
These modalities are currently conflated, but your PR suggests to give it some thought whether conflating these will work for your use.

rospoly: Hi Nikolaj, thank you for your comments. In order to have these features available in the release build, the _is_trace_enabled("causality")_ should become something like _m.has_quantifier_trace_stream()_. 
At the same time, the tags "causality", "bindings", etc. should become input options to Z3 (e.g. causality=true or bindings=true).
I would like to introduce a new "quantifiers" trace that is independent from the existing trace. This new quantifiers trace is enabled as soon as one of the options (e.g. causality=true or bindings=true) is given in the command line.
In case this direction sounds good I am happy to work on it.

src/smt/smt_quantifier.cpp
```diff
@@ -170,15 +170,21 @@ namespace smt {
         void display_stats(std::ostream & out, quantifier * q) {
             quantifier_stat * s     = get_stat(q);
             unsigned num_instances  = s->get_num_instances();
+            unsigned num_instances_simplify_true = s->get_num_instances_simplify_true();
+            unsigned num_instances_checker_sat  = s->get_num_instances_checker_sat();
             unsigned max_generation = s->get_max_generation();
             float max_cost          = s->get_max_cost();
-            if (num_instances > 0) {
+            if (num_instances > 0 or num_instances_simplify_true>0 or num_instances_checker_sat>0) {
```
NikolajBjorner, line 177: For compatibility: use "||" instead of "or".
The pull request currently does not build because of this.

src/smt/smt_quantifier.cpp
```diff
@@ -199,6 +205,33 @@ namespace smt {
             return m_plugin->is_shared(n);
         }
 
+        void log_causality(
+                fingerprint* f,
+                app * pat,
+                vector<std::tuple<enode *, enode *>> & used_enodes) {
+
+            if (pat != nullptr) {
+                if (used_enodes.size()>0){
```
NikolajBjorner, line 214: add space around >

src/smt/smt_quantifier.cpp
```diff
@@ -261,13 +294,17 @@ namespace smt {
...
                           vector<std::tuple<enode *, enode *>> & used_enodes) {
+
             max_generation = std::max(max_generation, get_generation(q));
             if (m_num_instances > m_params.m_qi_max_instances) {
                 return false;
             }
             get_stat(q)->update_max_generation(max_generation);
             fingerprint * f = m_context.add_fingerprint(q, q->get_id(), num_bindings, bindings, def);
             if (f) {
+                if (is_trace_enabled("causality")){
```
NikolajBjorner, line 305: add space between ) and {

src/smt/qi_queue.cpp
```diff
@@ -219,15 +234,18 @@ namespace smt {
...
                 display_instance_profile(f, q, num_bindings, bindings, pr ? pr->get_id() : 0, generation);
                 m.trace_stream() << "[end-of-instance]\n";
             }
 
+            disable_trace("coming_from_quant");
             return;
         }
         TRACE("qi_queue", tout << "simplified instance:\n" << s_instance << "\n";);
-        quantifier_stat * stat = m_qm.get_stat(q);
+        //quantifier_stat * stat = m_qm.get_stat(q);
```
NikolajBjorner, line 248: Just remove line

src/smt/qi_queue.cpp
```diff
@@ -202,11 +202,26 @@ namespace smt {
         ent.m_instantiated = true;
 
         TRACE("qi_queue_profile", tout << q->get_qid() << ", gen: " << generation << " " << *f << " cost: " << ent.m_cost << "\n";);
+        //NEVER remove coming_from_quant
```
NikolajBjorner, line 205: Add space between // and first character

src/smt/mam.cpp
NikolajBjorner: It is up to you. At present it will only work if you build with the debug trace enabled. You can do this for release builds. It results in a slight overhead.

src/smt/smt_quantifier.cpp
NikolajBjorner: Note that this is currently blocking the PR

src/ast/rewriter/rewriter_def.h
rospoly: Thanks, done.

src/smt/smt_quantifier.cpp
rospoly: done.

src/smt/smt_quantifier.cpp
rospoly: done

src/smt/smt_quantifier.cpp
rospoly: done

src/smt/qi_queue.cpp
rospoly: done

src/smt/qi_queue.cpp
rospoly: done

NikolajBjorner:
Merged for now.
I believe you want to sort out whether you are relying on 

*  ``Z3_ENABLE_TRACING_FOR_NON_DEBUG`` - BOOL. If set to ``TRUE`` enable tracing in non-debug builds, if set to ``FALSE`` disable tracing in non-debug builds. Note in debug builds tracing is always enabled.

and your own builds or a low-overhead mechanism for tracing relevant instantiations in the format and pipe of your choice.

rospoly:
Thank you Nikolaj, this is a precious suggestion for custom builds.

NikolajBjorner:
If by this you mean that our Releases and Nightly Releases should contain other build types (with tracing enabled)
I am afraid it is a non-starter. If you want your tracing to work from distributed release builds you would need to 
enable some means for an appropriate output stream to be supplied relevant logging.

rospoly:
I didn't know about `Z3_ENABLE_TRACING_FOR_NON_DEBUG` option for `cmake`.

For the moment I can play with the "custom" build:
`cmake -DCMAKE_BUILD_TYPE=Release -DZ3_ENABLE_TRACING_FOR_NON_DEBUG=TRUE ../`

Sure, in case we need the quantifier trace available in distributed release builds, I will prepare a new pull request where the trace is logged with an appropriate output stream (similarly to the existing trace dumped into z3.log).

Thank you for your support.

PR #4620, opened 2020-08-08
dgml output generation for regex state graphs

veanes:
Added support for directed graph markup language (dgml) generation support when debugging with option /tr:state_graph. The output will be saved in the file '.z3-state-graph.dgml' every time state_graph is changed in regex solver. The file can be viewed using VS 2019 and some earlier versions like VS 2017 and VS 2015 (dgml viewing is not supported in VS Code).


NikolajBjorner:
./src/util/state_graph.cpp:419:30: error: no ‘void state_graph::write_dgml()’ member function declared in class ‘state_graph’
 void state_graph::write_dgml() {
                              ^
Makefile:238: recipe for target 'util/state_graph.o' failed
make: *** [util/state_graph.o] Error 1
make: *** Waiting for unfinished jobs....

PR #4621, opened 2020-08-09
added support for saving state graph in dot format

veanes:
Calling  state_graph::write_dot() outputs the whole state graph in dot format into the file '.z3-state-graph.dot'
The function is called every time when state_graph is updated in debug mode when '/tr:state_graph' is specified.
The output is equivalent dot format of the dgml format.

Moved both state_graph::write_dot() and state_graph::write_dgml() under the Z3DEBUG flag so that 
they are not included in the release build.

### DIFF
--- a/src/util/state_graph.cpp
+++ b/src/util/state_graph.cpp
@@ -263,10 +263,9 @@ void state_graph::add_state(state s) {
     if (m_seen.contains(s)) return;
     STRACE("state_graph", tout << "[state_graph] adding state " << s << ": ";);
     add_state_core(s);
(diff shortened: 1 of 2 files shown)


src/util/state_graph.cpp
```diff
@@ -413,6 +421,7 @@ std::ostream& state_graph::display(std::ostream& o) const {
     return o;
 }
 
+#ifdef Z3DEBUG
```
NikolajBjorner, line 424: uses of write_dgml inside STRACE in release mode will not compile.

PR #4638, opened 2020-08-13
pp support for regex expressions in more-or-less standard syntax

veanes:
Pretty printing of regex expressions into standard format such as ((abc)|(def))* when this is possible. 
Also: using & for conjunction and ~ for complement of regexes; 
using if(cond, t, f) for the new conditional case where cond is a character condition.


src/ast/seq_decl_plugin.cpp
```diff
@@ -1463,3 +1463,191 @@ bool seq_util::re::is_epsilon(expr* r) const {
...
+  Provides a standard pretty printed view of the regex r when possible.
+*/
+std::string seq_util::re::pp(expr* r) {
+    SASSERT(u.is_re(r));
+    std::ostringstream buffer;
+    pp_compact_to_buffer(buffer, r);
+    return buffer.str();
+}
+
+void seq_util::re::pp_compact_to_buffer(std::ostringstream& buffer, expr* r) {
```
NikolajBjorner, line 1477: std::ostream& is more general than std::ostringstream&

src/ast/seq_decl_plugin.cpp
```diff
@@ -1463,3 +1463,191 @@ bool seq_util::re::is_epsilon(expr* r) const {
...
+    else if (is_loop(r, r1, lo))
+        if (pp_can_skip_parenth(r1))
+        {
+            pp_compact_to_buffer(buffer, r1);
+            buffer << "{" << std::to_string(lo) << ",}";
+        }
+        else {
+            buffer << "(";
+            pp_compact_to_buffer(buffer, r1);
+            buffer << "){" << std::to_string(lo) << ",}";
```
NikolajBjorner, line 1548: I think there is already printers to std::ostream& so no need for std::to_string after you change the parameter type

Also, rename buffer to "out"

veanes: ok

src/ast/seq_decl_plugin.cpp
veanes: will change

src/ast/seq_decl_plugin.h
```diff
@@ -410,6 +410,11 @@ class seq_util {
         seq_util&    u;
         ast_manager& m;
         family_id    m_fid;
+        void seq_util::re::pp_compact_helper_seq(std::ostream& out, expr* s);
```
NikolajBjorner, line 413: seq_util::re:: can be skipped in header

PR #4647, opened 2020-08-18
computing and memoizing info for regexes

veanes:
Started computing info for regex expressions, initially just computing min_length. 
Replaced also original implementation of min_length by using the info value.
Will use it to maintain misc information, such as being standard, being monadic, star-depth, etc.

### DIFF
--- a/src/ast/seq_decl_plugin.cpp
+++ b/src/ast/seq_decl_plugin.cpp
@@ -1675,7 +1675,7 @@ seq_util::re::info seq_util::re::mk_info_rec(app* e) const {
     if (u.m.is_ite(e, c, t, f)) {
         i1 = get_info_rec(t);
         i2 = get_info_rec(f);
-        return info(u.max_plus(1, std::min(i1.min_length, i2.min_length)));
+        return info(std::min(i1.min_length, i2.min_length));
     }
     return invalid_info;
 }
--- a/src/ast/seq_decl_plugin.cpp
+++ b/src/ast/seq_decl_plugin.cpp
@@ -1578,6 +1578,15 @@ std::ostream& seq_util::re::pp::display(std::ostream& out) const {
 }
 
 /*
+  Pretty prints the regex r into the output string
+*/
+std::string seq_util::re::to_str(expr* r) const {
+    std::ostringstream out;
+    out << pp(u.re, r);
+    return out.str();
+}
+
+/*
   Returns true iff info has been computed for the regex r
 */
 bool seq_util::re::has_valid_info(expr* r) const {
@@ -1617,8 +1626,8 @@ seq_util::re::info seq_util::re::get_info_rec(expr* e) const {
     if (!is_app(e))
         return invalid_info;
     result = mk_info_rec(to_app(e));
-    STRACE("re_info", tout << "compute_info(" << pp(u.re, e) << "): min_length=" << m_infos[e->get_id()].min_length << std::endl;);
     m_infos.setx(e->get_id(), result, invalid_info);
+    STRACE("re_info", tout << "compute_info(" << pp(u.re, e) << "): min_length=" << m_infos[e->get_id()].min_length << std::endl;);
     return result;
 }
 
--- a/src/ast/seq_decl_plugin.cpp
+++ b/src/ast/seq_decl_plugin.cpp
@@ -1312,7 +1312,7 @@ unsigned seq_util::str::max_length(expr* s) const {
     return result;
 }
 
-unsigned seq_util::re::min_length(expr* r) {
+unsigned seq_util::re::min_length(expr* r) const {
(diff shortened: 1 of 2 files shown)


src/ast/seq_decl_plugin.h
```diff
@@ -478,10 +493,11 @@ class seq_util {
         bool is_loop(expr const* n, expr*& body, unsigned& lo) const;
         bool is_loop(expr const* n, expr*& body, expr*& lo, expr*& hi) const;
         bool is_loop(expr const* n, expr*& body, expr*& lo) const;
-        unsigned min_length(expr* r) const;
+        unsigned min_length(expr* r);
```
NikolajBjorner, line 496: declare m_info and m_pinned with "mutable", then we can keep the const qualifier, which is probably reasonable.

src/ast/seq_decl_plugin.cpp
```diff
@@ -1595,3 +1576,108 @@ std::ostream& seq_util::re::pp::display(std::ostream& out) const {
...
+    case OP_RE_OPTION:
+        e_info = info(0);
+        break;
+    case OP_RE_RANGE:
+    case OP_RE_FULL_CHAR_SET:
+    case OP_RE_OF_PRED:
+        e_info = info(1);
+        break;
+    case OP_RE_CONCAT:
+        compute_info_rec(ea->get_arg(0));
```
NikolajBjorner, line 1631: if you have this return an info, then you can combine many of the lines that first compute info, then access it.

src/ast/seq_decl_plugin.cpp
```diff
@@ -1595,3 +1576,108 @@ std::ostream& seq_util::re::pp::display(std::ostream& out) const {
...
+        m_info_pinned.push_back(e);
+        compute_info_rec(e);
+    }
+    return m_infos[e->get_id()];
+}
+
+/*
+  Computes the info value for the given regex e recursively over the structure of e
+*/
+void seq_util::re::compute_info_rec(expr* e) {
```
NikolajBjorner, line 1603: thus with return value a seq_util::re::info you can use the result in the combiner.

src/ast/seq_decl_plugin.cpp
```diff
@@ -1595,3 +1576,108 @@ std::ostream& seq_util::re::pp::display(std::ostream& out) const {
...
+    if (has_valid_info(e))
+        return;
+
+    SASSERT(is_app(e));
+    app* ea = to_app(e);
+    SASSERT(!m.is_ite(e));
```
NikolajBjorner, line 1609: this is going to be too fragile to assume the expressions have some restricted form.

src/ast/seq_decl_plugin.cpp
```diff
@@ -1595,3 +1576,108 @@ std::ostream& seq_util::re::pp::display(std::ostream& out) const {
...
+    SASSERT(ea->get_decl()->get_decl_kind() != OP_RE_DERIVATIVE);
+    SASSERT(ea->get_decl()->get_decl_kind() != _OP_RE_ANTIMOROV_UNION);
+
+    info e_info(invalid_info);
+    unsigned k1, k2;
+    switch (ea->get_decl()->get_decl_kind())
```
NikolajBjorner, line 1615: the get_decl_kind() can only be accessed after you have ensured the family_id() of the decl belongs to seq. So there has to be a test whether the ea->get_family_id() is right.

src/ast/seq_decl_plugin.cpp
```diff
@@ -1595,3 +1576,108 @@ std::ostream& seq_util::re::pp::display(std::ostream& out) const {
...
+  Returns true iff info has been computed for the regex r
+*/
+bool seq_util::re::has_valid_info(expr* r) {
+    return r->get_id() < m_infos.size() && m_infos[r->get_id()].is_valid();
+}
+
+/*
+  Get the information value associated with the regular expression e
+*/
+seq_util::re::info seq_util::re::get_info(expr* e)
```
NikolajBjorner, line 1590: Seems cleaner to separate the memoization from the recursive calls.

```
info get_info(expr * e) {
     auto result = get_cached_info(e);
     if (result.is_valid())
           return result;
     m_info_pinned.push_back(e);
     return get_info_rec(e);
}

info get_info_rec(expr* e) {
     auto result = get_cached_info(e);
     if (result.is_valid())
           return result;
     if (!is_app(e))
           return unknown_info;
     result = mk_info(to_app(e));
     add_cache(e, result);
     return result;
}

info mk_info(app* r) {
       if (r->get_family_id() == u.get_family_id()) {
            switch (r->get_decl_kind()) {
            case ...
            
            }
            return unknown_info;
       }
       if (m.is_ite(r, c, t, e)) {

       }
     
}
```

Issue #4651, opened 2020-08-20
(seq and z3str) Issues in string solvers


rainoftime:
Hi, for the following formula
~~~~
(set-logic QF_SLIA)
(declare-const i7 Int)
(declare-const Str3 String)
(declare-const Str7 String)
(declare-const Str14 String)
(assert (str.in_re Str14 (str.to_re (str.substr Str14 0 (abs i7)))))
(assert (str.in_re (str.++ Str3 "pprtut" Str7 "") (re.++ (str.to_re (str.substr Str14 0 (abs i7))) (str.to_re "qbtciigmtwadmbzhrwrpusydtnouwhnukkdarqzjmqmwcitojz"))))
(check-sat)
~~~~

cvc4 yields `sat`, but both z3seq and z3str3 return `unsat` (commit 79aa3457c1)

A model given by cvc4
~~~~
(model
(define-fun i7 () Int 6)
(define-fun Str3 () String "")
(define-fun Str7 () String "qbtciigmtwadmbzhrwrpusydtnouwhnukkdarqzjmqmwcitojz")
(define-fun Str14 () String "pprtut")
)
~~~~

mtrberzi:
@NikolajBjorner if this is affecting both string solvers, maybe something is going on with `seq_rewriter`? I can take a look today if it's useful.

rainoftime:
At 4682b48d3a
~~~~
(set-logic QF_SLIA)
(set-option :model_validate true)
(set-option :smt.arith.solver 2)
(set-option :smt.string_solver z3str3)
(declare-const i7 Int)
(declare-const Str9 String)
(declare-const Str11 String)
(declare-const Str12 String)
(declare-const Str13 String)
(declare-const Str19 String)
(assert (>= (str.len Str12) 37))
(assert (= Str13 Str11 Str19 Str12 (str.substr Str9 0 i7)))
(check-sat)
~~~~

~~~~
sat
(error "line 13 column 10: an invalid model was generated")
(model 
  (define-fun i7 () Int
    (- 1))
  (define-fun Str9 () String
    "?????????????????????????????????????")
  (define-fun Str13 () String
    "")
  (define-fun Str11 () String
    "")
  (define-fun Str19 () String
    "")
  (define-fun Str12 () String
    "")
)
~~~~

rainoftime:
at 4682b48
~~~~
(set-logic QF_S)
(set-option :smt.phase_selection 1)
(declare-const v3 Bool)
[...]

PR #4666, opened 2020-08-27
fixing issue #4651

veanes:
Added rewrite rules to lift to_re from within ite-expressions.
Also disabled lift_ites_throttled that seems to cause performance issues (in this example 100x slowdown).
Updated also pp of regexes to indicate where a regex is not fixed.
The underlying issue of #4651 is fixed or at least does not show up with this example.


src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -728,9 +728,11 @@ br_status seq_rewriter::mk_app_core(func_decl * f, unsigned num_args, expr * con
     case _OP_STRING_STRIDOF: 
         UNREACHABLE();
     }
+#if 0
```
NikolajBjorner, line 731: this will create a perf regression for strings, where lifting ites is very useful.
So it could look at the type of the expression. If it is a seq, then lift, otherwise don't.
That can be accomplished inside lift_ites_throttled, not here.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3811,7 +3840,15 @@ br_status seq_rewriter::mk_re_star(expr* a, expr_ref& result) {
         result = re().mk_star(re().mk_union(b1, c1));
         return BR_REWRITE2;
     }
+    if (m().is_ite(a, c, b1, c1))
+    {
```
NikolajBjorner, line 3844: move { to after ), not on a new line as in C#

src/ast/seq_decl_plugin.cpp
```diff
@@ -1452,20 +1452,24 @@ app* seq_util::rex::mk_epsilon(sort* seq_sort) {
...
+                if (re.u.str.is_unit(e))
+                    seq_unit(out, e);
+                else
+                    out << mk_pp(e, re.m);
+            }
         }
+        else if (re.u.str.is_empty(s))
+            out << "()";
+        else
+            seq_unit(out, s);
```
NikolajBjorner, line 1469: this makes a "long distance" assumption about values.
Better to not make such assumptions deeply buried in code and have the routine more self-contained.
You can simply check for is_concat(s) regardless of whether it is a value, and you can check for unit, and empty as 3 separate cases. The is_value check only filters cases of concatenations of non-constants and constants, where the conventions in this routine are not uniform: in the default case you use (to_re ...), in the non-default case it doesn't use (to_re ..) around the expression.

So instead
```
if (is_empty(s))
    out << "();
else if is_unit(s))
      seq_unit(s, s);
else if (is_concat(s)) 
    for (expr* e : concats(s))
          concat_helper_sec(out, s);
else out << "(to_re " << pp(s) << ")";
```

veanes: I updated it similarly, but essentially I wanted to distinguish between 'A' vs uninterpreted const A, the latter is now printed as {A}
Another thing I wanted to distinguish is (in some cases) pp(to_re S) from pp(S), but in most cases the distiction is not needed or desired and just adds noise, so I did not do it in the end. e.g. pp(tore("ab") ++ tore("cd") ) and 
pp(tore("ab" ++ "cd")) are printed the same way

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -860,24 +860,41 @@ br_status seq_rewriter::mk_seq_length(expr* a, expr_ref& result) {
...
-    }
     return BR_FAILED;
 }
 
+/* returns false iff the ite must not be lifted */
+bool seq_rewriter::lift_ites_filter(func_decl* f, expr* ite)
+{
+    // do not lift over regexes
+    // for example DO NOT lift to_re(ite(c, s, t)) to ite(c, to_re(s), to_re(t))
+    if (is_sort_of(f->get_range(), get_fid(), RE_SORT))
```
NikolajBjorner, line 887: this is too low level, we have higher level checks, or add them if they don't already exist:

- u.is_re(f->get_range())
- u.is_seq(f->get_range())
- a.is_arith(m().get_sort(ite))

veanes:
While the branch is named  fix4651, it only partially addresses use of nonground regexes in the related issue. This is work in progress. But the PR does fix one crucial bug where state ids were off by one in lookup table, luckily this lookup table was not used in core algorithms (yet). Discovered this with the new feature I added to dgml generation that also generates a 'Value' field to each state that is the pretty printed regex corresponding to the state with the given label and is shown as tool-tip in VS when hovering over the node. I found it very useful for debugging. I added a callback method to the state_graph that asks for the string value of a state, so the state_graph is still not dependent and knows nothing of regexes, it can be sued for anu other graphs.

PR #4667, opened 2020-08-28
Allow to skip System.loadLibrary() calls from Java Native class

vlsergey:
It's not hard for client code to write a couple of lines for System.loadLibrary()

But it is very very hard for client code to embed Z3 code into JAR file where libraries are need to be loaded from JAR file (as resource). The problem here that System.loadLibrary() works only if libraries are placed in java.library.path and does not allow to load it from temporary folder/file.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -542,8 +542,13 @@ def mk_java(java_dir, package_name):
     java_native.write('  public static native void setInternalErrorHandler(long ctx);\n\n')
 
     java_native.write('  static {\n')
-    java_native.write('    try { System.loadLibrary("z3java"); }\n')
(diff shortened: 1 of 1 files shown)


kfriedberger:
@NikolajBjorner and @vlsergey 
Is it guaranteed that this behavior is stable?
Would it be possible to officially document the environment variable and specify the range of possible values?
Currently any (!) value for the environment variable "z3.skipLibraryLoad" will disable the loading of the native library of Z3, including a value like `false`, which might misleading.
Are there any plans to change the range of values to only allow a value `true` here to enable this functionality of not loading the library?

vlsergey:
not sure what is the best place for documenting this

PR #4710, opened 2020-09-25
Perf branch

levnach:


levnach:
There are performance issues - do not merge yet!

PR #4729, opened 2020-10-10
z3++: add missing fpa operator >= implementation

xlauko:


wintersteiger:
LGTM! Not sure whether the CI failure is just noise. @NikolajBjorner is the timeout expected?

NikolajBjorner:
this timeout has happened with increasing frequency. It only affects the 64 bit windows build with a particular build system.

PR #4741, opened 2020-10-19
mbp

NikolajBjorner:


NikolajBjorner:
Remove "model" keyword from model printout per SMTLIB2 standardization discussion

PR #4748, opened 2020-10-21
thread pool for scoped_timer

regehr:
creating a fresh thread for every scoped_timer has significant overhead
in some use cases. this patch creates a persistent pool of worker threads
to do this job, resulting in 20-30% speedup of some alive2 jobs on a
large multicore

### DIFF
--- a/src/util/scoped_timer.cpp
+++ b/src/util/scoped_timer.cpp
@@ -20,6 +20,7 @@ Revision History:
 --*/
 
 #include "util/scoped_timer.h"
+#include "util/mutex.h"
 #include "util/util.h"
 #include <chrono>
 #include <climits>
@@ -38,6 +39,30 @@ struct state {
     std::mutex cv_lock;
 };
 
+static void thread_func(state *s) {
+    while (true) {
+        s->cv_lock.lock();
+        s->cv.wait(s->cv_lock, [=]{return s->work > 0;});
+        s->cv_lock.unlock();
+
+        auto end = std::chrono::steady_clock::now() + std::chrono::milliseconds(s->ms);
+
+        while (!s->m_mutex.try_lock_until(end)) {
+            if (std::chrono::steady_clock::now() >= end) {
+                s->eh->operator()(TIMEOUT_EH_CALLER);
+                goto next;
+            }
+        }
+
+        s->m_mutex.unlock();
+
+    next:
+        s->cv_lock.lock();
+        s->work--;
+        s->cv_lock.unlock();
+    }
+}
+
 /*
  * NOTE: this implementation deliberately leaks threads when Z3
  * exits. this is preferable to deallocating on exit, because
@@ -45,36 +70,10 @@ struct state {
  * deadlock.
  */
 static std::vector<state *> available_workers;
-static std::mutex workers;
+static mutex workers;
 
 struct scoped_timer::imp {
 private:
-
-    static void thread_func(state *s) {
-        while (true) {
-
-            s->cv_lock.lock();
-            s->cv.wait(s->cv_lock, [=]{return s->work > 0;});
-            s->cv_lock.unlock();
-
-            auto end = std::chrono::steady_clock::now() + std::chrono::milliseconds(s->ms);
-
-            while (!s->m_mutex.try_lock_until(end)) {
-                if (std::chrono::steady_clock::now() >= end) {
-                    s->eh->operator()(TIMEOUT_EH_CALLER);
-                    goto next;
-                }
-            }
-
-            s->m_mutex.unlock();
-
-        next:
-            s->cv_lock.lock();
-            s->work--;
-            s->cv_lock.unlock();
-        }
-    }
-
     state *s;
 
 public:
--- a/src/util/scoped_timer.cpp
+++ b/src/util/scoped_timer.cpp
@@ -83,6 +83,7 @@ public:
         } else {
             s = available_workers.back();
             available_workers.pop_back();
+            workers.unlock();
         }
         s->ms = ms;
         s->eh = eh;
@@ -93,7 +94,6 @@ public:
             s->m_thread->detach();
         } else {
             s->cv.notify_one();
-            workers.unlock();
         }
     }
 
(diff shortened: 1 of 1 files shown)


nunoplopes:
thanks. I'll benchmark this tomorrow on my 8-core VM as well & review the patch.

regehr:
do you mind benchmarking using this script, in addition to whatever benchmarks you run yourself?
[AnghaBench.pl.txt](https://github.com/Z3Prover/z3/files/5419562/AnghaBench.pl.txt)
you'll need to download these: https://github.com/brenocfg/AnghaBench
you may need to change a couple of variables at the top of the perl script too

regehr:
here are some results from a 10-core machine. speedup is ~23% if we add up user+sys times.

before:
```
regehr@home:~/llvm-project/for-alive/bin$ time ~/svn/code/alive2/AnghaBench.pl
10 cores.
1045235 files found.
using 500 of them.

real	13m42.630s
user	41m53.956s
sys	7m29.556s
regehr@home:~/llvm-project/for-alive/bin$ time ~/svn/code/alive2/AnghaBench.pl
10 cores.
1045235 files found.
using 500 of them.

real	13m49.678s
user	42m0.444s
sys	7m34.632s
regehr@home:~/llvm-project/for-alive/bin$ time ~/svn/code/alive2/AnghaBench.pl
10 cores.
1045235 files found.
using 500 of them.

real	13m46.964s
user	42m12.235s
sys	7m35.660s
```

after:
```
regehr@home:~/llvm-project/for-alive/bin$ time ~/svn/code/alive2/AnghaBench.pl
10 cores.
1045235 files found.
using 500 of them.

real	13m42.630s
user	41m53.956s
sys	7m29.556s
regehr@home:~/llvm-project/for-alive/bin$ time ~/svn/code/alive2/AnghaBench.pl
10 cores.
1045235 files found.
using 500 of them.

real	13m49.678s
user	42m0.444s
sys	7m34.632s
regehr@home:~/llvm-project/for-alive/bin$ time ~/svn/code/alive2/AnghaBench.pl
10 cores.
1045235 files found.
using 500 of them.

real	13m46.964s
user	42m12.235s
sys	7m35.660s
```

regehr:
all numbers I've taken so far are on Linux. I should test this patch on my mac. it's pretty slow and I use it a lot so I might not be able to get to this before the weekend. I don't have a good Windows machine for testing.

nunoplopes:
In confirm perf on linux & windows with 8 cores improves quite a bit. Kernel time goes down by 3-4x.

NikolajBjorner:
seems to still be a source of memory leak on shutdown

nunoplopes:
True, it will leak memory on shutdown. That's hard/expensive to avoid. It's only on shutdown, so it's benign. Given the nice speedup, I think it's a fair tradeoff. Most leak checking tools allow one to give a file with things to ignore.

NikolajBjorner:
Other users complain about memory state of dynamically loaded libraries. 
It adds liability in a different side and is going to fall on me. 
Last time I checked one can join threads to properly collect workers.

regehr:
well, let me know if you folks want me to write this cleanup code, it's not difficult stuff, though it is a bit ugly.

nunoplopes:
The threads are blocked on conditional variables so they can't be joined straight away.
We would need a protocol to unblock the conditional variables and exit the threads. Not impossible, but non-trivial.
Also, we would prefer to keep the threads live between Z3 memory resets as applications need to reset Z3's memory frequently to keep memory consumption under control. There's atexit(), but do we really want to go that way?  The memory leak is really small.

regehr:
ok, the termination sequence is simple:
- add an atomic "terminating" flag to this module that gets set by someone
- add an atomic thread count variable
- modify the condvar wakeup condition to include terminating == true; when this happens the thread decrements the thread count and exits
- someone needs to loop over the cond vars signaling them, and then that needs to be wrapped in a second loop since these wakeup signals can get dropped. the outer loop finishes when thread count == 0.

the thing I don't know here is how to wire this into Z3's cleanup machinery. maybe we can talk about this offline, Nuno.

NikolajBjorner:
I added a stub that is wired up

NikolajBjorner:
> Not impossible, but non-trivial.

We do this all over the place elsewhere. This is a known art. For me, it is mainly an issue to properly tune into what you have done with detached threads and which locks you are using. So much easier for me to say you can add this facility. I added comment about requirement - you can leak on abnormal shutdown, that is fine. If there is a timer task running it is not a strong requirement to collect even though it is of course possible. This should make it even easier.

nunoplopes:
I think `ADD_FINALIZER` is called on every memory reset, not just on shutdown. Deleting all threads on every reset will impose a perf degradation. `atexit()` is the only way to cleanup on shutdown AFAICT.
I really don't see an issue with leaking a few bytes on program exit. And this is just for programs that use timeouts anyway.

tavianator:
I think the cleanest way to do the cleanup would be to wrap the global state in a struct with a destructor

NikolajBjorner:
So ALIVE2 uses reset_memory and the requirement is that it should not impose overhead based on deallocating and reallocating threads. 

The API exposes two methods: Z3_reset_memory and Z3_finalize_memory. The requirement for finalize memory is to clean up state and I am going to get support calls (and they tend to be repeated from different sources) if the requirement for finalize memory isn't satisfied. So an option is to special case the finalization of timers for Z3_finalize_memory and not touch the thread pool in reset_memory.
Since reset_memory touches the memory manager we will have to not use Z3's memory manager (as the original code didn't).

nunoplopes:
Yes, Alive2 calls `Z3_reset_memory` frequently to keep memory consumption under control. Otherwise memory consumption sky rockets due to unbounded caching in e.g. the small object allocator. Also, it resets the AST ids which is helpful for algorithms that are dense over AST ids. So any application creating & deleting a lot of Z3 expressions needs to call `Z3_reset_memory` frequently.
`Z3_reset_memory` calls `memory::finalize();` & `memory::initialize(0);`, so everything on Z3 side is destroyed.

NikolajBjorner:
> so everything on Z3 side is destroyed.

Let me try to spell it out: above I imply that Z3_finalize should use a separate method for finalizing the timer.

nunoplopes:
> > so everything on Z3 side is destroyed.
> 
> Let me try to spell it out: above I imply that Z3_finalize should use a separate method for finalizing the timer.

Ah ok, understood now. I'll work on it.

PR #4751, opened 2020-10-22
finalize scoped_thread workers instead of leaking them

regehr:
hello z3 folks,
of course I don't have an opinion about how (or whether) you integrate this patch, but I believe that it correctly finalizes the scoped_timer workers for both the case where finalization is called in the middle of execution, and when it is called during program termination.

### DIFF
--- a/src/util/scoped_timer.cpp
+++ b/src/util/scoped_timer.cpp
@@ -22,6 +22,7 @@ Revision History:
 #include "util/scoped_timer.h"
 #include "util/util.h"
 #include "util/vector.h"
+#include <atomic>
 #include <chrono>
 #include <climits>
 #include <condition_variable>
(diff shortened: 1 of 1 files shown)


src/util/scoped_timer.cpp
```diff
@@ -34,25 +35,27 @@ struct state {
...
 };
 
-/*
- * NOTE: this implementation deliberately leaks threads when Z3
- * exits. this is preferable to deallocating on exit, because
- * destructing threads blocked on condition variables leads to
- * deadlock.
- */
-static std::vector<state*> available_workers;
+static ptr_vector<state> available_workers;
```
NikolajBjorner, line 42: vector::clear will not call "delete", so finalize will not really clean up memory. This is easy for me or Nuno to fix.

Based on Nuno's input I would say we don't have the finalize_scoped_timer registered as a finalizer, but we call it explicitly in api_context.cpp to clean up state for DLLs. This is the only scenario that really matters (and the command-line versions to quiet down the dynamic tools that would start complaining).

nunoplopes:
Just found this. Will review tmr.

nunoplopes:
Ok, I've fixed this issue (valgrind clean!). This patch was still leaking state.
https://github.com/Z3Prover/z3/commit/4e9035d4b9f7b42b9cb1f07b2d901ae80fd83fba

PR #4753, opened 2020-10-25
Fixed ModelRef translate method in Python API

acw1251:
This fixes a bug in the translate method of the ModelRef class in the Python API. Without this fix, calling the translate method causes an exception because the wrong function is called.

Here's a simple test program I used. Without the fix, `s1.model().translate(c2)` raises an exception, with the fix, the program prints `sat`, `True`, and `True`.
```python
import z3

c1 = z3.Context()
c2 = z3.Context()
a1, b1 = z3.Bools('a b', c1)
a2, b2 = z3.Bools('a b', c2)

s1 = z3.Solver(ctx=c1)
s1.add(z3.And(a1,b1))
print(s1.check())
m2 = s1.model().translate(c2)
print(m2[a2])
print(m2[b2])
```


NikolajBjorner:
Thanks!

PR #4785, opened 2020-11-10
Fix in nla order

levnach:


NikolajBjorner:
great, seems we arrived to the same conclusion. See also the change in the function that deals with the equality case.

PR #4818, opened 2020-11-24
z3str3: reject certain unhandled expressions

mtrberzi:
Reject unhandled terms like `replace_all`, `replace_re`, etc. This is in line with the current behaviour of the sequence solver (`theory_seq.cpp:3172`)


mtrberzi:
@NikolajBjorner are the test errors to be expected here? I did not modify the arithmetic solver so seeing these tests fail is surprising

PR #4832, opened 2020-11-28
Java type generics

VTCAKAVSMoACE:
Using type generics like this makes it more feasible to implement operations in Java with fewer casts and much easier to use with the `var` keyword. There are a few breaking changes that I'd like to review with a contributor and a few questions about typing (as I am not familiar with some of the features of z3, e.g. enum types vs finite domains, Optimize, Fixedpoint, etc). Gonna go through and add comments on the PR for each of these.

### DIFF
--- a/examples/java/Java13Example.java
+++ b/examples/java/Java13Example.java
@@ -1,0 +1,2295 @@
+/*++
+ Copyright (c) 2012 Microsoft Corporation
+
+Module Name:
+
+    Program.java
+
+Abstract:
+
+    Z3 Java API: Example program
+
+Author:
+
+    Christoph Wintersteiger (cwinter) 2012-11-27
+
+Notes:
+    
+--*/
+
+import com.microsoft.z3.*;
+
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import static java.util.stream.Stream.concat;
+
+class Java13Example
+{
+    @SuppressWarnings("serial")
+    class TestFailedException extends Exception
+    {
+        public TestFailedException()
+        {
+            super("Check FAILED");
+        }
+    };
+
+    // / Create axiom: function f is injective in the i-th argument.
+
+    // / <remarks>
+    // / The following axiom is produced:
+    // / <code>
+    // / forall (x_0, ..., x_n) finv(f(x_0, ..., x_i, ..., x_{n-1})) = x_i
+    // / </code>
+    // / Where, <code>finv</code>is a fresh function declaration.
+
+    public <R extends Sort> BoolExpr injAxiom(Context ctx, FuncDecl<R> f, int i)
+    {
+        Sort[] domain = f.getDomain();
+        int sz = f.getDomainSize();
+
+        if (i >= sz)
+        {
+            System.out.println("failed to create inj axiom");
+            return null;
+        }
+
+        /* declare the i-th inverse of f: finv */
+        Sort finv_domain = f.getRange();
+        Sort finv_range = domain[i];
+        var finv = ctx.mkFuncDecl("f_fresh", finv_domain, finv_range);
+
+        /* allocate temporary arrays */
+        /* fill types, names and xs */
+        Sort[] types = domain.clone();
+        var xs = IntStream.range(0, sz).mapToObj(j -> ctx.mkBound(j, types[j])).collect(Collectors.toList());
+        Symbol[] names = IntStream.range(0, sz).mapToObj(j -> ctx.mkSymbol(String.format("x_%d", j))).toArray(Symbol[]::new);
+
+        var x_i = xs.get(i);
+
+        /* create f(x_0, ..., x_i, ..., x_{n-1}) */
+        var fxs = f.apply(xs.toArray(Expr[]::new));
+
+        /* create f_inv(f(x_0, ..., x_i, ..., x_{n-1})) */
+        var finv_fxs = finv.apply(fxs);
+
+        /* create finv(f(x_0, ..., x_i, ..., x_{n-1})) = x_i */
+        var eq = ctx.mkEq(finv_fxs, x_i);
+
+        /* use f(x_0, ..., x_i, ..., x_{n-1}) as the pattern for the quantifier */
+        Pattern p = ctx.mkPattern(fxs);
+
+        /* create & assert quantifier */
+
+        return ctx.mkForall(types, /* types of quantified variables */
+                names, /* names of quantified variables */
+                eq, 1, new Pattern[] { p } /* patterns */, null, null, null);
+    }
+
+    // / Create axiom: function f is injective in the i-th argument.
+
+    // / <remarks>
+    // / The following axiom is produced:
+    // / <code>
+    // / forall (x_0, ..., x_n) finv(f(x_0, ..., x_i, ..., x_{n-1})) = x_i
+    // / </code>
+    // / Where, <code>finv</code>is a fresh function declaration.
+
+    public <R extends Sort> BoolExpr injAxiomAbs(Context ctx, FuncDecl<R> f, int i)
+    {
+        Sort[] domain = f.getDomain();
+        int sz = f.getDomainSize();
+
+        if (i >= sz)
+        {
+            System.out.println("failed to create inj axiom");
+            return null;
+        }
+
+        /* declare the i-th inverse of f: finv */
+        var finv_domain = f.getRange();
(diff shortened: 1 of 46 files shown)


examples/java/JavaExample.java
```diff
@@ -599,7 +599,7 @@ void quantifierExample1(Context ctx)
         System.out.println("QuantifierExample");
         Log.append("QuantifierExample");
 
-        Sort[] types = new Sort[3];
+        IntSort[] types = new IntSort[3];
```
VTCAKAVSMoACE, line 602: Change is necessary because of cast to ArithExpr<?>, which is technically not strictly compatible with Expr<?> because ArithExpr<T extends ArithSort> specifies that the generic T is more specific than Sort. There are ways around this but it makes things a little clunky and removes the advantages that typing like this poses.

Provided that most users are specifically typing, this will likely not affect many users.

src/api/java/Context.java
```diff
@@ -217,22 +218,22 @@ public BitVecSort mkBitVecSort(int size)
...
-        return new ArraySort(this, domain, range);
+        return new ArraySort<>(this, domain, range);
     }
 
 
     /**
      * Create a new array sort.
      **/
-    public ArraySort mkArraySort(Sort[] domains, Sort range)
+    public <R extends Sort> ArraySort<?, R> mkArraySort(Sort[] domains, R range)
```
VTCAKAVSMoACE, line 232: No real way to extract the domain sort here. Java type generics aren't quite as powerful as C++ templates.

src/api/java/Context.java
```diff
@@ -548,28 +549,28 @@ public FuncDecl mkConstDecl(String name, Sort range)
...
-        return Expr.create(this,
+        return (Expr<R>) Expr.create(this,
                 Native.mkBound(nCtx(), index, ty.getNativeObject()));
     }
 
     /**
      * Create a quantifier pattern.
      **/
-    public Pattern mkPattern(Expr... terms)
+    public Pattern mkPattern(Expr<?>... terms)
```
VTCAKAVSMoACE, line 573: Is there anything we can say consistently about the types of the terms in a pattern? Not sure how this is restricted in z3.

src/api/java/Optimize.java
```diff
@@ -75,7 +76,7 @@ public void Add(BoolExpr ... constraints)
     /**
      * Handle to objectives returned by objective functions.
      **/
-    public static class Handle {
+    public static class Handle<R extends Sort> {
```
VTCAKAVSMoACE, line 79: Can this be more narrowly defined to `R extends ArithSort`?

src/api/java/Optimize.java
```diff
@@ -110,7 +111,7 @@ public Expr getUpper()
          * and otherwise is represented by the expression {@code value + eps * EPSILON},
          * where {@code EPSILON} is an arbitrarily small real number.
          */
-        public Expr[] getUpperAsVector()
+        public Expr<?>[] getUpperAsVector()
```
VTCAKAVSMoACE, line 114: Can this be typed to `R` instead of `?`?

src/api/java/Optimize.java
```diff
@@ -150,19 +150,19 @@ public String toString()
      * Return an objective which associates with the group of constraints.
      *
      **/
-    public Handle AssertSoft(BoolExpr constraint, int weight, String group)
+    public Handle<?> AssertSoft(BoolExpr constraint, int weight, String group)
```
VTCAKAVSMoACE, line 153: Requires explicit casting by user because there's no input to base the result type with; I think this is a reasonable requirement on the user. Won't break existing code.

VTCAKAVSMoACE (commented):
Because Expr\<FPSort\> and Expr\<RealSort\> have the same parameters after type erasure, it's not possible to have them permissively typed. There might be a workaround with a rename (e.g. mkFPtoFP_FP, mkFPtoFP_Real), but this is kind of gross and in most cases it's likely acceptable for it to just be cast by the user.

VTCAKAVSMoACE (commented):
Like the previous issue between Expr\<FPSort\> and Expr\<RealSort\>, these two conflict after type erasure. Same notes as above.

VTCAKAVSMoACE:
I think this is just about ready. Only breaking changes are where things are explicitly typed by a cast where they should not be. Things are made more permissive for different types of `Expr` and there's forced consistency between various operations, e.g. `mkAdd` will no longer accept Exprs of two different subtypes without cast (though this would only be breaking if someone was already using `var` keywords).

I'll add an example of why this makes life so much better for all of us later this evening, gotta get food.

VTCAKAVSMoACE:
You can diff JavaExample.java and Java13Example.java to see how auto-typing + functional with streams makes things a lot easier to use.

examples/java/Java13Example.java
```diff
@@ -1649,18 +1472,17 @@ public void treeExample(Context ctx) throws TestFailedException
     // / forest ::= nil | cons(tree, forest)
     // / tree ::= nil | cons(forest, forest)
     // / </remarks>
-    public void forestExample(Context ctx) throws TestFailedException
+    @SuppressWarnings({"unchecked", "unused", "UnusedAssignment"})
+    public <Tree, Forest> void forestExample(Context ctx) throws TestFailedException
```
VTCAKAVSMoACE, line 1476: I especially encourage you to look at this to get a better understanding of why these changes are such a quality of life improvement.

VTCAKAVSMoACE:
Looks like the pipelines are failing because it's [attempting to compile Java13Example with Java 8](https://github.com/actions/virtual-environments/blob/ubuntu16/20201116.1/images/linux/Ubuntu1604-README.md).

The code that modifies z3 proper is all Java 8 compatible, but the example code is theoretically 11+ (though I've only tested it with 13). The script or the build environment could be modified, but I think that's a little beyond my decision.

NikolajBjorner:
These updates look great.
There was a decision to have sorts of expressions reflected in sub-types. This is both manifested in Java and .Net.
It forces writing application code that is statically type correct, but then ends up requiring a bunch of coercions.
It is my understanding that this update retains the subtyping, but avoids some of the end-user hassle.

It would also break other client code that would have to be ported to handle this new discipline.
Given that it is a breaking change, I will wait a little week before merging in case someone using the Java bindings have substantial issues.

The Java13 example could be excluded from the build or the pipelines would somehow have to be configured to use the newer, non-default, version of Java. I am not sure how exactly to configure the pipelines with a non-default version of Java so I suggest to omit the Java13 example in the first round.

VTCAKAVSMoACE:
As far as I know, there are technically no breaking changes to *already compiled* code, as these casts are removed during type erasure, I did not modify the underlying un-generic types, and generically typed return values are strictly subtypes of what was originally returned. As for the code which will be compiled after this change, the only breaking changes are to when sorts are explicitly declared as wider than what they are, e.g. `Sort s = ctx.getIntSort()`, which are then used to type the return values of a `mkConst` or similar. Technically I think I can work around these too and make it have zero breaking changes (!), so let me see what I can do.

I'll just `/* */` out Java13Example.java.

NikolajBjorner:
> Technically I think I can work around these too and make it have zero breaking changes (!), so let me see what I can do.
I don't think it is worth it. Clients should either use "var" or the more specific types. I don't see a reason to support something that is a step back.

VTCAKAVSMoACE:
It boils down to whether you prefer completeness or correctness:
```
Sort s = ctx.getIntSort();
IntExpr e = (IntExpr) ctx.mkConst("a", s);
```

This fails to compile, even though it's sound. On the other hand, this also becomes permissible:
```
Sort s = ctx.getIntSort();
BitVecExpr e = (BitVecExpr) ctx.mkConst("a", s);
```

Which would compile with the change, but it would be incorrect. Note that, were `s` specifically typed to `IntSort`, this would not compile.

Which case would be more apt? With the change, you sacrifice some correctness, but it's compatible with older versions as unchecked type generic casting.

VTCAKAVSMoACE:
Actually, after testing this, that change makes more problems than it solves. Disregard.

VTCAKAVSMoACE:
Rather than outright deleting Java13Example, I replaced it with JavaGenericExample which is Java 8 compatible (used some scripting to undo var, then removed some API features introduced in Java 11). Let's see what happens in CI and hopefully that's that.

NikolajBjorner:
> It boils down to whether you prefer completeness or correctness:

I don't see any need to change your pull request: the changes required in JavaExample.java are benign and I consider them warranted. I think other clients will be well served with making similar changes.

VTCAKAVSMoACE:
Okay, that's everything. Let me know if you find anything that doesn't make perfect sense.

NikolajBjorner:
Super!

PR #4850, opened 2020-12-04
Fix finding .git directory in CMake when z3 is a submodule of some other repository

SweetVishnya:
Hi!

We use z3 as a submodule in our project. This commit fixes z3 build when its source is located in a submodule.


andrewvaughanj:
Are there edge-cases where `${GIT_WORKTREE_DIR}/commondir` doesn't exist and Z3 is not a submodule?

Your code has changed the build system to default to a submodule (I guess without _confirming_ it is a submodule) vs. the current behaviour to error-out.

SweetVishnya:
Added verification that z3 is actually a submodule, otherwise I error-out as current behavior.

PR #4887, opened 2020-12-11
Fix: Prove and Disprove methods are exactly the same

louthy:
Seems like a cut 'n' paste issue with Prove and Disprove.  

There's a spurious `MkNot` in the main boolean expression assertion for Prove.  This has been removed.

There are other build issues with this project, but this seems like a particularly misleading example, so I figured I'd fix it in isolation.


NikolajBjorner:
It is actually right. Maybe the fact that we have Prove and Disprove as very similar functions is the problem with the example.

PR #4958, opened 2021-01-21
fix #4956

### LINKED ISSUE #4956: 'utf-8' codec can't decode byte 0xe9 in position 2: invalid continuation byte.
When I'm using accented characters in constant names, I get the error in subject.

The following code in z3core.py (line 69) is highly dubious to me:

    def _str_to_bytes(s):
      if isinstance(s, str):
        try:
          return s.encode('latin-1')
        except:
          # kick the bucket down the road.  :-J
          return s
      else:
        return s

Maybe it's a left-over from python2.  
If I change `latin-1` to `utf-8`, it works... but that's more a hack than a good solution.

pcarbonn:
Fix dubious use of 'latin-1' encoding.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -337,7 +337,7 @@ def Z3_set_error_handler(ctx, hndlr, _elems=Elementaries(_lib.Z3_set_error_handl
   _elems.Check(ctx)
   return ceh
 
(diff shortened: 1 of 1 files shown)


pcarbonn:
Sorry that my linter removed some white space.
Also, the commented section is only for easy reference to the previous version.  It can be removed.

pcarbonn:
I don't have a make tool chain, nor a set of regression test, so I cannot guarantee non regression.
However, when I modify the code in my local z3core.py, my problem goes away.

PR #4959, opened 2021-01-22
Added the SUBMODULE_DIR_NAME variable for better submodule support

sillydan1:
Having `"modules"` as the submodules folder name is a bit restrictive. My personal projects use the name `"deps"` instead. 

This PR adds the `SUBMODULE_DIR_NAME` variable to the git cmake utility so that people can set the variable before the `add_subdirectory()` in cmake.


roastduck:
Perhaps this new `SUBMODULE_DIR_NAME` option should be documented into the options list in `README-CMake.md`.

PR #4981, opened 2021-01-28
Unicode for Z3str3

mtrberzi:
This PR enhances Z3str3 with support for Unicode and a more general theory of characters (see #4939).


src/smt/smt_setup.cpp
```diff
@@ -129,6 +129,8 @@ namespace smt {
             setup_QF_FPBV();
         else if (m_logic == "QF_S" || m_logic == "QF_SLIA")
             setup_QF_S();
+        else if (m_logic == "QF_CHAR")
```
NikolajBjorner, line 132: Can you revert changes in smt_setup.h and smt_setup.cpp?
It would introduce a path towards QF_CHAR support for an ad-hoc logic, which I would like to avoid.

src/smt/theory_str_mc.cpp
```diff
@@ -916,7 +909,7 @@ namespace smt {
 
         smt_params subsolver_params;
         smt::kernel subsolver(m, subsolver_params);
-        subsolver.set_logic(symbol("QF_BV"));
+        subsolver.set_logic(symbol("QF_CHAR"));
```
NikolajBjorner, line 912: You can do "QF_S"

src/smt/theory_str_mc.cpp
```diff
@@ -770,7 +771,7 @@ namespace smt {
                 // convert iValue to a constant
                 zstring iValue_str(iValue.to_string());
                 for (unsigned idx = 0; idx < iValue_str.length(); ++idx) {
-                    expr_ref chTerm(bitvector_character_constants.get(iValue_str[idx]), sub_m);
+                    expr_ref chTerm(u.mk_char(iValue_str[idx]), sub_m);
```
NikolajBjorner, line 774: sub_m has to be the same as m, otherwise this will crash. u.mk_char refers to m.
So suggestion is to get rid of sub_m.

NikolajBjorner:
I added the changes other than smt_setup

src/smt/theory_str_mc.cpp
mtrberzi: This is giving me trouble for some reason even after adding `setup_char()` and `setup_QF_BV()` to `setup_str()` in `smt_setup`. I'm taking a look

mtrberzi: Even doing `subsolver_params.m_string_solver=symbol("seq")` does not give me a working character theory

mtrberzi: For some reason if a string solver theory plugin is loaded at the same time as `theory_char` this does not work. I tried making a case for a "char" string solver in `setup_QF_S()` that only loads QF_BV and theory_char, and this works perfectly -- would this solution be acceptable since it doesn't create an ad_hoc `QF_CHAR` as before?

mtrberzi:
New commits to reflect the review comments. I'm happy to move the `m_string_solver='char'` to a separate option in `smt_params` that isn't exposed to the user. Either way, it seems that I need to control the setup a bit more precisely since having either string theory plugin loaded at the same time as the character theory is causing problems, and I also need to make sure bit-vectors are loaded since I can't know whether they will be used instead of Unicode characters.

NikolajBjorner:
Some feathers have been roughed.
Regression 3757.smt2 now fails. It is specific for z3str3.    

There is an issue with reference counted objects and the following signature points to a problem:

bool theory_str::fixed_length_reduce_string_term(smt::kernel & subsolver, expr * term,
            ptr_vector<expr> & eqc_chars, expr_ref & cex) {


The expressions in the returned vector eqc_chars are not accounted. So if they are supplied by some expression that gets garbage collected before return, code down the line crashes.
The systematic fix is to use expr_ref_vector& instead of ptr_vector<expr>.
Since this change requires touching a lot of code I am not going to implement it.

NikolajBjorner:
string9.smt2 is also toast

mtrberzi:
Taking a look and I will fix the reference counting problem you pointed out.

With regard to the existing open issues I will see where we stand after these changes and fixes. I am working with another developer right now who will be (most likely) continuing development of theory_str once I graduate. We have some code that uses the new seq_axioms in particular and I believe it fixes a few issues but needs to be cleaned up. We will pick this up in the next couple of weeks.

mtrberzi:
Release build produces the expected results for both of the indicated regressions but I will re-test after fixing refcount

mtrberzi:
Updated PR. There are other instances of `obj_map<expr, ptr_vector<expr>>` which I will need to update. This fixes the most critical ones for this patch. I followed the example done in e.g. `value_generator.cpp` which needs to maintain a similar map where the value is an `expr_ref_vector`. 

It would be fantastic if there was a specialization for `obj_map<expr, expr_ref_vector>` since the constructor of `expr_ref_vector` is non-trivial (as it takes `ast_manager`) and otherwise I have to resort to manually `alloc` and `dealloc`ing the objects. If this is useful elsewhere, I can add it.

PR #5003, opened 2021-02-07
switch expr_safe_replace to std::unordered_map

nunoplopes:
`Z3_substitute` is extremely slow. I profiled it and one of the culprits is hash collisions in `expr_safe_replace`'s hash table. Z3's hash table is not particularly smart in handling collisions and goes into linear scan very quickly.

I tried several alternatives and the best one gives a nice 35% speedup in some crazy Alive2 benchmarks:

speedup

This PR is the top performing alternative.
I know that Z3 has traditionally avoid using std:: containers, but there's precedence in the new lp code, for example. Moreover, it's unlikely any std:: implementation performs worse than Z3's 😅

### DIFF
--- a/src/ast/rewriter/expr_safe_replace.cpp
+++ b/src/ast/rewriter/expr_safe_replace.cpp
@@ -22,18 +22,12 @@ Revision History:
 #include "ast/rewriter/var_subst.h"
 #include "ast/ast_pp.h"
 
-#define ALIVE_OPT 0
-
 
 void expr_safe_replace::insert(expr* src, expr* dst) {
     SASSERT(src->get_sort() == dst->get_sort());
     m_src.push_back(src);
     m_dst.push_back(dst);
-#if ALIVE_OPT
-    cache_insert(src, dst);
-#else
     m_subst.insert(src, dst);
-#endif
 }
 
 void expr_safe_replace::operator()(expr_ref_vector& es) {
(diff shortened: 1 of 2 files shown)


src/ast/rewriter/expr_safe_replace.cpp
```diff
@@ -71,6 +55,11 @@ void expr_safe_replace::operator()(expr* e, expr_ref& res) {
         res = e;
         return;
     }
+
+    for (auto &[src,dst] : m_subst) {
```
NikolajBjorner, line 59: Are you also going to update our build chain from C++11 now that I updated the Centos build?

nunoplopes: nothing would make me happier than moving Z3 to C++20! 😁
Ok, fine, C++17 for now.

nunoplopes: I've removed the non-C++11 stuff from this commit + a few little tweaks for.. a nice 50% speedup 🚀

src/ast/rewriter/expr_safe_replace.cpp
```diff
@@ -46,76 +39,57 @@ void expr_safe_replace::operator()(expr_ref_vector& es) {
...
     expr* a, *b;
     
     while (!m_todo.empty()) {
         a = m_todo.back();
-        if (cache_find(a)) {
+        auto &cached = m_cache[a];
+        if (cached) {
             m_todo.pop_back();
         }
-#if !ALIVE_OPT
```
NikolajBjorner, line 82: why is this removed?

nunoplopes:
Regarding `ALIVE_OPT`, it consumed a lot of memory (we have a lot of expressions lying around) and the speedups weren't consistent. I think this version with unordered_map is a better tradeoff, which has lower memory consumption but still brings a good speedup.

Issue #4409, opened 2020-05-20
(z3str3) Soundness bug on QF_S formula with str.to_code, str.from_code


muchang:
Hi,
For this case, Z3  gives an incorrect answer:
```
[643] % z3 small.smt2
unknown
[644] % cvc4 -q small.smt2
unsat
[645] % z3 smt.string_solver=z3str3 small.smt2
sat
[646] % 
[646] % cat small.smt2
(declare-const a String)
(assert (distinct (str.to_code (str.from_code (str.to_code a))) (str.to_code a)))
(check-sat)
[647] %
```

OS: Ubuntu 18.04
Commit: f2d3160

mtrberzi:
This has returned `unknown` in Z3str3 for a while, but now that we properly support Unicode I think I can take another look at this and get support for `str.from_code / str.to_code`.

PR #5004, opened 2021-02-07
Fix is_from_code() and is_to_code()

mtrberzi:
While implementing `str.from_code` and `str.to_code` in Z3str3 I ran into some weird problems matching against these operators and wondered what was going on -- turns out they are backwards in `seq_util::str`!

As a bonus, `theory_seq` correctly solves the instance in #4409 with this patch (support for it in `theory_str` will be on the way)


NikolajBjorner:
Thanks, yes, this fixed it.

Issue #4984, opened 2021-01-30
Build failure on s390x with march != zEC12


LocutusOfBorg:
Hello, for some reasons, Ubuntu s390x doesn't default to zEC12, and the package FTBFS
See: 
https://launchpadlibrarian.net/520351419/buildlog_ubuntu-hirsute-s390x.z3_4.8.10-1_BUILDING.txt.gz

Unfortunately from the build log nothing obvious is seen, but doing this:
CXXFLAGS = -march=zEC12
works as a workaround
https://launchpad.net/ubuntu/+source/z3/4.8.10-1ubuntu1/+build/20954530

Do you have any idea?

NikolajBjorner:
There is something obvious:

/usr/bin/ld: ../math/lp/CMakeFiles/lp.dir/nla_core.cpp.o: in function `std::_Function_handler<void (unsigned int, bool, _scoped_interval<dep_intervals>&), nla::core::check_pdd_eq(dd::solver::equation const*)::{lambda(unsigned int, bool, _scoped_interval<dep_intervals>&)#1}>::_M_invoke(std::_Any_data const&, unsigned int&&, bool&&, _scoped_interval<dep_intervals>&)':
./obj-s390x-linux-gnu/src/math/lp/./src/math/lp/nla_core.cpp:1695: undefined reference to `void nla::intervals::set_var_interval<(dep_intervals::with_deps_t)1>(unsigned int, dep_intervals::im_config::interval&)'

NikolajBjorner:
These linking issues are vary between platforms and compilers. They boil down to instantiating certain templates to ensure that object code gets produced. I pushed an attempted fix.

LocutusOfBorg:
```
/<<PKGBUILDDIR>>/src/math/lp/nla_core.cpp: In instantiation of ‘void nla::intervals::set_var_interval(nla::lpvar, nla::intervals::interval&) [with dep_intervals::with_deps_t wd = dep_intervals::with_deps; nla::lpvar = unsigned int; nla::intervals::interval = dep_intervals::im_config::interval]’:
/<<PKGBUILDDIR>>/src/math/lp/nla_core.cpp:1694:74:   required from here
[...]

PR #5005, opened 2021-02-08
Add debian directory

LocutusOfBorg:
Simple travis-ci s390x test for #4984


debian/patches/00-intrinsics.patch
```diff
@@ -0,0 +1,71 @@
...
+---
+This patch header follows DEP-3: http://dep.debian.net/deps/dep3/
+--- a/src/util/hwf.cpp
++++ b/src/util/hwf.cpp
+@@ -48,7 +48,7 @@
+ // clear to the compiler what instructions should be used. E.g., for sqrt(), the Windows compiler selects
+ // the x87 FPU, even when /arch:SSE2 is on.
+ // Luckily, these are kind of standardized, at least for Windows/Linux/macOS.
+-#if defined(__clang__) || defined(_M_ARM) && defined(_M_ARM64)
++#if defined(__clang__) || (defined(__GNUC__) && !defined(__SSE2__)) || defined(_M_ARM) && defined(_M_ARM64)
```
NikolajBjorner, line 15: A change like this could be pushed upstream.

debian/tests/include-z3-test
```diff
@@ -0,0 +1,54 @@
+#!/bin/sh
```
NikolajBjorner, line 1: The builds that build and run examples are supposed to cover this test.

debian/patches/00-intrinsics.patch
andrewvaughanj: I think the copyright on this might be a problem. @FabianWolff opened PR #4504 for these changes, but they weren't merged due to CLA issues.

debian/changelog
```diff
@@ -0,0 +1,270 @@
...
+  * Refresh/update patches.
+  * Upgrade to Standards-Version 4.5.0 (no changes).
+  * Forward patches to upstream.
+
+ -- Fabian Wolff <fabi.wolff@arcor.de>  Sat, 06 Jun 2020 16:32:57 +0200
+
+z3 (4.8.7-4) unstable; urgency=medium
+
+  * Remove all :all and :native annotations from the build dependencies in
+    debian/control (Closes: #948763).
```
NikolajBjorner, line 42: So is the entire point to ensure builds for Debian work. 
This should be addressable by loading an appropriate Docker container similar to what we do for Centos.
There we use "container: "quay.io/pypa/manylinux2010_x86_64:latest"
What is the debian version? Is it something like "debian:bullseye"? 
My point of reference is: https://hub.docker.com/_/debian

debian/patches/00-intrinsics.patch
LocutusOfBorg: yes, in any case, this PR is not meant to be merged, but to trigger s390x builds

debian/tests/include-z3-test
LocutusOfBorg: This is done in the installed package as autopkgtest, not as build-time test.
This is useful to detect if something in the toolchain breaks the test of the package, without having to rebuild it by hand

debian/changelog
LocutusOfBorg: nack on this, the idea is to use an Ubuntu container, because debian defaults to zEC12 (and that architecture works)

FabianWolff:
I'm not sure what the trajectory of this conversation/PR is, but as the maintainer of the z3 Debian package, I would prefer if there was no upstream `debian/` subdirectory, because I would then have to repack every release tarball to remove `debian/*` before adding my own `debian/*`. Currently, I can just import your releases verbatim (so to speak), which is easier for me.

NikolajBjorner:
The point is to avoid build breaks downstream, so in the continuous build pipelines include a sufficient mix of target platforms. 
We used to release for Debian and can do so again if I can figure out how to set up the build pipeline. I don't believe it would not obscure efforts to create deeper packaging solutions.
Note that building for s390x seems still out of scope for what I can offer from my end.

NikolajBjorner:
I spot this thing in the Travis build: "std=gnu++11"
Would that entail that you are using GCC11? 
This is not yet standard available and also not tested in our pipelines.

andrewvaughanj:
> I spot this thing in the Travis build: "std=gnu++11"

That's for the "GNU extensions" of the C++11 standard (similar to `gnu99` for C99).

NikolajBjorner:
> > I spot this thing in the Travis build: "std=gnu++11"
> 
> That's for the "GNU extensions" of the C++11 standard (similar to `gnu99` for C99).

doh..

LocutusOfBorg:
@FabianWolff I just opened this pull request to do some travis PR and test-builds on s390x, not to get it merged :)

LocutusOfBorg:
Interestingly, on focal with gcc-10 it doesn't FTBFS. So it might be a change that broke on groovy and hirsute (probably glibc then?)

LocutusOfBorg:
I confirm Ubuntu 20.04 is good
https://launchpad.net/~costamagnagianfranco/+archive/ubuntu/locutusofborg-ppa/+build/21008969

And Ubuntu 20.10 is bad
https://launchpad.net/~costamagnagianfranco/+archive/ubuntu/locutusofborg-ppa/+build/21008953
According to the build logs I suspect glibc or something like binutils

NikolajBjorner:
I don't see a descriptive error message in the build log for 20.10. 
So at this point I guess it is a bug with tools used by 20.10 and outside of Z3.

PR #5008, opened 2021-02-09
Cleanup z3util python module

tpaviot:
pylint and codespell issues fixes: remove trailing whitespaces, remove extraneous newlines, unnecessary else/elif after return, dangerous [] as default argument value (https://satran.in/b/python--dangerous-default-value-as-argument), typo.

### DIFF
--- a/src/api/python/z3/z3util.py
+++ b/src/api/python/z3/z3util.py
@@ -1,20 +1,21 @@
 ############################################
 # Copyright (c) 2012 Microsoft Corporation
-# 
+#
 # Z3 Python interface
 #
 # Authors: Leonardo de Moura (leonardo)
 #          ThanhVu (Vu) Nguyen <tnguyen@cs.unm.edu>
 ############################################
 """
-Usage:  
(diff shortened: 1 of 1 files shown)


tpaviot:
Sorry, I did not realize I have to sign the MS CLA. I don't agree with some of the terms.

NikolajBjorner:
Sorry. It for sure would appear to be overkill for a pull request around python utilities.
An important reason why we have this in place is to give commercial companies assurance 
that if they ship with Z3 there isn't something out of the blue they get held liable about.
I don't think the intent is to irritate contributors.

tpaviot:
No problem, I understand, I should have carefully read the contributor guidelines. z3 is a great project, I wish I could contribute some code to the python api, it needs an overall cleanup to be consistent with python3 coding rules. But it works quite well as is, these are only cosmetic changes.

NikolajBjorner:
You are invited to point out what should be changed for Python 3 and I can try to follow the advise.

PR #5038, opened 2021-02-18
z3str3: add str.is_digit support

mtrberzi:


NikolajBjorner:
Try also to check out z3test/regressions/string4.smt2 in debug mode. I had to disable this unit test as theory_str hits an assert there.

mtrberzi:
@NikolajBjorner OK - that will be next. I've updated this PR against the one I just merged.

PR #5055, opened 2021-02-24
z3str3: use assert_axiom_rw more consistently

mtrberzi:
Fixes a few spurious `assertion failure: !is_true` cases in debug mode, and likely improves performance a little bit since we don't have to create a new `th_rewriter` everywhere.


mtrberzi:
I'm assuming this build failure is a timeout, so I will merge this today if there are no objections since all the other checks passed quickly.

PR #5098, opened 2021-03-12
Mixing Integers and Rational in the new Java API #5085

### LINKED ISSUE #5085: Mixing Integers and Rational in the new Java API
Hi, this is a question and comment on the new Java API as introduced in #4832 by @VTCAKAVSMoACE 

I saw that one cannot mix values of different sorts in operations anymore as it worked beforehand.
For instance the following is valid in the old API (4.8.9) but does not compile in the new API (4.8.10).
```java
var intNum = ctx.mkInt(-1);
var floatNum = ctx.mkReal(3.14);
var x = ctx.mkMul(intNum, floatNum);
```
What is the intended way of writing something like this in the new API?
(Obviously one could call `ctx.mkInt2Real` but this would shift work from the Z3 to the developer.)

akreuzer:
This should fix #5085 

The idea is to allow `Expr` to the have sort `ArithSort`, which could be integers or reals.
The problem in Java with this is that is does not support covariance, i.e., `Expr<ArithSort>` is not a supertype of `Expr<RealSort>`.
To overcome this, create the function `Expr.distillSort` which can do this cast.

Please let me know what you think.

### DIFF
--- a/src/api/java/Context.java
+++ b/src/api/java/Context.java
@@ -723,7 +723,7 @@ public class Context implements AutoCloseable {
     /**
      * Creates the equality {@code x = y}
      **/
-    public <R extends Sort> BoolExpr mkEq(Expr<R> x, Expr<R> y)
+    public BoolExpr mkEq(Expr<?> x, Expr<?> y)
     {
         checkContextMatch(x);
         checkContextMatch(y);
@@ -735,7 +735,7 @@ public class Context implements AutoCloseable {
      * Creates a {@code distinct} term.
      **/
     @SafeVarargs
-    public final <R extends Sort> BoolExpr mkDistinct(Expr<R>... args)
+    public final BoolExpr mkDistinct(Expr<?>... args)
     {
         checkContextMatch(args);
         return new BoolExpr(this, Native.mkDistinct(nCtx(), args.length,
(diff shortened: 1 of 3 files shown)


src/api/java/Context.java
```diff
@@ -723,7 +723,7 @@ public BoolExpr mkBool(boolean value)
     /**
      * Creates the equality {@code x = y}
      **/
-    public <R extends Sort> BoolExpr mkEq(Expr<R> x, Expr<R> y)
+    public BoolExpr mkEq(Expr<?> x, Expr<?> y)
```
VTCAKAVSMoACE, line 726: Now the issue is inverted; this will accept arbitrary exprs for comparison (e.g. a sequence and an integral sort). I don't see a good solution for fixing the mixed arith without adding a new method, though...

akreuzer: Indeed it is possible to compare expr that cannot be equal because of their sort. 
But I think it is better to allow this rather than not begin able to express valid comparisons.

I totally agree with you that this is suboptimal, but essentially this is a limitation of Java.

NikolajBjorner: I would support the new version because equality has some overloading properties that cannot be captured in the type system. At least Z3 throws a runtime exception.

PR #5104, opened 2021-03-15
Circular seq axioms node

andrewvaughanj:
In master (hash 909808421), the following:

```bash
./configure
make smt/seq_axioms.h.node
```

will give:

```
make: Circular smt/seq_axioms.h.node <- smt/seq_axioms.h.node dependency dropped.
```

This is because the functionality in `find_file` was not written to expect there to be multiple includes with the same base name, and where shared names include each other.

For example, this is a problem with `smt/seq_axioms.h` which itself includes `ast/rewriter/seq_axioms.h`. In the past, when trying to resolve `seq_axioms.h` **in the `src_dir` of `smt`** would give the same file, because it was only using `seq_axioms.h`.

This PR corrects this to endeavour to correct ambiguity by also storing the full `#include` directive (and not just the base name).

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -976,9 +976,17 @@ class Component:
     # That is, we were looking for fname when processing ownerfile
     def find_file(self, fname, ownerfile, orig_path=None):
         full_fname = os.path.join(self.src_dir, fname)
+
+        # Store all our possible locations
         possibilities = set()
+
+        # If the our file exists in the current directory, then we store it
         if os.path.exists(full_fname):
-            return self
+
+            # We cannot return here, as we might have files with the same
+            # basename, but different include paths
+            possibilities.add(self)
+
         for dep in self.deps:
             c_dep = get_component(dep)
             full_fname = os.path.join(c_dep.src_dir, fname)
@@ -1005,8 +1013,10 @@ class Component:
                         # ... use our new match
                         return possibility
 
-            # This means we didn't make an exact match, but we're no worse than
-            # before
+            # This means we didn't make an exact match ...
+            #
+            # We return any one possibility, just to ensure we don't break Z3's
+            # builds
             return possibilities.pop()
 
         raise MKException("Failed to find include file '%s' for '%s' when processing '%s'." % (fname, ownerfile, self.name))
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -974,7 +974,7 @@ class Component:
     # Find fname in the include paths for the given component.
     # ownerfile is only used for creating error messages.
     # That is, we were looking for fname when processing ownerfile
-    def find_file(self, fname, ownerfile, orig_path=None):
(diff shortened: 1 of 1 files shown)


scripts/mk_util.py
```diff
@@ -769,7 +769,7 @@ def parse_options():
 # Return a list containing a file names included using '#include' in
 # the given C/C++ file named fname.
 def extract_c_includes(fname):
```
NikolajBjorner, line 771: I wonder, this code dates back to a time when we didn't require include paths to be relative to src.

andrewvaughanj: Oh, like every dir in the source tree was specified as an include path?

NikolajBjorner: It used to be the case that includes were all relative to the paths given by the dependencies in mk_project.
This was changed so that include paths are relative to src directory. 
The python build project was written based on the earlier assumption. It couldn't deal with two header files with the same name. So now I added two occurrences of seq_axioms.h (and seq_eq_solver.h) but under different paths. 
This appears to the source of the warning and based on your changes it would seem that this original python code bakes in the unique name assumption.

andrewvaughanj:
So, not only has this fixed the warning as reported by @LeventErkok, but it has possibly removed an unrequired `.node` dependency:

```bash
avj@platypus ~/clones/z3$ diff <(grep -Eoh "^[^:]*\.node:" circular_seq_axioms_node/build/Makefile | sort | uniq) <(grep -Eoh "^[^:]*\.node:" master/build/Makefile | sort | uniq)
289a290
> math/lp/polynomial.h.node:
```

That is, my branch does not generate a rule for `math/lp/polynomial.h.node`, but that's good because it seems nothing:

* in `src/math/lp` `#includes` `polynomial.h`
* nothing in the rest of the Z3 tree `#includes` `lp/polynomial.h`

scripts/mk_util.py
andrewvaughanj: Yep, this all aligns with what I found when I made the change.

I wonder if a different approach would be just to gut this logic all together and just use the path, as it appears in the `#include`?

NikolajBjorner: Yes, better if it is possible to remove this customization from the build system.
Sorry, I haven't dug into this in more detail and tried yet.

andrewvaughanj: Just so I understand (and this isn't said in a grumpy way, it is to avoid wasted effort): are you saying you'd rather not merge this at all, and await a PR that fixes it "properly"?

That's totally fine, just want to make sure we're on the same page!

NikolajBjorner:
I don't have a definite opinion on this. My preference is to merge and move on. If this is ready for integration it can be done. The python build system has some technical debt that I really would like to sunset all-together given that cmake takes care of a lot (but unfortunately not all).

NikolajBjorner:
Seems the builds all passed except for MacOS that was canceled for other reasons

PR #5156, opened 2021-04-07
Add move semantics to c++ api

zwimer:
This PR:
1. Adds move semantics to every class expose by the C++ API.
2. Replaces constructor declarations with definitions (e.x. `array(array const & s);` -> `array(array const & s) = default;` (This avoids linkage errors as it promises a definition as well).
3. Updated all classes to obey the rule of 5. That is, if a move or copy constructor or operator is defined, or a destructor, all must be defined. This is because defining one causes the compiler to silently not-define others within that set. E.x. defining a destructor undefines the implicit move constructor, which if defined undefines the implict copy constructor.
4. Mark a few trivially non-exception throwing methods as `noexcept`
5. Fixed a bug or two (e.x. `optimize(optimize& o):object(o)` -> `optimize(optimize const& o):object(o)`
6. Removed trailing whitespace
7. Declared a few classes as `final`.

Features added / justifications:
1. Move semantics supported by the C++ API (all of which are noexcept leave the objects in states that are safe to destruct).
2. Move semantics can give an absolutely **massive** performance increase.
3. Expanded standard library support due to move semantics
4. Makes it easier to move API elements onto the heap (no longer requires an expensive copy that increments internal reference counts 'unnecessarily')
5. Construction and assignment can be used in `noexcept` contexts; since move constructors and assignment operators should always be coded to be noexcept, they can _safely_ be used within noexcept contexts. This is beneficial for anything that requires the `noexcept` specifier, and for functions that can now have it added (because it is a decent optimization).
6. `public` constructors / assignment operators. In the classes with `private` copy constructors / assignment operators, this enables public a public constructor / assignment operator; this should be save since they are `move` functions and not `copy` functions; that is they do not 'duplicate' the item, but rather 'relocate' it. For classes that didn't have these, moving them into a vector, or onto the heap would have been non-trivial.

For example, `vec.push_back(old)` is no longer a copy followed by a destruction, both of which edit reference counts and make internal z3 calls, instead if we replace `old` with `std::move(old)` it simply relocates the data contained by `old` into the vector, leaving `old` in a state that, when destructed, will not affect the data `old` used to contain. In fact, due to the move constructor we can even replace this with `vec.emplace_back(std::move(old))` which constructs the new object in-place by moving the internals of `old` into the new object. Reference counts etc will not need to be increased here because `old` will be left in a state where it no longer points to the internal objects.

Downsides:
1. Move semantics were introduced in `C++11`; this has to be compiled with `--std=c++11` or later.

### DIFF
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -110,8 +110,8 @@ namespace z3 {
     */
     class config {
         Z3_config    m_cfg;
-        config(config const &) = default;
-        config & operator=(config const &) = default;
+        config(config const &) = delete;
+        config & operator=(config const &) = delete;
     public:
         config() { m_cfg = Z3_mk_config(); }
         config(config && s) noexcept : m_cfg{std::move(s.m_cfg)} {
@@ -425,8 +425,8 @@ namespace z3 {
     class array {
         T *      m_array;
         unsigned m_size;
-        array(array const &) = default;
-        array & operator=(array const &) = default;
+        array(array const &) = delete;
+        array & operator=(array const &) = delete;
     public:
         array(unsigned sz):m_size(sz) { m_array = new T[sz]; }
         array(array && s) noexcept : m_array{s.m_array}, m_size{s.m_size} {
@@ -440,7 +440,7 @@ namespace z3 {
         }
         template<typename T2>
         array(ast_vector_tpl<T2> const & v);
-        ~array() { if(m_array != nullptr) delete[] m_array; }
+        ~array() { delete[] m_array; }
         void resize(unsigned sz) { delete[] m_array; m_size = sz; m_array = new T[sz]; }
         unsigned size() const { return m_size; }
         T & operator[](int i) { assert(0 <= i); assert(static_cast<unsigned>(i) < m_size); return m_array[i]; }
@@ -518,7 +518,7 @@ namespace z3 {
         }
         param_descrs& operator=(param_descrs && o) noexcept {
             object::operator=(std::forward<object>(o));
-             m_descrs = o.m_descrs;
+            m_descrs = o.m_descrs;
             o.m_descrs = nullptr;
             return *this;
         }
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -3766,6 +3766,10 @@ namespace z3 {
             scoped_cb(void* _p, Z3_solver_callback cb):p(static_cast<user_propagator_base*>(_p)) {
                 p->cb = cb;
             }
+			scoped_cb(scoped_cb const &) = default;
+			scoped_cb(scoped_cb &&) = default;
+			scoped_cb & operator=(scoped_cb const &) = default;
+			scoped_cb & operator=(scoped_cb &&) = default;
             ~scoped_cb() { 
                 p->cb = nullptr; 
             }
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -412,9 +412,9 @@ namespace z3 {
         context m_ctx;
     public:
         scoped_context(Z3_context c): m_ctx(c) {}
-        scoped_context(scoped_context const &) = default;
+        scoped_context(scoped_context const &) = delete;
         scoped_context(scoped_context &&) noexcept = default;
-        scoped_context & operator=(scoped_context const &) = default;
+        scoped_context & operator=(scoped_context const &) = delete;
         scoped_context & operator=(scoped_context &&) noexcept = default;
         ~scoped_context() { m_ctx.detach(); }
         context& operator()() { return m_ctx; }
@@ -453,17 +453,6 @@ namespace z3 {
     protected:
         context * m_ctx;
     public:
-        object(context & c):m_ctx(&c) {}
-        object(object const &) = default;
-        object(object && s) noexcept : m_ctx{s.m_ctx} {
-            s.m_ctx = nullptr;
-        }
-        object & operator=(object const &) = default;
-        object & operator=(object && s) noexcept {
-            m_ctx = s.m_ctx;
-            s.m_ctx = nullptr;
-            return *this;
-        }
         context & ctx() const { return *m_ctx; }
         Z3_error_code check_error() const { return m_ctx->check_error(); }
         friend void check_context(object const & a, object const & b);
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -425,9 +425,8 @@ namespace z3 {
             s.m_array.release();
         }
         array & operator=(array && s) noexcept {
-            m_array = s.m_array;
+            m_array.reset(s.m_array.release());
             m_size = s.m_size;
-            s.m_array.release();
             return *this;
         }
         template<typename T2>
@@ -492,7 +491,7 @@ namespace z3 {
             o.m_descrs = nullptr;
             return *this;
         }
-        ~param_descrs() { if(m_descrs != nullptr) Z3_param_descrs_dec_ref(ctx(), m_descrs); }
+        ~param_descrs() { if (m_descrs != nullptr) Z3_param_descrs_dec_ref(ctx(), m_descrs); }
         static param_descrs simplify_param_descrs(context& c) { return param_descrs(c, Z3_simplify_get_param_descrs(c)); }
 
         unsigned size() { return Z3_param_descrs_size(ctx(), m_descrs); }
@@ -510,7 +509,7 @@ namespace z3 {
         params(context & c):object(c) { m_params = Z3_mk_params(c); Z3_params_inc_ref(ctx(), m_params); }
         params(params const & s):object(s), m_params(s.m_params) { Z3_params_inc_ref(ctx(), m_params); }
         params(params && s) noexcept : object(std::forward<object>(s)), m_params(s.m_params) { s.m_params = nullptr; }
-        ~params() { if(m_params != nullptr) Z3_params_dec_ref(ctx(), m_params); }
+        ~params() { if (m_params != nullptr) Z3_params_dec_ref(ctx(), m_params); }
         operator Z3_params() const { return m_params; }
         params & operator=(params const & s) {
             Z3_params_inc_ref(s.ctx(), s.m_params);
@@ -585,7 +584,7 @@ namespace z3 {
         ast_vector_tpl(ast_vector_tpl && s) noexcept : object(std::forward<object>(s)), m_vector(s.m_vector) { s.m_vector = nullptr; }
         ast_vector_tpl(context& c, ast_vector_tpl const& src): object(c) { init(Z3_ast_vector_translate(src.ctx(), src, c)); }
 
-        ~ast_vector_tpl() { if(m_vector != nullptr) Z3_ast_vector_dec_ref(ctx(), m_vector); }
+        ~ast_vector_tpl() { if (m_vector != nullptr) Z3_ast_vector_dec_ref(ctx(), m_vector); }
         operator Z3_ast_vector() const { return m_vector; }
         unsigned size() const { return Z3_ast_vector_size(ctx(), m_vector); }
         T operator[](int i) const { assert(0 <= i); Z3_ast r = Z3_ast_vector_get(ctx(), m_vector, i); check_error(); return cast_ast<T>()(ctx(), r); }
@@ -2247,7 +2246,7 @@ namespace z3 {
         func_entry(context & c, Z3_func_entry e):object(c) { init(e); }
         func_entry(func_entry const & s):object(s) { init(s.m_entry); }
         func_entry(func_entry && s) noexcept : object(std::forward<object>(s)), m_entry{s.m_entry} { s.m_entry = nullptr; }
-        ~func_entry() { if(m_entry != nullptr) Z3_func_entry_dec_ref(ctx(), m_entry); }
+        ~func_entry() { if (m_entry != nullptr) Z3_func_entry_dec_ref(ctx(), m_entry); }
         operator Z3_func_entry() const { return m_entry; }
         func_entry & operator=(func_entry const & s) noexcept {
             Z3_func_entry_inc_ref(s.ctx(), s.m_entry);
@@ -2277,7 +2276,7 @@ namespace z3 {
         func_interp(context & c, Z3_func_interp e):object(c) { init(e); }
         func_interp(func_interp const & s):object(s) { init(s.m_interp); }
         func_interp(func_interp && s) noexcept : object(std::forward<object>(s)), m_interp{s.m_interp} { s.m_interp = nullptr; }
-        ~func_interp() { if(m_interp != nullptr) Z3_func_interp_dec_ref(ctx(), m_interp); }
+        ~func_interp() { if (m_interp != nullptr) Z3_func_interp_dec_ref(ctx(), m_interp); }
         operator Z3_func_interp() const { return m_interp; }
         func_interp & operator=(func_interp const & s) {
             Z3_func_interp_inc_ref(s.ctx(), s.m_interp);
@@ -2318,7 +2317,7 @@ namespace z3 {
         model(model const & s):object(s) { init(s.m_model); }
         model(model && s) noexcept : object(std::forward<object>(s)), m_model(s.m_model) { s.m_model = nullptr; }
         model(model& src, context& dst, translate) : object(dst) { init(Z3_model_translate(src.ctx(), src, dst)); }
-        ~model() { if(m_model != nullptr) Z3_model_dec_ref(ctx(), m_model); }
+        ~model() { if (m_model != nullptr) Z3_model_dec_ref(ctx(), m_model); }
         operator Z3_model() const { return m_model; }
         model & operator=(model const & s) {
             Z3_model_inc_ref(s.ctx(), s.m_model);
@@ -2455,7 +2454,7 @@ namespace z3 {
         solver(context & c, solver const& src, translate): object(c) { init(Z3_solver_translate(src.ctx(), src, c)); }
         solver(solver const & s):object(s) { init(s.m_solver); }
         solver(solver && s) noexcept : object(std::forward<object>(s)), m_solver(s.m_solver) { s.m_solver = nullptr; }
-        ~solver() { if(m_solver != nullptr) Z3_solver_dec_ref(ctx(), m_solver); }
+        ~solver() { if (m_solver != nullptr) Z3_solver_dec_ref(ctx(), m_solver); }
         operator Z3_solver() const { return m_solver; }
         solver & operator=(solver const & s) {
             Z3_solver_inc_ref(s.ctx(), s.m_solver);
@@ -2673,7 +2672,7 @@ namespace z3 {
         goal(context & c, Z3_goal s):object(c) { init(s); }
         goal(goal const & s):object(s) { init(s.m_goal); }
         goal(goal && s) noexcept : object(std::forward<object>(s)), m_goal(s.m_goal) { s.m_goal = nullptr; }
-        ~goal() { if(m_goal != nullptr) Z3_goal_dec_ref(ctx(), m_goal); }
+        ~goal() { if (m_goal != nullptr) Z3_goal_dec_ref(ctx(), m_goal); }
         operator Z3_goal() const { return m_goal; }
         goal & operator=(goal const & s) {
             Z3_goal_inc_ref(s.ctx(), s.m_goal);
@@ -2738,7 +2737,7 @@ namespace z3 {
         apply_result(context & c, Z3_apply_result s):object(c) { init(s); }
         apply_result(apply_result const & s):object(s) { init(s.m_apply_result); }
         apply_result(apply_result && s) noexcept : object(std::forward<object>(s)), m_apply_result(s.m_apply_result) { s.m_apply_result = nullptr; }
-        ~apply_result() { if(m_apply_result != nullptr) Z3_apply_result_dec_ref(ctx(), m_apply_result); }
+        ~apply_result() { if (m_apply_result != nullptr) Z3_apply_result_dec_ref(ctx(), m_apply_result); }
         operator Z3_apply_result() const { return m_apply_result; }
         apply_result & operator=(apply_result const & s) {
             Z3_apply_result_inc_ref(s.ctx(), s.m_apply_result);
@@ -2770,7 +2769,7 @@ namespace z3 {
         tactic(context & c, Z3_tactic s):object(c) { init(s); }
         tactic(tactic const & s):object(s) { init(s.m_tactic); }
         tactic(tactic && s) noexcept : object(std::forward<object>(s)), m_tactic(s.m_tactic) { s.m_tactic = nullptr; }
-        ~tactic() { if(m_tactic != nullptr) Z3_tactic_dec_ref(ctx(), m_tactic); }
+        ~tactic() { if (m_tactic != nullptr) Z3_tactic_dec_ref(ctx(), m_tactic); }
         operator Z3_tactic() const { return m_tactic; }
         tactic & operator=(tactic const & s) {
             Z3_tactic_inc_ref(s.ctx(), s.m_tactic);
@@ -2864,7 +2863,7 @@ namespace z3 {
         probe(context & c, Z3_probe s):object(c) { init(s); }
         probe(probe const & s):object(s) { init(s.m_probe); }
         probe(probe && s) noexcept : object(std::forward<object>(s)), m_probe(s.m_probe) { s.m_probe = nullptr; }
-        ~probe() { if(m_probe != nullptr) Z3_probe_dec_ref(ctx(), m_probe); }
+        ~probe() { if (m_probe != nullptr) Z3_probe_dec_ref(ctx(), m_probe); }
         operator Z3_probe() const { return m_probe; }
         probe & operator=(probe const & s) {
             Z3_probe_inc_ref(s.ctx(), s.m_probe);
@@ -2973,7 +2972,7 @@ namespace z3 {
             o.m_opt = nullptr;
             return *this;
         }
-        ~optimize() { if(m_opt != nullptr) Z3_optimize_dec_ref(ctx(), m_opt); }
+        ~optimize() { if (m_opt != nullptr) Z3_optimize_dec_ref(ctx(), m_opt); }
         operator Z3_optimize() const { return m_opt; }
         void add(expr const& e) {
             assert(e.is_bool());
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -453,6 +453,7 @@ namespace z3 {
     protected:
         context * m_ctx;
     public:
+		object(context & c) : m_ctx{&c} {}
         context & ctx() const { return *m_ctx; }
         Z3_error_code check_error() const { return m_ctx->check_error(); }
         friend void check_context(object const & a, object const & b);
@@ -462,7 +463,7 @@ namespace z3 {
     class symbol : public object {
         Z3_symbol m_sym;
     public:
-        symbol(context & c, Z3_symbol s):object(c), m_sym(s) {}
+        symbol(context & c, Z3_symbol s):object{c}, m_sym(s) {}
         operator Z3_symbol() const { return m_sym; }
         Z3_symbol_kind kind() const { return Z3_get_symbol_kind(ctx(), m_sym); }
         std::string str() const { assert(kind() == Z3_STRING_SYMBOL); return Z3_get_symbol_string(ctx(), m_sym); }
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -21,6 +21,7 @@ Notes:
 #pragma once
 
 #include<cassert>
+#include<utility>
 #include<iostream>
 #include<string>
 #include<sstream>
@@ -85,6 +86,10 @@ namespace z3 {
         std::string m_msg;
     public:
         exception(char const * msg):m_msg(msg) {}
+		exception(exception const &) = default;
+		exception(exception &&) = default;
+		exception & operator=(exception const &) = default;
+		exception & operator=(exception &&) = default;
         virtual ~exception() throw() {}
         char const * msg() const { return m_msg.c_str(); }
         char const * what() const throw() { return m_msg.c_str(); }
@@ -105,11 +110,23 @@ namespace z3 {
     */
     class config {
         Z3_config    m_cfg;
-        config(config const & s);
-        config & operator=(config const & s);
+        config(config const &) = default;
+        config(config && s) noexcept : m_cfg{std::move(s.m_cfg)} {
+			s.m_cfg = nullptr;
+		}
+        config & operator=(config const &) = default;
+        config & operator=(config && s) noexcept {
+			m_cfg = std::move(s.m_cfg);
+			s.m_cfg = nullptr;
+			return *this;
+		}
     public:
         config() { m_cfg = Z3_mk_config(); }
-        ~config() { Z3_del_config(m_cfg); }
+        ~config() {
+			if (m_cfg != nullptr) {
+				Z3_del_config(m_cfg);
+			}
+		}
         operator Z3_config() const { return m_cfg; }
         /**
            \brief Set global parameter \c param with string \c value.
@@ -168,9 +185,21 @@ namespace z3 {
             Z3_set_ast_print_mode(m_ctx, Z3_PRINT_SMTLIB2_COMPLIANT);
         }
 
-
-        context(context const & s);
-        context & operator=(context const & s);
+        context(context const &) = default;
+        context(context && s) noexcept :    
+			m_enable_exceptions{ s.m_enable_exceptions },
+			rounding_mode { s.rounding_mode },
+			m_ctx { std::move(s.m_ctx) } {
+			s.m_ctx = nullptr;
+		}
+        context & operator=(context const &) = default;
+        context & operator=(context && s) noexcept {
+			m_enable_exceptions = s.m_enable_exceptions;
+			rounding_mode = s.rounding_mode;
+			m_ctx = std::move(s.m_ctx);
+			s.m_ctx = nullptr;
+			return *this;
+		}
 
         friend class scoped_context;
         context(Z3_context c) { set_context(c); }
@@ -385,6 +414,10 @@ namespace z3 {
         context m_ctx;
     public:
         scoped_context(Z3_context c): m_ctx(c) {}
+        scoped_context(Z3_context const &) = default;
+        scoped_context(Z3_context &&) noexcept = default;
+		scoped_context & operator=(Z3_context const &) = default;
+		scoped_context & operator=(Z3_context &&) noexcept = default;
         ~scoped_context() { m_ctx.detach(); }
         context& operator()() { return m_ctx; }
     };
@@ -394,13 +427,22 @@ namespace z3 {
     class array {
         T *      m_array;
         unsigned m_size;
-        array(array const & s);
-        array & operator=(array const & s);
+        array(array const &) = default
+        array(array && s) noexcept : m_array{s.m_array}, m_size{s.m_size} {
+			s.m_array = nullptr;
+		}
+        array & operator=(array const &) = default;
+        array & operator=(array && s) noexcept {
+        	m_array = s.m_array;
+			m_size = s.m_size;
+			s.m_array = nullptr;
+			return *this;
+		}
     public:
         array(unsigned sz):m_size(sz) { m_array = new T[sz]; }
         template<typename T2>
         array(ast_vector_tpl<T2> const & v);
-        ~array() { delete[] m_array; }
+        ~array() { if(m_array != nullptr) delete[] m_array; }
         void resize(unsigned sz) { delete[] m_array; m_size = sz; m_array = new T[sz]; }
         unsigned size() const { return m_size; }
         T & operator[](int i) { assert(0 <= i); assert(static_cast<unsigned>(i) < m_size); return m_array[i]; }
@@ -414,7 +456,16 @@ namespace z3 {
         context * m_ctx;
     public:
         object(context & c):m_ctx(&c) {}
-        object(object const & s):m_ctx(s.m_ctx) {}
+		object(object const &) = default;
+		object(object && s) : m_ctx{s.m_ctx} noexcept {
+			s.m_ctx = nullptr;
+		}
+		object & operator=(object const &) = default;
+		object & operator=(object && s) noexcept {
+			m_ctx = s.m_ctx;
+			s.m_ctx = nullptr;
+			return *this;
+		}
         context & ctx() const { return *m_ctx; }
         Z3_error_code check_error() const { return m_ctx->check_error(); }
         friend void check_context(object const & a, object const & b);
@@ -425,8 +476,16 @@ namespace z3 {
         Z3_symbol m_sym;
     public:
         symbol(context & c, Z3_symbol s):object(c), m_sym(s) {}
-        symbol(symbol const & s):object(s), m_sym(s.m_sym) {}
-        symbol & operator=(symbol const & s) { m_ctx = s.m_ctx; m_sym = s.m_sym; return *this; }
+        symbol(symbol const & s) = default;
+        symbol(symbol && s) noexcept : object{std::forward<object>(s)}, m_sym{std::move(s.m_sym)} {
+			m_sym = nullptr;
+		}
+        symbol & operator=(symbol const & s) = default;
+        symbol & operator=(symbol && s) noexcept {
+			object::operator=(std::forward<object>(s));
+			m_sym = s.m_sym;
+			return *this;
+		}
         operator Z3_symbol() const { return m_sym; }
         Z3_symbol_kind kind() const { return Z3_get_symbol_kind(ctx(), m_sym); }
         std::string str() const { assert(kind() == Z3_STRING_SYMBOL); return Z3_get_symbol_string(ctx(), m_sym); }
@@ -448,14 +507,22 @@ namespace z3 {
     public:
         param_descrs(context& c, Z3_param_descrs d): object(c), m_descrs(d) { Z3_param_descrs_inc_ref(c, d); }
         param_descrs(param_descrs const& o): object(o.ctx()), m_descrs(o.m_descrs) { Z3_param_descrs_inc_ref(ctx(), m_descrs); }
+        param_descrs(param_descrs && o) noexcept : object{std::forward<object>(o)}, m_descrs{o.m_descrs} {
+			o.m_descrs = nullptr;
+		}
         param_descrs& operator=(param_descrs const& o) {
             Z3_param_descrs_inc_ref(o.ctx(), o.m_descrs);
             Z3_param_descrs_dec_ref(ctx(), m_descrs);
             m_descrs = o.m_descrs;
-            m_ctx = o.m_ctx;
+			object::operator=(o);
             return *this;
         }
-        ~param_descrs() { Z3_param_descrs_dec_ref(ctx(), m_descrs); }
+        param_descrs& operator=(param_descrs && o) noexcept {
+			object::operator=(std::forward<object>(o));
+		 	m_descrs = o.m_descrs;
+			o.m_descrs = nullptr;
+		}
+        ~param_descrs() { if(m_descrs != nullptr) Z3_param_descrs_dec_ref(ctx(), m_descrs); }
         static param_descrs simplify_param_descrs(context& c) { return param_descrs(c, Z3_simplify_get_param_descrs(c)); }
 
         unsigned size() { return Z3_param_descrs_size(ctx(), m_descrs); }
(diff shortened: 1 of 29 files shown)


src/api/c++/z3++.h
```diff
@@ -85,6 +86,10 @@ namespace z3 {
         std::string m_msg;
     public:
         exception(char const * msg):m_msg(msg) {}
```
zwimer, line 88: Rule of 5; the destructor was declared

src/api/c++/z3++.h
```diff
@@ -105,11 +110,23 @@ namespace z3 {
     */
     class config {
         Z3_config    m_cfg;
-        config(config const & s);
```
zwimer, line 108: These are just declarations, not definitions. This could lead to linker errors and other issues. Added `=default` so that the compiler will include a definition if it was silently implicitly deleted by something else (for example due to a rule of 5 violation)

src/api/c++/z3++.h
```diff
@@ -448,14 +506,23 @@ namespace z3 {
...
         param_descrs(context& c, Z3_param_descrs d): object(c), m_descrs(d) { Z3_param_descrs_inc_ref(c, d); }
         param_descrs(param_descrs const& o): object(o.ctx()), m_descrs(o.m_descrs) { Z3_param_descrs_inc_ref(ctx(), m_descrs); }
+        param_descrs(param_descrs && o) noexcept : object{std::forward<object>(o)}, m_descrs{o.m_descrs} {
+			o.m_descrs = nullptr;
+		}
         param_descrs& operator=(param_descrs const& o) {
             Z3_param_descrs_inc_ref(o.ctx(), o.m_descrs);
             Z3_param_descrs_dec_ref(ctx(), m_descrs);
             m_descrs = o.m_descrs;
-            m_ctx = o.m_ctx;
```
zwimer, line 455: Instead of manually copying over data from parent classes, invoke the parent classes' copy operator. This is less brittle and reduces the number of hassles in case the parent class is ever altered.

src/api/c++/z3++.h
```diff
@@ -559,32 +643,30 @@ namespace z3 {
         }
         */
 
-        class iterator {
+        class iterator final {
```
zwimer, line 646: Since `iterator` will not be extended, we can declare it `final`. This has optimization benefits.

src/api/c++/z3++.h
```diff
@@ -559,32 +643,30 @@ namespace z3 {
...
             ast_vector_tpl const* m_vector;
             unsigned m_index;
         public:
-            iterator(ast_vector_tpl const* v, unsigned i): m_vector(v), m_index(i) {}
-            iterator(iterator const& other): m_vector(other.m_vector), m_index(other.m_index) {}
-            iterator operator=(iterator const& other) { m_vector = other.m_vector; m_index = other.m_index; return *this; }
+            iterator(ast_vector_tpl const* v, unsigned i) noexcept : m_vector(v), m_index(i) {}
 
-            bool operator==(iterator const& other) const {
+            bool operator==(iterator const& other) const noexcept {
```
zwimer, line 652: This should never raise an exception, we optimize by marking the function as `noexcept`; this also allows this function/operator to be used safely in `noexcept` contexts

src/api/c++/z3++.h
```diff
@@ -593,23 +675,18 @@ namespace z3 {
     /**
        \brief A Z3 sort (aka type). Every expression (i.e., formula or term) in Z3 has a sort.
     */
-    class sort : public ast {
+    class sort final : public ast {
     public:
         sort(context & c):ast(c) {}
         sort(context & c, Z3_sort s):ast(c, reinterpret_cast<Z3_ast>(s)) {}
         sort(context & c, Z3_ast a):ast(c, a) {}
-        sort(sort const & s):ast(s) {}
```
zwimer, line 601: This, and the operator below, constitute a rule of 5 violation when alone. Since they are both 'default' (i.e. do nothing differently than if they were implicitly generated), we just delete them. The compiler will implicitly create these methods and since this fixes the rule of 5 violation, the other impilict methods will be safe to use.

src/api/c++/z3++.h
```diff
@@ -593,23 +675,18 @@ namespace z3 {
...
 
         /**
            \brief retrieve unique identifier for func_decl.
          */
         unsigned id() const { unsigned r = Z3_get_sort_id(ctx(), *this); check_error(); return r; }
 
-        /**
-           \brief Assign sort s to this
-        */
-        sort & operator=(sort const & s) { return static_cast<sort&>(ast::operator=(s)); }
```
zwimer, line 612: Returning a `static_cast` of the parent's operator is unnecessary compared to `return *this`. Any compiler should be able to able to optimize away the `*this` operation.

More important, this may introduce undefined behavior. Specifically, there is no guarantee that the parent `=` operator will return `this`; it is done by convention but not required. In case the operator returns something else, we are now statically downcasting an unknown pointer to a type it may not be, and one which we may not own.

src/api/c++/z3++.h
```diff
@@ -2830,19 +2966,21 @@ namespace z3 {
...
     public:
-        class handle {
+        class handle final {
             unsigned m_h;
         public:
             handle(unsigned h): m_h(h) {}
             unsigned h() const { return m_h; }
         };
         optimize(context& c):object(c) { m_opt = Z3_mk_optimize(c); Z3_optimize_inc_ref(c, m_opt); }
-        optimize(optimize& o):object(o)  {
```
zwimer, line 2840: `m_opt` can be initialized rather than set after initial construction, so it was move it before the body of the constructor.

zwimer: I suspect there should also be a dec_ref here for m_opt though. If that is the case I can add it in properly; this appears to be a referencing counting issue here.

src/api/c++/z3++.h
```diff
@@ -2935,7 +3079,26 @@ namespace z3 {
...
     public:
         fixedpoint(context& c):object(c) { m_fp = Z3_mk_fixedpoint(c); Z3_fixedpoint_inc_ref(c, m_fp); }
-        ~fixedpoint() { Z3_fixedpoint_dec_ref(ctx(), m_fp); }
+		fixedpoint(fixedpoint const & o) : object{o}, m_fp{o.m_fp} {
+        	Z3_fixedpoint_inc_ref(ctx(), m_fp);
+		}
+		fixedpoint(fixedpoint && o) noexcept : object{std::forward<object>(o)}, m_fp{o.m_fp} {
+			o.m_fp = nullptr;
+		}
+		fixedpoint & operator=(fixedpoint const & o) {
```
zwimer, line 3088: These were implicitly generated before and did not do reference counting as a result. I assume that was a bug.

src/api/c++/z3++.h
```diff
@@ -2830,19 +2966,21 @@ namespace z3 {
...
-        class handle {
+        class handle final {
             unsigned m_h;
         public:
             handle(unsigned h): m_h(h) {}
             unsigned h() const { return m_h; }
         };
         optimize(context& c):object(c) { m_opt = Z3_mk_optimize(c); Z3_optimize_inc_ref(c, m_opt); }
-        optimize(optimize& o):object(o)  {
+        optimize(optimize const& o) : object{o}, m_opt{o.m_opt}  {
```
zwimer, line 2976: Made the copy construct take in the argument as `const`; I assume this was a typo from before

zwimer:
For each notable type of change I made, I provided a comment on adjacent to a line in the diff showing the change.

src/api/c++/z3++.h
```diff
@@ -2935,7 +3079,26 @@ namespace z3 {
         Z3_fixedpoint m_fp;
     public:
         fixedpoint(context& c):object(c) { m_fp = Z3_mk_fixedpoint(c); Z3_fixedpoint_inc_ref(c, m_fp); }
-        ~fixedpoint() { Z3_fixedpoint_dec_ref(ctx(), m_fp); }
+        fixedpoint(fixedpoint const & o) : object{o}, m_fp{o.m_fp} {
+            Z3_fixedpoint_inc_ref(ctx(), m_fp);
+        }
+        fixedpoint(fixedpoint && o) noexcept : object{std::forward<object>(o)}, m_fp{o.m_fp} {
```
zwimer, line 3085: These were implicitly generated before and did not do reference counting as a result. I assume that was a bug.

src/api/c++/z3++.h
```diff
@@ -529,9 +606,10 @@ namespace z3 {
         ast_vector_tpl(context & c):object(c) { init(Z3_mk_ast_vector(c)); }
         ast_vector_tpl(context & c, Z3_ast_vector v):object(c) { init(v); }
         ast_vector_tpl(ast_vector_tpl const & s):object(s), m_vector(s.m_vector) { Z3_ast_vector_inc_ref(ctx(), m_vector); }
+        ast_vector_tpl(ast_vector_tpl && s) noexcept : object{std::forward<object>(s)}, m_vector{s.m_vector} { s.m_vector = nullptr; }
         ast_vector_tpl(context& c, ast_vector_tpl const& src): object(c) { init(Z3_ast_vector_translate(src.ctx(), src, c)); }
 
-        ~ast_vector_tpl() { Z3_ast_vector_dec_ref(ctx(), m_vector); }
+        ~ast_vector_tpl() { if(m_vector != nullptr) Z3_ast_vector_dec_ref(ctx(), m_vector); }
```
NikolajBjorner, line 612: this is actually a bug in api_ast_vector.cpp. It should just check for null

src/api/c++/z3++.h
```diff
@@ -743,12 +818,10 @@ namespace z3 {
        \brief A Z3 expression is used to represent formulas and terms. For Z3, a formula is any expression of sort Boolean.
        Every expression has a sort.
     */
-    class expr : public ast {
+    class expr final : public ast {
```
NikolajBjorner, line 821: could a user of the library complain that they can't extend this?

src/api/c++/z3++.h
```diff
@@ -2255,14 +2342,21 @@ namespace z3 {
         model(context & c):object(c) { init(Z3_mk_model(c)); }
         model(context & c, Z3_model m):object(c) { init(m); }
         model(model const & s):object(s) { init(s.m_model); }
+        model(model && s) noexcept : object{std::forward<object>(s)}, m_model{s.m_model} { s.m_model = nullptr; }
         model(model& src, context& dst, translate) : object(dst) { init(Z3_model_translate(src.ctx(), src, dst)); }
-        ~model() { Z3_model_dec_ref(ctx(), m_model); }
+        ~model() { if(m_model != nullptr) Z3_model_dec_ref(ctx(), m_model); }
```
NikolajBjorner, line 2347: Z3_model_dec_ref also works with null

src/api/c++/z3++.h
NikolajBjorner: yes, it was a bug

NikolajBjorner:
So clients should use at least C++11 or higher?
We have been compiling C++ examples with down-level versions.
But this changes too, just with some delay our overall dependencies.

src/api/c++/z3++.h
zwimer: Indeed. I could make this either a build option or remove it. It is an optimization _if_ `expr` is not going to be extended. Would you prefer me remove it or make a macro called something like `MAYBE_FINAL` that the user can define before including this file?

src/api/c++/z3++.h
zwimer: I'll take a look at which `dec_ref` functions work with `nullptr`. _Should_ they all? I.e. any that do not, should I change the `dec_ref` source code? Or are they all correct, except for the one in `api_ast_vector.cpp` as you said above?

zwimer:
Correct. Move semantics require C++11 or later, as does `noexcept` and `final`. I figured it would be since it is a decade old and pretty much every compiler on every supported version of every operating system supports it by now.

If backwards compatibility is a _must_ we could litter the code with `#ifdef CPP11` statements and replace `noexcept` with a macro `MAYBE_NOEXCEPT` or something like that; then the user could set the macro before including; but personally I see this as an ugly solution.

What would need to be done to move the API into using C++11 / more modern versions of C++?

NikolajBjorner:
It should now be fine to insist on new clients use at least C++11. 
Littering code with ifdefs is a non-starter.

src/api/c++/z3++.h
zwimer: @NikolajBjorner should this constructor have a `dec_ref` in it as well? I assume this is a bug and not intentional? I can fix it in this PR if so.

NikolajBjorner: It shouldn't have dec-ref. The reference counting discipline is similar to the "ast" object.

src/api/c++/z3++.h
```diff
@@ -3676,42 +3843,42 @@ namespace z3 {
...
            \brief register a callback on final-check.
            During the final check stage, all propagations have been processed.
            This is an opportunity for the user-propagator to delay some analysis
            that could be expensive to perform incrementally. It is also an opportunity
-           for the propagator to implement branch and bound optimization. 
+           for the propagator to implement branch and bound optimization.
         */
 
-        void final(final_eh_t& f) { 
+        void final_(final_eh_t& f) {
```
zwimer, line 3869: This function was named `final`, which is a C++11+ keyword. I renamed it `final_` so there are no issues.

zwimer:
After looking at the code, I think it makes sense to leave the null-check in the API (it could also be in the `dec_ref` as well)  for  two main reasons:
1. With move semantics, this will be a fairly common case. For example, consider moving an `expr` onto the heap then wrapping it in an `std::shared_ptr`; or even something as simple as `vec.emplace_back(std::move( - expr ))`. Or map emplacement / insertion, etc. If the null-check is only in the `dec_ref`, I noticed there were quite a few function calls before the if. If the `if(p)` can be moved to encompass the entire function then that is another story, but as is, it is within a try-catch block, which already is a few CPU cycles and memory cost, not to mention the other statements preceding it.
2. Clarity. If it isn't documented that each of the `dec_ref` statements can accept a null, I think it is just clearer that there are no errors like this. If it is documented, or will be, then this is also moot.

I don't have terribly strong opinions though, that's just my two cents. Let me know what you prefer and I can update the PR.

NikolajBjorner:
build failure due to regression pushed to master. Need to retrigger builds against updated master

NikolajBjorner:
OK to leave nullptr check in API, but convention for AST object is to allow nullptr at the C API level. So it makes sense to be consistent for C API. There are other consumers at this level.

NikolajBjorner:
@nunoplopes OK to merge?

zwimer:
I believe so. There are improvements that can be made in the future, like adding `emplace` methods to things like solver rather than just `add`; but I think those are distinct future PRs.

src/api/c++/z3++.h
nunoplopes: these definitions are useless

src/api/c++/z3++.h
nunoplopes: these changes are wrong. if anything, they would be marked with = delete. That's the way of doing things

src/api/c++/z3++.h
```diff
@@ -168,16 +185,26 @@ namespace z3 {
             Z3_set_ast_print_mode(m_ctx, Z3_PRINT_SMTLIB2_COMPLIANT);
         }
 
-
-        context(context const & s);
-        context & operator=(context const & s);
+        context(context const &) = default;
```
nunoplopes, line 188: = delete.

src/api/c++/z3++.h
```diff
@@ -385,6 +412,10 @@ namespace z3 {
         context m_ctx;
     public:
         scoped_context(Z3_context c): m_ctx(c) {}
+        scoped_context(scoped_context const &) = default;
```
nunoplopes, line 415: these are useless

src/api/c++/z3++.h
```diff
@@ -394,13 +425,22 @@ namespace z3 {
     class array {
         T *      m_array;
         unsigned m_size;
-        array(array const & s);
-        array & operator=(array const & s);
+        array(array const &) = default;
```
nunoplopes, line 428: =delete

src/api/c++/z3++.h
```diff
@@ -394,13 +425,22 @@ namespace z3 {
...
+        array & operator=(array && s) noexcept {
+            m_array = s.m_array;
+            m_size = s.m_size;
+            s.m_array = nullptr;
+            return *this;
+        }
         template<typename T2>
         array(ast_vector_tpl<T2> const & v);
-        ~array() { delete[] m_array; }
+        ~array() { if(m_array != nullptr) delete[] m_array; }
```
nunoplopes, line 443: wrong. delete can take a nullptr

src/api/c++/z3++.h
```diff
@@ -414,7 +454,16 @@ namespace z3 {
         context * m_ctx;
     public:
         object(context & c):m_ctx(&c) {}
-        object(object const & s):m_ctx(s.m_ctx) {}
+        object(object const &) = default;
+        object(object && s) noexcept : m_ctx{s.m_ctx} {
```
nunoplopes, line 458: this change and the assignment operator are useless.

src/api/c++/z3++.h
```diff
@@ -425,8 +474,17 @@ namespace z3 {
         Z3_symbol m_sym;
     public:
         symbol(context & c, Z3_symbol s):object(c), m_sym(s) {}
-        symbol(symbol const & s):object(s), m_sym(s.m_sym) {}
-        symbol & operator=(symbol const & s) { m_ctx = s.m_ctx; m_sym = s.m_sym; return *this; }
+        symbol(symbol const & s) = default;
+        symbol(symbol && s) noexcept : object{std::forward<object>(s)}, m_sym{std::move(s.m_sym)} {
```
nunoplopes, line 478: useless changes

src/api/c++/z3++.h
```diff
@@ -448,14 +506,23 @@ namespace z3 {
     public:
         param_descrs(context& c, Z3_param_descrs d): object(c), m_descrs(d) { Z3_param_descrs_inc_ref(c, d); }
         param_descrs(param_descrs const& o): object(o.ctx()), m_descrs(o.m_descrs) { Z3_param_descrs_inc_ref(ctx(), m_descrs); }
+        param_descrs(param_descrs && o) noexcept : object{std::forward<object>(o)}, m_descrs{o.m_descrs} {
```
nunoplopes, line 509: the object bit is redundant

src/api/c++/z3++.h
```diff
@@ -448,14 +506,23 @@ namespace z3 {
...
             Z3_param_descrs_dec_ref(ctx(), m_descrs);
             m_descrs = o.m_descrs;
-            m_ctx = o.m_ctx;
+            object::operator=(o);
             return *this;
         }
-        ~param_descrs() { Z3_param_descrs_dec_ref(ctx(), m_descrs); }
+        param_descrs& operator=(param_descrs && o) noexcept {
+            object::operator=(std::forward<object>(o));
+             m_descrs = o.m_descrs;
```
nunoplopes, line 521: bad ident

src/api/c++/z3++.h
```diff
@@ -448,14 +506,23 @@ namespace z3 {
...
             Z3_param_descrs_inc_ref(o.ctx(), o.m_descrs);
             Z3_param_descrs_dec_ref(ctx(), m_descrs);
             m_descrs = o.m_descrs;
-            m_ctx = o.m_ctx;
+            object::operator=(o);
             return *this;
         }
-        ~param_descrs() { Z3_param_descrs_dec_ref(ctx(), m_descrs); }
+        param_descrs& operator=(param_descrs && o) noexcept {
+            object::operator=(std::forward<object>(o));
```
nunoplopes, line 520: std::move

src/api/c++/z3++.h
```diff
@@ -472,13 +539,20 @@ namespace z3 {
     public:
         params(context & c):object(c) { m_params = Z3_mk_params(c); Z3_params_inc_ref(ctx(), m_params); }
         params(params const & s):object(s), m_params(s.m_params) { Z3_params_inc_ref(ctx(), m_params); }
-        ~params() { Z3_params_dec_ref(ctx(), m_params); }
+        params(params && s) noexcept : object{std::forward<object>(s)}, m_params{s.m_params} { s.m_params = nullptr; }
```
nunoplopes, line 542: the object bit is redundant

src/api/c++/z3++.h
```diff
@@ -1259,7 +1332,7 @@ namespace z3 {
         friend expr bvneg_no_overflow(expr const& a);
         friend expr bvmul_no_overflow(expr const& a, expr const& b, bool is_signed);
         friend expr bvmul_no_underflow(expr const& a, expr const& b);
-
```
nunoplopes, line 1262: these changes just annoy the reviewer. I've spotted the review here. Too long.

nunoplopes (changes requested):

src/api/c++/z3++.h
zwimer: Adding noexcept changes the function signature / where it can be use safely, and change behavior as they affect whether an exception will be propogated, or the program terminated; it is also an optimization.

The `object{std::forward<object>(s)}` change was just, in my opinion, good practice as they allow easier maintainability in case the superclass changes at all; but they are functionally identical.

src/api/c++/z3++.h
zwimer: They allow different methods of catching the exception. Since the destructor was declared, that prevents the move operator and constructor from being implicitly generated, thus we default them. But since we do, that prevents the compiler from generating the copy operator / constructor. This is known as the rule of 5. Since the message is a string and member variable, it will benefit from having move construction / assignment.

src/api/c++/z3++.h
zwimer: In the original source they were:
```
        config(config const & s);
        config & operator=(config const & s);
```
Are you saying that's a bug?

src/api/c++/z3++.h
zwimer: This is a rule of 5 violation. Since a member variable is a `context`, i.e. not a singular primitive, not including them would lead to the copy operator / constructor to not be created, due to the explicit definitions of the corresponding move functions.

src/api/c++/z3++.h
zwimer: Yup, I'll fix that.

src/api/c++/z3++.h
zwimer: The noexcept is functional, but the `= default` is mostly just a coding practice to allow easier extension / editing in the future; they shouldn't really lead to code changes.

src/api/c++/z3++.h
zwimer: What do you mean? The word? That is just the template argument the standard library's `std::forward` requires, since we are only forwarding the `object` piece of the class.

src/api/c++/z3++.h
zwimer: Fixing now

src/api/c++/z3++.h
zwimer: `std::forward`'s advantage here is that it also explicitly up-casts, but functionally that should be the same; this is just not-implict. If it's a sticking point though I can change it, I just prefer explicit casting to implicit casting as a convention.

src/api/c++/z3++.h
zwimer: Responded to in a different comment with the same message.

src/api/c++/z3++.h
zwimer: Looking over the code it does seem that these should be deleted to avoid the reference counting error that would occur via double deletion. I will change them to `= delete;`, both the copy constructor and assignment operator.

zwimer: Oh, I see, they were declared private. Got it. I will rectify this in all related comments.

src/api/c++/z3++.h
nunoplopes: See rule of zero here: https://en.cppreference.com/w/cpp/language/rule_of_three
More code is not better code.

src/api/c++/z3++.h
nunoplopes: more is not always more.

nunoplopes: let me expand: I've never seen anyone writing code like this nor I'm aware of any coding standard that suggests this.

src/api/c++/z3++.h
zwimer: Yeah, I'll make sure to turn off my trailing whitespace zapper next time; sorry about that.

src/api/c++/z3++.h
zwimer: Although now that `context` is marked as delete, these probably should be as well. I'll update them now.

src/api/c++/z3++.h
zwimer: Ah, indeed there is no destructor, I can remove them all then since there is no need to set `m_ctx` to null then. Good catch.

src/api/c++/z3++.h
zwimer: `noexcept` is recommended by isocpp for move semantics, as it allows moves in noexcept contexts. If you mean `std::move` though I can change it. I typically use it for forwarding arguments of different types, since that is the difference between it and move.

zwimer: Actually, I think the rule of zero applies here as well, so I will update this.

nunoplopes:
@zwimer If you want me to continue to review this patch you have to undo the whitespace changes. The patch is too big.
it's a good practice to save reviewers' time when you want a patch to get in.

zwimer:
I removed the commit that deleted trailing whitespace to make the diff easier @nunoplopes

src/api/c++/z3++.h
nunoplopes: the destructor can be removed; it's already declared in std::exception

src/api/c++/z3++.h
```diff
@@ -274,7 +301,7 @@ namespace z3 {
         /**
            \brief Return a floating point sort.
            \c ebits is a number of exponent bits,
-           \c sbits	is a number of significand bits,
+           \c sbits    is a number of significand bits,
```
nunoplopes, line 304: random change

src/api/c++/z3++.h
```diff
@@ -448,14 +485,23 @@ namespace z3 {
     public:
         param_descrs(context& c, Z3_param_descrs d): object(c), m_descrs(d) { Z3_param_descrs_inc_ref(c, d); }
         param_descrs(param_descrs const& o): object(o.ctx()), m_descrs(o.m_descrs) { Z3_param_descrs_inc_ref(ctx(), m_descrs); }
+        param_descrs(param_descrs && o) noexcept : object{std::forward<object>(o)}, m_descrs{o.m_descrs} {
```
nunoplopes, line 488: std::move is so much more concise. forward is for templates

src/api/c++/z3++.h
nunoplopes: I don't see any advantage. It takes more bytes and gives no benefit.

src/api/c++/z3++.h
```diff
@@ -3615,13 +3757,17 @@ namespace z3 {
...
 
-        struct scoped_cb {
+        struct scoped_cb final {
             user_propagator_base* p;
             scoped_cb(void* _p, Z3_solver_callback cb):p(static_cast<user_propagator_base*>(_p)) {
                 p->cb = cb;
             }
-            ~scoped_cb() { 
-                p->cb = nullptr; 
+            scoped_cb(scoped_cb const &) = default;
```
nunoplopes, line 3765: this is a private class. doesn't need any change.

src/api/c++/z3++.h
nunoplopes: let's maybe call it `register_final`

src/api/c++/z3++.h
```diff
@@ -394,10 +425,19 @@ namespace z3 {
     class array {
         T *      m_array;
```
nunoplopes, line 426: we can maybe change this to a unique_ptr and let C++ do its magic.

src/api/c++/z3++.h
zwimer: Ah, that was a tab before, I'll make it one space.

src/api/c++/z3++.h
zwimer: Makes sense. Changing.

src/api/c++/z3++.h
zwimer: I agree but I think we can do better.

Is there a reason we can't just replace this class with `std::vector`?  It seems to do the same thing (well, the resize function looses its daya when resizing so, but maybe that is a bug?). The only thing is that this has a custom constructor for `ast_vector_tpl`, but that is either something we could switch to a custom constructor for vector (sort of like how std encourages people to make custom `std::hash` functors), or we could just make it into a function. I currently do not see the benefit of using `array` over `std::vector`.

src/api/c++/z3++.h
zwimer: Normally, but in this case we are not moving the entire instance. Applying move semantics to the entire thing leads to the implication that the entire instance is to-be consumed. But here we want only the `object` part of the instance to be consumed; in other words the cast is a promise that only the desired components of the instance will be consumed rather than all of it which move semantics traditionally imply.

src/api/c++/z3++.h
nunoplopes: agreed. I checked the users and they are all super simple. Actually many of them can be replaced with std::vector. Can we keep this class alone (ie, revert changes), and then do another PR for array pls?

src/api/c++/z3++.h
```diff
@@ -385,6 +412,10 @@ namespace z3 {
         context m_ctx;
     public:
         scoped_context(Z3_context c): m_ctx(c) {}
+        scoped_context(scoped_context const &) = delete;
```
nunoplopes, line 415: These are not needed. context already forbids copying. And the move constructors aren't useful. No one will move this thing.

src/api/c++/z3++.h
zwimer: Ok. After this PR I will make a separate PR replacing classes that could be replaced by stl classes like `std::vector`

src/api/c++/z3++.h
zwimer: Ok, I will apply the rule of 0 then.

src/api/c++/z3++.h
zwimer: I have the copy functions in just because of the rule of 5, even if we happen to be doing what it would be implicit in the first place. As for move semantics, this is an API, I see no reason to disallow a feature that has almost no downsides to permitting.

src/api/c++/z3++.h
zwimer: I'm resolving this and the other issues that are duplicates of the one above to consolidate the conversation.

src/api/c++/z3++.h
zwimer: Ok

src/api/c++/z3++.h
nunoplopes: I take back what I said. This is an internal class and using std::vector is not better. Thus adding move constructors & etc is not needed.
If anything, I would switched this to a unique_ptr.

src/api/c++/z3++.h
nunoplopes: std::move doesn't move. There's no implication.
std::forward makes the code harder to maintain. If you change the base class you need to update all forwards.

src/api/c++/z3++.h
```diff
@@ -3615,13 +3752,13 @@ namespace z3 {
...
-        struct scoped_cb {
+        struct scoped_cb final {
             user_propagator_base* p;
             scoped_cb(void* _p, Z3_solver_callback cb):p(static_cast<user_propagator_base*>(_p)) {
                 p->cb = cb;
             }
-            ~scoped_cb() { 
-                p->cb = nullptr; 
+            ~scoped_cb() {
+                if (p != nullptr) p->cb = nullptr;
```
nunoplopes, line 3761: this change should be reverted

zwimer: Is p guaranteed never to be `nullptr`? If so then I agree, if not this is needed.

src/api/c++/z3++.h
zwimer: I might have been unclear: `std::move` casts to a r-value reference; the implication of passing a move reference is that the function accepting it will consume it. As for the base class, if it changes then this line must be updated anyway since `object{` would no longer be valid.

src/api/c++/z3++.h
NikolajBjorner: changed to &

src/api/c++/z3++.h
zwimer: Interesting. Can you expand on why `std;:vector` is not preferred? I seem to be lacking the knowledge as to why it being internal means we want `class array` as opposed to just `std::vector` or `template <typename T> using array = std::vector<T>;` if the desire is to maintain the name 'array'.

Somewhat related, but if we want to keep array, is the resize function buggy? It does resize, but it does not keep the data it had during the resize.

src/api/c++/z3++.h
zwimer: Fixed by commit #54f04a5

src/api/c++/z3++.h
zwimer: If nothing else, why not `class array{ std::vector<T> arr; /* ... */ };` or something like that? What is the benefit of maintaining this class as is (or with the unique pointer).

src/api/c++/z3++.h
nunoplopes: std::move doesn't move. There are no implications.

src/api/c++/z3++.h
nunoplopes: std::vector wastes space vs a simple array. Plus given that this is an internal API, there's no need for move constructors and etc.
If anything, change this to unique_ptr, otherwise leave it as-is.

src/api/c++/z3++.h
nunoplopes: this is an internal class. there's no need for this stuff

zwimer: I see. If we don't care about extensibility of the superclass I can delete them.

src/api/c++/z3++.h
zwimer: It might not explicitly move, but it has strong implications. Consider `foo(std::move(bar))`. This should imply that `foo` consumes `bar`. If not, `foo` should take `bar` in by reference, pointer, or value. Except with forwarding const references in templates, where you don't use `std::move` anyway, r-value references' primary use is moving an item. `std::move` doesn't move, but it is designed to indicate something can be move; it is an explicit cast that says 'this object may be moved', otherwise it wouldn't be called `std::move`. In most contexts, `foo(std::move(x))` implies that `foo` consumes `x`, because otherwise you shouldn't invoke `std::move`. Even here in the constructor that is what we are doing, `object`'s constructor consumes the `object` portion of the instance we are passing it. Substituting `std::move` is an example, if we have `object{std::move(s)}` object is clearly consuming `s`, that is why we use `std::move`. Thus things wrapped in `std::move`, after passed to a function, should be considered 'already moved / altered'. The difference here is that with `std::forward<object>(s)` we are not saying 'consume s', but rather 'consume the object part of s'. In other words, 'the non-object part of s will not be touched by the function and we can safely use the rest of `s` after this function call'.

src/api/c++/z3++.h
zwimer: I suppose `class array` indeed does not contain a size. It doesn't seem to be used as a member variable, so I'm not sure that justifies this over a vector. But that would be another PR if it is worth it. I will make this one a `unique_ptr` for now.

NikolajBjorner:
what to do?
There are many reasonable fixes and even bug fixes, but even after the review some more controversial changes.
How about a small pull request with non-controversial changes:

- private constructors without definitions -> add the "delete" keyword.
- final -> register_final
- fix reference counting bug in copy constructor for optimize

Some left over { } instead of ( ) in constructor lists are still dangling around. 
Maybe it is legal syntax, but it isn't consistent with conventions used everywhere else in the code base. 
The move semantics updates can go separately.

zwimer:
I'll make a PR that does:
1. private constructors without definitions -> add the "delete" keyword.
2. final -> register_final
3. fix reference counting bug in copy constructor for optimize
Using the code from this PR. I will then update this PR to remove initializer list syntax for anything that doesn't invoke the big pars problem, which in this case is probably everywhere. When I make the PR, I will post a link below

NikolajBjorner:
Great. You put in a lot of effort. Not all updates may end up being taken, but it is better to make sure that the clear cases get in so your work isn't lost to the changes that could be punted.

zwimer:
We can put this PR on hold for now. I am making a series of small PRs each containing some of the changes from this. Once those are done we can come back to this one which, at that point, should mostly be move semantics as originally intended.

zwimer:
For `array` and `unique_ptr`: https://github.com/Z3Prover/z3/pull/5171

zwimer:
final: https://github.com/Z3Prover/z3/pull/5172

zwimer:
Bug fixes: https://github.com/Z3Prover/z3/pull/5173

src/api/c++/z3++.h
```diff
@@ -168,16 +180,26 @@ namespace z3 {
             Z3_set_ast_print_mode(m_ctx, Z3_PRINT_SMTLIB2_COMPLIANT);
         }
 
-
-        context(context const & s);
-        context & operator=(context const & s);
+        context(context const &) = delete;
```
NikolajBjorner, line 183: using = delete is a good change.
This could go into separate pull request to merge.

zwimer:
Deletion of constructors: https://github.com/Z3Prover/z3/pull/5174

src/api/c++/z3++.h
```diff
@@ -2199,13 +2239,20 @@ namespace z3 {
     public:
         func_entry(context & c, Z3_func_entry e):object(c) { init(e); }
         func_entry(func_entry const & s):object(s) { init(s.m_entry); }
-        ~func_entry() { Z3_func_entry_dec_ref(ctx(), m_entry); }
+        func_entry(func_entry && s) noexcept : object(std::forward<object>(s)), m_entry{s.m_entry} { s.m_entry = nullptr; }
+        ~func_entry() { if(m_entry != nullptr) Z3_func_entry_dec_ref(ctx(), m_entry); }
         operator Z3_func_entry() const { return m_entry; }
-        func_entry & operator=(func_entry const & s) {
+        func_entry & operator=(func_entry const & s) noexcept {
```
NikolajBjorner, line 2245: noexcept could also go into a separate pull request

src/api/c++/z3++.h
```diff
@@ -2935,7 +3046,26 @@ namespace z3 {
...
     public:
         fixedpoint(context& c):object(c) { m_fp = Z3_mk_fixedpoint(c); Z3_fixedpoint_inc_ref(c, m_fp); }
-        ~fixedpoint() { Z3_fixedpoint_dec_ref(ctx(), m_fp); }
+        fixedpoint(fixedpoint const & o) : object{o}, m_fp{o.m_fp} {
+            Z3_fixedpoint_inc_ref(ctx(), m_fp);
+        }
+        fixedpoint(fixedpoint && o) noexcept : object{std::forward<object>(o)}, m_fp{o.m_fp} {
+            o.m_fp = nullptr;
+        }
+        fixedpoint & operator=(fixedpoint const & o) {
```
NikolajBjorner, line 3055: It is definitely a miss that = isn't provided and not declared as del

src/api/c++/z3++.h
```diff
@@ -2935,7 +3046,26 @@ namespace z3 {
...
+            Z3_fixedpoint_inc_ref(o.ctx(), o.m_fp);
+            Z3_fixedpoint_inc_ref(ctx(), m_fp);
```
NikolajBjorner, line 3057: Z3_fixedpoint_dec_ref

zwimer:
Other improvements made by this PR: https://github.com/Z3Prover/z3/pull/5175
I.e. `noexcept`, `final`, and deleting a few copy functions that would be generated implicitly, but if defined explicitly have side effects.

src/api/c++/z3++.h
zwimer: I fixed that in the other PR, I will merge them into this soon.

src/api/c++/z3++.h
zwimer: I am failing to read that sentence. What do you mean?

zwimer:
I can also make a PR for copy operator= to use `operator=` instead of manually copying data. That should make this PR primarily about move semantics and rule of 5 things.

src/api/c++/z3++.h
```diff
@@ -2847,7 +2967,13 @@ namespace z3 {
...
             return *this;
         }
-        ~optimize() { Z3_optimize_dec_ref(ctx(), m_opt); }
+        optimize& operator=(optimize && o) noexcept {
+            object::operator=(std::forward<object>(o));
+            m_opt = o.m_opt;
+            o.m_opt = nullptr;
+            return *this;
+        }
+        ~optimize() { if(m_opt != nullptr) Z3_optimize_dec_ref(ctx(), m_opt); }
```
NikolajBjorner, line 2976: convention used elsewhere (unfortunately we don't have a way to enforce conventions): "if (", not "if("

zwimer: I'll fix it. If it matters `clang-format` is pretty easy to set up if you want linting / convention enforcement. `cpp-check` as well.

src/api/c++/z3++.h
```diff
@@ -567,8 +623,7 @@ namespace z3 {
             ast_vector_tpl const* m_vector;
             unsigned m_index;
         public:
-            iterator(ast_vector_tpl const* v, unsigned i): m_vector(v), m_index(i) {}
-            iterator(iterator const& other): m_vector(other.m_vector), m_index(other.m_index) {}
+            iterator(ast_vector_tpl const* v, unsigned i) noexcept : m_vector(v), m_index(i) {}
```
zwimer, line 625: @NikolajBjorner I left this constructor in, but constructors like this are not needed. C++11 introduces initializer lists which have a few benefits. For example, `A a{1, 2}` would correctly initialize `struct A { int a; int b };` The compiler will implicitly generate a constructor with `noexcept` and everything it can. Having explicit constructors for pure data copying like this is actually against cpp guidelines. That said if we are upping to C++11 just now, people using the older `()` syntax for constructors might run into a harder time; that's why I left it in. If you have any objections to this though since it is bad practice let me know.

zwimer:
Ok, merges and updated post-merges are complete.

src/api/c++/z3++.h
```diff
@@ -82,10 +83,9 @@ namespace z3 {
     /**
        \brief Exception used to sign API usage errors.
     */
-    class exception : public std::exception {
+    class exception final : public std::exception {
```
zwimer, line 86: We can either have this be `final` or it should define a virtual destructor then obey the rule of 5 by defaulting everything else.

src/api/c++/z3++.h
```diff
@@ -567,8 +622,7 @@ namespace z3 {
             ast_vector_tpl const* m_vector;
             unsigned m_index;
         public:
-            iterator(ast_vector_tpl const* v, unsigned i): m_vector(v), m_index(i) {}
-            iterator(iterator const& other): m_vector(other.m_vector), m_index(other.m_index) {}
```
zwimer, line 571: This should be implicit.

src/api/c++/z3++.h
NikolajBjorner: It seems like nit. I say leave it at what it is now and move on to more consequential issues

src/api/c++/z3++.h
```diff
@@ -398,7 +420,15 @@ namespace z3 {
         array(array const &) = delete;
         array & operator=(array const &) = delete;
     public:
-        array(unsigned sz):m_array(new T[sz]),m_size(sz) {}
+        array(unsigned sz) : m_array(new T[sz]), m_size(sz) {}
+        array(array && s) noexcept : m_array(s.m_array), m_size(s.m_size) {
```
nunoplopes, line 424: these are not needed; array is internal.

src/api/c++/z3++.h
NikolajBjorner: The sentence says "it is a bug that fixedpoint object didn't come with operator=.

NikolajBjorner:
@nunoplopes do you mind taking the call if remaining pull request should be further split or taken, and then merge when appropriate?
I don't have a take on what to do with supporting move semantics for API classes. 
Only the expr class can be expected to be in a critical loop, but if it helps for uniformity or other purposes, very fine.

nunoplopes:
> @nunoplopes do you mind taking the call if remaining pull request should be further split or taken, and then merge when appropriate?
> I don't have a take on what to do with supporting move semantics for API classes.
> Only the expr class can be expected to be in a critical loop, but if it helps for uniformity or other purposes, very fine.

I agree that most of these classes are never moved in practice. Not in any code I've written. And if they are, the cost of an incref is zero. Agreed that expr are the only where move make sense.
Adding all these move constructors has a cost: an extra null-ptr comparison on destruction (extra code size for everybody) + extra maintenance burden.

Let's not merge this. Some changes while ok (eg noexcept), don't make any difference.

zwimer:
@NikolajBjorner @nunoplopes 
If we remove the redundant nullcheck, there is 0 overhead performance wise. I also decided to benchmark the speed of the move/copy constructor and I got above an _22.3615%_ speed increase between the two. Here is my quick benchmark script if you want to verify this yourself.

```
#include <chrono>
using namespace std::chrono;

[[gnu::optimize("-O0")]] void benchmark() {

    context c;

    auto s0 = high_resolution_clock::now();
    for(long long i = 0; i < 1E9; ++i) 
        expr five = c.int_val(5);
    auto e0 = high_resolution_clock::now();

    auto s1 = high_resolution_clock::now();
    for(long long i = 0; i < 1E9; ++i) {
        expr five = c.int_val(5);
        expr _ {five};
    }    
    auto e1 = high_resolution_clock::now();

    auto s2 = high_resolution_clock::now();
    for(long long i = 0; i < 1E9; ++i) {
        expr five = c.int_val(5);
        expr _ {std::move(five)};
    }    
    auto e2 = high_resolution_clock::now();

    auto d0 = (e0-s0).count();
    auto d1 = (e1-s1).count() - d0;
    auto d2 = (e2-s2).count() - d0;

    std::cout << "copy: " << d1 << std::endl;
    std::cout << "move: " << d2 << std::endl;
    std::cout << "speedup " << (((float)d1/(float)d2)-1) << std::endl;
}
```

To adjust for possible cache discrepancies, I ran this with copy first, then with move first and averaged the two.

The long and short of it is that move semantics are much faster than copy semantics in this situation. With the nullcheck in it is 1-2 CPU instructions of overhead, but since `z3_dec_ref` already nullchecks we can remove that entirely. Also, the overhead code-wise is literally 2 lines of code in the class `ast` if you desire to only provide this for `expr`.

If nothing else, move semantics should be added to `expr` as it is a very large optimization with almost no downsides.

NikolajBjorner:
A pull request with move on expr and symbol is welcome. Other classes are not expected to be in tight loops and if they are, something would be odd.

PR #5163, opened 2021-04-09
Test and memoize pdd factoring

JakobR:


nunoplopes:
failed windows build:
```
D:\a\1\s\src\test\pdd.cpp(415): error C3493: 'va' cannot be implicitly captured because no default capture mode has been specified
D:\a\1\s\src\test\pdd.cpp(415): error C3493: 'vb' cannot be implicitly captured because no default capture mode has been specified
D:\a\1\s\src\test\pdd.cpp(415): error C3493: 'vc' cannot be implicitly captured because no default capture mode has been specified
D:\a\1\s\src\test\pdd.cpp(415): error C3493: 'vd' cannot be implicitly captured because no default capture mode has been specified
D:\a\1\s\src\test\pdd.cpp(422): error C2064: term does not evaluate to a function taking 1 arguments
D:\a\1\s\src\test\pdd.cpp(423): error C2064: term does not evaluate to a function taking 1 arguments
D:\a\1\s\src\test\pdd.cpp(424): error C2064: term does not evaluate to a function taking 1 arguments
D:\a\1\s\src\test\pdd.cpp(425): error C2064: term does not evaluate to a function taking 1 arguments
D:\a\1\s\src\test\pdd.cpp(426): error C2064: term does not evaluate to a function taking 1 arguments
D:\a\1\s\src\test\pdd.cpp(427): error C2064: term does not evaluate to a function taking 1 arguments
```

JakobR:
That's funny, I had removed `va` etc. from the capture list because clang (on macOS) complained about it… Thanks, I'll fix it.

src/math/dd/dd_pdd.cpp
```diff
@@ -733,6 +743,8 @@ namespace dd {
                 rest = p;
             }
         }
+        e.m_lc = lc.root;
```
NikolajBjorner, line 746: factor_entry could be invalidated if there is a gc between cache lookup and the update here.
Instead of resetting the entire cache you can mark the entry as busy and avoid removing busy entries during gc.

Issue #5127, opened 2021-03-26
(rewriter.hoist_mul=true rewriter.flat=false) refutation soundness bug in string formula


jiwonparc:
commit: b918f121e
```
$ z3release small.smt2
sat
$ z3-4.8.10 rewriter.hoist_mul=true rewriter.flat=false small.smt2
sat
$ z3release rewriter.hoist_mul=true rewriter.flat=false small.smt2
unsat
$ cat small.smt2
(declare-fun a () String)
(declare-fun b () String)
(assert (str.in_re (str.++ a "z" b) (re.++ (re.* (str.to_re "z")) (str.to_re "a"))))
(assert (not (str.in_re (str.++ "a" a "z" a "z" b "b" b) (re.++ (re.* (str.to_re "aa")) (str.to_re "a")))))
(check-sat)
```

levnach:
solved by https://github.com/Z3Prover/z3/pull/5165

PR #5165, opened 2021-04-09
B5127

levnach:
fix for issue https://github.com/Z3Prover/z3/issues/5127


levnach:
The failure at regressions/smt2/eq.smt2 does not execute the touched code.

PR #5170, opened 2021-04-12
Fix PDD factor cache in case GC happens while factoring

JakobR:
I used a counter instead of marking active entries as busy, because then I can keep storing entries directly in the hashtable instead of separately allocating them and storing pointers.

### DIFF
--- a/src/math/dd/dd_pdd.cpp
+++ b/src/math/dd/dd_pdd.cpp
@@ -696,8 +696,8 @@ namespace dd {
     void pdd_manager::factor(pdd const& p, unsigned v, unsigned degree, pdd& lc, pdd& rest) {
         unsigned level_v = m_var2level[v];
         if (degree == 0) {
-            lc = p;
-            rest = zero();
+            lc = zero();
+            rest = p;
             return;
         }
         if (level(p.root) < level_v) {
@@ -785,7 +785,8 @@ namespace dd {
             // no reduction
             return false;
         }
-        if (l == 0) {
+        if (m == 0) {
+            // no reduction (result would still contain v^l)
             return false;
         }
         pdd a = zero();
@@ -795,6 +796,7 @@ namespace dd {
         p.factor(v, l, a, b);
         q.factor(v, m, c, d);
         unsigned const j = std::min(max_pow2_divisor(a), max_pow2_divisor(c));
+        SASSERT(j != UINT_MAX);  // should only be possible when both l and m are 0
         rational const pow2j = rational::power_of_two(j);
         auto div_pow2j = [&pow2j](rational const& r) -> rational {
             rational result = r / pow2j;
--- a/src/math/dd/dd_pdd.cpp
+++ b/src/math/dd/dd_pdd.cpp
@@ -51,6 +51,7 @@ namespace dd {
     void pdd_manager::reset(unsigned_vector const& level2var) {
         reset_op_cache();
         m_factor_cache.reset();
+        m_factor_cache_generation++;
         m_node_table.reset();
         m_nodes.reset();
         m_free_nodes.reset();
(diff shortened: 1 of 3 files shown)


src/math/dd/dd_pdd.cpp
```diff
@@ -707,12 +708,13 @@ namespace dd {
         }
         // Memoize nontrivial cases
         auto* et = m_factor_cache.insert_if_not_there2({p.root, v, degree});
-        factor_entry& e = et->get_data();
-        if (e.is_valid()) {
-            lc = pdd(e.m_lc, this);
-            rest = pdd(e.m_rest, this);
+        factor_entry* e = &et->get_data();
```
NikolajBjorner, line 711: it is the same

src/math/dd/dd_pdd.cpp
```diff
@@ -743,10 +745,122 @@ namespace dd {
...
+        rational const pow2j = rational::power_of_two(j);
+        auto div_pow2j = [&pow2j](rational const& r) -> rational {
+            rational result = r / pow2j;
+            SASSERT(result.is_int());
+            return result;
+        };
+        pdd aa = map_coefficients(a, div_pow2j);
+        pdd cc = map_coefficients(c, div_pow2j);
+        pdd vv = one();
+        for (unsigned deg = l - m; deg > 0; --deg) {
```
NikolajBjorner, line 817: I have gone over to use this pattern:

   for (unsigned deg = m; deg-- > 0; )

NikolajBjorner: add as stand-alone routine to pdd_manager: lift polynomial to the n'th power.

src/math/dd/dd_pdd.cpp
```diff
@@ -51,6 +51,7 @@ namespace dd {
     void pdd_manager::reset(unsigned_vector const& level2var) {
         reset_op_cache();
         m_factor_cache.reset();
+        m_factor_cache_generation++;
```
NikolajBjorner, line 54: yes, makes sense

src/math/dd/dd_pdd.cpp
```diff
@@ -743,10 +745,122 @@ namespace dd {
...
+            auto* et = m_factor_cache.insert_if_not_there2({p.root, v, degree});
+            e = &et->get_data();
+        }
+        e->m_lc = lc.root;
+        e->m_rest = rest.root;
+    }
+
+    template <class Fn>
+    pdd pdd_manager::map_coefficients(pdd const& p, Fn f)
+    {
```
NikolajBjorner, line 760: convention: { on previous line.

src/math/dd/dd_pdd.cpp
NikolajBjorner: would make sense to call it gc_generation. Note that m_factor_cache is reset during gc. This is the reset that could be problematic. This reset is not used during operations.

src/math/dd/dd_pdd.cpp
NikolajBjorner: this could also be memoized with the op-cache. 
map-coefficients isn't nested with different Fn so the same op-code can be used.

NikolajBjorner: op-cache use has to be very careful: it persists across calls and the calls could use different divisors.
So the divisor itself needs to be on the cache object. This is possible as op-cache takes 2 arguments + op.
The constant node corresponding to pow2j would be one of the arguments.

JakobR: And map_coefficients needs to be less generic as well (not allow arbitrary functions like now).

NikolajBjorner: of course, it would be a different function, similar to minus and minus_rec

PR #5171, opened 2021-04-12
array uses unique_ptr

zwimer:
In the C++ API, the array class now uses `unique_ptr` instead of manual memory management

### DIFF
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -1991,8 +1991,7 @@ namespace z3 {
 
     template<typename T>
     template<typename T2>
-    array<T>::array(ast_vector_tpl<T2> const & v) {
-        m_array.reset(new T[v.size()]);
+    array<T>::array(ast_vector_tpl<T2> const & v):m_array(new T[v.size()]) {
         m_size  = v.size();
         for (unsigned i = 0; i < m_size; i++) {
             m_array[i] = v[i];
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -405,8 +405,8 @@ namespace z3 {
         unsigned size() const { return m_size; }
         T & operator[](int i) { assert(0 <= i); assert(static_cast<unsigned>(i) < m_size); return m_array[i]; }
         T const & operator[](int i) const { assert(0 <= i); assert(static_cast<unsigned>(i) < m_size); return m_array[i]; }
-        T const * ptr() const { return & m_array[0]; }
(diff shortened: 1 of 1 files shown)


src/api/c++/z3++.h
```diff
@@ -392,21 +393,20 @@ namespace z3 {
...
-        ~array() { delete[] m_array; }
-        void resize(unsigned sz) { delete[] m_array; m_size = sz; m_array = new T[sz]; }
+        void resize(unsigned sz) { m_array.reset(new T[sz]); m_size = sz; }
         unsigned size() const { return m_size; }
         T & operator[](int i) { assert(0 <= i); assert(static_cast<unsigned>(i) < m_size); return m_array[i]; }
         T const & operator[](int i) const { assert(0 <= i); assert(static_cast<unsigned>(i) < m_size); return m_array[i]; }
-        T const * ptr() const { return m_array; }
-        T * ptr() { return m_array; }
+        T const * ptr() const { return & m_array[0]; }
+        T * ptr() { return & m_array[0]; }
```
NikolajBjorner, line 409: I have seen debug runtimes assert on these kinds of tricks. 
Introduces maintenance hassles. Is this really best practice to use &v[0]?

NikolajBjorner: Don't you want https://en.cppreference.com/w/cpp/memory/unique_ptr/get

zwimer: According to the docs, the array version does not support raw pointer stuff, so this is how it has to be done. Here:
```
Single-object version, unique_ptr<T>
operator*operator-> | dereferences pointer to the managed object(public member function)
Array version, unique_ptr<T[]>
operator[] | provides indexed access to the managed array
```
Source: https://en.cppreference.com/w/cpp/memory/unique_ptr

zwimer: Regardless, since it is a smart pointer, the best we could reduce this to would be: `&*m_array`, if it did support `*` in array mode.

zwimer: I can update it to `get`.

zwimer:
Regarding:
```
   406         T & operator[](int i) { assert(0 <= i); assert(static_cast<unsigned>(i) < m_size); return m_array[i]; }
   407         T const & operator[](int i) const { assert(0 <= i); assert(static_cast<unsigned>(i) < m_size); return                 m_array[i]; }
   ```
Is there a reason we don't just take in `i` as `unsigned` in the first place, rather than `int`?

PR #5177, opened 2021-04-12
prefer parent operator= to manually copying parent data for extensibi…

zwimer:
…lity reasons


src/api/c++/z3++.h
```diff
@@ -562,7 +569,6 @@ namespace z3 {
         public:
             iterator(ast_vector_tpl const* v, unsigned i): m_vector(v), m_index(i) {}
             iterator(iterator const& other): m_vector(other.m_vector), m_index(other.m_index) {}
-            iterator operator=(iterator const& other) { m_vector = other.m_vector; m_index = other.m_index; return *this; }
```
zwimer, line 565: This should be implicit rather than explicit because of the rule of 5.

PR #5180, opened 2021-04-14
Add more PDD utilities (div, pow)

JakobR:


NikolajBjorner:
do a pull and you can start testing test-z3 polysat to see how the first two examples behave (and add fixes/diagnostics).

PR #5198, opened 2021-04-19
Add functionality for BDD vectors

JakobR:


src/math/dd/dd_bdd.cpp
```diff
@@ -896,87 +911,54 @@ namespace dd {
...
+        unsigned var_idx = bits.size();
         while (!is_const(b)) {
-            unsigned const var = m_level2var[level(b)];
-            if (var >= w)
-                b = lo(b);
-            else
-                b = val.get_bit(var) ? hi(b) : lo(b);
+            VERIFY(var_idx-- > 0);
+            SASSERT(var(b) <= bits[var_idx]);
+            while (var(b) < bits[var_idx]) {
```
NikolajBjorner, line 920: is this invariant under re-ordering?

src/math/dd/dd_bdd.h
```diff
@@ -303,17 +299,102 @@ namespace dd {
...
 
+    class bddv {
+        friend bdd_manager;
+
+        // TODO: currently we repeat the pointer to the manager in every entry of bits.
+        // Should use BDDV instead (drawback: the functions in bdd_manager aren't as nice and they need to be careful about reference counting if they work with BDD directly).
+        vector<bdd>  bits;
+        bdd_manager& m;
+
+        bddv(vector<bdd> bits, bdd_manager* m): bits(bits), m(*m) { inc_bits_ref(); }
```
NikolajBjorner, line 329: The bdd class is already reference counted.

src/math/dd/dd_bdd.h
```diff
@@ -303,17 +299,102 @@ namespace dd {
...
+        }
+        ~bddv() {
+            // NOTE: necessary if we switch to BDDV as storage
+            // for (BDD b : bits) { m.dec_ref(b); }
+        }
+
+        unsigned size() const { return bits.size(); }
+        vector<bdd> const& get_bits() const { return bits; }
+
+        bdd operator<=(bddv const& other) const { return m.mk_ule(*this, other); }
```
NikolajBjorner, line 369: Just expose them as sle, ule, sgt, uge, ... instead of opeators. We don't have sign information in the bits.

src/math/dd/dd_bdd.cpp
```diff
@@ -896,87 +911,54 @@ namespace dd {
...
+                VERIFY(var_idx-- > 0);
+            }
+            SASSERT(var(b) == bits[var_idx]);
+            b = val.get_bit(var_idx) ? hi(b) : lo(b);
         }
         return is_true(b);
     }
 
-    find_int_t bdd_manager::find_int(BDD b, unsigned w, rational& val) {
+    find_result bdd_manager::find_num(BDD b, unsigned_vector bits, rational& val) {
```
NikolajBjorner, line 929: Instead of all of this inside the bdd manager, the polysat solver object could maintain a layer (a class) that maintains the association of variables. Pushing this functionality into bdd_manager is so far awkard. For a BDD native version you would basically want to pass in a function that maps variables to power of two. Together with some accounting for how many variables were fixed.

src/math/dd/dd_bdd.h
NikolajBjorner: Or keep these for niceness of unit tests, but add signed versions using sle, sgt.

src/math/dd/dd_bdd.h
```diff
@@ -26,13 +26,17 @@ Revision History:
...
-    enum class find_int_t { empty, singleton, multiple };
+    enum class find_result { empty, singleton, multiple };
+    std::ostream& operator<<(std::ostream& out, find_result x);
 
     class bdd_manager {
         friend bdd;
+        friend bddv;
 
         typedef unsigned BDD;
+        typedef svector<BDD> BDDV;
```
NikolajBjorner, line 39: I bet you never use this version. It can be deleted

src/math/dd/dd_bdd.h
```diff
@@ -303,17 +299,102 @@ namespace dd {
...
+        friend bdd_manager;
+
+        // TODO: currently we repeat the pointer to the manager in every entry of bits.
+        // Should use BDDV instead (drawback: the functions in bdd_manager aren't as nice and they need to be careful about reference counting if they work with BDD directly).
+        vector<bdd>  bits;
+        bdd_manager& m;
+
+        bddv(vector<bdd> bits, bdd_manager* m): bits(bits), m(*m) { inc_bits_ref(); }
+        bddv(vector<bdd>&& bits, bdd_manager* m): bits(std::move(bits)), m(*m) { inc_bits_ref(); }
+        void inc_bits_ref() {
```
NikolajBjorner, line 331: not worth it.

src/math/dd/dd_bdd.cpp
```diff
@@ -1014,14 +996,133 @@ namespace dd {
...
+        bddv result(this);
+        for (unsigned i = 0; i < num_bits; ++i) {
+            result.push_back(n.get_bit(i) ? mk_true() : mk_false());
+        }
+        return result;
+    }
+
+    bddv bdd_manager::mk_ones(unsigned num_bits) {
+        bddv result(this);
+        for (unsigned i = 0; i < num_bits; ++i) {
```
NikolajBjorner, line 1084: after years of instisting on { for single statements I have abandoned this style. New code can avoid extra { }.
The benefits are too rare: accidentially adding an extra statement to be grouped inside of the for loop or having a poorly defined macro that has more than one statement.

src/math/dd/dd_bdd.cpp
```diff
@@ -1014,14 +996,133 @@ namespace dd {
...
+        bddv result = mk_zero(a.size());
+        for (unsigned i = 0; i < b.size(); ++i) {
+#if 1
+            bddv s = a_shifted;
+            for (unsigned j = i; j < b.size(); ++j) {
+                s[j] &= b[i];
+            }
+            result = mk_add(result, s);
+#else
+            // From BuDDy's bvec_mul. It seems to compute more intermediate BDDs than the version above?
```
NikolajBjorner, line 1039: not sure it is worth spending time on general bdd multiplication. It is too expensive to be used upstream for anything with larger bit-widths.

src/math/dd/dd_bdd.cpp
```diff
@@ -896,87 +911,54 @@ namespace dd {
...
-            rational inv_aa;
-            VERIFY(aa.mult_inverse(j, inv_aa));
-            rational const cc = mod(inv_aa * -bb, rational::power_of_two(j));
-            return mk_int(cc, j);
-        }
-   }
-
 
     bdd bdd_manager::mk_eq(bddv const& a, bddv const& b) {
+        SASSERT(a.size() == b.size());
```
NikolajBjorner, line 961: we might consider to run the loop from below instead of above to follow common case variable ordering (in the case this leads to fewer overall operations to build bdd). Maybe easy to test by implementing two versions and counting statistics in a test.

src/math/dd/dd_bdd.cpp
JakobR: Ok; I was just wondering why BuDDy is using a seemingly more complicated version. In any case, both versions seem to work correctly.

src/math/dd/dd_bdd.cpp
JakobR: Ok. I'm trying to follow your style but often the extra { } sneak in out of habit :)

src/math/dd/dd_bdd.cpp
JakobR: Yes, counting from below seems to be slightly better. To compare, I printed `m_node_table.size()` as an approximation for the number of operations (idea: counting the "real" operations that don't just look up from the cache).

src/math/dd/dd_bdd.cpp
JakobR: Good point, it wasn't. I added a unit test to make sure the new version works under re-ordering.

src/math/dd/dd_bdd.cpp
NikolajBjorner: We can compare several sources for encoding multipliers. Also CUDD. In the end the BDDs are unique modulo ordering and there is never a good ordering. Just stay with the simple grade school encoding.

PR #5200, opened 2021-04-19
Cpp api add missing fp methods (return type correction)

zwimer:
Bug fix

### DIFF
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -891,6 +891,16 @@ namespace z3 {
         }
 
         /**
+           \brief Convert this fpa into an IEEE BV
+        */
+        bool to_ieee_bv() const {
+            assert(is_fpa());
+            Z3_ast r = Z3_mk_fpa_to_ieee_bv(ctx(), m_ast),
+            check_error();
+            return expr(ctx(), r);
+        }
+
+        /**
            \brief Return string representation of numeral or algebraic number
            This method assumes the expression is numeral or algebraic
 
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -873,7 +873,7 @@ namespace z3 {
         /**
            \brief Return true if this expression is an fpa and is inf
         */
-        bool is_inf() const {
+        expr is_inf() const {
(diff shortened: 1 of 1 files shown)


src/api/c++/z3++.h
```diff
@@ -873,7 +873,7 @@ namespace z3 {
         /**
            \brief Return true if this expression is an fpa and is inf
         */
-        bool is_inf() const {
+        expr is_inf() const {
```
wintersteiger, line 876: @zwimer @NikolajBjorner Something is off here. I don't think I added these functions, so I'm not sure what the intention was, but as it is, the docstring and the implementation disagree on what these functions do (`mk_is_inf` vs `is_inf`). I think these should be the  bool versions, like all the functions around them.

NikolajBjorner: should be called mk_is_inf.

zwimer: I will update the docstring in a new PR. I can rename it `mk_is_inf`. I was looking to implement the ones that returned a Z3 expr, not a bool though. Specifically I was trying to create this in the API:
https://github.com/angr/claripy/blob/e84355f64bfbb802715872a2b88fea84362fd37e/claripy/backends/backend_z3.py#L1147

That said, `is_nan` is in the same bucket; would you rather the `mk_` version for both or not?

I left it without because of things like:
https://github.com/Z3Prover/z3/blob/a52b485d9c27c4c7cd26f47fdf878b264900a0a6/src/api/c%2B%2B/z3%2B%2B.h#L1229
```C++
        friend expr is_int(expr const& e);
```

Though I figured it should be a member instead of a friend. If you'd rather it be a friend and be placed beside `is_int` as `is_inf` though, that would be be fine too.

Just let me know what you prefer @NikolajBjorner @wintersteiger

NikolajBjorner: line 1229 should likely have been mk_is_int even though I would say these "mk_" have become a bit tiresome.
See also https://github.com/Z3Prover/z3/commit/44d77a978a97fbadefe0cc964c53f24c3679b6d7

PR #5218, opened 2021-04-26
Cpp api fp to bv

zwimer:
Add missing API methods `fpa_to_ubv` and `sbv`.
Bug fix of missing semi-colon.


zwimer:
Added the `fpa_fp` method as well to the API.

PR #5231, opened 2021-04-30
WiP: Add get-models command

ambiso:
Hi,

I've made a terrible mistake in creating this horrible hack, however it could solve a problem I think [many face](https://stackoverflow.com/q/11867611).

For me, it is mainly an issue of convenience - writing an `smt2` file it is unwieldy to manually block models to obtain the next possible one.

The modifications I made add a `get-models` command (not to be confused with the `get-model` command) to the `smt2parser`.

The `get-models` command returns all possible models.
It essentially performs a `check-sat`, `get-model` then blocks the model and repeats until the formula is unsat.

Here is an example:

```smt2
./test.smt2
(declare-const x Bool)
(declare-const y Bool)

(assert (or x y))
(get-models)
```

This outputs the following

```smt2
z3 ./test.smt2
sat
(
  (define-fun y () Bool
    false)
  (define-fun x () Bool
    true)
)
sat
(
  (define-fun y () Bool
    true)
  (define-fun x () Bool
    false)
)
sat
(
  (define-fun y () Bool
    true)
  (define-fun x () Bool
    true)
)
unsat
```

Is this something others may be interested in? Did I solve a problem nobody has? Is this problem already solved by other means?

I suppose there are also better ways of integrating this - possibly not directly inside the `smt2parser`, and it's not unlikely that I've made some terrible mistakes in the implementation.

Looking forward to hearing some feedback on this!

Best,
ambiso

### DIFF
--- a/src/parsers/smt2/smt2parser.cpp
+++ b/src/parsers/smt2/smt2parser.cpp
@@ -94,6 +94,7 @@ namespace smt2 {
 
         symbol               m_assert;
         symbol               m_check_sat;
+        symbol               m_get_models;
         symbol               m_define_fun;
         symbol               m_define_const;
         symbol               m_model_add;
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
The biggest issue is the aggregate cost of maintaining various features out of the beaten path. 
It solves a problem for a very specific use case and it isn't even efficient because adding blocking clauses
ends up creating overhead. 
A more scalable approach is to use backtracking (not backjumping), disable some versions of restarts, during search and close leaves using backtracking. Developing such solutions is well beyond a pull request that touches front-end matters.

There are existing ways users can get this effect from programmatic API or if they really want to use SMTLIB2, then write scripts around it. There are even some wrappers around SMT solvers that use just SMTLIB2.

Issue #4673, opened 2020-09-02
Floating point exponent saturates rather than becoming infinite


mpdn:
It seems like if the exponent of a floating point overflows, it just saturates instead of becoming (positive or negative) infinite.

Example:
```python
import numpy as np
from z3 import *

# This does not find a solution
x = FP('x', Float32())
solve(x > 0, (x * 2) * 0.5 > x)

# But this *is* a solution and prints `True`
x = np.finfo(np.float32).max
print((x * np.float32(2)) * np.float32(0.5) > x)
```

LeventErkok:
It seems the Python interface is using the default rounding mode of round-towards-zero, which causes the problem to be `unsat`. If you set the regular defaulting mode of nearest-ties-to-even, you get a model:

```python
from z3 import *

print(get_default_rounding_mode())
set_default_rounding_mode(Z3_OP_FPA_RM_NEAREST_TIES_TO_EVEN)
print(get_default_rounding_mode())

x = FP('x', Float32())
solve(x > 0, (x * 2) * 0.5 > x)
```

This prints:

```
RTZ()
RNE()
[x = 1.00000095367431640625*(2**127)]
```

This is set in:

  https://github.com/Z3Prover/z3/blob/daf7e9e3ef9b5c9584b516673e505e17511d2bdf/src/api/python/z3/z3.py#L8679

So, it does the "right" thing per this definition, but I do agree that the default rounding mode should be set to `Z3_OP_FPA_RM_NEAREST_TIES_TO_EVEN`

mpdn:
Ah, thanks a lot. Although I don't quite get how round-to-zero causes it to not become infinite, but maybe this is me misunderstanding IEEE-754.

LeventErkok:
@mpdn When you multiply `max` with `2` in RTZ, the value remains `max` and does not become `Infinity`. The easiest way to see this is to consider where the infinitely precise value remains: It's clearly between `max` and `infinity`, and `RTZ` says go towards zero, i.e., the result remains `max`. (It's funny to talk about a value being "close" to max than infinity, but the definition of overflow is very precisely defined in IEEE-754.)

You can double-check this with a C program:

```C
#include <stdio.h>
#include <fenv.h>

int main()
{
    float f = 3.4028235e38;

    float g1 = f*2;
    printf("f = %g, g1 = %g\n", f, g1);

    fesetround(FE_TOWARDZERO);
    float g2 = f*2;
[...]

PR #5242, opened 2021-05-04
Implemented missing methods to the C++ API

mikhailramalho:
(inspired by PR #5234)

I've added the following methods to the C++ API:
* `std::ostream & operator<<(std::ostream & out, sort const & s)` to print sorts
* `mk_is_normal` checks if a fp is normal
* `mk_is_subnormal` checks if a fp is subnormal
* `mk_is_zero` checks if a fp is zero
* `sbv_to_fpa` converts from signed bv to fp
* `ubv_to_fpa` converts from unsigned bv to fp
* `fpa_to_fpa` converts from fp to fp
* `round_fpa_to_closest_integer` rounds fp to closest fp
* `mk_from_ieee_bv` converts from bv (in IEEE format) to fp
* `bvredor` calls bv reduction `or`
* `bvredand` calls bv reduction `and`
* `fp_eq` calls `fp.eq`
* `fpa_nan` creates a fp NaN
* `fpa_inf` creates a fp +/-inf

### DIFF
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -1337,6 +1337,9 @@ namespace z3 {
         expr rotate_right(unsigned i) { Z3_ast r = Z3_mk_rotate_right(ctx(), i, *this); ctx().check_error(); return expr(ctx(), r); }
         expr repeat(unsigned i) { Z3_ast r = Z3_mk_repeat(ctx(), i, *this); ctx().check_error(); return expr(ctx(), r); }
 
+        friend expr bvredor(expr const & a);
+        friend expr bvredand(expr const & a);
+
         friend expr abs(expr const & a);
         friend expr sqrt(expr const & a, expr const & rm);
 
@@ -1848,6 +1851,18 @@ namespace z3 {
             r = Z3_mk_fpa_max(a.ctx(), a, b); 
         }
         return expr(a.ctx(), r); 
+    }
+    inline expr bvredor(expr const & a) {
+        assert(a.is_bv());
+        Z3_ast r = Z3_mk_bvredor(a.ctx(), a);
+        a.check_error();
+        return expr(a.ctx(), r);
+    }
+    inline expr bvredand(expr const & a) {
+        assert(a.is_bv());
+        Z3_ast r = Z3_mk_bvredor(a.ctx(), a);
+        a.check_error();
+        return expr(a.ctx(), r);
     }
     inline expr abs(expr const & a) { 
         Z3_ast r;
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -694,6 +694,8 @@ namespace z3 {
             \pre is_array()
         */
         sort array_range() const { assert(is_array()); Z3_sort s = Z3_get_array_sort_range(ctx(), *this); check_error(); return sort(ctx(), s); }
+
+        friend std::ostream & operator<<(std::ostream & out, sort const & s) { return out << Z3_sort_to_string(s.ctx(), Z3_sort(s.m_ast)); }
     };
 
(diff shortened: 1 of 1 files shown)


mikhailramalho:
I still have a couple of methods that I want in the C++ API but I'm not sure how to implement them properly:

* `Z3_mk_fpa_rounding_mode_sort`, `Z3_mk_fpa_rna` (and the others rounding mode ones)

There is `context::fpa_rounding_mode` which also returns a rounding mode sort, however, it calls `Z3_mk_fpa_rna` (and the others) directly. 

My question is here is that `Z3_mk_fpa_rna` and the others return `Z3_ast` but the C++ API is creating a sort out of it. Is that the expected behaviour? 

The documentation states that it is indeed returning a sort `Create a numeral of RoundingMode sort which represents the TowardPositive rounding mode` even though it returns a `Z3_ast`, but I have been using it for the past few years... I always thought that, e.g., `Z3_mk_fpa_rna` creates an expression which has sort `Z3_mk_fpa_rounding_mode_sort`.

wintersteiger:
Thanks for pointing out these issues! Those definitions in the C++ API are indeed wrong. As you suggest, `Z3_mk_fpa_rounding_mode_sort` returns a rounding-mode sort and `Z3_mk_fpa_rna` and friends return a numeral of rounding-mode sort. I think that for pretty much all of these and similar issues, when in doubt, rely on the C API and assume that the C++ is wrong. 

`m_rounding_mode` is the default rounding mode that's applied in the overloaded operators (so you can use `+`, `-`, ... without specifying a rounding mode every time). To be honest, I don't like those overloads for floats, but as long as users understand the semantics it's convenient for them. Speaking of which, the default rounding mode in the C++ API is apparently `RNA`, while many users would expect `RNE` (see also https://github.com/Z3Prover/z3/issues/4673). 

Please go ahead and add those fixes, I'm happy to take another look at this PR or answer more questions after that!

mikhailramalho:
> To be honest, I don't like those overloads for floats, but as long as users understand the semantics it's convenient for them. 

I'm not a fan either, especially because you have to set the rounding mode for every operation in SMT... BUT it follows what most programming languages expect you to do, right? Either you explicitly set the rounding mode before the operation(s) or you go with what is already set.

> 
> Please go ahead and add those fixes, I'm happy to take another look at this PR or answer more questions after that!

Cool, I've just updated the PR. 

I removed `expr::fpa_rounding_mode` since all operations get the rounding mode from the context, please let me know if it makes sense for you.

NikolajBjorner:
@wintersteiger I am under the impression that you would comment on this and merge when ready.

>  I'm happy to take another look at this PR or answer more questions after that!

wintersteiger:
Yep, that all looks good to me, many thanks for cleaning this up!

wintersteiger:
@NikolajBjorner the logs of the failed CI builds are not available anymore. Are those expected and be ignored?

NikolajBjorner:
Seems build retention timeout was met. Since change is to add new code mostly and the main breaking change is rounding mode, which is discussed above, I will merge.

NikolajBjorner:
Also, I couldn't successfully restart the failed builds. Probably because of retention. As there is no obvious way to start a new build I merged.

wintersteiger:
Yes, I saw that too. You can manually submit jobs in the pipeline UI, but I forgot why that didn't work for me last week (maybe permission problems, SC-alt accounts etc).

NikolajBjorner:
The build failures were due to breaks in Julia bindings. I ended up addressing one of the issues by removing the method that broke from Julia. I am not familiar with the use case for floating points from Julia bindings and guessing that it is very low opted for this approach to move forward.

PR #5290, opened 2021-05-21
Fix NuGet package name and glibc rid.

Choc13:


Choc13:
@NikolajBjorner thanks for merging #5288. I managed to test this locally by downloading the NuGet from the nightly release on here. It works fine for Mac. However, I noticed that the deps.json for the linux lib was now showing as glibc-x64, which isn't a valid linux RID according to [here](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json). I also asked a colleague on Linux to try the nightly and it broke for him. 

I think this is possibly only an issue in the nightlies because it's just in `mk_nuget_task.py` that it mentions `glibc-x64`. 

Second to this I noticed that the NuGet package name had changed to `Microsoft.Z3` instead of being `Microsoft.Z3.x64`. If this was intentional then apologies and I can revert that.

NikolajBjorner:
Thanks for checking up on the ubuntu vs glibc. It was introduced when our hosted build machines started using this naming convention so I assumed it was a Linux-wide convention change. 

For x64: the reason is that someone took ownership of the nuget package Microsoft.Z3 namespace and the party didn't respond to releasing the name for some time as I was setting up a native distribution pipeline. 
I believe now it was revoked and can be used by z3 distributions. We can try with removing the x64 suffix which is by now misleading given the requirement by Don to add the x86 Windows.

PR #5295, opened 2021-05-23
Fix some PEP-8 violations in Python code

orsinium:
[PEP-8](https://www.python.org/dev/peps/pep-0008/) is an optional (but strongly recommended) code style advised by the Python core dev team for Python code. This PR fixes some of the PEP-8 violations. It's hard to fix all of them, though. For instance, PEP-8 recommends avoiding wildcard imports but Z3 heavily relies on them.

Anyway, this is what is fixed:

1. Bare `except:` block replaced by more precise exceptions handled (`except Exception:`)
2. Wrap too long lines. PEP-8 advises 80 chars line length limit but I decided to be less strict and wrap everything to 120 chars.
3. Make some `isinstance` check shorter: use `isinstance(a, (int, str))` instead of `isinstance(a, int) or isinstance(a, str)`
4. Autoformat code using [autopep8](https://github.com/hhatto/autopep8/) in aggressive mode: drop trailing whitespaces, separate top-level declarations by 2 blank lines, use `a is None` instead of `a == None`, wrap binary operations with single whitespaces, use 4 whitespaces for indentation, start comments from single `#`.
5. Autoformat code using [unify](https://github.com/myint/unify): use double quotes for all string literals for consistency.
6. Use `sys.version_info` instead of `sys.version`.

There are still remain a few violations reported by [flake8](https://github.com/PyCQA/flake8/) but they are potential bugs and so require a more thoughtful approach. This PR is already huge, so I want to keep it simple and safe, with mostly stylistic changes.

### DIFF
--- a/src/api/python/z3/z3.py
+++ b/src/api/python/z3/z3.py
@@ -1822,10 +1822,10 @@ def And(*args):
     """
     last_arg = None
     if len(args) > 0:
-        last_arg = args[len(args)-1]
+        last_arg = args[len(args) - 1]
     if isinstance(last_arg, Context):
-        ctx = args[len(args)-1]
-        args = args[:len(args)-1]
+        ctx = args[len(args) - 1]
+        args = args[:len(args) - 1]
     elif len(args) == 1 and isinstance(args[0], AstVector):
         ctx = args[0].ctx
         args = [a for a in args[0]]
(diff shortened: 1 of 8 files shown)


orsinium:
Thank you!

PR #5310, opened 2021-05-28
Use more generic linux-x64 for NuGet rid.

iscottb122:
The more generic 'linux-x64' should be used as the linux rid for NuGet packages, because at the moment only 'ubuntu-x64' is included in the package. Therefore, loading in dotnet on e.g. 'debian-x64' does not work, even though there seems to be no fundamental reason it should not. As a minimal test, manually changing the rid and the associated directory path in the build output to 'linux-x64' makes everything work as expected on 'debian-x64'.

It does not appear that different shared objects (per specific distribution) are being produced anyway, so this change should have no regressions, as far as I understand it. At the very least, the rid should be changed to 'debian-x64' since this is a parent of 'ubuntu-x64' in the [hierarchy](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json), but I don't see any reason not to go to the further ancestor 'linux-x64'.

### DIFF
--- a/scripts/mk_nuget_task.py
+++ b/scripts/mk_nuget_task.py
@@ -22,14 +22,14 @@ def mk_dir(d):
         os.makedirs(d)
 
 
-os_info = {"z64-ubuntu-14" : ('so', 'ubuntu.14.04-x64'),
-           'ubuntu-18' : ('so', 'ubuntu-x64'),
-           'ubuntu-20' : ('so', 'ubuntu-x64'),
-           'glibc-2.31' : ('so', 'ubuntu-x64'),
+os_info = {"z64-ubuntu-14" : ('so', 'linux-x64'),
+           'ubuntu-18' : ('so', 'linux-x64'),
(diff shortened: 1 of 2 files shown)


scripts/mk_nuget_task.py
```diff
@@ -22,14 +22,14 @@ def mk_dir(d):
         os.makedirs(d)
 
 
-os_info = {"z64-ubuntu-14" : ('so', 'ubuntu.14.04-x64'),
-           'ubuntu-18' : ('so', 'ubuntu-x64'),
-           'ubuntu-20' : ('so', 'ubuntu-x64'),
-           'glibc-2.31' : ('so', 'ubuntu-x64'),
+os_info = {"z64-ubuntu-14" : ('so', 'linux-x64'),
```
NikolajBjorner, line 25: ubuntu-14 is probably long gone and I will remove on next occasion.
I use either 18 or default 20 these days. I understood there are some incompatibilities with 20 on downlevel platforms.

PR #5311, opened 2021-05-28
Polysat: disjunctive lemmas

JakobR:


src/util/ref_vector.h
```diff
@@ -20,6 +20,7 @@ Revision History:
```
NikolajBjorner, line 20: push to master

src/util/ref.h
```diff
@@ -66,10 +66,14 @@ class ref {
         return m_ptr;
```
NikolajBjorner, line 66: push to master together with spacer_context

src/math/polysat/constraint.h
```diff
@@ -15,58 +15,70 @@ Module Name:
...
     enum csign_t : bool { neg_t = false, pos_t = true };
 
     class constraint;
+    class constraint_manager;
     class clause;
     class scoped_clause;
     class eq_constraint;
     class var_constraint;
     class ule_constraint;
+    using constraint_ref = ref<constraint>;
```
NikolajBjorner, line 35: Good, seems very sensible.

src/math/polysat/solver.cpp
```diff
@@ -20,6 +20,9 @@ Module Name:
 #include "math/polysat/log.h"
 #include "math/polysat/forbidden_intervals.h"
 
+// For development; to be removed once the linear solver works well enough
+#define ENABLE_LINEAR_SOLVER 0
```
NikolajBjorner, line 24: good

src/math/polysat/solver.cpp
```diff
@@ -682,46 +716,105 @@ namespace polysat {
      * We add 'p == 0' as a lemma. The lemma depends on the dependencies used
      * to derive p, and the level of the lemma is the maximal level of the dependencies.
      */
-    void solver::learn_lemma(pvar v, scoped_clause&& lemma) {
+    void solver::learn_lemma(pvar v, clause_ref lemma) {
```
NikolajBjorner, line 719: clause_ref& ?

JakobR: I would prefer to keep passing by value. To avoid unnecessary reference counting, we can use `std::move` at the callsite. This also makes it clear at the callsite where the ownership will be "stolen".

PR #5327, opened 2021-06-03
Shell api

chadbrewbaker:
@regehr needed to shell out to z3.

This is the minimal Rust. I didn't test it on windows.


andrewvaughanj:
Just to help people understand: why do you feel this belongs in the Z3 repo vs. your own repo or a gist?

I don’t see anything Z3-specific in this PR; it looks more like a “this is how you communicate over stdin/stdout via Rust” than anything tied to Z3 (apart from the command you run).

chadbrewbaker:
Calling Z3 from the shell interface in a systems language is a common use case. Cross platform shell from C is gnarly - this was cleaner.

I thought about a more featureful calling convention.  Specify push/pop vs reset to stream queries - specify how many z3 shell processes to spin up - benchmark on large multicore boxes to tune the z3 shell process count and optimal pipe buffer sizes.

Depends if this is something that should live in the API directory alongside an eventual C shell header, something for the examples directory, or something that should be it's own rust crate.

andrewvaughanj:
Sure, but this isn't an "API", right? This is taking [this example](https://doc.rust-lang.org/rust-by-example/std_misc/process/pipe.html) and changing from `wc` to `z3` (and removing the comments).

My point here would be that, if this was a response to a Z3 discussion or StackOverflow post (e.g., "how do I call Z3 from Rust?"), it is great; as a standalone Rust file inside of the Z3 source tree, I don't get the "value add" -- maybe I'm missing something?

NikolajBjorner:
I have to resist adding dependencies. They add liability down the road. 
I am not familiar with the Stackoverflow discussions, sorry can't follow everything.
Also new dependencies should be built and tested in the azure-pipeline infrastructure on the supported platforms. 

The Travis pipelines will go away soon (when Travis shuts down .org, as I don't have access to disabling the builds there). 
They have some features not integrated in the azure pipelines, but it is going to be necessary to scope the number of moving parts.

There is a separate project that adds Rust runtime bindings that was initiated some years ago and revived. If maintainers want it in the z3 build cycle they are welcome to consider adding it, but it has to be buildable in the CI process. There are also separate shell bindings for wasm.

chadbrewbaker:
Agreed that the api root directory should stay clean as possible. Perhaps an api/shell/ or examples/shell directory with language specific shell access? Setting up the pipes can take hours of frustration - many times leading to developers giving up and forking a new z3 shell process every query.

chadbrewbaker:
@NikolajBjorner agreed the CI shouldn't support every language toolchain. I would only add a CI test for the C shell interface. 

I have shell workers running at scale on AWS Lambda. Users can shim their own z3 shell binary that calls out to cloud workers.  On AWS Lambda itself the z3 shell binary can be uploaded as a "layer" without modification.

One other reason to have a C shell interface is test rigs, https://twitter.com/alastair_d_reid/status/1400824710234005505 .  Platforms like AWS Lambda don't have locked down worker cgroups, requiring removal/whitelisting of several system calls in the z3 shell binary to run untrusted SMT-LIB. I would like some automated guarantees.

NikolajBjorner:
> agreed the CI shouldn't support every language toolchain

That's not what I am trying to say. 
What I have in mind is that when adding code to z3 it should be part of the CI process.
You are adding Rust code. There is no build or test trigger for this in the pull request. 
It is easier to not integrate this dependency into z3prover/z3.git.

chadbrewbaker:
Closing for now. Thanks for the input.

PR #5368, opened 2021-06-24
scoped_timer: deadlock in thread pool based implementation since 4.8.10

zhouzhenghui:
Unlock immediately, not at destruct time, to avoid interactions between locks.


zhouzhenghui:
The PR is closed because I misunderstand the purpose of scoped_timer. I will propose another one soon.

PR #5371, opened 2021-06-25
scoped_timer: deadlock in thread pool based implementation since 4.8.10

zhouzhenghui:
When scoped_timer expires, the lock isn't released before it is returned to the thread pool, which could cause reenter deadlock and data race.

### DIFF
--- a/src/util/scoped_timer.cpp
+++ b/src/util/scoped_timer.cpp
@@ -51,6 +51,7 @@ static int working = WORKING;
 static void thread_func(scoped_timer_state *s) {
     workers.lock();
     while (true) {
+    start:
         s->cv.wait(workers, [=]{ return s->work > 0; });
         workers.unlock();
 
@@ -66,7 +67,7 @@ static void thread_func(scoped_timer_state *s) {
                     s->eh->operator()(TIMEOUT_EH_CALLER);
                     s->work = IDLE;
                     workers.lock();
-                    goto end;
+                    goto start;
                 }
                 goto next;
             }
@@ -77,8 +78,6 @@ static void thread_func(scoped_timer_state *s) {
     next:
         workers.lock();
         available_workers.push_back(s);
-
-    end:
     }
 }
 
--- a/src/util/scoped_timer.cpp
+++ b/src/util/scoped_timer.cpp
@@ -51,7 +51,6 @@ static atomic<unsigned> num_workers(0);
 static void thread_func(scoped_timer_state *s) {
     workers.lock();
     while (true) {
-    start:
         s->cv.wait(workers, [=]{ return s->work > IDLE; });
         workers.unlock();
 
@@ -71,8 +70,8 @@ static void thread_func(scoped_timer_state *s) {
         s->m_mutex.unlock();
 
     next:
+        s->work = IDLE;
         workers.lock();
-        available_workers.push_back(s);
     }
 }
 
--- a/src/util/scoped_timer.cpp
+++ b/src/util/scoped_timer.cpp
@@ -33,7 +33,7 @@ Revision History:
 #include <pthread.h>
 #endif
 
-enum scoped_timer_work_state { EXPIRED = -1, IDLE = 0, WORKING = 1, EXITING = 2 };
+enum scoped_timer_work_state { IDLE = 0, WORKING = 1, EXITING = 2 };
 
 struct scoped_timer_state {
     std::thread m_thread;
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
@nunoplopes can you check? (@regehr)

regehr:
I don't understand the patch yet, but it's definitely not acceptable style to mix enum values and regular integers in the "work" member variable.

zhouzhenghui:
The names of enum are used for improving the readability of the code, and I would like to keep the modifications to a minimum.
A new commit has been added to satisfy the style requirements, the specified values of enum are kept to conform to the common usage.

nunoplopes:
I'll review this patch next week. Sorry, really busy now.

A few buildbots have failed; these need to be fixed.

zhouzhenghui:
> 
> 
> I'll review this patch next week. Sorry, really busy now.
> 
> A few buildbots have failed; these need to be fixed.

Thanks, the build errors I checked seem irrelevant with the patch.

nunoplopes:
This patch is trying to fix 3 things at once:
1. adding threads back to the thread pool with a locked mutex. This is not ideal, but not problematic in practice. The scoped_timer constructor will hang a bit until the previous destructor is called.
2. Then we have a race, which I believe is real, in the scoped_timer destructor. A thread is added back to the pool with `work==0`, but as soon as the destructor releases the lock, the constructor could set `work` back to 1, and the destructor will hang "forever".
3. Finally, there's a race between calling the destructor & executing the timeout handler. The handler may run even if the destructor was already called. I don't think this is problematic.

Do you agree with this assessment or did I miss something? If you agree, then we should fix 2. only. The state machine may indeed require a new state, but the solution should be simpler that this patch.

zhouzhenghui:
Thanks for the response,

The original problem I met is deadlock in constructor of scoped_timer::imp at line 99 when try to lock a work thread taken from the thread pool. Here is my understand to the 3 things:

1.  As mentioned, if the constructor and the destructor are called in order or by different threads everything goes well. But deadlock may existed in nested scoped_timers, when an inner scoped_timer is constructing and it try to reuse a locked work thread of the outer scoped_timer which has expired and the lock will never be released.

2. I agree though I don't know if it really happened.

3. I'm not sure here, but it will be more safe since the less assumptions are requested.

src/util/scoped_timer.cpp
```diff
@@ -33,42 +33,50 @@ Revision History:
...
     workers.lock();
     while (true) {
-        s->cv.wait(workers, [=]{ return s->work > 0; });
+    start:
+        s->cv.wait(workers, [=]{ return s->work > IDLE; });
         workers.unlock();
 
         // exiting..
-        if (s->work == 2)
+        if (s->work == EXITING)
```
NikolajBjorner, line 60: @nunoplopes you have got to admit that using enums is so much better than those non-descriptive ints.

NikolajBjorner:
@nunoplopes - should we look into this?

nunoplopes:
Sorry for the delay. I've now reviewed this patch.
I like it in general, thanks.

However, I would like to ask you to revert the addition of the EXPIRED state. It is trying to solve the race between a timeout being triggered and the timeout handler being destroyed. This gives you a race where Z3 may have finished the computation and still trigger a timeout. I think it's not worth it to fix this race as it's benign; if you hit it, it just means that  some runs will timeout and some won't as the run time is around the timeout value. Hence it's not worth to slow down or complicate this code to fix this issue.
Moreover, the proposed fix for that issue is still racy. Between reading the state value and calling the handler the value could change. So the behavior is still racy and will always be.

The rest of the patch is good. If you could just remove this EXPIRED state, I'll merge this PR. Thank you!

NikolajBjorner:
@nunoplopes - seems there is no activity on the party that provided the pull request. Since it fixes a bug in your code, could you adapt it with your changes and integrate it?

nunoplopes:
> @nunoplopes - seems there is no activity on the party that provided the pull request. Since it fixes a bug in your code, could you adapt it with your changes and integrate it?

sure, I'll take over. Will test the fix & benchmark & commit.

zhouzhenghui:
> Sorry for the delay. I've now reviewed this patch. I like it in general, thanks.
> 
> However, I would like to ask you to revert the addition of the EXPIRED state. It is trying to solve the race between a timeout being triggered and the timeout handler being destroyed. This gives you a race where Z3 may have finished the computation and still trigger a timeout. I think it's not worth it to fix this race as it's benign; if you hit it, it just means that some runs will timeout and some won't as the run time is around the timeout value. Hence it's not worth to slow down or complicate this code to fix this issue. Moreover, the proposed fix for that issue is still racy. Between reading the state value and calling the handler the value could change. So the behavior is still racy and will always be.
> 
> The rest of the patch is good. If you could just remove this EXPIRED state, I'll merge this PR. Thank you!

Sorry for not paying attention in time. The patch was updated, thanks for the review!

nunoplopes:
Merged, thank you for the code & report and apologies for the long turnaround.

PR #5383, opened 2021-07-02
Remove size argument in OCaml's `Z3.mk_re_intersect`

Drup:
Passing the size as argument is unnecessary in OCaml, and that argument is abridged in all similar `Seq` functions (see `concat` a bit higher). This applies the same pattern.


Drup:
As a drive-by, I also fixed the documentation comments of `Z3.Seq` to actually be documentation comments.

PR #5386, opened 2021-07-04
fix #5384 (char.<= being applied to integers)

### LINKED ISSUE #5384: possible bug with regular expressions and (Seq Int)
Hello, I'm getting an unexpected error for the following code:
```
(define-fun good () (RegEx (Seq Int)) (re.++ (seq.to.re (seq.unit 3)) (as re.all (RegEx (Seq Int)))))
(assert (seq.in.re (seq.unit 3) good))
(check-sat)

(define-fun bad () (RegEx (Seq Int)) (re.++ (seq.to.re (seq.unit 4)) (as re.all (RegEx (Seq Int)))))
(assert (seq.in.re (seq.unit 3) bad))
(check-sat)
```
Z3 output:
```
sat
(error "line 6 column 36: incorrect first argument type Int")
sat
```
Note that if I change everything to `String` instead of `(Seq Int)`, and use the literals `"3"` and `"4"`, Z3 gives me the output `sat` followed by `unsat`, as expected.

jrmcclurg:


NikolajBjorner:
Based on what we do elsewhere, it should have used sort not ast.

PR #5389, opened 2021-07-06
fix is_char(s) being applied to non-sort s

jrmcclurg:
@NikolajBjorner one minor addition to your fix in
https://github.com/Z3Prover/z3/commit/bdcfba1324ec834bd194d28f017798fdbb390322
(`is_char` was being applied to an `expr*` rather than a `sort*`)


NikolajBjorner:
It is overloaded. I moved up the overload definition up to be next to the one on sort*.

Issue #1540, opened 2018-03-16
Duplicate __deepcopy__ function in python API


fmgoncalves:
The `ASTRef` class contains two definitions of the `__deepcopy__` function.

The first one, which is incidentally the oldest _and_ the correct one, is at [z3.py#L286](https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L286)

The second one, which does not conform to the deepcopy spec, is at 
[z3.py#L371](https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L371) and was introduced by 8dadd30db5e1f09739bcb877e9dc5fb3240b1cf9 

The consequence is only the second (wrong) one is actually registered with the class, so errors like [this](https://github.com/MAIAN-tool/MAIAN/issues/6) can happen.

If it is any help, this is a class of problems pylint identifies and warns against so it could be useful to run it against your python API.

PR to follow.

PR #5416, opened 2021-07-17
added regex simplification rules ~() = .+ and .+* = .*

veanes:
This fixes the wrong unsat result in the following example reported in issue #1540, 
here the condition 'a in (~())star' of the ite becomes trivially true: (ite (a in .star) (a = "") ...)
and the correct model a="" is generated

(declare-fun a () String)
(assert (str.in_re a (re.* (str.to_re "a"))))
(assert (str.in_re a (re.* (str.to_re "b"))))
(assert
 (ite (str.in_re a (re.* (re.comp (str.to_re ""))))
  (= a "")
  (str.in_re (str.++ "aaab" a)
   (re.*
    (re.++ (re.comp (re.* (str.to_re "b")))
     (str.to_re (str.from_int (str.len a))))))))
(check-sat)
(get-model)

### DIFF
--- a/src/ast/rewriter/seq_rewriter.cpp
+++ b/src/ast/rewriter/seq_rewriter.cpp
@@ -3950,6 +3950,7 @@ br_status seq_rewriter::mk_re_union(expr* a, expr* b, expr_ref& result) {
     comp(none) -> all
     comp(all) -> none
     comp(comp(e1)) -> e1
+    comp(epsilon) -> .+
 */
 br_status seq_rewriter::mk_re_complement(expr* a, expr_ref& result) {
     expr *e1 = nullptr, *e2 = nullptr;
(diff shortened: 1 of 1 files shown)


veanes:
The underlying bug is still there, the following variant gives also unsat, 
but is equivalent to the one above because ("u"*)&("v"*) = ""
Investigating it.
(declare-fun x () String)
(assert (str.in_re x (re.* (str.to_re "a"))))
(assert (str.in_re x (re.* (str.to_re "b"))))
(assert
 (ite (str.in_re x (re.* (re.comp (re.inter (re.* (str.to_re "u")) (re.* (str.to_re "v"))))))
  (= x "")
  (str.in_re (str.++ "aaab" x)
   (re.*
    (re.++ (re.comp (re.* (str.to_re "b")))
     (str.to_re (str.from_int (str.len x))))))))
(check-sat)
(get-model)

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3973,6 +3974,10 @@ br_status seq_rewriter::mk_re_complement(expr* a, expr_ref& result) {
         result = e1;
         return BR_DONE;
     }
+    if (re().is_to_re(a, e1) && str().is_empty(e1)) {
+        result = re().mk_plus(re().mk_full_char(a->get_sort()));
+        return BR_DONE;
```
NikolajBjorner, line 3979: other rewrites convert R+ to R R* so maybe not BR_DONE

Issue #5140, opened 2021-03-30
Invalid model in string formula


jiwonparc:
commit: c629f09f2
```
$ z3release model_validate=true small.smt2
sat
(error "line 13 column 10: an invalid model was generated")
$ cat small.smt2
(declare-fun a () String)
(assert
 (str.in_re a
 (re.++
  (re.diff
  (re.++ (re.* (re.union (str.to_re "a") (str.to_re "b"))) (str.to_re "b"))
  (re.++
   (re.inter
   (re.++ (re.++ (re.* (re.union (str.to_re "a") (str.to_re "b"))) (str.to_re "b")) (str.to_re a))
   (re.++ (re.* (re.union (str.to_re "a") (str.to_re "b")))))
   (str.to_re "a")))
  (str.to_re (ite (str.in_re a (re.union (str.to_re "b"))) "b" "a""b")))))
(check-sat)
```

jiwonparc:
commit: c629f09f2
simpler case
```
$ z3release model_validate=true small.smt2
sat
(error "line 7 column 10: an invalid model was generated")
(
 (define-fun b () String
  "")
 (define-fun a () String
  "z")
)
$ z3-4.8.10 model_validate=true small.smt2
sat
(
 (define-fun b () String
  "")
 (define-fun a () String
  "b")
)
$ cat small.smt2
(declare-fun a () String)
(declare-fun b () String)
(assert (str.in_re a
     (re.diff (re.union (re.* (str.to_re "b")) (str.to_re "z"))
     (re.++ (re.* (str.to_re "z")) (str.to_re b)))))
(assert (= 0 (str.len b)))
(check-sat)
(get-model)
```

NikolajBjorner:
@veanes @cdstanford : maybe connected with symbolic regexes. (str.to_re b) is symbolic.
Could you check?

jiwonparc:
commit: 1fc9a7b
```
$ z3release model_validate=true small.smt2
sat
(error "line 8 column 10: an invalid model was generated")
$ z3-4.8.10 model_validate=true small.smt2
sat
$ cat small.smt2
(declare-fun a () String)
(assert
 (str.in_re a
 (re.++
  (str.to_re
  (ite (str.in_re a (re.++ (str.to_re "\u\u\u\u\u") (re.* re.allchar))) a "\u\u\u\u\u"))
  (re.* re.allchar))))
(check-sat)
```

cdstanford:
[...]

PR #5431, opened 2021-07-25
added derivative support for (str.to_re (str.from_int ...))

veanes:
This fixes an unsound unsat result in one case discussed in issue #5140 
involving the use of (str.to_re (str.from_int (str.len x))) where x is uninterpreted.

### DIFF
--- a/src/ast/rewriter/seq_rewriter.cpp
+++ b/src/ast/rewriter/seq_rewriter.cpp
@@ -3339,6 +3339,26 @@ expr_ref seq_rewriter::mk_derivative_rec(expr* ele, expr* r) {
             //while mk_empty() = [], because deriv(epsilon) = [] = nothing
             return mk_empty();
         }
+        else if (str().is_itos(r1, r2))
+        {
+            //here r1 = (str.from_int r2) and r2 is non-ground 
+            //or else the expression would have been simplified earlier
+            //so r1 must be nonempty and must consists of decimal digits 
+            //'0' = char(48) <= elem <= char(57) = '9'
+            m_br.mk_and(u().mk_le(m_util.mk_char(48), ele), u().mk_le(ele, m_util.mk_char(57)), result);
+            if (m().is_false(result))
+                return mk_empty();
+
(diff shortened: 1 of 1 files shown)


src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3339,6 +3339,26 @@ expr_ref seq_rewriter::mk_derivative_rec(expr* ele, expr* r) {
             //while mk_empty() = [], because deriv(epsilon) = [] = nothing
             return mk_empty();
         }
+        else if (str().is_itos(r1, r2))
```
NikolajBjorner, line 3342: I don't see how it fixes the underlying issue. The default behavior is to return a derivative expression. It should be sound no matter what it is applied to. Adding this code just bypasses the default case so it seems to mask the bug, not outright address it for other scenarios. Suppose I introduce an uninterpreted function that has the same effect as itos. This code would go to the default case.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3339,6 +3339,26 @@ expr_ref seq_rewriter::mk_derivative_rec(expr* ele, expr* r) {
...
+                return mk_empty();
+
+            hd = str().mk_nth_i(r1, m_autil.mk_int(0));
+            m_br.mk_and(result, m().mk_eq(hd, ele), result);  // result := result & ele = (hd r1)
+            if (m().is_false(result))
+                return mk_empty();
+
+            tl = str().mk_substr(r1, m_autil.mk_int(1), m_autil.mk_sub(str().mk_length(r1), m_autil.mk_int(1)));
+            //if ((isdigit ele) and (ele = (hd r1))) then (to_re (tl r1)) else []
+            result = re_and(result, re().mk_to_re(tl));
```
NikolajBjorner, line 3359: re_and already performs constant folding when the condition is true/false. So having the caller perform constant folding is at best a micro-optimization with the effect of making the code more elaborate.
I

PR #5440, opened 2021-07-28
Need `-thread` option to compile OCaml example

master-q:
I can compile the OCaml example with `-thread` option at Linux.

```
$ ocaml --version
The OCaml toplevel, version 4.05.0
```


NikolajBjorner:
Thanks!

PR #5442, opened 2021-07-29
Add Ubuntu CMake Coverage CI step

0152la:
Adds an extra step to CI jobs which executes the Z3 test suite with
coverage enabled, and additionally executes coverage-enhancing tests
added to z3test.

The coverage test already integrated in z3test might fail without the other PR [1].

[1] https://github.com/Z3Prover/z3test/pull/28

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,6 +5,7 @@ variables:
   cmakeNet: '-DZ3_BUILD_DOTNET_BINDINGS=True'
   cmakePy:  '-DZ3_BUILD_PYTHON_BINDINGS=True'
   cmakeStdArgs: '-DZ3_BUILD_DOTNET_BINDINGS=True -DZ3_BUILD_JAVA_BINDINGS=True -DZ3_BUILD_PYTHON_BINDINGS=True -G "Ninja" ../'
+  cmakeCovArgs: '-DCMAKE_INSTALL_PREFIX=./install -G "Ninja" ../'
   asanEnv: 'CXXFLAGS="${CXXFLAGS} -fsanitize=address -fno-omit-frame-pointer" CFLAGS="${CFLAGS} -fsanitize=address -fno-omit-frame-pointer"'
   ubsanEnv: 'CXXFLAGS="${CXXFLAGS} -fsanitize=undefined" CFLAGS="${CFLAGS} -fsanitize=undefined"'
   msanEnv: 'CC=clang LDFLAGS="-L../libcxx/libcxx_msan/lib -lc++abi -Wl,-rpath=../libcxx/libcxx_msan/lib" CXX=clang++ CXXFLAGS="${CXXFLAGS} -stdlib=libc++ -fsanitize-memory-track-origins -fsanitize=memory -fPIE -fno-omit-frame-pointer -g -O2" CFLAGS="${CFLAGS} -stdlib=libc -fsanitize=memory -fsanitize-memory-track-origins -fno-omit-frame-pointer -g -O2"'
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
It runs and exposes a first issue with a memory leak.
It would be an issue with the C++ interface.

The repro is a bit shorter than cov1.cpp

```
    // We construct two equivalent expressions, by leveraging that fact that
    // `mod(x, y) = x - y * x / y` when using integer division
    z3::expr out_0 = z3::ite(in_1 == 0, in_1, z3::mod(in_1, in_1));
    out_0 = z3::ite(in_0 == 0, out_0, out_0); 
    return 0;
```

If removing the assignment to out_0 or assigning it into a different expression I don't get the leak.

The source of the bug seems to be a desire to use C++17 move semantics.
```
        ast & operator=(ast && s) noexcept {
            if (this != &s) {
                object::operator=(std::forward<object>(s));
                m_ast = s.m_ast;
                s.m_ast = nullptr;
            }
            return *this;
        }
```
s.m_ast is set to null and the reference counts for the two occurrences of out_0 are no longer accounted for.
I am going to remove this use of move and other uses should probably be purged. 

If you have bandwidth to either delta-debug or root cause such interface issues it would be great. 
The use of the move assignment was introduced by a different user who had a drive to modernize the interface.

NikolajBjorner:
Given that the coverage pipeline takes almost an hour to complete and is another moving part, 
I am attempting to migrate the coverage pipeline to a self-contained Github action.
The conventions for creating github actions are somewhat different and I am a newbie to these and probably will get paths and dependencies wrong for a while. 

The result of coverage should also be published in a format they can be inspected. I am unfamiliar with using the tool being invoked but presume it produces a file (or can be made to produce a file). 
Without doubt, but possibly with some hoops, the file can be published as an artifact or published with one of the mechanisms listed along with the beginner guide for GitHub actions.

It would be great if you can also lend a hand in the self-contained action and setting up an appropriate way to publish the results.

NikolajBjorner:
The GitHub action now runs the coverage test, so other functionality will be put / migrated to this.

I guess these are instructions for extracting reports:
https://github.com/mapbox/cpp/blob/master/docs/coverage.md

0152la:
Hi, Nikolaj. Sorry for the delay in responding, I am in the process of moving among other things, so things might be a bit slow at times until that finishes.

Regarding the move issue that was exposed, I will admit I don't have too much experience with C++17 move semantics right now, and as I am currently writing up my thesis, I'm a bit apprehensive to start looking at it right now. I will add it on my log, just in case some time frees up in the mid-term.

Regarding the github action, your initial commit seems reasonable. I can implement the remainder of producing the coverage repot on Monday. I haven't personally used `llvm-cov` or `codecov` as is indicated in the link you posted, but an option I am familiar with is `gcovr`. We can use with `--html-details` to produce something similar to [CVC5's coverage reports](https://cvc4.cs.stanford.edu/downloads/builds/coverage/cvc5-2021-07-31/) (I'm unsure of the exact command line, but I think `gcovr --html-details <output-folder> <z3-repo-dir>` might be sufficient). From my experience, these will take a fair amount to produce, maybe even as long as running the tests themselves, at least with `gcovr`; it could be worth trying this `llvm-cov` or `lcov`.

Alternatively, in the interest of time, just the front page, without per-file coverage could be produced with `gcovr --html` (similar to [this](https://www.doc.ic.ac.uk/~al2510/coverage/cvc52.html)), which should mean much less time spent producing the coverage artifact, as well as it being contained within a single file, rather than a folder.

If there will be no further changes, I'll submit the PR either today or tomorrow to produce the coverage report in the single-file variant and potentially mark  it as an actions artifact, if I can figure that out. Hopefully won't be to hard.

NikolajBjorner:
> Regarding the move issue that was exposed, I will admit I don't have too much experience with C++17 move semantics right now, and as I am currently writing up my thesis, I'm a bit apprehensive to start looking at it right now. 

Actually, this is taken care of: your coverage test exposed the issue. The fix I applied was to remove the code that used std::forward. What I meant to imply is that coverage tests may expose bugs. Some of these bugs have fairly simple repros and simple fixes.

> Regarding the github action, y ... If there will be no further changes, I'll submit the PR either today or tomorrow to produce th

Great! You own the token on what magical flags and parameters to add to the script. The first order is to simply save a file on the workflow locally. Uploading it is a separate step that I can probably RTFM on.

0152la:
> Actually, this is taken care of: your coverage test exposed the issue. The fix I applied was to remove the code that used std::forward. What I meant to imply is that coverage tests may expose bugs. Some of these bugs have fairly simple repros and simple fixes.

I understand now. I didn't notice that the test was exposing such an issue in my runs, that's why I didn't mention anything. I will definitely raise an issue if I notice something going wrong.

PR #5451, opened 2021-08-02
Improve coverage CI script

0152la:
* Due to the long duration of the CI execution, execute it at a set time
  daily (currently 11 UTC / 4 PDT)
* Use `Ninja` to build instead of Makefile, due to better compilation
  time
* Execute all the available Z3 tests and examples: `test-z3 -a`,
  `z3test` regression suites (`smt2`, `smt2-debug`, and `smt2-extra`),
  `z3test` coverage tests, and the 4 provided examples
* Upload `gcovr` report as an artifact associated with the CI run

TODOs:
* Fix `gcovr` emitting an empty report
* Potentially take the artifact and upload it somewhere accessible

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,34 +1,90 @@
 name: Code Coverage
 
 on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+  #push:
+    #branches: [ master ]
+  schedule:
+    - cron: "0 11 * * *"
 
 env:
   BUILD_TYPE: Debug
+  CMAKE_GENERATOR: Ninja
 
 jobs:
   build:
     runs-on: ubuntu-latest
 
+    env:
+      CC: clang
+      CXX: clang++
+
     steps:
     - uses: actions/checkout@v2
 
-    - name: Configure CMake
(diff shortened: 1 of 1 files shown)


0152la:
Unfortunately the core issue of emitting the coverage report I was unable to solve, as for some reason `gcovr` emits an empty report, when it works on a machine with the copy-pasted commands. As it's getting late, and I have to wait for a full compilation process for every small change, I'll have to leave this for tomorrow.

NikolajBjorner:
From what I can tell it compiles using clang, but coverage report says using gcc.

0152la:
Thanks for the pointer in your last message. That *should* be made compatible with `gcovr` by passing `--gcov-executable "llvm-cov gcov"`, or by setting the `GCOV` env variable to the same, as per the `gcovr` documentation [1]. I was able to replicate this on a local docker container, but not on the CI machine.

I've been trying various things on my fork here and there, including the instructions you mentioned [2], but at the moment, nothing seems to work to produce the report. Compiling with `clang` specific instrumentation seems to hang the Z3 build process for some reason. I'll be trying more stuff here and there as time permits, and will update if I have a breakthrough.

[1] https://gcovr.com/en/stable/guide.html#cmdoption-gcovr-gcov-executable
[2] https://github.com/mapbox/cpp/blob/master/docs/coverage.md

PR #5459, opened 2021-08-05
report which operator bit-blast failed on

jameysharp:
I was trying to use the bit-blast tactic without simplifying first and got an error message about using an unsupported operator, but the message didn't tell me which operator. I was only using a half-dozen different operators so I ended up digging through the z3 source code to figure out which one was the problem. A little more information in the error message would have saved me quite a bit of time, so here's a patch to enhance the error message in this case.

I've tested this patch using the following input (without my patch in #5458 applied):

```
(declare-const x (_ BitVec 2))
(assert (= (bvand x x) (_ bv3 2)))
(check-sat-using (and-then bit-blast sat))
(get-info :reason-unknown)
```


NikolajBjorner:
Thanks!

PR #5463, opened 2021-08-08
updates related to issue #5140

veanes:
Fixed some bugs in issue #5140 (work in progress, not all are fixed yet).

Added functionality for dealing with derivatives over nonground sequences s without introducing the mkderivative OP in the ast.
Essentially D(e,s) = ite(s!=()& first(s)=e, rest(s), [])
This avoids stuck-situations -- that I believe lead to unsound unsat results in some cases.

Also, if s in uninterpreted, then 
D(e,reverse(s)) = ite(s!=()& last(s)=e, reverse(butlast(s)), [])

Currently:
butlast(s) = substr(s,0,|s|-1) 
and the pattern butlast(butlast(s)) is simplified to substr(s,0,|s|-2)

Similarly:
rest(s) = substr(s,1,|s|-1) 
rest(rest(s)) = ..TBD.. =  substr(s,2,|s|-2)

first(rest(s)) = nth(2,s) 

etc..

Fixed also some rex:pp issues (still some TBD).

### DIFF
--- a/src/ast/rewriter/seq_rewriter.cpp
+++ b/src/ast/rewriter/seq_rewriter.cpp
@@ -14,6 +14,7 @@ Author:
     Nikolaj Bjorner (nbjorner) 2015-12-5
     Murphy Berzish 2017-02-21
     Caleb Stanford 2020-07-07
+    Margus Veanes 2021
 
 --*/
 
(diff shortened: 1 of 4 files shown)


src/ast/seq_decl_plugin.cpp
```diff
@@ -1068,6 +1069,10 @@ std::ostream& seq_util::rex::pp::compact_helper_seq(std::ostream& out, expr* s)
         for (expr* e : es)
             compact_helper_seq(out, e);
     }
+    else if (re.u.str.is_string(s, z)) {
+        for (int i=0; i < z.length(); i++)
```
NikolajBjorner, line 1073: space i = 0

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -1912,11 +1967,63 @@ br_status seq_rewriter::mk_seq_replace_all(expr* a, expr* b, expr* c, expr_ref&
...
+        return false;
+    return true;
+}
+
+/*
+* Replace all subvectors of b in a by c
+*/
+void seq_rewriter::replace_all_subvectors(expr_ref_vector const& a, expr_ref_vector const& b, expr* c, expr_ref_vector& result) {
+    int i = 0;
+    int k = b.size();
```
NikolajBjorner, line 2010: use unsigned for i, k, j, b.size() is unsigned.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -1912,11 +1967,63 @@ br_status seq_rewriter::mk_seq_replace_all(expr* a, expr* b, expr* c, expr_ref&
         result = str().mk_concat(strs, a->get_sort());
         return BR_REWRITE_FULL;
     }
-    // TBD: add case when a, b are concatenation of units that are values.
-    // in this case we can use a similar loop as for strings.
+    expr_ref_vector a_vals(m());
+    expr_ref_vector b_vals(m());
+    if (try_get_unit_values(a, a_vals) && try_get_unit_values(b, b_vals))
+    {
```
NikolajBjorner, line 1973: { after if is not on separate line (contrast to typical C# conventions).

veanes: fixed this and the ones above in the following commit

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -872,6 +873,65 @@ br_status seq_rewriter::mk_seq_length(expr* a, expr_ref& result) {
...
 }
 
+/*
+*  In general constructs nth(t,0) but if t = substring(s,j,..) then simplifies to nth(s,j)
+*  This method assumes that |t| > 0.
+*/
+expr_ref seq_rewriter::mk_seq_first(expr* t) {
+    expr_ref result(m());
+    expr* s, * j, * k;
+    if (str().is_extract(t, s, j, k))
```
NikolajBjorner, line 883: Is this sound?

if j < 0, then extract evaluates to an empty string, then the result should be the same as nth("",0). But you pass nth(s, j) which is under-specified. The rewrite would then restrict interpretations for nth(s, j) for j negative to nth("",0).

A further potential issue is the use of nth_i, which is the interpreted version of nth. It gets introduced as a rewrite nth(s, k) = if 0 <= k < len(s) then nth_i(s, k) else nth_u(s, k).

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -872,6 +873,65 @@ br_status seq_rewriter::mk_seq_length(expr* a, expr_ref& result) {
...
+}
+
+/*
+*  In general constructs substring(t,1,|t|-1) but if t = substring(s,j,k) then simplifies to substring(s,j+1,k-1)
+*  This method assumes that |t| > 0.
+*/
+expr_ref seq_rewriter::mk_seq_rest(expr* t) {
+    expr_ref result(m());
+    expr* s, * j, * k; 
+    expr* one = m_autil.mk_int(1);
```
NikolajBjorner, line 897: Pin the reference for uniformity:

expr_ref one(m_autil.mk_int(1), m());

src/ast/rewriter/seq_rewriter.cpp
veanes: the constraint of the use of mk_seq_first (in the derivation) is that the context implies is that |t|>0 which should imply that 0 <= j < |t| and also that j + k <= |t|. 
This is because mk_seq_first(t) always occurs in conjunction with not(t = ()).
This is why I dropped the additional checks I had originally.

src/ast/rewriter/seq_rewriter.cpp
veanes: ok, done

veanes:
In the last commit, I added also a missing case for get_re_head_tail_reversed but disabled it for now because it should not affect correctness but hides a bug I'm chasing. 
The bug I'm looking at right now is incorrect model generated for 
x in "BA".(ite (x in .*"AB") "B"  "A")"AA"
the bug happens after (check-sat) which seems to mess around with if constants remain uninterpreted.

veanes:
the last commit fixes 2 bugs
1) ite concat re-order bug in seq_rewriter::mk_der_op_rec
2) max_length bug in str::max_length

Also enables a missing case in get_re_head_tail_reversed that trteats e.g. (.*A)B 
same as .*(AB) by lifting out the tail "AB" instead of just the tail "B"

PR #5475, opened 2021-08-12
Add AssertSoft String overload for Java Api

intrigus-lgtm:
This adds a String overload for AssertSoft.
Previously only integer weights could have been used,
limiting the user. With Strings the user can now use
e.g. Java's BigInteger class converted to a String instead.


intrigus-lgtm:
Note that I'm not really familiar with Z3, but it looks like non-`int` weights should work.
I also did not try to compile everything of Z3 to test my change. I hope any potential problems will be caught by CI.

PR #5483, opened 2021-08-16
Update coverage github action

0152la:
I've managed to fix up the issue. I was using pre-installed `clang`, which seems to have been incompatible with `llvm` from the repos. In addition, I've added the generation of a detailed report to be produced and uploaded. You can check a sample on my fork [1]. They're zipped, as that's what the action does, but they could be further unzipped and placed somewhere accessible, via a subsequent dependent job and using the `download-artifact` action [2], as they're `html` files.

The total execution time seems to be around the hour mark, and I've added everything that I found sensible to be executed for coverage. This means: `z3-test -a`, the four examples, `smt2`, `smt2-debug`, and `smt2-extra` regression suites from `z3test`, and the coverage tests from `z3test` (to which I'll add more PRs for shortly).

[1] https://github.com/0152la/z3/actions/runs/1124141648
[2] https://github.com/actions/download-artifact

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -6,10 +6,6 @@ on:
   schedule:
     - cron: "0 11 * * *"
 
-env:
-  BUILD_TYPE: Debug
-  CMAKE_GENERATOR: Ninja
-
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
It is great. The artifacts are accessible from the same place the action is specified. This makes it transparent.
There are files that aren't covered. For example there are no entry points into smt_induction. So I removed this from the build.
Other files are heavily hit and then others are sparsely. The tools that automatically produce inputs for higher coverage in the sparser sets are useful and it would be great to accumulate the effects of these tools in the code that gets run by Coverage. 
The timeout is longer than an hour for Azure functions...

PR #5496, opened 2021-08-22
rewrite bit-blaster multiplier

jameysharp:
This version handles more special cases efficiently, with less code, and introduces some basic unit tests.

I deleted a bunch of code that wasn't being invoked, specifically the Wallace tree multiplier and Booth constant multiplier. As far as I can tell, there was no existing way to turn either of these on without recompiling z3, and they were not turned on at either use site.

Instead of the inline Wallace adder, I isolated into a separate class the choice of how to add together many terms. I hope that class will be reused by the rest of the bit-blaster at some point, because it could produce a smaller expression graph for bit-vector problems containing linear operations, especially when some bits are constant.

I think it was good that the Booth constant multiplier was not turned on, because it is sometimes worse than long multiplication. For example, multiplying by 010101... adds a term for every bit after Booth encoding, when it only needs a term every other bit in long multiplication. The commented-out radix-4 Booth multiplier isn't better, because although it would handle that example correctly, in a case like 001001001... it still needs three terms per six bits, instead of two.

But using non-adjacent form (NAF) instead is always at least as good as long multiplication and also at least as good as either form of Booth encoding, so it's worth doing unconditionally.

Optimizing multiplication by constants also subsumes the check for multiplication by -1, so I deleted that too. That optimization falls out automatically from either NAF or Booth encoding, along with other cases such as the negation of any power of two.

I also deleted the constant case multiplier, which was called but its result was ignored: it always returned false, indicating the caller should skip it and complete the multiply a different way.

I tried just fixing up the constant case multiplier, and did some experiments with different implementations. I got worse results in some cases, and in other cases I think more general optimizations in either the adder or the boolean rewriter will get the same result. So I think at best it's not worth the extra code complexity; at worst it might be actively harmful.

### DIFF
--- a/src/ast/rewriter/bit_blaster/bit_blaster_adder.cpp
+++ b/src/ast/rewriter/bit_blaster/bit_blaster_adder.cpp
@@ -1,0 +1,112 @@
+#include "ast/rewriter/bit_blaster/bit_blaster_adder.h"
+
+bit_blaster_adder::bit_blaster_adder(bool_rewriter & rewriter, unsigned sz, numeral const & value):
+    m_rewriter(rewriter),
+    m_constant(value)
+{
+    reduce();
+    expr_ref_vector empty(m());
+    m_variable.resize(sz, empty);
+}
+
+bit_blaster_adder::bit_blaster_adder(bool_rewriter & rewriter, unsigned sz, expr * const * bits):
+    bit_blaster_adder(rewriter, sz)
+{
+    for (unsigned i = 0; i < sz; i++)
+        add_bit(i, bits[i]);
+}
+
+expr_ref bit_blaster_adder::sum_bits(vector< expr_ref_vector > & columns, expr_ref_vector & out_bits) const {
+    SASSERT(out_bits.empty());
+
+    expr_ref_vector carries(m());
+    expr_ref tmp1(m()), tmp2(m()), tmp3(m());
+
+    for (auto & column : columns) {
+        column.append(carries);
+        carries.reset();
+
+        while (column.size() >= 3) {
+            expr * a = column.back();
+            column.pop_back();
+            expr * b = column.back();
+            column.pop_back();
+            expr * c = column.back();
+            column.pop_back();
+
+            m_rewriter.mk_xor(a, b, tmp1);
+            m_rewriter.mk_xor(tmp1, c, tmp2);
+            column.push_back(tmp2);
+
+            m_rewriter.mk_and(a, b, tmp2);
+            m_rewriter.mk_and(tmp1, c, tmp3);
+            // tmp2 and tmp3 can't be true at the same time, so use
+            // whichever of mk_or vs mk_xor makes the most sense here.
+            m_rewriter.mk_or(tmp2, tmp3, tmp1);
+            carries.push_back(tmp1);
+        }
+
+        if (column.size() == 2) {
+            expr * a = column.back();
+            column.pop_back();
+            expr * b = column.back();
+            column.pop_back();
+
+            m_rewriter.mk_xor(a, b, tmp1);
+            column.push_back(tmp1);
+
+            m_rewriter.mk_and(a, b, tmp1);
+            carries.push_back(tmp1);
+        }
+
+        out_bits.push_back(column.get(0, m().mk_false()));
+    }
+
+    SASSERT(out_bits.size() == size());
+
+    // We return the carry separately in case the caller wants it.
+    tmp1 = m().mk_false();
+    for (auto & carry : carries) {
+        m_rewriter.mk_xor(tmp1, carry, tmp2);
+        tmp1 = tmp2;
+    }
+    return tmp1;
+}
+
+expr_ref bit_blaster_adder::variable_bits(expr_ref_vector & out_bits) const {
+    vector< expr_ref_vector > columns(m_variable);
+    return sum_bits(columns, out_bits);
+}
+
+expr_ref bit_blaster_adder::total_bits(expr_ref_vector & out_bits) const {
+    vector< expr_ref_vector > columns(m_variable);
+
+    expr_ref one(m());
+    one = m().mk_true();
+    for (unsigned i = 0; i < size(); i++)
+        if (m_constant.get_bit(i))
+            columns[i].push_back(one);
+
+    expr_ref carry(m());
+    carry = sum_bits(columns, out_bits);
+    if (m_constant.get_bit(size())) {
+        m_rewriter.mk_not(carry, one);
+        carry = one;
+    }
+    return carry;
+}
+
+bit_blaster_adder & bit_blaster_adder::add_shifted(bit_blaster_adder const & other, unsigned shift) {
+    add_shifted(other.m_constant, shift);
+    for (unsigned i = 0; shift + i < size(); i++)
+        m_variable[shift + i].append(other.m_variable[i]);
+    return *this;
+}
+
+bit_blaster_adder & bit_blaster_adder::add_shifted(unsigned sz, expr * const * bits, unsigned shift) {
+    if (sz > size() - shift)
+        sz = size() - shift;
+    for (unsigned i = 0; i < sz; i++)
+        add_bit(shift + i, bits[i]);
+    return *this;
+}
--- a/src/ast/rewriter/bit_blaster/bit_blaster_adder.h
+++ b/src/ast/rewriter/bit_blaster/bit_blaster_adder.h
@@ -1,0 +1,98 @@
+#pragma once
+
+#include "ast/ast.h"
+#include "ast/rewriter/bool_rewriter.h"
+#include "util/rational.h"
+
+class bit_blaster_adder {
+public:
+    typedef rational numeral;
+
+    bit_blaster_adder(bool_rewriter & rewriter, unsigned sz, numeral const & value = numeral::zero());
+    bit_blaster_adder(bool_rewriter & rewriter, unsigned sz, expr * const * bits);
+
+    bit_blaster_adder(bool_rewriter & rewriter, expr_ref_vector & value):
+        bit_blaster_adder(rewriter, value.size(), value.data()) {}
+
+    bit_blaster_adder(bit_blaster_adder const & other):
+        bit_blaster_adder(other.m_rewriter, other.size()) {
+        *this += other;
+    }
+
+    bit_blaster_adder(bit_blaster_adder &&) noexcept = default;
+
+    // The *_bits functions all return the final carry bit out of the addition;
+    // just ignore it if you don't need it.
+
+    // Return the sum of the known-constant inputs to this adder.
+    bool constant_bits(numeral & value) const {
+        value = m_constant % power(size());
+        return m_constant.get_bit(size());
+    }
+
+    // Return the sum of the non-constant inputs to this adder.
+    expr_ref variable_bits(expr_ref_vector & out_bits) const;
+
+    // Return the sum of all inputs to this adder.
+    expr_ref total_bits(expr_ref_vector & out_bits) const;
+
+    unsigned size() const {
+        return m_variable.size();
+    }
+
+    ast_manager & m() const {
+        return m_rewriter.m();
+    }
+
+    bit_blaster_adder & operator+=(bit_blaster_adder const & other) {
+        SASSERT(size() == other.size());
+        return add_shifted(other, 0);
+    }
+
+    bit_blaster_adder & add_bit(unsigned idx, bool bit) {
+        SASSERT(idx < size());
+        if (bit)
+            m_constant += power(idx);
+        return *this;
+    }
+
+    bit_blaster_adder & add_bit(unsigned idx, expr * bit) {
+        SASSERT(idx < size());
+        if (m().is_true(bit))
+            m_constant += power(idx);
+        else if (!m().is_false(bit))
+            m_variable[idx].push_back(bit);
+        return *this;
+    }
+
+    bit_blaster_adder & add_bit(unsigned idx, expr_ref & bit) {
+        return add_bit(idx, bit.get());
+    }
+
+    bit_blaster_adder & add_shifted(bit_blaster_adder const & other, unsigned shift);
+    bit_blaster_adder & add_shifted(unsigned sz, expr * const * bits, unsigned shift);
+
+    bit_blaster_adder & add_shifted(expr_ref_vector const & bits, unsigned shift) {
+        return add_shifted(bits.size(), bits.data(), shift);
+    }
+
+    bit_blaster_adder & add_shifted(numeral const & value, unsigned shift) {
+        m_constant += value * power(shift);
+        reduce();
+        return *this;
+    }
+
+protected:
+    bool_rewriter & m_rewriter;
+    numeral m_constant;
+    vector< expr_ref_vector > m_variable;
+
+    numeral power(unsigned n) const { return numeral::power_of_two(n); }
+
+    void reduce() {
+        // keep one extra bit in case somebody wants the final carry bit
+        m_constant %= power(size() + 1);
+    }
+
+    expr_ref sum_bits(vector< expr_ref_vector > & columns, expr_ref_vector & out_bits) const;
+};
--- a/src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
+++ b/src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
@@ -20,6 +20,7 @@ Revision History:
 #include "util/common_msgs.h"
 #include "ast/rewriter/bit_blaster/bit_blaster_tpl.h"
 #include "ast/ast_pp.h"
+#include "ast/rewriter/bool_rewriter.h"
 #include "ast/rewriter/rewriter_types.h"
 
 
@@ -29,17 +30,6 @@ void bit_blaster_tpl<Cfg>::checkpoint() {
         throw rewriter_exception(Z3_MAX_MEMORY_MSG);
     if (!m().inc())
         throw rewriter_exception(m().limit().get_cancel_msg());
-}
-
-/**
-   \brief Return true if all bits are true or false.
-*/
-template<typename Cfg>
-bool bit_blaster_tpl<Cfg>::is_numeral(unsigned sz, expr * const * bits) const {
-    for (unsigned i = 0; i < sz; i++)
-        if (!is_bool_const(bits[i])) 
-            return false;
-    return true;
 }
 
 /**
@@ -55,17 +45,6 @@ bool bit_blaster_tpl<Cfg>::is_numeral(unsigned sz, expr * const * bits, numeral 
         else if (!m().is_false(bits[i]))
             return false;
     }
-    return true;
-}
-
-/**
-   \brief Return true if all bits are true.
-*/
-template<typename Cfg>
-bool bit_blaster_tpl<Cfg>::is_minus_one(unsigned sz, expr * const * bits) const {
-    for (unsigned i = 0; i < sz; i++)
-        if (!m().is_true(bits[i]))
-            return false;
     return true;
 }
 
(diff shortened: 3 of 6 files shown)


src/ast/rewriter/bit_blaster/bit_blaster_adder.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    expr_ref_vector carries(m());
+    expr_ref tmp1(m()), tmp2(m()), tmp3(m());
+
+    for (auto & column : columns) {
+        column.append(carries);
+        carries.reset();
+
+        while (column.size() >= 3) {
+            expr * a = column.back();
```
NikolajBjorner, line 30: in principle, the column owns the reference count on expressions. Since you pop and push expressions on the column the reference count gets stale at times.
That is, after popping expr * a, you add tmp2 to the back of column, in the next iteration you pop tmp2 and shrink the column. 
The reference to tmp2, or a, is then stale after the pop_back(). 
It might be valid still because tmp2 is declared outside of the loop. 
Though this seems too brittle and subtle when later having to revisit this code.

We could simply delay popping the column and update it in place by maintaining a pointer to the current virtual size.

So:

unsigned sz = columns.size();
while (sz >= 3) {
    expr* a = columns.get(sz -1);
    expr* b = columns.get(sz - 2);
    expr* c = columns.get(sz - 3);
    tmp1 = mk_xor(a, b);
    tmp2 = mk_xor(tmp1, c);
    carries.push_back(mk_or(mk_and(a, b), mk_and(tmp1, c)));
    columns[sz-3] = tmp2;
    sz -= 3;
}

etc.

src/ast/rewriter/bit_blaster/bit_blaster_adder.cpp
```diff
@@ -0,0 +1,112 @@
+#include "ast/rewriter/bit_blaster/bit_blaster_adder.h"
```
NikolajBjorner, line 1: Even as the Contribution license already covers it some OSS users ask for copyright in all files. 
Would be useful to also include acknowledgement of you as author.

src/ast/rewriter/bit_blaster/bit_blaster_adder.cpp
```diff
@@ -0,0 +1,112 @@
...
+            column.pop_back();
+
+            m_rewriter.mk_xor(a, b, tmp1);
+            column.push_back(tmp1);
+
+            m_rewriter.mk_and(a, b, tmp1);
+            carries.push_back(tmp1);
+        }
+
+        out_bits.push_back(column.get(0, m().mk_false()));
```
NikolajBjorner, line 62: correct, but subtle.
Why not explicit case on sz?

switch (sz) {
case 0:
case 1:
case 2:
default:
....
}

src/ast/rewriter/bit_blaster/bit_blaster_adder.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        while (column.size() >= 3) {
+            expr * a = column.back();
+            column.pop_back();
+            expr * b = column.back();
+            column.pop_back();
+            expr * c = column.back();
+            column.pop_back();
+
+            m_rewriter.mk_xor(a, b, tmp1);
```
NikolajBjorner, line 37: this format is tedious. I finally added ability to tmp1 = m_rewriter.mk_or(a, b);

src/ast/rewriter/bit_blaster/bit_blaster_adder.h
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    bit_blaster_adder & add_shifted(numeral const & value, unsigned shift) {
+        m_constant += value * power(shift);
+        reduce();
+        return *this;
+    }
+
+protected:
+    bool_rewriter & m_rewriter;
```
NikolajBjorner, line 86: hopefully the m_rewriter that gets passed in is not configured with "flat" set to "true".
There was a long latent bug where the bit-blaster would convert into circuits but break all structure sharing by setting flat to true.

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
```diff
@@ -178,175 +157,70 @@ void bit_blaster_tpl<Cfg>::mk_subtracter(unsigned sz, expr * const * a_bits, exp
 
 template<typename Cfg>
 void bit_blaster_tpl<Cfg>::mk_multiplier(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & out_bits) {
-    SASSERT(sz > 0);
```
NikolajBjorner, line 181: yes, we tried several things over time.
It is right, we might as well kick this out.

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
```diff
@@ -178,175 +157,70 @@ void bit_blaster_tpl<Cfg>::mk_subtracter(unsigned sz, expr * const * a_bits, exp
...
+    bit_blaster_adder b(rewriter, sz, b_bits);
+
+    numeral a_const, b_const;
+    expr_ref_vector a_var(m()), b_var(m());
+    a.constant_bits(a_const);
+    b.constant_bits(b_const);
+    a.variable_bits(a_var);
+    b.variable_bits(b_var);
+
+    // Then this multiplication can be split into four terms:
```
NikolajBjorner, line 178: @msoeken can you check this out?

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
```diff
@@ -178,175 +157,70 @@ void bit_blaster_tpl<Cfg>::mk_subtracter(unsigned sz, expr * const * a_bits, exp
...
-            expr_ref_vector & pp = pps[i];
-            expr_ref t(m());
-            for (unsigned j = 0; j < i; j++)
-                pp.push_back(zero); // left shift by i bits
-            for (unsigned j = 0; j < (sz - i); j++) {
-                mk_and(a_bits[j], b_bits[i], t);
-                pp.push_back(t);
-            }
+    // TODO: share the bool_rewriter with the rest of the bit-blaster
+    bool_rewriter rewriter(m());
```
NikolajBjorner, line 163: we should really check if setting the parameter "flat" to false helps 
in this case:

```
params_ref p;
p.set_bool("flat", false);
rewriter.updt_params(p);
```

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
```diff
@@ -178,175 +157,70 @@ void bit_blaster_tpl<Cfg>::mk_subtracter(unsigned sz, expr * const * a_bits, exp
...
+    // TODO: allow callers to pass adders instead of just bits
+    bit_blaster_adder a(rewriter, sz, a_bits);
+    bit_blaster_adder b(rewriter, sz, b_bits);
+
+    numeral a_const, b_const;
+    expr_ref_vector a_var(m()), b_var(m());
+    a.constant_bits(a_const);
+    b.constant_bits(b_const);
+    a.variable_bits(a_var);
+    b.variable_bits(b_var);
```
NikolajBjorner, line 176: a_var, b_var seem not just variables + entries with constants replaced by false. The carry bit that gets returned is ignored. A comment on what it produces would be useful here.

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
```diff
@@ -178,175 +157,70 @@ void bit_blaster_tpl<Cfg>::mk_subtracter(unsigned sz, expr * const * a_bits, exp
...
-            pps.shrink(save_inx);
-        }        
+        // Once we see a prefix of b_var that's all 0, later prefixes will be
+        // shorter and therefore also all 0.
+        if (b_zero)
+            break;
+    }
 
-        SASSERT(pps.size() == 2);
+    // TODO: return total so further linear operations can constant-fold more
```
NikolajBjorner, line 216: intriguing

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
```diff
@@ -1181,164 +1055,43 @@ void bit_blaster_tpl<Cfg>::mk_comp(unsigned sz, expr * const * a_bits, expr * co
...
-void bit_blaster_tpl<Cfg>::mk_carry_save_adder(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr * const * c_bits, expr_ref_vector & sum_bits, expr_ref_vector & carry_bits) {    
-    expr_ref t(m());
-    for (unsigned i = 0; i < sz; i++) {            
-        mk_xor3(a_bits[i], b_bits[i], c_bits[i], t);
-        sum_bits.push_back(t);
-        mk_carry(a_bits[i], b_bits[i], c_bits[i], t);
-        carry_bits.push_back(t);
-    }
-}
+void bit_blaster_tpl<Cfg>::mk_const_multiplier(numeral & a, expr_ref_vector & b_bits, bit_blaster_adder & result) {
```
NikolajBjorner, line 1058: maybe name could be improved. The function _accumulates_ on to result.

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
NikolajBjorner: Though, for a careful merge it is better to have side-by-side with old version and run QF_BV benchmarks fully to evaluate effect.
I can do this easily on our regression infrastructure.

nunoplopes:
Do you have perf number for a collection of benchmarks?

NikolajBjorner:
> Do you have perf number for a collection of benchmarks?

I don't understand this question:
- you have benchmarks from Alive2
- we test against QF_BV and other categories and can compare
Comparing against QF_BV always gates changes with wide consequences like these

nunoplopes:
> > Do you have perf number for a collection of benchmarks?
> 
> I don't understand this question:
> 
> * you have benchmarks from Alive2
> * we test against QF_BV and other categories and can compare
>   Comparing against QF_BV always gates changes with wide consequences like these

Agreed, but there's also a burden on the author side to show that it's useful to waste my time (and money, as VMs aren't free) running benchmarks.

NikolajBjorner:
This is a serious pull request and the point about separating constant from non-constant bits is very sweet.

The biggest downstream issue with contributions is that authors disappear and cant be bothered when bug reports trickle in. 
The author of this pull request already included what appears to be adequate unit testing.
It is not much overhead to set up QF_BV comparison. It is currently running.

There are still last mile issues, such as feature interaction that authors of external pull requests can't know about up front (such as flattening).

jameysharp:
Thank you for the detailed review comments! Also I see the regression tests failed in CI, so I need to dig into that.

src/test/bit_blaster.cpp
```diff
@@ -36,38 +39,153 @@ void mk_bits(ast_manager & m, char const * prefix, unsigned sz, expr_ref_vector
...
-//     expr_ref_vector b(m);
-//     expr_ref_vector c(m);
-//     mk_bits(m, "a", sz, a);
-//     mk_bits(m, "b", sz, b);
-//     bool t = true;
-//     bit_blaster blaster(m, t);
-//     blaster.mk_adder(sz, a.c_ptr(), b.c_ptr(), c);
-//     TRACE("bit_blaster", display(tout, c););
+static unsigned to_int(model_core & mdl, expr_ref_vector & out) {
+    SASSERT(out.size() <= sizeof unsigned * 8);
```
NikolajBjorner, line 52:
```
  ../src/test/bit_blaster.cpp:52:5: note: in expansion of macro ‘SASSERT’
   52 |     SASSERT(out.size() <= sizeof unsigned * 8);

```

NikolajBjorner:
Build breaks here

```
../src/test/bit_blaster.cpp:52:5: note: in expansion of macro ‘SASSERT’
   52 |     SASSERT(out.size() <= sizeof unsigned * 8);

```

missing parenthesis in debug build

NikolajBjorner:
And regression, yes

```
Testing z3test/regressions/smt2/3570.smt2
EXPECTED
unsat

======================
PRODUCED
sat
(error "line 4 column 22: an invalid model was generated")

```

and some crash/craches

NikolajBjorner:
The regression tests already establish that there are bugs.
To give an idea, this is what the perf evaluation looks like

image

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
msoeken: @NikolajBjorner Sure, I can check this out until the end of this week.

wintersteiger:
The Wallace tree and Booth multipliers were experiments, they didn't have a significant impact on the competition benchmarks though, so they were never enabled by default. It's fine to remove them, but we should be extra careful that the new code doesn't introduce new bugs and that it at least maintains the average level of performance on all benchmarks we have.  @NikolajBjorner can you extract a CSV from our benchmarking infrastructure so @jameysharp can see which benchmarks throw errors?

jameysharp:
I wasn't going to worry about benchmark errors until I have the regression test errors fixed, since I don't need your internal infrastructure to run those. I haven't actually figured out how to run them myself yet, but at minimum I can push and let CI run them for me.

Relatedly, one request: could you document how to run the various tests? I only figured out the unit tests after digging through the CI configuration, which I didn't know how to read, so I did a lot of guesswork and reading generated makefiles.

I also didn't realize that I should be testing a Debug build, which is how the syntax error in that `SASSERT` slipped past, so that would be nice to have documented too. `README-CMake.md` explains how to get a Debug build, but doesn't explain that it's important during testing.

And I'd meant to ask about the comment block at the beginning of each file; it wasn't clear to me what copyright holder to declare in it. I guess the CLA is not a copyright assignment, just a license agreement, so I'm still the copyright holder, not Microsoft?

You could cover all this stuff in a [`CONTRIBUTING.md`](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) file so GitHub reminds new contributors like me to look at it when opening an issue or pull request.

src/ast/rewriter/bit_blaster/bit_blaster_adder.cpp
jameysharp: I've fixed this by changing a/b/c from `expr *` to `expr_ref`. That seems to have fixed many of the regression tests that failed before, so I guess the reference counting bug was a key source of the failures.

src/ast/rewriter/bit_blaster/bit_blaster_adder.h
jameysharp: I've added a call to `rewriter.set_flat(false)`. I'm not sure how to evaluate what impact that had on memory usage, but the unit tests I wrote went from reporting `:after-memory 0.13` to `:after-memory 0.02`, so I hope that's a positive sign?

NikolajBjorner:
crashes are gone, but there is a correctness bug or bugs.
This should be possible to catch with unit tests that try all 3^16 multiplier combinations by multiplying two 8 bit vectors.
4 or so bits also likely will do.

NikolajBjorner:
> @NikolajBjorner can you extract a CSV from our benchmarking infrastructure so @jameysharp can see which benchmarks throw errors?

I think the regression tests are more than sufficient at this point.
It is too bad we don't already have a good unit test to catch it.

jameysharp:
It turns out that I accidentally had the constructor for `bit_blaster_adder` zeroing the constant part when it was initialized from a constant. The culprit was something I intended as a memory optimization but it just made the code harder to reason about, so I've deleted it. At this point the regression tests pass for me both locally and in CI.

I wrote a bunch of fancy code in the process of creating this pull request, including a Dadda tree adder and a rewritten constant-case multiplier (those are in other branches on my fork), but at this point I think the implementation is approximately the simplest thing that could do what I set out to do.

@NikolajBjorner, could you do another benchmarking run on the `QF_BV` benchmarks? I've made two changes which I'd expect to affect performance: `rewriter.set_flat(false)` which is roughly what you suggested; and eliminating some vector copies that I'd used to make the adder API more purely functional, which it doesn't need to be.

If this still has worse performance on the benchmarks, then I would appreciate a list of some troublesome cases. I'm happy to go do some profiling.

I'm not sure how to approach the future work I was alluding to in the comment at the end of `mk_multiplier`, but the general idea is this: Currently the bit-blaster rewrites operand expressions into vectors of bits, bit-blasts the operation, and then clumps the resulting vector back into a temporary expression, to be extracted again when bit-blasting another operator that needs the result. If the intermediate clump were instead a `bit_blaster_adder`, then sequences of additive operations like `bvadd`, `bvmul`, and `bvneg` could 1) preserve more information about which bits are known constants, or sums of constants and expressions, and 2) sometimes generate more efficient adder networks by choosing the best structure to add three or more terms at once. Non-additive ops like `bvxor` would still have to collapse the sums into a vector of bit expressions, but that's no worse than now; it just defers performing the addition until you definitely need the result.

NikolajBjorner:
I will benchmark some variations and report back

NikolajBjorner:
Here are a few instances where the old blaster reduces the formula to a verdict, but the new blaster does not:

<HTML>
<BODY>
<TABLE><TR><TD>smtlib-latest/QF_BV/20190311-bv-term-small-rw-Noetzli/bv-term-small-rw_433.smt2</TD><TD>600</TD><TD>0.031</TD><TD>587.915690866511</TD><TD>0.0347185342538503</TD><TD>600</TD><TD>0.949</TD><TD>Timeout</TD><TD>Success</TD><TD>&nbsp;</TD><TD>0</TD><TD>599.969</TD><TD>587.880972332257</TD><TD>599.051</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>1</TD><TD>0</TD></TR><TR><TD>smtlib-latest/QF_BV/20190311-bv-term-small-rw-Noetzli/bv-term-small-rw_802.smt2</TD><TD>600</TD><TD>0.063</TD><TD>579.769053117783</TD><TD>0.0619081439393939</TD><TD>600</TD><TD>0.136</TD><TD>Timeout</TD><TD>Success</TD><TD>&nbsp;</TD><TD>0</TD><TD>599.937</TD><TD>579.707144973844</TD><TD>599.864</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>1</TD><TD>0</TD></TR><TR><TD>smtlib-latest/QF_BV/20190311-bv-term-small-rw-Noetzli/bv-term-small-rw_1220.smt2</TD><TD>600</TD><TD>0.125</TD><TD>567.963800904977</TD><TD>0.117159498207885</TD><TD>600</TD><TD>0.333</TD><TD>Timeout</TD><TD>Success</TD><TD>&nbsp;</TD><TD>0</TD><TD>599.875</TD><TD>567.846641406769</TD><TD>599.667</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>1</TD><TD>0</TD></TR><TR><TD>smtlib-latest/QF_BV/20190311-bv-term-small-rw-Noetzli/bv-term-small-rw_773.smt2</TD><TD>600</TD><TD>0.156</TD><TD>549.802890932983</TD><TD>0.178230643402398</TD><TD>600</TD><TD>0.251</TD><TD>Timeout</TD><TD>Success</TD><TD>&nbsp;</TD><TD>0</TD><TD>599.844</TD><TD>549.624660289581</TD><TD>599.749</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>1</TD><TD>0</TD></TR><TR><TD>smtlib-latest/QF_BV/20190311-bv-term-small-rw-Noetzli/bv-term-small-rw_1473.smt2</TD><TD>600</TD><TD>0.172</TD><TD>544.08322496749</TD><TD>0.179691404297851</TD><TD>600</TD><TD>0.24</TD><TD>Timeout</TD><TD>Success</TD><TD>&nbsp;</TD><TD>0</TD><TD>599.828</TD><TD>543.903533563192</TD><TD>599.76</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>1</TD><TD>0</TD></TR><TR><TD>smtlib-latest/QF_BV/20190311-bv-term-small-rw-Noetzli/bv-term-small-rw_466.smt2</TD><TD>600</TD><TD>0.172</TD><TD>635.866261398174</TD><TD>0.190648197242841</TD><TD>600</TD><TD>0.252</TD><TD>Timeout</TD><TD>Success</TD><TD>&nbsp;</TD><TD>0</TD><TD>599.828</TD><TD>635.675613200931</TD><TD>599.748</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>0</TD><TD>1</TD><TD>0</TD></TR></TABLE>

</BODY>
</HTML>

Example

```
C:\jsz3\release>z3 C:/QF_BV/20190311-bv-term-small-rw-Noetzli/bv-term-small-rw_433.smt2 /v:10
(simplifier :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(propagate-values :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(solve_eqs :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(elim-uncnstr :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(reduce-bv-size :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(simplifier :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(simplifier :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(max-bv-sharing :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(ackermannize_bv :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(bit-blast :num-exprs 8762 :num-asts 9079 :time 0.01 :before-memory 17.27 :after-memory 17.76)
(simplifier :num-exprs 6350 :num-asts 13053 :time 0.01 :before-memory 17.76 :after-memory 18.45)
(solve_eqs :num-exprs 6347 :num-asts 5993 :time 0.00 :before-memory 18.05 :after-memory 18.10)
(aig :num-exprs 6430 :num-asts 6627 :time 0.01 :before-memory 18.10 :after-memory 18.21)
(ast-table :prev-capacity 532480 :capacity 532480 :size 289)
(sat-status
  :inconsistent    false
  :vars            3876
  :elim-vars       0
  :lits            36457
  :assigned        0
  :binary-clauses  3722
  :ternary-clauses 9661
  :clauses         1
  :del-clause      0
  :avg-clause-size 2.72
  :memory          19.74)
(sat.solver)
(sat.stats   :conflicts  :restarts        :learned/bin    :gc          :time)
(sat.stats          :decisions :clauses/bin      :units       :memory       )
(sat.stats      0      0    0 13384/3722     0/0   0       0    19.74   0.00)
(sat.stats    193    397    0 13384/3722   193/1   0       0    19.84   0.01)
(sat.stats    193    397    0 13384/3722   193/1   0       0    19.84   0.01)
(sat.stats    193    397    0 13384/3722   193/1   0       0    19.84   0.01)
(sat.simplify :simplifications 1)
 (sat-scc :elim-vars 0 :time 0.00)
 (sat-subsumer :subsumed 0 :subsumption-resolution 0 :threshold 9984795520.78 :time 0.00)
 (sat-resolution :elim-vars 1837 :threshold 49929469221.73 :time 0.01)
 (sat-subsumer :subsumed 0 :subsumption-resolution 6 :threshold 9954522721.75 :time 0.00)
 (sat-probing :cost 14475221.35 :time 0.00)
(sat.stats    193    397    0 12065/1101    24/0   0       0    21.35   0.06)
(sat.stats    205    468    1 12065/1101    36/0   0       0    21.35   0.06)
(sat.stats :conflicts    :restarts      :learned/bin      :gc          :time)
(sat.stats        :decisions :clauses/bin        :units       :memory       )
(sat.stats    291    608    2 12065/1101   122/0   0       0    21.35   0.06)
(sat.stats    294    613    3 12065/1101   125/0   0       0    21.35   0.07)
(sat.stats    300    618    4 12065/1101   131/0   0       0    21.35   0.07)
(sat.stats    303    622    5 12065/1101   134/0   0       0    21.35   0.07)
(sat.stats    307    626    6 12065/1101   138/0   0       0    21.35   0.07)
(sat.stats    312    636    7 12065/1101   143/0   0       0    21.35   0.07)
(sat.stats    315    639    8 12065/1101   146/0   0       0    21.35   0.07)
(sat.stats    319    648    9 12065/1101   150/0   0       0    21.35   0.07)
(sat.stats    646   1059   10 12065/1101   477/0   0       0    21.47   0.08)
(sat.stats    650   1067   11 12065/1101   481/0   0       0    21.47   0.08)
(sat.stats    656   1074   12 12065/1101   487/0   0       0    21.47   0.08)
(sat.stats    660   1078   13 12065/1101   491/0   0       0    21.47   0.08)
(sat.stats    666   1083   14 12065/1101   497/0   0       0    21.47   0.08)
(sat.stats    673   1089   15 12065/1101   504/0   0       0    21.47   0.08)
(sat.stats    719   1143   16 12065/1101   550/0   0       0    21.47   0.09)
(sat.stats :conflicts   :restarts      :learned/bin       :gc          :time)
(sat.stats       :decisions  :clauses/bin        :units       :memory       )
(sat.stats   1250   1814   69 12065/1101  1081/0   0       0    21.67   0.10)
(sat.stats   1600   2339  142 12065/1101  1431/0   0       0    21.77   0.12)
(sat.stats   2397   3522  281 12065/1101  2228/0   0       0    22.02   0.15)
(sat.stats   3669   5634  456 12065/1101  3500/0   0       0    22.53   0.22)
(sat.stats   5395   8147  600 12065/1101  5226/0   0       0    23.08   0.29)
(sat.stats   8091  13942 1230 12065/1101  7922/0   0       0    24.78   0.54)
(sat.stats :conflicts :restarts       :learned/bin        :gc          :time)
(sat.stats      :decisions   :clauses/bin        :units       :memory       )
(sat.stats  12136  21817 1740 12065/1101 11967/0   0       0    27.02   0.92)
(sat.stats  18205  29468 2026 12065/1101 18036/0   0       0    29.74   1.23)
(sat.gc)
(sat-gc :strategy glue-psm :deleted 9914)
(sat.stats  28188  41849 3091 12065/1101 18105/0   0    9914    28.55   1.79)
(tactic-exception "canceled")
unknown

C:\jsz3\release>C:\z3\release\z3 C:/QF_BV/20190311-bv-term-small-rw-Noetzli/bv-term-small-rw_433.smt2 /v:10
(simplifier :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(propagate-values :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(solve_eqs :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(elim-uncnstr :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(reduce-bv-size :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(simplifier :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(simplifier :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(max-bv-sharing :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(ackermannize_bv :num-exprs 11 :num-asts 157 :time 0.00 :before-memory 17.27 :after-memory 17.27)
(bit-blast :num-exprs 12061 :num-asts 11419 :time 0.00 :before-memory 17.27 :after-memory 17.89)
(simplifier :num-exprs 1 :num-asts 14720 :time 0.01 :before-memory 17.89 :after-memory 18.53)
unsat
```

NikolajBjorner:
At this point it looks like the new blaster results in some measurable speedup of nearly 20% across QF_BV.
On the other hand, I get 75 more timeouts within 600s with these runs.
I would have to re-measure the overall effects with different experiments because the TO and speedup numbers can fluctuate a bit based on random seeds.

The first thing to look after are the few benchmarks listed above.
It suggests there is a regression when it comes to detecting some identities.
For example, it could be that the order of operands should be swapped to break symmetries some places, though it requires digging into the details to know.


image

NikolajBjorner:
There are no real differences on these benchmarks between flat = true and flat = false.
This doesn't say too much because the differences have in the past shown up on user instances not in QF_BV.

jameysharp:
I'm excited about that 20% improvement :grin:

I spent today investigating `bv-term-small-rw_433.smt2`. It doesn't have any constant inputs to the multiplier, so ideally both implementations should produce identical results. The differences come down to the choice of arrangement of full and half adders, and the specifics of how those full and half adders are implemented.

I simplified that example instance to this, with much smaller bit-vectors so it's a little easier to see what it's doing:

```
(set-logic QF_BV)
(declare-fun s () (_ BitVec 3))
(declare-fun t () (_ BitVec 3))
(assert (not (= (bvmul s (bvand t (bvor s t))) (bvmul s t))))
(apply (and-then simplify bit-blast))
```

Locally, I've made my full and half adders identical to the original ones, and set `rewriter.set_elim_and(true)` to match what `bit_blaster_rewriter.cpp` does. I also changed how I wire the adders together, but I haven't figured out how to exactly match the old implementation there.

Then I wrote a script to inline let-bindings and extract common subexpressions from s-expressions, so by feeding it the output from both versions at once it highlights which parts of the graph are the same and which are different. Here's the difference between Z3 4.8.10 and my current version on the above input:

<details>
<summary>differences: 4.8.10, then local</summary>

```
(let ((c!0 (not k!0)))
(let ((c!1 (not k!3)))
(let ((c!4 (or c!1 (not (or k!0 k!3)))))
(let ((c!8 (= (not (or c!0 c!4)) (or c!0 c!1))))
(let ((c!9 (not k!4)))
(let ((c!12 (or c!9 (not (or k!1 k!4)))))
(let ((c!13 (or c!0 c!12)))
(let ((c!14 (not k!1)))
(let ((c!15 (or c!14 c!4)))
(let ((c!17 (or c!0 c!9)))
(let ((c!18 (or c!14 c!1)))
(let ((c!21 (= (= c!13 c!15) (= c!17 (not c!18)))))
(let ((c!22 (not k!2)))
(let ((c!24 (not (or c!22 c!4))))
(let ((c!25 (not k!5)))
(let ((c!28 (or c!0 c!25 (not (or k!2 k!5)))))
(let ((c!29 (or c!14 c!12)))
(let ((c!32 (not (or c!13 c!15))))
(let ((c!35 (or c!22 c!1)))
(let ((c!36 (or c!0 c!25)))
(let ((c!37 (or c!14 c!9)))
(let ((c!40 (not (or c!17 c!18))))
((goals
  (goal
   (not
    (not
     (or
      c!8
      c!21
      (=
       (= c!24 (= (= c!28 c!29) c!32))
       (= c!35 (= (= c!36 c!37) c!40))))))
   :precision
   precise
   :depth
   2))
 (goals
  (goal
   (not
    (not
     (or
      c!8
      c!21
      (=
       (= (= c!28 (= c!29 c!24)) c!32)
       (= (= (not c!36) (= c!37 (not c!35))) c!40)))))
   :precision
   precise
   :depth
   2)))
))))))))))))))))))))))
```
</details>

That shows that most of the expression graph is identical, and indeed it's just a difference in which order bits are added together. Also there are two extra `(not)` nodes from my version, but they cancel since they're in a big XOR. Somehow those small differences are enough to keep the simplifier from turning this into a trivial solution on bigger instances.

I don't have any ideas right now about how to proceed. I can keep thinking about how to match the adder structure from the existing code, but I'd appreciate any suggestions you might have.

NikolajBjorner:
So current blaster just happens to be lucky. Understanding what makes it lucky is probably well worth it.
As noted above, the 20% speed improvement comes with 75 more timeouts so I have to read this data from different angles and produce some more measures to know what to say.

I took your example with (check-sat) instead of (apply ...) and ran the current z3 with the following options in debug mode:

 /v:10 /tr:tactic /tr:rewriter /tr:reduce_app /tr:th_rewriter_step

(use - instead of / on Linux/Mac).

It illustrates the rewrite steps where simplifications happen.
Noteworthy rewrites are probably:

```
or
(not k!3)
(not (or k!0 k!3))
---------->
(not k!3)
```
So it has a way to perform shallow lookahead rewriting when it can detect negations. 
It begs the issue that a good AIG reducer pass should subsume this (but maybe doesn't in Z3).


It ends rewriting the top level to

```
=
(not (or (not k!0) (not k!3)))
(or (not k!0) (not k!3))
---------->
false
```

and then 
```
=
(let ((a!1 (= (= (or (not k!0) (not k!5)) (or (not k!1) (not k!4)))
              (not (or (not k!3) (not k!4) (not k!0) (not k!1))))))
  (= (not (or (not k!2) (not k!3))) a!1))
(let ((a!1 (= (= (or (not k!0) (not k!5)) (or (not k!1) (not k!4)))
              (not (or (not k!3) (not k!4) (not k!0) (not k!1))))))
  (= (or (not k!2) (not k!3)) a!1))
---------->
false
------------------------------------------------
-------- [reduce_app] rewriter_tpl<struct `anonymous namespace'::th_rewriter_cfg>::process_app C:\z3\src\ast/rewriter/rewriter_def.h:312 ---------
(= (= (not #63) (= #65 #57)) (= (or #62 #30) (= #76 #69)))
st: 4 --->
false
------------------------------------------------
-------- [th_rewriter_step] `anonymous-namespace'::th_rewriter_cfg::reduce_app ..\..\..\src\ast\rewriter\th_rewriter.cpp:674 ---------
or
false
false
false
---------->
false

```

With your new blaster we can probably identify some places of divergence and missed rewrites / lookaheads or encode the xor/ors to to enable the current lookaheads.

NikolajBjorner:
I would also note that the legacy blaster uses mk_full_adder, mk_half_adder that use m_carry, mk_xor3 that perform inline rewriting.

Logging the original benchmarks indicate also that the rewriter doesn't eliminate "and" directly as the original does (it uses set_elim_and(true) somewhere). Though this omission doesn't make a difference (when adding     rewriter.set_elim_and(true);).

PR #5512, opened 2021-08-27
either fix up or remove the constant-case multiplier

jameysharp:
There's been a `mk_const_case_multiplier` in the bit-blaster since 2015, but in all that time its output has never been used. With this PR I'm trying to give this multiplier the best opportunity to show its value over straightforward long multiplication. If it doesn't actually help, then instead of merging this PR I propose to remove the implementation.

@NikolajBjorner, would you mind running a performance comparison of this patch versus master?

### DIFF
--- a/src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
+++ b/src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
@@ -194,7 +194,11 @@ void bit_blaster_tpl<Cfg>::mk_multiplier(unsigned sz, expr * const * a_bits, exp
         SASSERT(sz == out_bits.size());
         return;
     }
-   
+
+    if (mk_const_case_multiplier(sz, a_bits, b_bits, out_bits)) {
+        SASSERT(sz == out_bits.size());
+        return;
+    }
     if (mk_const_multiplier(sz, a_bits, b_bits, out_bits)) {
         SASSERT(sz == out_bits.size());
         return;
@@ -203,7 +207,7 @@ void bit_blaster_tpl<Cfg>::mk_multiplier(unsigned sz, expr * const * a_bits, exp
         SASSERT(sz == out_bits.size());
         return;
     }
-    out_bits.reset();
+
     if (!m_use_wtm) {
 #if 0
     static unsigned counter = 0;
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
with the current branch I get

image

The trend seems to be very clearly in favor of this change. 
We will need to merge with the master and I will retest to make sure there isn't something regressed from my changes.

GIven that this has the clearest signal, I would merge this first.

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
```diff
@@ -1247,20 +1253,16 @@ void bit_blaster_tpl<Cfg>::mk_const_case_multiplier(bool is_a, unsigned i, unsig
 
 template<typename Cfg>
 bool bit_blaster_tpl<Cfg>::mk_const_multiplier(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & out_bits) {
+    if (!m_use_bcm) {
```
NikolajBjorner, line 1256: I removed m_use_bcm. It would always be false, but my change almost for sure changes the behavior which I would have to revise or check.

src/ast/rewriter/bit_blaster/bit_blaster_tpl_def.h
```diff
@@ -1212,7 +1216,7 @@ bool bit_blaster_tpl<Cfg>::mk_const_case_multiplier(unsigned sz, expr * const *
     ptr_buffer<expr, 128> nb_bits;
     nb_bits.append(sz, b_bits);
     mk_const_case_multiplier(true, 0, sz, na_bits, nb_bits, out_bits); 
-    return false;
```
NikolajBjorner, line 1215: this looks like a terrible mistake

NikolajBjorner:
Since I introduced a bug in my previous push (a code path ignores that m_use_bcm is now false) I am porting your pull request directly and trying it out. There seems to be no longer a use for mk_const_multiplier either so this can be removed. 
After testing I would push this today or tomorrow. No need to update this pull request but I will reference it on the push.

NikolajBjorner:
The changes from this pull request have now been integrated.

NikolajBjorner:
image

jameysharp:
One note about the version you merged: you only need to call the const case multiplier once. If calling it with the arguments one way around returns false, the other way will too. In the previous version it had to be called twice just because it was called only when its first argument was entirely constant, but that's not true in the current version.

PR #5513, opened 2021-08-27
either use non-adjacent form instead of Booth encoding, or delete it

jameysharp:
There's been a Booth encoding implementation of multiplying by a constant in the bit blaster for as long as there's git history, but I don't think it's been turned on in any release during that time. Booth encoding is sometimes worse than straightforward long multiplication, but non-adjacent form is always at least as good. So in this PR I'm trying to give non-adjacent form every opportunity to shine, by using it unconditionally but making an effort to produce the same circuit as long multiplication in case NAF is only equally as good. If it doesn't actually help, then instead of merging this PR I propose to remove the implementation.

@NikolajBjorner, would you mind running a performance comparison of this patch versus master?

### DIFF
--- a/src/ast/rewriter/bit_blaster/bit_blaster_tpl.h
+++ b/src/ast/rewriter/bit_blaster/bit_blaster_tpl.h
@@ -116,7 +116,6 @@ public:
     void mk_smul_no_underflow(unsigned sz, expr * const * a_bits,  expr * const * b_bits, expr_ref & out);
     void mk_comp(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & out_bits);
 
-    void mk_carry_save_adder(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr * const * c_bits, expr_ref_vector & sum_bits, expr_ref_vector & carry_bits);
     bool mk_const_case_multiplier(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & out_bits);
     void mk_const_case_multiplier(bool is_a, unsigned i, unsigned sz, ptr_buffer<expr, 128>& a_bits, ptr_buffer<expr, 128>& b_bits, expr_ref_vector & out_bits);
 
--- a/src/ast/rewriter/bit_blaster/bit_blaster_tpl.h
+++ b/src/ast/rewriter/bit_blaster/bit_blaster_tpl.h
@@ -71,7 +71,6 @@ public:
 
     bool is_numeral(unsigned sz, expr * const * bits) const;
     bool is_numeral(unsigned sz, expr * const * bits, numeral & r) const;
-    bool is_minus_one(unsigned sz, expr * const * bits) const;
     void num2bits(numeral const & v, unsigned sz, expr_ref_vector & out_bits) const;
     
     void mk_half_adder(expr * a, expr * b, expr_ref & out, expr_ref & cout);
@@ -80,6 +79,7 @@ public:
     void mk_adder(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & out_bits);
     void mk_subtracter(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & out_bits, expr_ref & cout);
     void mk_multiplier(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & out_bits);
+    void mk_multiplier_core(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & out_bits);
     void mk_udiv_urem(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & q_bits, expr_ref_vector & r_bits);
     void mk_udiv(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & q_bits);
     void mk_urem(unsigned sz, expr * const * a_bits, expr * const * b_bits, expr_ref_vector & r_bits);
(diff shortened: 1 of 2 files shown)


jameysharp:
Okay, sounds good. I've rebased this branch against master and it still passes the regression tests for me locally.

NikolajBjorner:
updates to master have made a mess around this experiment, sorry.

jameysharp:
No worries! I'll fix it up again sometime in the next few days.

NikolajBjorner:
I am already running updated experiments with a merge so not urgent.
There are still several details around how the constant multipliers apply and the very 
ad-hoc metric for counting number of non-constant bits.

NikolajBjorner:
When added on top of current master the adjacency encoding is about the same and on the run I have loses 50 benchmarks to timeouts (this is still within the margin of randomness so not a strong indicator). So without a clear advantage I would punt on this change.

image

jameysharp:
I half expected it wouldn't make a big difference, but I'm confused. What changed since the performance comparison you posted at the start of this thread that would reduce the effect from a 25% improvement to basically 0%? I'd like to understand that before writing it off completely. Aside from that I have no complaints if you want to just close this PR.

NikolajBjorner:
I share the confusion, but run into such often enough to judge that the way I prepare the test is also imperfect (different compilers/flags/small differences in code that we overlook). There isn't bandwidth to figure this out, but it is more useful to simply try again based on your pull request when it is updated. At that point we have a current frontier and a delta with respect to the frontier instead of some incompatible set of changes.

We also don't have a clear example benchmark that illustrates the effect of the code change. 
This is also useful in these cases aside the bird's eye overview.

I still believe the bigger takeaway is that we can gain even more uniformly by post-processing the circuits with simplification.
The effect of fixing the propagation of negation is stunning. At some point we called out into Mathias Soeken's circuit rewriting tool to mine for simplifications. It ran into its own scalability issues and turned out to be also expensive. There is a built in circuit rewriting tool but exposed for SAT clauses (it is in src/sat/ in the "cut" files). It also has overhead that isn't necessarily counter-weighted by gains.

jameysharp:
Okay, I've updated this pull request.

I'd like to read up on cut-set enumeration. After some quick searching I've found [Applying Logic Synthesis for Speeding Up SAT](http://gauss.ececs.uc.edu/Courses/c626/lectures/AIG/synth_in_sat.pdf) by Niklas Een et al, but are there any papers you'd recommend on the topic?

NikolajBjorner:
A few tests runs:

The pull request with two different random seeds:

image

timeouts fluctuate and there is overall 2-3% speedup/slowdown differences.

The pull request against master:

image

similar differences / sameness with a slight indication of a regression, but it is possible this can fluctuate based on random seeds.

Overall, I don't think we should take the change.

NikolajBjorner:
References on cut-set enumeration:

- it is implemented and is integral to ABC
- there is an implementation in Yices, with some demonstrated speedups
- the mockturtle tool implements cut enumeration ( M. Soeken, H. Riener, W. Haaswijk, and G. De Micheli, “The EPFL
logic synthesis libraries,” May 2018, arXiv:1805.05121) @msoeken - are your lecture notes online?
- there is an implementation in Z3, that is not enabled by default at this point.

Disclaimer on my side is that I haven't identified the sweet spot for this. Implementing it at the level of the SAT solver also revealed some complexities both regarding correctness and performance. From my recollection, once we start creating hash-tables for every small cut the rewriting overhead goes up.

jameysharp:
Quick status update, as I'm still trying to catch up with a whole field of literature I wasn't aware of. Which is huge fun, of course. 😁

I guess looking up an optimal circuit and its cost from a precomputed table is O(1) with small constants, and the table can be quite small. I've written up what I've learned about that: https://jamey.thesharps.us/2021/09/05/z3-boolean-matching/

My current guess is that most of the bool rewriter simplifications can be replaced by a single-pass bottom-up cut-set thing that does O(1) work at each node, using space proportional to the depth of the graph. It might only need additional memory proportional to the number of shared nodes, or it might be important to keep the whole graph structurally hashed, although I feel like that's a good idea regardless. At any rate, some heuristics I've read about seem plausible for greedily picking good-enough cuts using only information propagated bottom-up.

The papers I've read so far seem to be focused on finding optimal cuts, so I feel like reframing the problem as "it just has to do at least as well as a bottom-up algebraic rewriter" might open up possibilities for fast implementations.

I find the thought of integrating this into Z3 efficiently to be rather daunting, though, as I don't think it fits in the current bottom-up rewriting infrastructure. It seems to me that you'd need to augment the framework to propagate analysis results, like the best cuts for the current node, upward alongside each rewritten expression.

Okay, that's what I've got so far. If you have any suggestions I'd love to hear them.

NikolajBjorner:
Some quick comments:

- there could very well be some embarrassing simplifications left to be done in the traditional rewriter. For example, gleaning at Bool-rewriter, it doesn't look like we sort arguments to or/and. Since these are typically binary they could be sorted by an optional swap. 
- another possibility at the level of the bool_rewriter is to enumerate all 2-3 input trees of limited depth and their simplest form and synthesize rewrites for cases that create proper sub-expressions. This kind of code:
```
        expr *la, *lb, *ra, *rb;
        // fold (iff (iff a b) (iff (not a) b)) to false
        if (m().is_eq(lhs, la, lb) && m().is_eq(rhs, ra, rb)) {
            expr *n;
            if ((la == ra && ((m().is_not(rb, n) && n == lb) ||
                (m().is_not(lb, n) && n == rb))) ||
                (lb == rb && ((m().is_not(ra, n) && n == la) ||
                    (m().is_not(la, n) && n == ra)))) {
                result = m().mk_false();
                return BR_DONE;
            }
        }
```
is a giveaway that the rewriter is hand crafted and therefore likely misses some more cases.
-  it is true that the bottom-up rewriter is ill suited for this kind of work: the reality is that we are rewriting a network, not a tree, not a DAG. The network has multiple roots and multiple leaves. When replacing one sub-expression inside the network, it could increase the overall size because internal nodes point to the old expression (below the root). This is not theoretical, but hits, seen often in the case of if-then-else expression rewriting.
- we typically use tactics to lump in functionality that can be self-contained, global, and used in a pre-processing pass. We even wrapped a version of mock-turtle as a tactic until abandoning that path based on non-scalability. 
- Regarding scale/speed, this is where glorified network rewriting may turn out to not give enough ROI. I don't know, but the efforts so far have not turned into wins. But as you noted, these are rather powerful methods for general purpose circuit minimization and are integral in other tools where the tradeoffs are in favor of generating minimal circuits (vs. throwing a CDCL solver at the result to check satisfiability of some formulas).

PR #5525, opened 2021-08-31
remove scoped

NikolajBjorner:
@JakobR can you check if changes can be merged?
The forbidden interval code uses scoped constraints that are probably not dedup'ed so this seems broken with the change but then I take the idea is to revise forbidden interval creation.


JakobR:
Looks good to me. Since the constraints from forbidden intervals are created through the constraint_manager, they are already dedup'ed. (Yes, forbidden interval implementation will be revised later.)

PR #5600, opened 2021-10-14
tweak GC in OCaml bindings

c-cube:
at Imandra we've had issues with recent versions of OCaml, where Z3 objects wouldn't be collected and would accumulate in RAM, eventually taking multiple GBs. There's some related issues for .NET I believe. This patch uses a different version of custom blocks (OCaml heap values that wrap foreign objects, such as a Z3 solver, expr, or context) with more explicit resource limits. I've hand picked limits that seem reasonable but it's subject to discussion. Pinging @ivg if he's interested.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,15 +43,15 @@ jobs:
     - ${{if eq(variables['runRegressions'], 'True')}}:
       - template: scripts/test-regressions.yml
 
-- job: "Ubuntu18Python"
-  displayName: "Ubuntu 18 with ocaml"
+- job: "Ubuntu20OCaml"
+  displayName: "Ubuntu 20 with ocaml"
   pool:
-    vmImage: "Ubuntu-18.04"
+    vmImage: "Ubuntu-20.04"
   steps:
     - script: sudo apt-get install ocaml opam libgmp-dev
     - script: opam init -y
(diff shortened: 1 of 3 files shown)


wintersteiger:
Conceptually this seems like a good idea, but I'd add some way for applications to tweak those numbers without making changes to the API code. I'm also slightly concerned that this breaks compatibility with older versions of OCaml, which may be a problem for some existing projects.

Minor: Why the change to Ubuntu 20? Does this hint at the fact that it won't work on Ubuntu 18 anymore? Also, I don't know what the guarantees of `GC.finalise` are, but in other systems (like .NET) this does not mean that an object is actually finalised at the request time (or indeed at all).

NikolajBjorner:
Just open another PR.

The bare bones C API has it that expressions are pointers without pointers to contexts.
Contexts maintain hash-tables of expressions so need the reference decrement to point to a context. This is consistent with how dec_ref is exposed over the API, where it takes a context argument.

The "region" notion is already there in one way: you can allocate a context in a mode where expression lifetimes come with the context.

c-cube:
> Just open another PR.

I will!

> The "region" notion is already there in one way: you can allocate a context in a mode where expression lifetimes come with the context.

In that case, what I can use to refer to an expression through a context? Is there an integer key or something like that? That would be optimal for me.

PR #5625, opened 2021-10-26
Added user propagator example

CEisenhofer:


NikolajBjorner:
You should look at the output of the hosted builds. Click on "Details"

NikolajBjorner:
There are also many warnings, such as

```

../examples/c++/userPropagator.cpp:258:31: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<z3::expr>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  258 |         for (int j = i + 1; j < queens.size(); j++) {
      |                             ~~^~~~~~~~~~~~~~~
../examples/c++/userPropagator.cpp: In function ‘int test23(unsigned int, bool)’:
../examples/c++/userPropagator.cpp:326:23: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<z3::expr>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  326 |     for (int i = 0; i < queens.size(); i++) {
      |                     ~~^~~~~~~~~~~~~~~
../examples/c++/userPropagator.cpp:337:12: warning: deleting object of polymorphic class type ‘user_propagator’ which has non-virtual destructor might cause undefined behavior [-Wdelete-non-virtual-dtor]
  337 |     delete propagator;
      |            ^~~~~~~~~~
../examples/c++/userPropagator.cpp: In function ‘int main()’:
../examples/c++/userPropagator.cpp:408:27: warning: comparison of integer expressions of different signedness: ‘int’ and ‘const long unsigned int’ [-Wsign-compare]
  408 |         for (int i = 0; i < SIZE(permutation); i++) {
      |                         
```

These are easy to fix

examples/c++/userPropagator.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+class model {
+
+public:
+
+    unsigned *values;
+    unsigned cnt;
+
+    model(unsigned cnt) : cnt(cnt), values(nullptr) {
+        values = (unsigned*) malloc(sizeof(unsigned) * cnt);
```
NikolajBjorner, line 48: why use malloc from c++?

examples/c++/userPropagator.cpp
```diff
@@ -0,0 +1,420 @@
...
+        }
+        return hash;
+    }
+};
+
+class user_propagator : public z3::user_propagator_base {
+
+protected:
+
+    std::unordered_map<unsigned, unsigned> *id_mapping;
```
NikolajBjorner, line 102: I think you want & instead of *. It saves a character on dereference (. instead of ->) and captures invariant of being non-null

examples/c++/userPropagator.cpp
```diff
@@ -0,0 +1,420 @@
...
+        return true;
+    }
+
+    bool operator!=(const model &other) const {
+        return !this->operator==(other);
+    }
+
+    model* clone() const {
+        model *newModel = new model(cnt);
+        memcpy(newModel->values, values, sizeof(unsigned) * cnt);
```
NikolajBjorner, line 82: maybe just have a constructor for model that performs clone.

examples/c++/userPropagator.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+        currentModel = new model(board);
+
+        this->register_fixed();
+        this->register_final();
+    }
+
+    ~user_propagator() {
+        delete currentModel;
+        currentModel = nullptr;
```
NikolajBjorner, line 156: setting to nullptr is redundant. C++ has scoped ptr that does the free automatically.

examples/c++/userPropagator.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+    void push() override {
+        fixedCnt.push((unsigned) fixedValues.size());
+    }
+
+    void pop(unsigned num_scopes) override {
+        for (unsigned i = 0; i < num_scopes; i++) {
+            unsigned lastCnt = fixedCnt.top();
+            fixedCnt.pop();
+            for (unsigned j = (unsigned) fixedValues.size(); j > lastCnt; j--) {
```
NikolajBjorner, line 167: for (auto j = fixedValues.size(); j > lastCnt; j--)

examples/c++/userPropagator.cpp
```diff
@@ -0,0 +1,420 @@
...
+#define WriteEmptyLine std::cout << std::endl
+#define WriteLine(x) std::cout << (x) << std::endl
+#define Write(x) std::cout << x
+#else
+#define WriteEmptyLine
+#define WriteLine(x)
+#define Write(x)
+#endif
+
+class model {
```
NikolajBjorner, line 40: I don't get why you need a model abstraction.
Why not class model : public std::vector<unsigned> {};

CEisenhofer:
Thanks for your comments! I've refactored the code accordingly. However, I've no clue why the build tries to link the cpp_example.o and the userPropagator.o. (From Azure's log: `g++ -o cpp_example  example.o userPropagator.o  libz3.so -lpthread`)
In the CMakeList.txt I've specified to build 2 executables and not to link those two files. (On my PC there is no problem.)  
What does Azure execute to test the commit?

PR #5695, opened 2021-12-05
Handle correctly cancelled run 

levnach:


src/math/lp/int_solver.cpp
```diff
@@ -365,8 +365,6 @@ bool int_solver::get_freedom_interval_for_column(unsigned j, bool & inf_l, impq
             m = lcm(m, denominator(a));
 
         if (!inf_l && !inf_u) {
-            if (l > u)
```
NikolajBjorner, line 368: why are you removing this?

src/math/lp/lar_core_solver_def.h
```diff
@@ -179,11 +179,17 @@ void lar_core_solver::solve() {
...
-    if (m_r_solver.get_status() == lp_status::INFEASIBLE) {
+    switch (m_r_solver.get_status())
+    {
+    case lp_status::INFEASIBLE:
         fill_not_improvable_zero_sum();
-    } 
-    else if (m_r_solver.get_status() != lp_status::UNBOUNDED) {
+        break;
+    case lp_status::CANCELLED:
+    case lp_status::UNBOUNDED: // do nothing in these cases
```
NikolajBjorner, line 188: theory_lra only checks against lp_status::OPTIMAL when deciding whether to re-check feasiblity in final_check_eh. Now UNBOUNDED would likely leak too. So shouldn't theory_lra condition the status on being feasible, where feasible is one of OPTIMAL, UNBOUNDED or so?

src/math/lp/int_solver.cpp
levnach: It cannot happen if the column is feasible. And we assume here that it is feasible. We have l <= 0 <= u during this loop. Then we adjust l and u by shifting to xj.

NikolajBjorner: It seems to me that you are baking in some assumptions about the calling context. Maybe the current calling context is that the tableau is feasible, but this code would be easier to maintain if it doesn't rely on strong assumptions.

src/math/lp/lar_core_solver_def.h
levnach: Yes, it seems the right approach. I can fix it later.

src/math/lp/int_solver.cpp
levnach: The calculation will be incorrect if at least one of the basic columns 'i ' is infeasible. The feasibility assumption is already backed in to the code as I understand it.

src/math/lp/lar_core_solver_def.h
NikolajBjorner: The question is how theory_lra handles the return codes.

Can we always assume that if the status is OPTIMAL, FEASIBLE, UNBOUNDED that all bounds are satisfied?

If so, then there are two places to change in theory_lra. It makes no assumptions about feasibility with UNBOUNDED.

src/math/lp/int_solver.cpp
levnach: Actually, the code seems to be correct even when some i is infeasible. Going to reverse the change related to get_freedom...

levnach: From the other side, if we consider the infeasible case, we cannot just "continue" the loop on l==u, expecting that further on we might encounter another infeasible variable that will create l < u!

levnach: It looks like we have some optimizations for the feasible case that are not available in general.

NikolajBjorner: at minimum add an assertion to enforce whatever contract you assume

src/math/lp/lar_core_solver_def.h
levnach: Yes. Or, lp can have a boolean function, smth like, feasible_status(lp_status), providing this abstraction.

NikolajBjorner: yes please

src/math/lp/lar_solver.cpp
```diff
@@ -797,7 +797,17 @@ namespace lp {
         lp_assert(((stats().m_make_feasible% 100) != 0) || m_status != lp_status::OPTIMAL || all_constraints_hold());
     }
 
-
+    bool lar_solver::status_is_feasible(lp_status st) {
```
NikolajBjorner, line 800: just call it is_feasible() and use get_status() internally.

NikolajBjorner: The use case is here:

https://github.com/Z3Prover/z3/blob/0242566792d858275b39175936ad35e1faa22c63/src/smt/theory_lra.cpp#L1532

The new code would be

```
  if (!lp().is_feasible() || lp().has_changed_columns()) 
```
or even the has_changed_columns() really implies that the solver is not in the feasible state.

PR #5717, opened 2021-12-16
Update regex union and intersection to maintain ANF

veanes:
Updates to ensure correct ANF.

### DIFF
--- a/src/ast/rewriter/seq_rewriter.cpp
+++ b/src/ast/rewriter/seq_rewriter.cpp
@@ -3112,7 +3112,7 @@ void seq_rewriter::mk_antimirov_deriv_rec(expr* e, expr* r, expr* path, expr_ref
         else
             // D(e,r1)r2|(ite (r1nullable) (D(e,r2)) [])
             // observe that (mk_ite_simplify(true, D(e,r2), []) = D(e,r2)
-            result = mk_antimirov_deriv_union(c1, re().mk_ite_simplify(r1nullable, mk_antimirov_deriv(e, r2, path), nothing()));
+            result = mk_regex_union_normalize(c1, re().mk_ite_simplify(r1nullable, mk_antimirov_deriv(e, r2, path), nothing()));
     }
     else if (m().is_ite(r, c, r1, r2)) {
         c1 = simplify_path(e, m().mk_and(c, path));
@@ -3171,7 +3171,7 @@ void seq_rewriter::mk_antimirov_deriv_rec(expr* e, expr* r, expr* path, expr_ref
             result = re().mk_ite_simplify(range, epsilon(), nothing());
     }
     else if (re().is_union(r, r1, r2))
-        result = mk_antimirov_deriv_union(mk_antimirov_deriv(e, r1, path), mk_antimirov_deriv(e, r2, path));
+        result = mk_regex_union_normalize(mk_antimirov_deriv(e, r1, path), mk_antimirov_deriv(e, r2, path));
     else if (re().is_intersection(r, r1, r2))
         result = mk_antimirov_deriv_intersection(e, 
             mk_antimirov_deriv(e, r1, path),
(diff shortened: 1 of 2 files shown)


src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -4515,6 +4624,11 @@ br_status seq_rewriter::mk_re_union(expr* a, expr* b, expr_ref& result) {
         result = mk_full();
         return BR_DONE;
     }
+
+    //just keep the union normalized
+    result = mk_regex_union_normalize(a, b);
+    return BR_DONE;
+
```
NikolajBjorner, line 4631: dead code?

NikolajBjorner:
Mac build failure can be ignored

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3344,13 +3320,137 @@ expr_ref seq_rewriter::mk_antimirov_deriv_restrict(expr* e, expr* d, expr* cond)
...
+// Assumes r1 and r2 are both ordered lists in right-associative form without duplicates when unions
+// Merges the two lists as a new ordered list where duplicates are removed
+// an element and its complement have the same merge-order and result in true
+expr_ref seq_rewriter::mk_regex_union_merge(expr* r1, expr* r2) {
+    expr_ref result(m());
+    expr* r1first, * r1rest, * r2first, * r2rest;
+    if (re().is_union(r1, r1first, r1rest)) {
+        if (re().is_union(r2, r2first, r2rest)) {
+            if (r1first->get_id() < r2first->get_id())
+                result = re().mk_union(r1first, mk_regex_union_merge(r1rest, r2));
```
NikolajBjorner, line 3353: someone will fuzz this to stack overflow.

Also, can't you assume that r2, r1 are already in normal form?

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3344,13 +3320,137 @@ expr_ref seq_rewriter::mk_antimirov_deriv_restrict(expr* e, expr* d, expr* cond)
...
+    else if (re().is_dot_plus(r2) && re().get_info(r1).min_length > 0)
+        result = r1;
+    else
+        result = mk_regex_inter_merge(r1, r2);
+    return result;
+}
+// Assumes r1 and r2 are both ordered lists in right-associative form without duplicates when intersections
+// Merges the two lists as a new ordered list where duplicates are removed
+// an element and its complement have the same merge-order and result in false
+expr_ref seq_rewriter::mk_regex_inter_merge(expr* r1, expr* r2) {
```
NikolajBjorner, line 3416: this code is similar to union.
Instead of duplicating code, have common routine that takes decomposition + composition as function arguments?

expr_ref seq_rewriter::merge_ac(expr* r1, expr* r2, 
std::function<bool(expr*,expr*&, expr*&)>& decompose,
std::function<expr*(expr*, expr*)>& compose) {
     ptr_vector<expr> sorted;
     expr* a, *b, *c, *d;
     while (true) {
         if (r1 == r2) {
             sorted.push_back(r1);
             break;
         }
         if (test(r1, a, b) && r2->get_id() < a->get_id()) {
             sorted.push_back(r2);
             sorted.push_back(r1);
             break;
         }
         // same for test(r2, c, d) and r1 is smaller than a
         // other cases: test(r1, a, b) and r1 == a
         // !test(r1, a, b) && !test(r2, c, d)
         if (test(r1, a, b) && test(r2, c, d) && a->get_id() < c->get_id()) {
              sorted.push_back(a);
              r1 = b;
              continue;
         
     }

     // walk sorted from back to front and apply "compose"
}

veanes: OK, I'll rewrite the merge using a single merge routine that is non-recursive, covering both union and intersection.

src/ast/rewriter/seq_rewriter.cpp
veanes: Here r2 and r1 are assumed to be in normal form (ordered lists essentially).
So merge makes their union keeping the result ordered. I'll change that to use a loop.

src/ast/rewriter/seq_rewriter.cpp
veanes: Yes, will remove. The previous code was using rewrite rules, now a and b are assumed in normal form and are merged, 
the code after is dead code, will remove.

PR #5762, opened 2022-01-10
Add WebAssembly/TypeScript bindings

bakkot:
[As discussed](https://github.com/Z3Prover/z3/discussions/5738#discussioncomment-1934018). See the files named `src/js/api/README.md` and `src/js/api/PUBLISHED_README.md` for more details.

This includes two new GitHub Actions:
- one which builds and tests the new bindings (replacing the existing [WASM Build](https://github.com/Z3Prover/z3/blob/f1bf660adc6f40cfdbd1c35de58c49b5f9960a9c/.github/workflows/wasm.yml) workflow, which wasn't doing much) on every PR and every commit to master,
- one which builds, tests, and publishes them on npm, which must be triggered [manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow). It's easy to change to run [whenever a release is published on GitHub](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release), though.

I confirmed these work on my fork.

The version number for the release is read (with a regex) from [`release.yml`](https://github.com/Z3Prover/z3/blob/f1bf660adc6f40cfdbd1c35de58c49b5f9960a9c/scripts/release.yml).

In order to use the publish action, I'll need to add your npm account as a maintainer of [the package](https://www.npmjs.com/package/z3-solver), and then you'll need to [create an `automation` token](https://docs.npmjs.com/creating-and-viewing-access-tokens) and [add it as a secret for the repository](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository) named `NPM_TOKEN`. Please let me know the account you'd like to use.

npm will not let you publish over an existing version, but I published the version from my fork as `-pre`. So you should be able to trigger the publish job once this lands.

also, cc @cpitclaudel @philzook58 @bramvdbogaerde as people who have previously expressed interest in having wasm bindings.


NikolajBjorner:
>  Please let me know the account you'd like to use: nbjorner@microsoft.com

.github/workflows/wasm.yml
```diff
@@ -1,43 +1,51 @@
-name: WASM Build
+name: WebAssembly Build
```
NikolajBjorner, line 1: I see this replaces the old action.

.gitignore
```diff
@@ -75,6 +75,13 @@ src/api/ml/z3enums.ml
 src/api/ml/z3native.mli
 src/api/ml/z3enums.mli
 src/api/ml/z3.mllib
+src/api/js/node_modules/
```
NikolajBjorner, line 78: Not sure why they are in .gitignore.

.github/workflows/wasm-release.yml
```diff
@@ -0,0 +1,61 @@
...
+          which clang++
+          npm run build-wasm
+
+      - name: Test
+        run: npm test
+
+      - name: Publish
+        run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
NikolajBjorner, line 61: I have installed an NPM token with my account that has write access to z3-solver.
I didn't find the z3-solver organization. Thanks if you deleted it.

.github/workflows/wasm-release.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 'lts/*'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Prepare for publish
+        run: |
+          npm version $(node -e 'console.log(fs.readFileSync("../../../scripts/release.yml", "utf8").match(/ReleaseVersion:\s*\x27(\S+)\x27/)[1])')
```
NikolajBjorner, line 29: I guess this is the version without "-pre" because I can't find a reference to "-pre" in the pull request.

src/api/js/scripts/parse-api.js
```diff
@@ -0,0 +1,354 @@
...
+  Z3_sort_opt: 'Z3_sort',
+  Z3_ast_opt: 'Z3_ast',
+  Z3_func_interp_opt: 'Z3_func_interp',
+};
+
+// parse type declarations
+let types = {
+  __proto__: null,
+
+  // these are function types I can't be bothered to parse
```
NikolajBjorner, line 49: I will likely create other such callbacks so if this list is required the more sustainable way is to mark them in z3_api as callbacks using some no-op macro. Based on the code below it looks like you match for "typedef enum" and other typedefs are not considered.

src/api/js/scripts/parse-api.js
```diff
@@ -0,0 +1,354 @@
...
+let functions = [];
+let enums = Object.create(null);
+for (let file of files) {
+  let contents = fs.readFileSync(path.join(__dirname, '..', '..', file), 'utf8');
+
+  // we _could_ use an actual C++ parser, which accounted for macros and everything
+  // but that's super painful
+  // and the files are regular enough that we can get away without it
+
+  // we could also do this by modifying the `update_api.py` script
```
NikolajBjorner, line 70: I don't see adding things to update_api as a goal. It doesn't serve a purpose in itself. At best one can reuse some feature from update_api, but this seems needless now.
The goals are to ensure the bindings can be built on every push and the bindings can be exported (to npm).

src/api/js/scripts/parse-api.js
```diff
@@ -0,0 +1,354 @@
...
+
+let defApis = Object.create(null);
+let functions = [];
+let enums = Object.create(null);
+for (let file of files) {
+  let contents = fs.readFileSync(path.join(__dirname, '..', '..', file), 'utf8');
+
+  // we _could_ use an actual C++ parser, which accounted for macros and everything
+  // but that's super painful
+  // and the files are regular enough that we can get away without it
```
NikolajBjorner, line 68: The files have to be regular.

.github/workflows/wasm.yml
bakkot: Yeah, as mentioned in the PR message. The old one didn't do much useful - you couldn't use the artifact without having the same version of the compiler used to build it.

.gitignore
bakkot: This is the usual practice for JS projects. This folder holds the dependencies used to build the project.

.github/workflows/wasm-release.yml
bakkot: Yup, I deleted it just before opening this PR. Sorry that wasn't clearer.

bakkot:
Thanks for merging! You should try the new wasm-publish action to confirm you can publish the package.

src/api/js/scripts/parse-api.js
NikolajBjorner: I added macros so the closures can be discovered instead of hard-wired

bakkot: I can add support for parsing these properly, but to be sure I've done it right I'd like to have at least _a_ test of them. Right now I can't find any uses of any of these types to compare against. Can you write a brief example (in C/C++) demonstrating how any one of the consumers of the `Z3_whatever_eh` types is intended to work, such that I could tell if my implementation has the right behavior?

NikolajBjorner: https://github.com/Z3Prover/z3/blob/56d3718cde7260d54c34e1420ef11a3436671139/src/api/python/z3/z3.py#L212

Other callbacks are not exercised yet in examples I have published for python. They are used from C++
https://github.com/Z3Prover/z3/blob/56d3718cde7260d54c34e1420ef11a3436671139/src/api/c%2B%2B/z3%2B%2B.h#L3966
and then for a self-contained example:
https://github.com/Z3Prover/z3/tree/master/examples/userPropagator

bakkot: https://github.com/Z3Prover/z3/tree/master/examples/userPropagator looks like something I can re-implement, thanks. That PDF is very helpful.

For `Z3_set_error_handler`, what's an example of a case where it should be called? For example, should I expect it to be called if I perform `Z3_query_constructor` on a `Z3_constructor` which has not previously been passed to `Z3_mk_datatypes`?

NikolajBjorner: Yes, the error handler gets invoked whenever there is some mistake in the API call. 
For example, if you add two arrays. The addition is not well typed and it fails to create the expression.
Without the error handler it should return null. With the error handler you can raise an exception from the handler assuming stack unwinding works (which I guess for WebAssembly it should).

bakkot: Another question: this line:

 https://github.com/Z3Prover/z3/blob/56d3718cde7260d54c34e1420ef11a3436671139/src/api/z3_api.h#L6700

refers to a `Z3_propagate_solver_declare`. I can't find any other mention of that. Is it just a typo?

Issue #4869, opened 2020-12-07
Spacer CHC solver and pattern-matching


BFarinier:
I am facing the following issue when using pattern-matching with the Spacer CHC solver:
```
(declare-datatype OptionInt
 ((SomeInt (valueInt Int))
  (NoneInt)))

(define-fun getValueDefault ((opt OptionInt) (default Int)) Int
 (match opt
  (((SomeInt int) int)
   (NoneInt default))))

(declare-var opt1 OptionInt)
(declare-var opt2 OptionInt)

(declare-var default Int)

(declare-rel equalOptionInt (OptionInt OptionInt))

(rule
 (=>
  (= (getValueDefault opt1 default) (getValueDefault opt2 default))
  (equalOptionInt opt1 opt2)))

(query equalOptionInt)
```
```
(error "query failed: Uninterpreted 'valueInt' in <null>:
equalOptionInt(#1,#0) :- 
 (= (ite ((_ is (SomeInt (Int) OptionInt)) (:var 1))
        (valueInt (:var 1))
        (:var 2))
   (ite ((_ is (SomeInt (Int) OptionInt)) (:var 0))
        (valueInt (:var 0))
        (:var 2))).
")
unknown
```
As far as I understand, pattern-matching in `getValueDefault` is deconstructed as an `if then else` in which selectors appear. But according to https://github.com/Z3Prover/z3/issues/3070 selectors are partially-defined, leading to the error. This is regrettable as the original pattern-matching was fully-specified.

Do you think there is a chance that the Spacer CHC solver could support fully-specified pattern-matching? I know that I could use the workaround detailed in https://github.com/Z3Prover/z3/issues/3070, but my real case involves nested pattern-matching, and doing so will result in the introduction of hundreds variables, an extent that I would like to avoid...

gernst:
Maybe a related comment (not sure if this behavior is intended though):
```
(set-logic HORN)

(declare-datatype OptionInt
 ((SomeInt (valueInt Int))
  (NoneInt)))

(declare-const v OptionInt)
(assert ((_ is SomeInt) v))
```
results in `(error "line 8 column 25: invalid use of indexed identifier, unknown builtin function is")`, but works fine with the default logic (or `ALL`).

PR #5821, opened 2022-02-07
Handle additional cases in rule_properties::check_accessor

wert310:
This PR adds additional safe accessor patterns to the previously accepted ones.

Instead of only allowing guarded accessors, this patch allows for an accessor to be used unguarded if all other recognizers are used in the other paths (so the only possible value at that point of the expression is the last remaining constructor). These new accepted patterns should cover all possible ways a match expression is desugared:

- guarded branch with correct recognizer (as before)
- if all other recognizes have been used negated in a rule the remaining unguarded accessor is safe
- `and`: if all branches except for one have negated recognizers, then the remaining one is safe
- `or`: if all branches except for one have recognizers the remaining one is safe
- `or`: if one branch has a negated recognizer (for the current case) access in another branch is safe
- `ite`: if the recognizer is in the condition, access in "then" branch is safe; the "else" branch is safe if all recognizers except for one have been used in the condition of an `ite` in the path to the current accessor
- any combination of the above

This PR improves the fix of https://github.com/Z3Prover/z3/issues/4869 by allowing pattern matching expressions where the last branch accesses the constructor arguments. The previous version of `check_accessor` incorrectly returned false for the example of https://github.com/Z3Prover/z3/issues/4869 if the cases of the match expression are swapped:
```
(define-fun getValueDefault ((opt OptionInt) (default Int)) Int
 (match opt
  ((NoneInt default)
   ((SomeInt int) int))))
```

In that example, the match is desugared as:
```
(ite ((_ is (NoneInt () OptionInt)) (:var 0)) (:var 2) (valueInt (:var 0)))
```
where `(valueInt (:var 0))` is a safe (but unguarded) accessor because it is in the else branch and it is the only remaining constructor. The previous check would not allow this accessor because it is unguarded.

(Edit: clarified second case in the above list)

### DIFF
--- a/src/muz/base/rule_properties.cpp
+++ b/src/muz/base/rule_properties.cpp
@@ -210,19 +210,19 @@ bool rule_properties::check_accessor(app* n) {
     SASSERT(m_dt.is_datatype(s));
     if (m_dt.get_datatype_constructors(s)->size() <= 1)
         return true;
-
     
     func_decl* f = n->get_decl();
     func_decl * c = m_dt.get_accessor_constructor(f);
     unsigned ut_size = m_rule->get_uninterpreted_tail_size();
     unsigned t_size  = m_rule->get_tail_size();
+    ptr_vector<expr> recognizers;
 
     auto is_recognizer_base = [&](expr* t) {
         return m_dt.is_recognizer(t) &&
             to_app(t)->get_arg(0) == n->get_arg(0) &&
             m_dt.get_recognizer_constructor(to_app(t)->get_decl()) == c;
     };
-    
+    // t is a recognizer for n
     auto is_recognizer = [&](expr* t) {
         if (m.is_and(t))
             for (expr* arg : *to_app(t))
@@ -231,11 +231,13 @@ bool rule_properties::check_accessor(app* n) {
         return is_recognizer_base(t);
     };
 
-    
-    for (unsigned i = ut_size; i < t_size; ++i)
-        if (is_recognizer(m_rule->get_tail(i)))
+    for (unsigned i = ut_size; i < t_size; ++i) {
+        auto *tail = m_rule->get_tail(i);
+        if (is_recognizer(tail))
             return true;
-
+        if (m.is_not(tail) && m_dt.is_recognizer(to_app(tail)->get_arg(0)))
+            recognizers.push_back(to_app(tail)->get_arg(0));
+    }
 
     // create parent use list for every sub-expression in the rule
     obj_map<expr, ptr_vector<expr>> use_list;
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
         if (!use_list.contains(e))
             return false;
         for (expr* parent : use_list[e]) {
-            if (!parent)
-                return false; // top-level expressions are not guarded
-            if (is_recognizer(parent)) 
+            if (!parent) { // top-level expression
+                // check if n is an unguarded "else" branch
+                ptr_vector<func_decl> ctors;
+                for (auto *rc : recognizers)
+                    if (to_app(rc)->get_arg(0) == n->get_arg(0))
+                        ctors.push_back(m_dt.get_recognizer_constructor(to_app(rc)->get_decl()));
+                ptr_vector<func_decl> diff;
+                for (auto *dtc : *m_dt.get_datatype_constructors(s))
+                    if (std::find(ctors.begin(), ctors.end(), dtc) == ctors.end())
+                        diff.push_back(dtc);
+                // the only unguarded constructor for s is c:
+                // all the others are guarded and we are in an "else" branch so the accessor is safe
+                if (diff.size() == 1 && diff[0] == c)
+                    continue;
+                return false; // the accessor is not safe
+            }
+            if (is_recognizer_base(parent))
                 continue;
-            if (m.is_ite(parent) && to_app(parent)->get_arg(1) == e && is_recognizer(to_app(parent)->get_arg(0))) 
-                continue;
-            todo.push_back(parent);                
+            if (m.is_ite(parent)) {
+                if (to_app(parent)->get_arg(1) == e && is_recognizer(to_app(parent)->get_arg(0)))
+                    continue; // e is guarded
+                else
+                    if (m_dt.is_recognizer(to_app(parent)->get_arg(0)))
+                        recognizers.push_back(to_app(parent)->get_arg(0)); // e is in the else branch of the ite
+            }
+            if (m.is_or(parent)) {
+                for (expr *arg : *to_app(parent)) {
+                    // if all branches are guarded the remaining one might be safe
+                    if (m_dt.is_recognizer(arg))
+                        recognizers.push_back(arg);
+                    // if one branch is not(recognizer) then the accessor is safe
+                    if (m.is_not(arg)) {
+                        if (is_recognizer(to_app(arg)->get_arg(0)))
+                            goto _continue;
+                    }
+                }
+            }
+            if (m.is_and(parent)) {
+                for (expr *arg : *to_app(parent)) {
+                    // if one branch is guarded the accessor is safe
+                    if (is_recognizer(arg))
+                        goto _continue;
+                    // if all branches are not(recognizer) the remaining one might be safe
+                    if (m.is_not(arg)) {
+                        auto *neg_rec = to_app(arg)->get_arg(0);
+                        if (m_dt.is_recognizer(neg_rec))
+                            recognizers.push_back(neg_rec);
+                    }
+                }
+            }
+            todo.push_back(parent);
+        _continue:;
         }
     }
-    
+
     return true;
-    
 }
 
 void rule_properties::operator()(app* n) {
--- a/src/muz/base/rule_properties.cpp
+++ b/src/muz/base/rule_properties.cpp
@@ -212,17 +212,33 @@ bool rule_properties::check_accessor(app* n) {
         return true;
     
     func_decl* f = n->get_decl();
-    func_decl * c = m_dt.get_accessor_constructor(f);
+    func_decl* c = m_dt.get_accessor_constructor(f);
     unsigned ut_size = m_rule->get_uninterpreted_tail_size();
     unsigned t_size  = m_rule->get_tail_size();
-    ptr_vector<expr> recognizers;
+    ptr_vector<func_decl> ctors;
 
+    // add recognizer constructor to ctors
+    auto add_recognizer = [&](expr* r) {
+        if (!m_dt.is_recognizer(r))
+            return;
+        if (n->get_arg(0) != to_app(r)->get_arg(0))
+            return;
+        auto* c2 = m_dt.get_recognizer_constructor(to_app(r)->get_decl());
+        if (c == c2)
+            return;
+        ctors.push_back(c2);
+    };
+    auto add_not_recognizer = [&](expr* r) {
+        if (m.is_not(r, r))
+            add_recognizer(r);
+    };
+
+    // t is a recognizer for n
     auto is_recognizer_base = [&](expr* t) {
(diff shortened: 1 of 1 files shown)


src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
...
+            }
+            if (is_recognizer_base(parent))
                 continue;
-            todo.push_back(parent);                
+            if (m.is_ite(parent)) {
+                if (to_app(parent)->get_arg(1) == e && is_recognizer(to_app(parent)->get_arg(0)))
+                    continue; // e is guarded
+                else
+                    if (m_dt.is_recognizer(to_app(parent)->get_arg(0)))
+                        recognizers.push_back(to_app(parent)->get_arg(0)); // e is in the else branch of the ite
```
hgvk94, line 284: I don't see why `e` is in the else branch. Can't it be that the if conditional failed because of the second condition?

src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
...
+            }
+            if (m.is_or(parent)) {
+                for (expr *arg : *to_app(parent)) {
+                    // if all branches are guarded the remaining one might be safe
+                    if (m_dt.is_recognizer(arg))
+                        recognizers.push_back(arg);
+                    // if one branch is not(recognizer) then the accessor is safe
+                    if (m.is_not(arg)) {
+                        if (is_recognizer(to_app(arg)->get_arg(0)))
```
hgvk94, line 293: This allows formulas like `(==> (is_cons x) (= y (tail x))` . I don’t consider this guarded because the second literal could be set to true even when the first is set to false. Please comment on why this is considered guarded.

src/muz/base/rule_properties.cpp
```diff
@@ -231,11 +231,13 @@ bool rule_properties::check_accessor(app* n) {
...
 
-    
-    for (unsigned i = ut_size; i < t_size; ++i)
-        if (is_recognizer(m_rule->get_tail(i)))
+    for (unsigned i = ut_size; i < t_size; ++i) {
+        auto *tail = m_rule->get_tail(i);
+        if (is_recognizer(tail))
             return true;
-
+        if (m.is_not(tail) && m_dt.is_recognizer(to_app(tail)->get_arg(0)))
```
NikolajBjorner, line 238: if (m.is_not(tail, ntail) && m_dt.is_recognizer(ntail))
   recognizers.push_back(ntail);

src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
...
-            if (m.is_ite(parent) && to_app(parent)->get_arg(1) == e && is_recognizer(to_app(parent)->get_arg(0))) 
+            if (!parent) { // top-level expression
+                // check if n is an unguarded "else" branch
+                ptr_vector<func_decl> ctors;
+                for (auto *rc : recognizers)
+                    if (to_app(rc)->get_arg(0) == n->get_arg(0))
+                        ctors.push_back(m_dt.get_recognizer_constructor(to_app(rc)->get_decl()));
+                ptr_vector<func_decl> diff;
+                for (auto *dtc : *m_dt.get_datatype_constructors(s))
+                    if (std::find(ctors.begin(), ctors.end(), dtc) == ctors.end())
```
NikolajBjorner, line 269: if (!ctx.contains(dtx))

NikolajBjorner: Shorter:

```
obj_hashtable<func_decl> ctors;
for (auto *rc : recognizers)
        if (to_app(rc)->get_arg(0) == n->get_arg(0))
             ctors.insert(m_dt.get_recognizer_constructor(to_app(rc)->get_decl()));
if (m_dt.get_datatype_constructors(s)->size() != ctors.size() + 1)
    return false;
continue;
```

src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
...
+                    if (m_dt.is_recognizer(to_app(parent)->get_arg(0)))
+                        recognizers.push_back(to_app(parent)->get_arg(0)); // e is in the else branch of the ite
+            }
+            if (m.is_or(parent)) {
+                for (expr *arg : *to_app(parent)) {
+                    // if all branches are guarded the remaining one might be safe
+                    if (m_dt.is_recognizer(arg))
+                        recognizers.push_back(arg);
+                    // if one branch is not(recognizer) then the accessor is safe
+                    if (m.is_not(arg)) {
```
NikolajBjorner, line 292: I tend to use m.is_not(arg, narg) for expr* narg; to avoid having to write awkaward to_app(arg)->get_arg(0). Poor man's pattern matching...

src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
         if (!use_list.contains(e))
             return false;
         for (expr* parent : use_list[e]) {
-            if (!parent)
-                return false; // top-level expressions are not guarded
-            if (is_recognizer(parent)) 
-                continue;
-            if (m.is_ite(parent) && to_app(parent)->get_arg(1) == e && is_recognizer(to_app(parent)->get_arg(0))) 
+            if (!parent) { // top-level expression
```
NikolajBjorner, line 261: I think you want to process top-level expressions _after_ "recognizers" have been populated.
So 'e' to a top-level todo list and process after use_list is handled.

src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
...
+                for (auto *dtc : *m_dt.get_datatype_constructors(s))
+                    if (std::find(ctors.begin(), ctors.end(), dtc) == ctors.end())
+                        diff.push_back(dtc);
+                // the only unguarded constructor for s is c:
+                // all the others are guarded and we are in an "else" branch so the accessor is safe
+                if (diff.size() == 1 && diff[0] == c)
+                    continue;
+                return false; // the accessor is not safe
+            }
+            if (is_recognizer_base(parent))
```
NikolajBjorner, line 277: just keep is_recognizer(parent) here. Then the handling of negated recognizers in a conjunction is separate and without the goto _continue clutter.

src/muz/base/rule_properties.cpp
NikolajBjorner: Also, seems instead of first collecting recognizers, then add them to ctors, just collect ctors directly.

auto add_not_recognizer = [&](expr* r) {
    if (!m.is_not(r, r))
        return;
    if (!m_dt.is_recognizer(r))
        return;
    if (n->get_arg(0) != to_app(r)->get_arg(0))
        return;
     ctors.insert(m_dt.get_recognizer_constructor(to_app(r)->get_decl()); 
};

src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
...
                 continue;
-            todo.push_back(parent);                
+            if (m.is_ite(parent)) {
+                if (to_app(parent)->get_arg(1) == e && is_recognizer(to_app(parent)->get_arg(0)))
+                    continue; // e is guarded
+                else
+                    if (m_dt.is_recognizer(to_app(parent)->get_arg(0)))
```
NikolajBjorner, line 283: Leaves some room as the ite condition could be conjunctive. So an abstraction of this test could be to extract 'must' recognizers from an expression. Overall, a more satisfactory approach would address complete patterns for nested cases which this code doesn't address.

```
match (x, y) with
    a::as, [] -> 
    [], b::bs -> 
    [], [] -> 
    _, _ -> body with hd(x), tl(y)....
```

src/muz/base/rule_properties.cpp
NikolajBjorner: On a second thought, recognizers _should not_ be global. Their validity is context sensitive. So it depends on each expression being processed. So the entire setup with global recognizers or ctrs has to change.

src/muz/base/rule_properties.cpp
NikolajBjorner: There is another bug: e could be in both branches of the ite. 
Then checking that it is in the if-branch and continuing masks that it is in the else branch.

src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
         if (!use_list.contains(e))
             return false;
         for (expr* parent : use_list[e]) {
-            if (!parent)
-                return false; // top-level expressions are not guarded
-            if (is_recognizer(parent)) 
-                continue;
-            if (m.is_ite(parent) && to_app(parent)->get_arg(1) == e && is_recognizer(to_app(parent)->get_arg(0)))
```
NikolajBjorner, line 263: bug with e in the else branch also present in old code

src/muz/base/rule_properties.cpp
NikolajBjorner: this should be fine. It is de-morgan dual to conjunction and context based.
dually, there could be a conjunction (not (and (is_cons x) (not (= y (tail x))))
The context-sensitive analysis determines that the conjunction is true only if (is_cons x)

src/muz/base/rule_properties.cpp
NikolajBjorner: These occurrences of recognizers are of course global and apply to all paths.
So when fixing for context sensitivity within paths there appears to be a use for both global and local context awareness. Alternatively, just delay adding global recognizers to the case where hitting !parent. It is a single pass over the tail.

wert310:
Thanks for your comments!

I've modified the loop to do a full depth first search over all parents, keeping track of the size of the `ctors` array across each expression (so that I can remove the elements when backtracking). The search starts with `ctors` containing all global constructors, then for each path it should contain only the "local" constructors.

I'm using an `std::tuple` in the `todo` stack to avoid creating multiple vectors/maps. In particular I use the last element of the tuple to keep track of the visited nodes in the tree. I noticed that elsewhere in the code you are using an `ast_mark`: do you prefer that I edit the current solution to use an additional `ast_mark` for consistency, or is this solution ok?

src/muz/base/rule_properties.cpp
```diff
@@ -256,18 +258,62 @@ bool rule_properties::check_accessor(app* n) {
...
+            }
+            if (m.is_or(parent)) {
+                for (expr *arg : *to_app(parent)) {
+                    // if all branches are guarded the remaining one might be safe
+                    if (m_dt.is_recognizer(arg))
+                        recognizers.push_back(arg);
+                    // if one branch is not(recognizer) then the accessor is safe
+                    if (m.is_not(arg)) {
+                        if (is_recognizer(to_app(arg)->get_arg(0)))
+                            goto _continue;
```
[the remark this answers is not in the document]
wert310: I need to skip the `todo.push_back`, so just `continue` from the inner for is not enough.
I'm not really happy about using `goto`, to remove it I could:
1. use `std::any_of` and `continue`
   ```c++
   if (m.is_or(parent)) {
       for (expr* arg : *to_app(parent))
           add_recognizer(arg);
       if (std::any_of(to_app(parent)->begin(), to_app(parent)->end(),
                       [&](expr *arg) { return m.is_not(arg, arg) && is_recognizer(arg); }))
           continue;
   }
   ```
   But this would loop twice over args
2. use a boolean flag and then continue if the flag is true
3. include the "or not" case in `is_recognizer`

Probably (1) is the cleaner one, even if it requires an additional loop/lambda.
I noticed that `goto` is used more than `std::any_of` in the codebase, so let me know which one you prefer for consistency.

Issue #4977, opened 2021-01-26
Assertion violation ../src/ast/ast.cpp line 432


deut-erium:
Following code crashes on z3py, I crashed it unintentionally
```
ASSERTION VIOLATION
File: ../src/ast/ast.cpp
Line: 432
UNREACHABLE CODE WAS REACHED.
Z3 4.8.9.0
```
```python
from z3 import *
s = Optimize()
f = Function('f',IntSort(),IntSort())
s.maximize(f)
```

NikolajBjorner:
the argument to maximize should be an expression, not a function. For example

```
from z3 import *
s = Optimize()
f = Function('f',IntSort(),IntSort())
s.maximize(f(1))
```

I have added checks in the front-end so it now throws a more informative exception

PR #5845, opened 2022-02-16
Propagator

NikolajBjorner:


delixus-valliammal1:
[> _ No description provided. _

](https://github.com/Z3Prover/z3/issues/4977)
I have this problem. It is Mythril Library uses z3 and in that it is occurring while threads are at parallel processing.
the line 414 and 432 are error in ast.cpp. Can it be resolved. Or can it be avoided by some means by avoiding program to exit.?

NikolajBjorner:
you need to use a separate context per thread. Two threads cannot operate on expressions/solvers from the same context.

delixus-valliammal1:
Now this flow instead of 414 and 432 in ast.cpp, giving error in rewrite_def.h in the line 224 and give error as "Unexpected Error"
Can you let us know what else to be changed?. Multiprocessing of threads and parallel is enabled now

Issue #5807, opened 2022-02-02
Algebraic numbers, index problem


makkiato83:
Hello, I am experimenting with z3 in Python.

I am not sure, but it looks like the index() method of the AlgebraicNumRef class is not working properly:

```
x, y = Reals('x y')
s = Solver()
s.add( x*x == 2 )
s.add( x>0)
s.add( y*y == 2 )
s.add( y<0)
s.check()
m = s.model()
val_x = m.get_interp(x)
val_y = m.get_interp(y)

print(f"The solutions are: x={val_x} and y={val_y}")
print(f"The defining polynomial of x has integer coefficients", val_x.poly(), " and root index ", val_x.index())
print(f"The defining polynomial of y has integer coefficients", val_y.poly(), " and root index ", val_y.index())
```

the output is:

```
The solutions are: x=1.4142135623? and y=-1.4142135623?

the defining polynomial of x has integer coefficients [-2, 0, 1]  and root index  0

the defining polynomial of y has integer coefficients [-2, 0, 1]  and root index  0
```

The polynomial defining the solutions is correct:  polynomial $1x^2 + 0x -2 = 0$

But the two roots are both getting index 0.

NikolajBjorner:
You are the only user of this feature. I am afraid you would have to fix this on your own.

janvrany:
Fix in PR #5888  should fix the problem.

Issue #2650, opened 2019-10-22
Incorrect result in NRA formula 2


muchang:
Hi,
For this formula,
```
(set-logic NRA)
(declare-fun a () Real)
(declare-fun b () Real)
(declare-fun c () Real)
(declare-fun d () Real)
(declare-fun e () Real)
(declare-fun f () Real)
(declare-fun g () Real)
(declare-fun h () Real)
(declare-fun i () Real)
(declare-fun j () Real)
(declare-fun k () Real)
(declare-fun l () Real)
(declare-fun m () Real)
(declare-fun n () Real)
(declare-fun o () Real)
(declare-fun p () Real)
(declare-fun q () Real)
(declare-fun r () Real)
(declare-fun s () Real)
(assert
  (not
    (exists
      ((t Real))
      (=>
        (and
          (or
            (and
              (=>
                (< 0 j)
                (and
                  (or
                    (<= 0 (+ (* n s) m))
                    (<= 0 p)
                  )
                  (> s 0)
                )
              )
              (<
                (*
                  o
                  (* (* n h) (/ o (- b h)))
                )
                0
              )
            )
            (< 0 n)
          )
          (< 0 p)
          (< 0 (- b h))
        )
        (<
          (/
            (* (+ (* n j) m) (+ (* n j) m))
            (* 2 o)
          )
          (* a e)
        )
      )
    )
  )
)
(assert
  (not
    (exists
      ((u Real))
      (=>
        (and
          (or (= (* c r) 0) (= l 0))
          (= e 2)
          (< h i (/ d r))
        )
        (=> (= 0 k) (not (<= u k 0 r)))
      )
    )
  )
)
(assert (= a (+ f o) (+ g o)))
(assert (= b (/ h q)))
(check-sat)
(get-model)
```
Z3 will incorrectly report sat, and give a model:
```
(define-fun h () Real
    (- 1.0))
  (define-fun b () Real
    0.0)
  (define-fun o () Real
    0.0)
  (define-fun q () Real
    0.0)
  (define-fun d () Real
    1.0)
  (define-fun m () Real
    1.0)
  (define-fun n () Real
    1.0)
  (define-fun j () Real
    1.0)
  (define-fun p () Real
    1.0)
  (define-fun s () Real
    (- 1.0))
  (define-fun g () Real
    (/ 1.0 2.0))
  (define-fun f () Real
    (/ 1.0 2.0))
  (define-fun a () Real
    (/ 1.0 2.0))
[...]

PR #5888, opened 2022-03-08
Care for root index being undefine while calling Z3_algebraic_get_i()

janvrany:
In some cases, Z3_algebraic_get_i() returned 0. For example, in the following
Python snippet, the last assert would fail:

    import z3
    x = z3.Real('x')
    s = z3.Solver()
    s.add( (x * x) - 2 == 0, x <= 0)
    s.check()
    val_x = s.model().get_interp(x)
    assert val_x.index() == 1

The problem was that `algebraic_numbers::manager::imp::get_i()` did not
check whether the root index was properly initialized.

This commit fixes this issue by checking whether root index is initialized
the same way various other routines do.

Fixes issue #5807.

### DIFF
--- a/src/math/polynomial/algebraic_numbers.cpp
+++ b/src/math/polynomial/algebraic_numbers.cpp
@@ -2013,6 +2013,11 @@ namespace algebraic_numbers {
             }
             else {
                 algebraic_cell * c = a.to_algebraic();
+                if (c->m_i == 0) {
+                    // undefined
+                    c->m_i = upm().get_root_id(c->m_p_sz, c->m_p, lower(c)) + 1;
+                }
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
@leodemoura - I could use a hand reviewing this. 
The index can be 0 when a numeral is 0, but not otherwise?
While having looked at bugs in this part of the code it also comes out of my cache. The long standing #2650 and related may take a couple of weeks just for warm-up to find the presumably isolated issue.

NikolajBjorner:
Based on what I can see roots are 1-indexed (so the first root has index 1). The pull request therefore seems correct. It would still be useful to have an additional evaluation.

Issue #5900, opened 2022-03-16
Disposing of Z3Objects in the dotnet API


Choc13:
Our application uses Z3 via the dotnet API. Recently we've been addressing issues with large memory consumption when we're running large programs and we wondered if someone could clarify for us the way that a `Z3Object` should be disposed of.

The `Z3Object` implements `IDisposable` and so usually we would expect to wrap these in a `using` statement to make sure that their `Dispose` method is called once they go out of scope. Maybe we've just not looked in the right place, but all of the usages of the Z3 dotnet API we have seen do *not* follow this pattern and instead just rely on the fact that the finalizer for `Z3Object` also calls `Dispose`. My first question therefore is, was it intended that users of the dotnet API should wrap each `Z3Object` in a `using` statement or not? 

Partially due to my initial naivety around the fact that I did not know that the finalizer for a `Z3Object` will call `Dispose`, I undertook the task of changing our code base such that each time a `Z3Object` was created it was wrapped in a `using` statement, as I believed we would be leaking each `Z3Object` that wasn't. 

It is my understanding from reading #5330 and #5043 that the native memory that a `Context` points to has allocated a large hash map and we were therefore keen to ensure that each `Context` was deterministically disposed to free this. Looking inside the [`Dispose`](https://github.com/Z3Prover/z3/blob/master/src/api/dotnet/Context.cs#L4934) method for `Context` the `refCount` must be `0` before it will free the native memory, and the `refCount` is only decremented when a `Z3Object` created by that `Context` is disposed. Hence my desire to ensure that each `Z3Object` created from that `Context` was also deterministically `Disposed` via a `using` statement.

This led us to some interesting findings that I wanted to share:

1. Using `using` statements gave us a much more stable memory profile, without this we found that the memory would jump around a lot, presumably because the GC can't actually see the native memory's size and so when it does finally do a GC collect it's already under a lot of pressure and the finalizer queue is full of a lot of pointers.
This was especially true when we created a fresh `Context` each time we needed one due to the fact that the native memory for the `Context` had allocated a large hash map. We are aware that, as mentioned in the issues linked above, the recommendation is to reuse them, but we were interested in whether this was still true when they were being deterministically disposed.

[...]

PR #5901, opened 2022-03-17
Dispose of intermediate Z3Objects created in dotnet api.

Choc13:
Addresses #5900 by ensuring that any intermediate objects created in the dotnet API that inherit from `Z3Object` are wrapped in a using statement so that they are deterministically disposed.

### DIFF
--- a/src/api/dotnet/Context.cs
+++ b/src/api/dotnet/Context.cs
@@ -312,9 +312,18 @@ namespace Microsoft.Z3
         public EnumSort MkEnumSort(string name, params string[] enumNames)
         {
             Debug.Assert(enumNames != null);
-	    using var sname = MkSymbol(name);
-	    using var snames = MkSymbols(enumNames);
-            return new EnumSort(this, sname, snames);
+
+            using var symbol = MkSymbol(name);
+            var symbols = MkSymbols(enumNames);
+            try 
+            {
+                return new EnumSort(this, symbol, symbols);
+            }
+            finally
+            {
+                foreach (var symbol in symbols)
+                    symbol.Dispose();
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


src/api/dotnet/Context.cs
```diff
@@ -312,9 +312,18 @@ public EnumSort MkEnumSort(Symbol name, params Symbol[] enumNames)
...
+
+            using var symbol = MkSymbol(name);
+            var symbols = MkSymbols(enumNames);
+            try 
+            {
+                return new EnumSort(this, symbol, symbols);
+            }
+            finally
+            {
+                foreach (var symbol in symbols)
```
Choc13, line 324: These collections of `Z3Object` were a bit annoying in that we can't just call `using var symbols = MkSymbols(enumNames)` because that will only dispose of the collection, which I do not believe dispose of the elements within the collection. It seems the only way to do it is to desugar the `using` can add an explicit `finally` block where enumerate the collection and manually `Dispose` each item.

src/api/dotnet/Context.cs
```diff
@@ -956,17 +985,15 @@ public BoolExpr MkXor(IEnumerable<BoolExpr> ts)
...
-                if (r == null)
-                    r = t;
-                else
-                    r = MkXor(r, t);
-            }
-            if (r == null)
-                r = MkTrue();
-            return r;
+
+            return ts.Any()
```
Choc13, line 989: This was probably the trickiest one to get right. I believe what the original code was doing is essentially a fold / aggregate (although the empty case was returning `MkTrue()`, which isn't the initial value for Xor, but I've preserved that behaviour here).

The trickiness with the disposing is that we only want to dispose of the temporary objects created as part of the fold. To ensure this is the case I seeded `Aggregate` with `MkFalse` so that in the first call to the lambda it will dispose of this temp object once the `MkXor` has been called. Each subsequent call to the lambda will dispose of `MkXor` created in the previous invocation. Therefore only objects explicitly created in this method (the intermediate values of the fold) are passed to the `using` statement. The final value is not disposed of as it is passed back to the caller and therefore they own it.

In the trivial case of an empty list therefore is nothing to dispose because we just return `MkTrue()`.

PR #5916, opened 2022-03-22
Jfleisher/nightlynuget

jfleisher:
publish a nightly nuget to private Z3Build feed
 https://dev.azure.com/Z3Build/Z3Build/_artifacts/feed/Z3-Nightly-Builds
 
 There is also a wee fix for a Debug.Assert in NativeContext


src/api/dotnet/NativeSolver.cs
```diff
@@ -405,7 +405,7 @@ public override string ToString()
 
         internal NativeSolver(NativeContext nativeCtx, Z3_solver z3solver)
         {
-            Debug.Assert(nCtx != IntPtr.Zero);
+            Debug.Assert(nativeCtx != null);
```
jfleisher, line 408: This is a fix for the arguments asserts in NativeContext

PR #5956, opened 2022-04-08
Added rewriting distinct with bitvectors to false if bit-size is too low

CEisenhofer:


src/ast/rewriter/bv_rewriter.cpp
```diff
@@ -2803,7 +2803,26 @@ br_status bv_rewriter::mk_ite_core(expr * c, expr * t, expr * e, expr_ref & resu
     return BR_FAILED;
 }
 
-br_status bv_rewriter::mk_bvsmul_no_overflow(unsigned num, expr * const * args, bool is_overflow, expr_ref & result) {
+br_status bv_rewriter::mk_distinct(unsigned num_args, expr * const * args, expr_ref & result) {
+    if (num_args == 0) {
+        return BR_FAILED;
```
NikolajBjorner, line 2808: result = true

NikolajBjorner: and if num_args = 1 then result = false.
OTOH, these rewrites are already handled by bool_rewriter

src/ast/rewriter/bv_rewriter.cpp
```diff
@@ -2803,7 +2803,26 @@ br_status bv_rewriter::mk_ite_core(expr * c, expr * t, expr * e, expr_ref & resu
...
+    }
+    unsigned sz = get_bv_size(args[0]);
+    // check if num_args > 2^sz
+    bool exact = true;
+    do {
+        exact &= (num_args % 2) == 0;
+        num_args /= 2;
+        sz--;
+    } while (num_args > 1 && sz > 0);
```
NikolajBjorner, line 2817: convention: while on the next line after }. 
I know this is contrary to some automatic formatting modes

src/ast/rewriter/th_rewriter.cpp
```diff
@@ -210,16 +207,25 @@ struct th_rewriter_cfg : public default_rewriter_cfg {
...
             }
-            if (k == OP_EQ && m_seq_rw.u().has_seq() && is_app(args[0]) && 
+            if (k == OP_EQ && m_seq_rw.u().has_seq() && is_app(args[0]) &&
                 to_app(args[0])->get_family_id() == m_seq_rw.get_fid()) {
                 st = m_seq_rw.mk_eq_core(args[0], args[1], result);
                 if (st != BR_FAILED)
                     return st;
             }
+            if (k == OP_DISTINCT) {
+                SASSERT(num > 0);
```
NikolajBjorner, line 221: just add it to the if statement
if (k == OP_DISTINCT && num > 0)

and for that matter m_bv_rw.is_bv_sort(args[0]->get_sort())

src/ast/rewriter/bv_rewriter.cpp
CEisenhofer: Do you mean result = true for num_args = 1?

NikolajBjorner: yes, right, true for both 0 and 1 arguments.

src/ast/rewriter/bv_rewriter.cpp
```diff
@@ -2803,6 +2803,27 @@ br_status bv_rewriter::mk_ite_core(expr * c, expr * t, expr * e, expr_ref & resu
     return BR_FAILED;
 }
 
+br_status bv_rewriter::mk_distinct(unsigned num_args, expr * const * args, expr_ref & result) {
+    if (num_args <= 1) {
+        result = m().mk_true();
+        return BR_DONE;
+    }
+    unsigned sz = get_bv_size(args[0]);
+    // check if num_args > 2^sz
```
NikolajBjorner, line 2812: this may be clever code, but not easy to read.
Here is something simpler:

```
if (sz >= 32) 
    return BR_FAILED;
if (num_args <= 1u << sz)
    return BR_FAILED;
result = m().mk_false();
return BR_DONE;        
```

src/ast/rewriter/th_rewriter.cpp
```diff
@@ -210,16 +207,23 @@ struct th_rewriter_cfg : public default_rewriter_cfg {
...
                     return st;
             }
-            if (k == OP_EQ && m_seq_rw.u().has_seq() && is_app(args[0]) && 
+            if (k == OP_EQ && m_seq_rw.u().has_seq() && is_app(args[0]) &&
                 to_app(args[0])->get_family_id() == m_seq_rw.get_fid()) {
                 st = m_seq_rw.mk_eq_core(args[0], args[1], result);
                 if (st != BR_FAILED)
                     return st;
             }
+            if (k == OP_DISTINCT && num > 0) {
```
NikolajBjorner, line 220: Why two nested ifs?

Issue #5926, opened 2022-03-25
pip install z3-solver: bad path or elf header


zwimer:
In the newest release of the z3 pip package, I've run into what I think is a bug on both macOS and Linux.

On Linux:

In `/lib` there is `libz3.so`; this is the shared library, not a symlink. There are no other libs.
If you run `readelf -a libz3.so | grep SONAME` you will fine the output is:
```
 0x000000000000000e (SONAME)             Library soname: [libz3.so.4.8]
 ```

When I try to link to the library the link command succeeds but it when I try to run the linked binary it looks for `libz3.so.4.8`, which does not exist. On macOS this same thing happens.

For more details, here is my exact linking command:
```
[ 83%] Linking CXX executable backend-z3-solver.test
cd /claripy/native/build/tests/unit/backend/z3 && /venv/lib/python3.8/site-packages/cmake/data/bin/cmake -E cmake_link_script CMakeFiles/backend-z3-solver.test.dir/link.txt --verbose=1
/usr/bin/c++ -O2 -g -DNDEBUG CMakeFiles/backend-z3-solver.test.dir/solver.cpp.o -o backend-z3-solver.test  -Wl,-rpath,/claripy/native/build/tests/unit/testlib:/claripy/native/build:/venv/lib/python3.8/site-packages/z3/lib ../../../../../gmp/build/.libs/libgmp.a ../../testlib/libtestlib.so ../../../../libclaricpp.so ../../../../../gmp/build/.libs/libgmp.a /venv/lib/python3.8/site-packages/z3/lib/libz3.so -ldl
```
Specifically, that last line; as you can see it links to `/venv/lib/python3.8/site-packages/z3/lib/libz3.so`; as one would expect. Then running `ldd` on the resulting binary gives:
```
root@93638fc1d8d2:/claripy/native/build# ldd libclaricpp.so
	linux-vdso.so.1 (0x0000ffff8f616000)
	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffff8f41c000)
	/lib/ld-linux-aarch64.so.1 (0x0000ffff8f5e6000)
	libdw.so.1 => /lib/aarch64-linux-gnu/libdw.so.1 (0x0000ffff8f3b8000)
	libz3.so.4.8 => not found
	libstdc++.so.6 => /lib/aarch64-linux-gnu/libstdc++.so.6 (0x0000ffff8f1d3000)
[...]

PR #5974, opened 2022-04-12
Add a hacky patch so that Z3 on M1 hardware can link to libs properly

zwimer:
This is a hacky patch which should ease the effect of: https://github.com/Z3Prover/z3/issues/5926

This is not a fix, it is a hacky patch. It should allow pip installed z3 to be linked to on M1 and unknown hardware (on both macOS and Linux). This will not necessarily work for all situations, but should work for some. It simply creates symlinks to pretend the library is named differently, to match the invalid SONAME in the library header.

### DIFF
--- a/src/api/python/setup.py
+++ b/src/api/python/setup.py
@@ -184,6 +184,18 @@ def _copy_bins():
                 continue
             shutil.copy(os.path.join(header_dir, fname), os.path.join(HEADERS_DIR, fname))
 
+    # This hack lets z3 installed libs link on M1 macs; it is a hack, not a proper fix
+    # @TODO: Linked issue: https://github.com/Z3Prover/z3/issues/5926
+    major_minor = '.'.join(_z3_version().split('.')[:2])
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
Can you also try the arm64 wheel on the nightly build? I hope I got the package path conventions right so that we can just distribute native ARM64 for the growing base on M1 users.

NikolajBjorner:
I will hold off on merging this until a second set of eyes (if possible, @rhelmot) agree or I get to double check

zwimer:
@NikolajBjorner Where would I find the nightly wheels? If you can point me to them I can test them. If those work, (and the bug does not appear) this PR can be ignored.

rhelmot:
This is a fine change so far as hacky patches go, the only thing I want to see changed is that the windows case shouldn't use a `return`, since if someone adds code after the symlink they may expect it'll be run on windows.

zwimer:
@NikolajBjorner With the merge of this PR, would it be possible to create a new pip release?

PR #5977, opened 2022-04-13
User propagator decide

CEisenhofer:


NikolajBjorner:
split into two parts and replay/squash opt changes

PR #5978, opened 2022-04-13
Added decide-callback to user-propagator

CEisenhofer:
Now just for the legacy solver

### DIFF
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -3964,18 +3964,23 @@ namespace z3 {
             }
         };
 
-        static void push_eh(void* p) {
+        static void push_eh(void* _p, Z3_solver_callback cb) {
+            user_propagator_base* p = static_cast<user_propagator_base*>(_p);
+            scoped_cb _cb(p, cb);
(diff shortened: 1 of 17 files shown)


src/api/c++/z3++.h
```diff
@@ -4009,8 +4011,15 @@ namespace z3 {
...
         }
-
-
+        
+        static void decide_eh(void* _p, Z3_solver_callback cb, Z3_ast& _val, unsigned& bit, Z3_lbool& is_pos) {
+            user_propagator_base* p = static_cast<user_propagator_base*>(_p);
+            scoped_cb _cb(p, cb);
+            expr val(p->ctx(), _val);
+            p->m_decide_eh(val, bit, is_pos);
+            _val = val;
```
NikolajBjorner, line 4020: the expression val is declared locally. The reference count is invalid when thedecide_eh goes out of scope. The expression _val is passed to the solver. 
If the reference count is not owned by the solver, this call will crash.
It can be expected the reference count isn't fully owned locally because the expression should be already registered. So the issue shouldn't manifest. OTOH, it is fragile and makes assumptions.

Issue #5970, opened 2022-04-12
OCaml: GC tweaks are wrong


giltho:
Hello,

Following an update of Z3 from 4.8.7 to 4.8.15, the execution of our verification got around ***4 times slower***.
Given that calls to Z3 are around 30-40% of our execution, I'm expecting that means calls to Z3 got ~8 times slower.

I spent a few hours finding the issue, and I found out that [this is the PR that made everything slower](https://github.com/Z3Prover/z3/pull/5600).

Agreed our usage of Z3 is very suboptimal (non-incremental, having one solver that we reuse by resetting on every call), but such a slowdown means that the change is wrong.

@c-cube, any idea? This change leaves us unable to upgrade past 4.8.12
I'd be happy running our benchmark on any change you make

c-cube:
Ah, interesting, yes. We also have a slowdown, but it did save a lot of RAM (:weary:). It was done with a 4.08 switch I think, and the GC has changed since in the ways it manages off-heap items. If you have time to tweak things yourself and see if it improves runtime I think it'd be very useful.

cc @Bronsa on that one, too :)

PR #5982, opened 2022-04-15
tweak GC 

c-cube:
Related to #5970. This improves times for us (backported on an older version); I haven't checked the memory usage yet. @giltho would you be interested in trying this branch?

### DIFF
--- a/src/api/ml/z3native_stubs.c.pre
+++ b/src/api/ml/z3native_stubs.c.pre
@@ -294,7 +294,8 @@ static struct custom_operations Z3_ast_plus_custom_ops = {
   Z3_ast_compare_ext
 };
 
-MK_CTX_OF(ast, 16) // let's say 16 bytes per ast
+// FUDGE
+MK_CTX_OF(ast, 8) // let's say 16 bytes per ast
 
 #define MK_PLUS_OBJ_NO_REF(X, USED)                                \
   typedef struct {                                                      \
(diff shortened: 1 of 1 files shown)


src/api/ml/z3native_stubs.c.pre
```diff
@@ -294,7 +294,8 @@ static struct custom_operations Z3_ast_plus_custom_ops = {
   Z3_ast_compare_ext
 };
 
-MK_CTX_OF(ast, 16) // let's say 16 bytes per ast
+// FUDGE
+MK_CTX_OF(ast, 8) // let's say 16 bytes per ast
```
NikolajBjorner, line 298: "let's say?" 8 is the new 16?

src/api/ml/z3native_stubs.c.pre
```diff
@@ -410,25 +411,26 @@ MK_CTX_OF(ast, 16) // let's say 16 bytes per ast
...
-MK_PLUS_OBJ(tactic, 128)
-MK_PLUS_OBJ(probe, 128)
-MK_PLUS_OBJ(apply_result, 128)
-MK_PLUS_OBJ(solver, 20 * 1000 * 1000) // pretend a solver is 20MB
-MK_PLUS_OBJ(stats, 128)
-MK_PLUS_OBJ(ast_map, 1024 * 2)
-MK_PLUS_OBJ(ast_vector, 128)
-MK_PLUS_OBJ(fixedpoint, 20 * 1000 * 1000)
-MK_PLUS_OBJ(optimize, 20 * 1000 * 1000)
+// FUDGE
```
NikolajBjorner, line 414: vanilla size?

giltho:
Hi! Thank you very much for looking into this!
With these changes, it looks like now times have gone up only 10%, instead of 330% for our set of benchmarks. That's definitely an acceptable time regression, especially if the gains in terms of memory are interesting

src/api/ml/z3native_stubs.c.pre
c-cube: ah, need to remove the comment :). These are size estimations, that the GC uses to guestimate how heavy a foreign value is. Seems like giving OCaml smaller estimations makes it faster here, it might be that it started garbage collecting like crazy with the previous change.

src/api/ml/z3native_stubs.c.pre
c-cube: it's something I stole from Jasmin B, it's just to indicate a potential tweaking point for heuristics :).

PR #5994, opened 2022-04-23
Document gotcha with z3-js

ViRb3:
Took me a lot of time to troubleshoot this issue, since the error that is thrown is very deceiving and completely unrelated to z3. Please have a read of the diff to see more details.


NikolajBjorner:
Thanks for augmenting the doc based on your experience!

PR #6003, opened 2022-04-26
ocaml: fix is_arithmetic_numeral and is_bv_numeral

percontation:


percontation:
Currently, newly created numerals aren't APP_AST, they're NUMERAL_AST, so this was happening:

```
utop # let ctxt = Z3.mk_context [];;
val ctxt : Z3.context = <abstr>
utop # let num = Z3.BitVector.mk_numeral ctxt "1" 1;;
val num : Z3.Expr.expr = <abstr>
utop # Z3.BitVector.is_bv_numeral num;;
- : bool = false
```

One Z3 internals question I'm unclear about: Is it somehow still possible to see OP_ANUM or OP_BNUM as an APP_AST? If so, I suppose `is_arithmetic_numeral` and `is_bv_numeral` should be checking for either `AST.is_app x || AST.is_numeral x`.

PR #6025, opened 2022-05-09
Generate bdist wheels for musllinux_1_1

liath:
This PR adds steps to the release scripts to package bdist wheels for musllinux, saving users of musl based distros the hassle of waiting for z3 to compile. I expect this is largely identical to the efforts of the Alpine `z3` and `py3-z3` packages but fully contained in pip instead of relying on system packages. The setup.py changes are probably the sketchiest part of this as they require Python 3.10+. I wasn't sure how big of a concern that is and the other solutions that came to mind were uglier so I left it as is, please let me know if there's a better approach. :)

### DIFF
--- a/scripts/release.yml
+++ b/scripts/release.yml
@@ -129,13 +129,23 @@ stages:
         artifactName: 'UbuntuDoc'
         targetPath: $(Build.ArtifactStagingDirectory)
 
-  - job: ManyLinuxBuild
-    displayName: "ManyLinux build"
+  - job: LinuxBuilds
+    strategy:
+      matrix:
+        manyLinux:
+          name: ManyLinux
+          image: "quay.io/pypa/manylinux2010_x86_64:latest"
+          python: "/opt/python/cp37-cp37m/bin/python"
+        muslLinux:
+          name: MuslLinux
+          image: "quay.io/pypa/musllinux_1_1_x86_64:latest"
+          python: "/opt/python/cp310-cp310/bin/python"
+    displayName: "$(name) build"
     pool:
       vmImage: "ubuntu-latest"
-    container: "quay.io/pypa/manylinux2010_x86_64:latest"
+    container: $(image)
     variables:
-      python: "/opt/python/cp37-cp37m/bin/python"
+      python: $(python)
     steps:
     - task: PythonScript@0
       displayName: Build
@@ -160,7 +170,7 @@ stages:
         targetFolder: $(Build.ArtifactStagingDirectory)
     - task: PublishPipelineArtifact@0
       inputs:
-        artifactName: 'ManyLinuxBuild'
+        artifactName: '$(name)Build'
         targetPath: $(Build.ArtifactStagingDirectory)
   
   - template: build-win-signed.yml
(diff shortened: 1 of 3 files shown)


scripts/release.yml
```diff
@@ -129,13 +129,23 @@ stages:
...
-  - job: ManyLinuxBuild
-    displayName: "ManyLinux build"
+  - job: LinuxBuilds
+    strategy:
+      matrix:
+        manyLinux:
+          name: ManyLinux
+          image: "quay.io/pypa/manylinux2010_x86_64:latest"
+          python: "/opt/python/cp37-cp37m/bin/python"
+        muslLinux:
```
NikolajBjorner, line 139: How stable is this?

I am trying to patch the script to revive muslLinux, but now get

```
Digest: sha256:5120c468fd2abee7ff03fa3743fa583ea2723236c8b98c6351c05c9d6ade1def
Status: Downloaded newer image for quay.io/pypa/musllinux_1_1_x86_64:latest
quay.io/pypa/musllinux_1_1_x86_64:latest
/usr/bin/docker info -f "{{range .Plugins.Network}}{{println .}}{{end}}"
bridge
host
ipvlan
macvlan
null
overlay
/usr/bin/docker network create --label f4ce27 vsts_network_fb5d973063f84550b52a9374e742e232
5b212dd2563d549e0361d47cfbf77945c117c84964cea1fb58c62243ec5d4900
/usr/bin/docker inspect --format="{{index .Config.Labels \"com.azure.dev.pipelines.agent.handler.node.path\"}}" quay.io/pypa/musllinux_1_1_x86_64:latest
/usr/bin/docker create --name f0fe55fa190646f9b49677ca60639c44_quayiopypamusllinux_1_1_x86_64latest_3f1385 --label f4ce27 --network vsts_network_fb5d973063f84550b52a9374e742e232  -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/vsts/work/1":"/__w/1" -v "/home/vsts/work/_temp":"/__w/_temp" -v "/home/vsts/work/_tasks":"/__w/_tasks" -v "/opt/hostedtoolcache":"/__t" -v "/home/vsts/agents/2.204.0/externals":"/__a/externals":ro -v "/home/vsts/work/.taskkey":"/__w/.taskkey" quay.io/pypa/musllinux_1_1_x86_64:latest "/__a/externals/node/bin/node" -e "setInterval(function(){}, 24 * 60 * 60 * 1000);"
a927978a15f94fa5e7fb0e75683b6f961214718b6b2618d6025d8681be164c4a
/usr/bin/docker start a927978a15f94fa5e7fb0e75683b6f961214718b6b2618d6025d8681be164c4a
a927978a15f94fa5e7fb0e75683b6f961214718b6b2618d6025d8681be164c4a
/usr/bin/docker ps --all --filter id=a927978a15f94fa5e7fb0e75683b6f961214718b6b2618d6025d8681be164c4a --filter status=running --no-trunc --format "{{.ID}} {{.Status}}"
a927978a15f94fa5e7fb0e75683b6f961214718b6b2618d6025d8681be164c4a Up Less than a second
/usr/bin/docker exec  a927978a15f94fa5e7fb0e75683b6f961214718b6b2618d6025d8681be164c4a sh -c "command -v bash"
Error response from daemon: Container a927978a15f94fa5e7fb0e75683b6f961214718b6b2618d6025d8681be164c4a is not running
##[error]Docker exec fail with exit code 1
Finishing: Initialize containers
```

NikolajBjorner:
I merged this way too early.
The nightly was perhaps broken for a different reason, but it has now been broken since thanks to at least one issue: that the variable $(name) doesn't get resolved to the value in the matrix in the command that specifies the "container".
I would expect the scope of $(name) to apply to everything below the "strategy" stub, but apparently this was not so in Azure pipelines.
A second issue could be stability of the muslLinux container as above.

A third issue was uncovered with the ManyLinux-latest container where the version of the compiler used there has a bug.
It does not like 

```
for (auto& [k, coeff] : xyzuv) {
     if (abcd)
        return coeff;
}
```
where the return type is a reference.
The compiler segfaults.

PR #6026, opened 2022-05-09
Spacer Global Guidance

agurfinkel:
Global guidance for Spacer

Based on the paper:

Hari Govind VK, Yuting Chen, Sharon Shoham, Arie Gurfinkel: Global Guidance for
Local Generalization in Model Checking. CAV (2) 2020: 101-125

Adds rules to infer patterns in lemmas and generate conjectures based on them.

To enable, use `fp.spacer.global=true`, which is disabled by default to preserve
existing behavior. To control individual strategies, see options under
`fp.spacer.gg.`.


agurfinkel:
@NikolajBjorner this is ready for another look. 

Lightly tested, so don't merge yet. There are issues with formatting (will fix). 
If everything looks ok, I'll try to do another restructuring pass to hide some of the code better in cpp files.

I left many of the original comments. They don't all make sense now, but they are indicative. 
Will do a pass to clean them up before this is merged.

agurfinkel:
@NikolajBjorner provided that the CI succeeds, this is ready to be merged.

PR #6048, opened 2022-05-23
Add high level bindings for js

ritave:
I've started writing high level bindings for javascript for my own needs based on z3py, but I've noticed discussion #6046.

Let's decide how to proceed, but I'm creating this draft PR for posterity sake.


bakkot (commented):
Some small comments at a quick glance.

bakkot:
Re this comment comment:

> //               We could use templated string literals Solver<'contextName'> but that would
> //               be cumbersome for the end user

I don't think it actually does end up cumbersome, if you provide a default and encourage users to use that as the name if they only have one context. That's what I did [here](https://github.com/bakkot/z3/blob/a669ddc6b95fa8201c7b5620da208def15b58df3/src/api/js/example-api.ts#L61-L66) and it ends up working [just fine](https://github.com/bakkot/z3/blob/a669ddc6b95fa8201c7b5620da208def15b58df3/src/api/js/example-api.ts#L64-L66), without the end user ever having to name the generic parameter explicitly (unless of course they're working with multiple contexts, but then you _want_ to distinguish between `Solver<'a'>` and `Solver<'b'>`).

src/api/js/src/high-level.ts
```diff
@@ -0,0 +1,1362 @@
...
+  Z3_probe,
+  Z3_sort,
+  Z3_sort_kind,
+  Z3_symbol,
+  Z3_symbol_kind,
+  Z3_tactic
+} from "../build/wrapper";
+
+type ConfigParams = {
+  proof: boolean;
```
NikolajBjorner: This should either be auto-generated or obtained dynamically. The options can change and should only be updated in the main source.

src/api/js/src/high-level/types.ts
```diff
@@ -0,0 +1,415 @@
...
+  getFullVersion(): string;
+  openLog(filename: string): boolean;
+  appendLog(s: string): void;
+  setParam(key: string, value: any): void;
+  setParam(key: Record<string, any>): void;
+  resetParams(): void;
+  getParam(name: string): string | null;
+
+  // Operations that require context
+  createContext(): Z3WithContext<string>;
```
bakkot, line 313: `string` is the supertype of all strings, so making this default to `string` will prevent any useful context-based typechecking. E.g., it means that `api.createContext().And(api.createContext('other').Bool('x'))` will typecheck. The type system can and should prevent that from working.

If the name isn't required then no one is ever going to know it exists, whereas if the name is required it's a lot more obvious and not any harder to use. Do you think it's important to allow unnamed contexts?

src/api/js/src/high-level/types.ts
```diff
@@ -0,0 +1,415 @@
...
+  Z3_pattern,
+  Z3_probe,
+  Z3_solver,
+  Z3_sort,
+  Z3_sort_kind,
+  Z3_symbol,
+  Z3_tactic,
+} from '../low-level';
+
+export type AnySort<Name extends string = string> = SortRef<Name> | BoolSortRef<Name>;
```
bakkot, line 19: You almost certainly don't want to default this to `string` either, because it means the types will be too permissive if you ever forget the type parameter. For example, [here](https://github.com/Z3Prover/z3/blob/3a5255deeae65c713326f8ad2142285098eed527/src/api/js/src/high-level/types.ts#L127) you've dropped the type parameter and so accepts anything which satisfies `ExprRef<string>`,  even stuff from other contexts, as in

```ts
const one = api.createContext('one');
const two = api.createContext('two');

let sort: SortRef<'one'> = one.BoolSort();
sort.cast(two.Bool('a')); // typechecks, but shouldn't
```

I think it makes sense to pick a particular string to be the default name and use that as the default value for the type parameter.

(This isn't just about maintaining the library - users who are working with multiple contexts are going to run into this also. If you only have one context, having that use the default name will mean never needing to think about the type parameter; if you have multiple contexts, you really want to make sure the type system is enforcing that they can't mix, which means not having permissive defaults.)

src/api/js/src/high-level/types.ts
ritave: You're right, skipping the whole type system is not a great idea, I'll backtrack on that.

src/api/js/src/high-level/types.ts
ritave: Since we have to do run-time Contexts checks anyways, I've decided to specifically optimize for user convenience rather than strict type correctness. 

If we go for correctness, all functions would have to be generic to work correctly.
```typescript
// This is nice to write
async function allSolutions(api: Z3WithContext, solver: Solver): Promise<Model[]> {
  solver.push();
  const results = [];
  while ((await solver.check()) === sat) {
    const model = solver.model();
    results.push(model);
    solver.add(api.Or(...model.decls().map(decl => decl.neq(model.get(decl)))));
  }
  solver.pop();
  return results;
}

// This is harder
async function allSolutionsGeneric<ContextName extends string>(
  api: Z3WithContext<ContextName>,
  solver: Solver<ContextName>,
): Promise<Model<ContextName>[]> {
  // ...
}
```

It gets even more unwieldy with complicated types
```typescript
function createMapping(api: Z3WithContext, bools: BoolRef[], ints: ArithRef[]): AstMap<BoolRef, ArithRef> {
  const map = new api.AstMap<BoolRef, ArithRef>();
  for (const [bool, int] of zip(bools, ints)) {
    map.set(bool, int);
  }
  return map;
}

function createMappingGeneric<ContextName extends string>(
  api: Z3WithContext<ContextName>,
  bools: BoolRef<ContextName>[],
  ints: ArithRef<ContextName>[],
): AstMap<BoolRef<ContextName>, ArithRef<ContextName>, ContextName> {
  const map = new api.AstMap<BoolRef<ContextName>, ArithRef<ContextName>>();
  for (const [bool, int] of zip(bools, ints)) {
    map.set(bool, int);
  }
  return map;
}
```

Having a default name for context will not alleviate the problem either.
```typescript
export interface Solver<Name extends string = 'main'> {
  /* ... */
}

declare function allSolutions(api: Z3WithContext, solver: Solver): Promise<Model[]>;

// This will now fail to compile, you have to make all of your functions generic anyways
allSolutions(new createContext('myContext').Solver())
```

Also should we have a default name, the same omission problem you've noticed could occur as well.

bakkot: > If we go for correctness, all functions written by the users would have to be generic to work correctly.

Only if working with multiple contexts. If you're only dealing with a single context, and you name it `'main'`, then you can just write `Solver` and it'll work. And programs dealing with multiple contexts are precisely the ones which benefit from having accurate type checks, so only the programs which benefit will have to pay that cost.

> Having a default name for context will not alleviate the problem either.

It will alleviate it as long as users who only need a single context are encouraged to use the default name as the name of their context, right?

> Also should we have a default name, the same omission problem you've noticed could occur as well.

I'm not following, say more?

src/api/js/scripts/parse-api.ts
```diff
@@ -32,35 +31,49 @@ let primitiveTypes = {
...
-    Z3_created_eh: 'Z3_created_eh',
-    Z3_decide_eh: 'Z3_decide_eh'
-};
+const types = {
+  __proto__: null,
 
-let defApis = Object.create(null);
-let functions = [];
-let enums = Object.create(null);
+  // these are function types I can't be bothered to parse
```
NikolajBjorner, line 48: this comment predates the macro Z3_DECLARE_CLOSURE.
You can use Z3_DECLARE_CLOSURE to initialize this table.

src/api/js/src/high-level.ts
NikolajBjorner: I added a function to the API to get descriptions of global parameters. 

https://github.com/Z3Prover/z3/blob/dd46224a1dd91c5bb7a3e12a52a23cf87441e6c1/src/api/z3_api.h#L1694

If you decide to bring ConfigParams back, then hopefully this feature can be used to populate contents dynamically.

bakkot: @NikolajBjorner is there any way of knowing more precisely which strings are accepted for a given parameter, or should they all just be typed as "string"?

Also, is there any way to get the default values for parameters? A way of getting the current value for the given context would be sufficient, since you can just make a context without setting any parameters and then read off the values for that context.

---

The fact that it has to be invoked at runtime rather than parsed from source means it will add another step to the build process, but we do have a thing we can invoke after calling `build:wasm`, so:

```js
'use strict';

(async () => {
  let mod = await require('./build/z3-built.js')();
  let ctx = mod._Z3_mk_context();
  let paramDescriptors = mod._Z3_get_global_param_descrs(ctx);
  mod._Z3_param_descrs_inc_ref(ctx, paramDescriptors);

  let configType = 'type ConfigParams = {';
  let size = mod._Z3_param_descrs_size(ctx, paramDescriptors);
  for (let i = 0; i < size; ++i) {
    let nameSymbol = mod._Z3_param_descrs_get_name(ctx, paramDescriptors, i);
    let name = mod.UTF8ToString(mod._Z3_get_symbol_string(ctx, nameSymbol));
    let comment = mod.UTF8ToString(mod._Z3_param_descrs_get_documentation(ctx, paramDescriptors, nameSymbol));
    configType += `\n  ${name}: string, // ${comment}`
  }
  configType += `\n};`;

  console.log(configType);
  mod.PThread.terminateAllThreads();
})().catch(e => {
  console.error(e);
  process.exit(1);
});
```
produces
```ts
type ConfigParams = {
  proof: string, // proof generation, it must be enabled when the Z3 context is created
  smtlib2_compliant: string, // enable/disable SMT-LIB 2.0 compliance
  timeout: string, // (default: infty) timeout in milliseconds.
  dump_models: string, // dump models whenever check-sat returns sat
  model_validate: string, // validate models produced by solvers
  model: string, // model generation for solvers, this parameter can be overwritten when creating a solver
  debug_ref_count: string, // debug support for AST reference counting
  encoding: string, // string encoding used internally: unicode|bmp|ascii
  trace: string, // trace generation for VCC
  trace_file_name: string, // trace out file name (see option 'trace')
  well_sorted_check: string, // type checker
  rlimit: string, // default resource limit used for solvers. Unrestricted when set to 0.
  unsat_core: string, // unsat-core generation for solvers, this parameter can be overwritten when creating a solver, not every solver in Z3 supports unsat core generation
  dot_proof_file: string, // file in which to output graphical proofs
  type_check: string, // type checker (alias for well_sorted_check)
  auto_config: string, // use heuristics to automatically select solver and configure it
  stats: string, // enable/disable statistics
  memory_max_alloc_count: string, // set hard upper limit for memory allocations, if 0 then there is no limit
  memory_high_watermark: string, // set high watermark for memory consumption (in megabytes), if 0 then there is no limit
  verbose: string, // be verbose, where the value is the verbosity level
  warning: string, // enable/disable warning messages
  memory_max_size: string, // set hard upper limit for memory consumption (in megabytes), if 0 then there is no limit
};
```

NikolajBjorner: > @NikolajBjorner is there any way of knowing more precisely which strings are accepted for a given parameter, or should they all just be typed as "string"

Z3 doesn't expose set of legal string options.

NikolajBjorner: > Also, is there any way to get the default values for parameters? A way of getting the current value for the given context would be sufficient, since you can just make a context without setting any parameters and then read off the values for that context.

You can get the parameter descriptions, then for each enumerated parameter retrieve the (current) parameter value.
The parameter defaults are not accessible as a separate value.

bakkot: How do you retrieve the current parameter value?

NikolajBjorner: > 
The fact that it has to be invoked at runtime rather than parsed from source means it will add another step to the build process, but we do have a thing we can invoke after calling build:wasm, so:

Right, this is not great. I thought that instead of exposing the parameters as a type "ConfigParams" as a map, it would be exposed as a closure that can be queried. Or if attributes can be created dynamically.

NikolajBjorner: I learned last week that we also have to update our documentation build to create an index of parameters and their descriptions.
The option z3.exe -pm (and similar) is available from the command-line for users, but I learned that a (power) user spent quite some time first browsing the documentation (that needs other facelift) for a description of parameters.
My own behavior is far from representative: I just browse the source code.

bakkot:
@ritave I have a few more comments on the design, but if you're still iterating I'm happy to wait until things settle down, if you'd prefer.

ritave:
@bakkot Now's a good time for feedback. The design is settling and I'll be adding BitVectors next

scripts/mk_util.py
```diff
@@ -246,10 +246,13 @@ def rmf(fname):
 
 def exec_compiler_cmd(cmd):
     r = exec_cmd(cmd)
-    if is_windows() or is_cygwin_mingw() or is_cygwin() or is_msys2():
-        rmf('a.exe')
-    else:
-        rmf('a.out')
+    # Windows
```
NikolajBjorner, line 249: these changes are fine, but could you add them in a self-contained pull request?
They seem to work regardless of the new work on js/ts.
It is better to get such independent pieces in earlier instead of in a single bulk.

src/api/js/src/high-level/high-level.ts
```diff
@@ -0,0 +1,1636 @@
...
+      Object.entries(params).forEach(([key, value]) => Z3.set_param_value(cfg, key, value.toString()));
+      const context = Z3.mk_context_rc(cfg);
+
+      this.ptr = context;
+      this.name = name;
+
+      Z3.set_ast_print_mode(this.ptr, Z3_ast_print_mode.Z3_PRINT_SMTLIB2_COMPLIANT);
+      Z3.del_config(cfg);
+
+      this.AstRef = AstRefImpl.bind(AstRefImpl, this) as any;
```
NikolajBjorner, line 173: You have "Solver" and "AstRef".
Why not just "Ast"?
The Python API defines Ast as a primitive Python object. The AstRef class wraps around a pointer object. It contains an Ast field.
I don't see where the js bindings refer to an Ast, so this could be used instead of AstRef.

src/api/js/src/high-level/high-level.test.ts
```diff
@@ -0,0 +1,383 @@
+import assert from 'assert';
```
NikolajBjorner, line 1: FWIW: We are in the process of creating a supported version of the z3 guide that uses the SMTLIB front-end. It could also mix in js code, not previously possible.

src/api/js/src/high-level/types.ts
```diff
@@ -0,0 +1,520 @@
...
+export type AnyExpr<Name extends string = any> =
+  | ExprRef<Name>
+  | BoolRef<Name>
+  | PatternRef<Name>
+  | QuantifierRef<Name>
+  | ArithRef<Name>
+  | IntNumRef<Name>
+  | RatNumRef<Name>
+  | BitVecRef<Name>;
+export type AnyAst<Name extends string = any> = AnyExpr<Name> | AnySort<Name> | AstRef<Name> | FuncDeclRef<Name>;
```
NikolajBjorner, line 33: Not sure how AstRef plays in here as it should be one of Sort, Expr, FuncDecl.
We have a very few functions that work polymorphically.

src/api/js/src/high-level/high-level.ts
```diff
@@ -0,0 +1,1636 @@
...
+      this.ArithRef = ArithRefImpl.bind(ArithRefImpl, this);
+      this.IntNumRef = IntNumRefImpl.bind(IntNumRefImpl, this);
+      this.RatNumRef = RatNumRefImpl.bind(RatNumRefImpl, this);
+      this.AstVector = AstVectorImpl.bind(AstVectorImpl, this) as any;
+      this.AstMap = AstMapImpl.bind(AstMapImpl, this) as any;
+
+      // We want to bind functions and operations to `this` inside Context
+      // So that the user can write things like this and have it work:
+      // ```
+      // const { And, Or } = new Context('main');
```
bakkot, line 195: If you make everything closed over rather than accessing properties of `this`, then you don't need to `bind` anything and this code will still work. That will be a _much_ nicer implementation.

src/api/js/src/high-level/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+  do {
+    for (const key of Reflect.ownKeys(obj)) {
+      properties.add([obj, key]);
+    }
+  } while ((obj = Reflect.getPrototypeOf(obj)!) && obj !== Object.prototype);
+  return properties;
+}
+
+// https://github.com/sindresorhus/auto-bind, MIT license
+// We copy it because it uses ESM instead of CommonJS
```
bakkot, line 14: `autoBind` is MIT licensed and so cannot be copy-pasted without including the license. But this shouldn't be necessary anyway, as mentioned above.

src/api/js/src/jest.ts
```diff
@@ -0,0 +1,62 @@
...
+        } else if (result !== undefined) {
+          resolve(result);
+        }
+        resolve();
+      }, ms)),
+  );
+  return { ...promise, cancel: () => clearTimeout(handle) };
+}
+
+function spinlock(premise: () => boolean, pollMs: number = 100): Promise<void> & { cancel(): void } {
```
bakkot, line 37: This is not a spinlock, since it doesn't block the thread it's on. Maybe `waitFor`?

src/api/js/src/high-level/types.ts
```diff
@@ -0,0 +1,520 @@
...
+  readonly __typename: 'BoolSortRef';
+
+  cast(expr: BoolRef<Name> | boolean): BoolRef<Name>;
+  cast(expr: CoercibleToExpr<Name>): never;
+}
+
+export interface BoolRef<Name extends string = any> extends ExprRef<Name, BoolSortRef<Name>, Z3_ast> {
+  readonly __typename: 'BoolRef';
+
+  mul(other: BoolRef | boolean): BoolRef<Name>;
```
bakkot, line 322: It looks like this class wasn't finished?

Also this is missing the type parameter on the argument type.

src/api/js/src/high-level/high-level.test.ts
ritave: Interesting, do you mean a guide with interactive examples?

src/api/js/src/high-level/high-level.ts
ritave: Could you explain more on what you mean?

The API should still work like this
```typescript
const ctx = new Context('main');
```

src/api/js/src/high-level/types.ts
```diff
@@ -0,0 +1,520 @@
...
+  Z3_pattern,
+  Z3_probe,
+  Z3_solver,
+  Z3_sort,
+  Z3_sort_kind,
+  Z3_symbol,
+  Z3_tactic,
+} from '../low-level';
+
+export type AnySort<Name extends string = any> =
```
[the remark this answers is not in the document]
bakkot: > I played around with some examples and I also strongly feel having a default context name is confusing for end users.

Can you say more about why? And do you think there's a case where it's confusing even if you're only ever using one context, or is the confusion something which only arises when using multiple contexts?

src/api/js/src/high-level/high-level.test.ts
```diff
@@ -0,0 +1,383 @@
...
+    expect(getParam('pp.decimal')).toStrictEqual('false');
+    expect(getParam('timeout')).toStrictEqual('4');
+
+    resetParams();
+    expect(getParam('pp.decimal')).toStrictEqual('false');
+    expect(getParam('timeout')).toStrictEqual('4294967295');
+  });
+
+  it('proves x = y implies g(x) = g(y)', async () => {
+    const { Solver, Int, IntSort, Function, Implies, Not } = new api.Context('main');
```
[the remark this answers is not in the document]
bakkot: C# is a different language with different idioms than JavaScript. In JavaScript, it is very unusual to use a class for producing a bag of names. For example, `new Proxy` gives you an instance, but there is a nearby API called `Proxy.revokable` which returns two named things, and it is not called with `new`.

`const { Solver, Int, IntSort, Function, Implies, Not } = api.creatContext('main')` is definitely cleaner than using `new api.Context` in that place.

If there is a reason to treat a `context` object as something _other_ than a bag of names, maybe that's a reason to make it a class. Is there such a reason?

src/api/js/src/high-level/high-level.ts
bakkot: `bind` is only necessary for methods which refer to `this`, and there is no reason to do so for this API.

That is, if instead of having

```js
class ContextImpl {
  _assertContext(...ctxs: (Context | { ctx: Context })[]) {
    ctxs.forEach(other => assert('ctx' in other ? this === other.ctx : this === other, 'Context mismatch'));
  }

  isAst(obj: unknown): obj is Ast {
    const r = obj instanceof AstImpl;
    r && this._assertContext(obj);
    return r;
  }
}
return { Context: ContextImpl };
```
you had
```js
function createContext(name) {
  let ctx = new ContextImpl(name);
  function _assertContext(...ctxs: (Context | { ctx: Context })[]) {
    ctxs.forEach(other => assert('ctx' in other ? ctx === other.ctx : ctx === other, 'Context mismatch'));
  }
  return {
    isAst(obj: unknown): obj is Ast {
      const r = obj instanceof AstImpl;
      r && _assertContext(obj); // note lack of `this`
      return r;
    }
  };
}
return { Context: createContext };
```
then everything would work without `bind`. And, in fact, you could still write `const ctx = new Context('main')`, if you really wanted to.

src/api/js/src/high-level/high-level.test.ts
NikolajBjorner: right.
Several years ago there were two guides exposed: one using SMTLIB2 formulas written into a text box. The other wrapped around Python. They used different backends and required hosting. This had to be discontinued.
With wasm bindings there need not be two different tutorials. The self-contained web pages can host examples in text and examples in code.

ritave:
@bakkot @NikolajBjorner This is as much as I could write during my evenings and weekends. I believe it would be good to merge this PR in current state, even if the high-level is not fully complete. What's included works and is somewhat documented. More PRs should be added to continue development of high-level APIs.

Personally in the future I'd like to add support for Arrays when I have more time.

Ready for review.

NikolajBjorner:
very cool!

doc/mk_api_doc.py
```diff
@@ -47,6 +49,11 @@ def parse_options():
         default=False,
         help='Include ML/OCaml API documentation'
     )
+    parser.add_argument('--js',
```
NikolajBjorner, line 52: was this tested?
I am trying to reproduce it for the documentation build. See #6155.
At this point trying to produce the --js documentation on top of an emsdk build (not sure if this dependency is required).

An alternative approach may be to decentralize the documentation build and include it in the build-wasm GitHub action to produce an artifact (and wishful thinking to copy it automatically to z3prover.github.io, though I have myself been entirely unsuccessful in getting any pre-cooked action to work properly).

ritave: @NikolajBjorner I did test it beforehand, yes.

The emsdk is not required to build the documentation, and as you've noticed - it's actually a detriment due to shadowing of Node version. Running separate from emsdk build would be easiest, but if you want to do everything inside one build, the documentation should be built before `emsdk activate` is called. Same way the Typescript is built before emsdk in the [Github Action](https://github.com/Z3Prover/z3/blob/527914db05d80da4d0a070485ddd87720f514c66/.github/workflows/wasm.yml#L38)

@bakkot has written a correct solution to the second problem in #6155 - currently there is an unwritten expectation that the developer has run `npm install` in `src/api/js` directory. `npm --prefix=src/api/js ci` also works and is preferred. Another solution is to quick fail inside the `mk_api_doc.py` script by checking whether `src/api/js/node_modules` folder exists (it's added during `npm install`).

NikolajBjorner: OK, thanks it is getting closer now.
Currently some path discrepancy: the relative path used in jest.ts doesn't work with the setup in mk_api_doc.py.

NikolajBjorner:
Z3_solver_to_string and related methods for displaying the state of objects seem to be cool to expose over high-level.

https://github.com/Z3Prover/z3/blob/59d47e30559d90a35fabb2c1f6fdfadf8ee11fd1/src/api/js/src/high-level/types.ts#L383

As a JS novice plan to be hitting on the API in coming days. 

https://microsoft.github.io/z3guide/docs/Z3%20JavaScript/test

bakkot:
`Z3_solver_to_string` [actually is exposed](https://github.com/Z3Prover/z3/blob/9cd3b9cad76a70f057600012eac716a0bb81464d/src/api/js/src/high-level/high-level.ts#L1004): if you call `solver.toString()` on a Solver object, that calls `Z3_solver_to_string`.

(`toString` isn't listed in the types because it's on every JS object.)

Any other things you recommend exposing? I don't know enough about the API to know which things are most helpful.

NikolajBjorner:
OK, I found a first epsilon 907dc2c2d262b44aa210044af30a34daf898dac1

bakkot:
Nice! Feel free to ping if you find other major deficiencies (other than the user propagators, which will require more effort to get working in the JS bindings).

By the way, towards making the bindings more useful, a question for you: if a thread is killed (from the outside, not with `interrupt`) while something like `Z3_eval_smtlib2_string` is executing on that thread, will that leave the rest of Z3 (in other threads) in a bad state? In this build `interrupt` won't work, I think, so forcibly killing threads from the outside is the only way to terminate a long-running computation (other than reloading the page).

If killing a thread is an OK thing to do, I can add a helper for cancelling long-running work.

NikolajBjorner:
if killing a thread doesn't interfere with locks held by the thread :-), then it would be fine modulo memory leaks.

NikolajBjorner:
user propagators aren't even done for Python. It is better to leave those aside for some time as the API matures from experiences. The main experiences now are over C++ and .Net. The python API is usable except for "fresh()" (required for quantifiers) and some newer features (decision callbacks).

NikolajBjorner:
single threaded is an oxymoron: for timeouts there is a threadpool that invokes a callback when the timeout expires.
It uses threads. The way I recollect, this caused the pthread bug with the main window.
I also vaguely recollect asking if we should just have a different implementation for timeouts in case of JS.
The old implementations used OS specialized code (very ugly), an intermediary way spawned a thread with every timeout and this was found to be super expensive for applications that needed lots of small calls after each other. They could just use the same thread. Thus began a saga of a thread pool optimized timeout implementation. Several deadlocks and race conditions later (and 25 years of conference papers with logics for proving concurrent programs correct) it seems stable.

bakkot:
Yes, the JS build does have multiple threads in the case of timeouts, as you point out. But I was hoping that building with `--single-threaded` means it won't have locks which would need to get released if a thread was forcibly killed.

I guess I'll just implement it and see if it works.

NikolajBjorner:
As someone embracing panic age, I can say it isn't that timer queues are any new. But they are timeless.
A timer queue implementation had perf bug fund and fixed about 20 years ago: 
insert n entries each with timeout 0 into a queue and get O(n^2) overhead thanks to the marvels of insertion sort.

NikolajBjorner:
There is invariably going to be "minor" deficiencies that get exposed when the API is used by a novice.
I added an expression to solve and get a funny error message.

const x = Z3.Int.const('x');
const y = Z3.Int.const('y');
(await Z3.solve(x.add(y).sexpr()))
Output:
Error: 
Cannot use 'in' operator to search for 'ctx' in (+ x y)

Overal, I would like to see if I can add a good set of examples to this page:

https://microsoft.github.io/z3guide/docs/Z3%20JavaScript/more%20tests/

and then establish tutorial material that is interactive for a more polished set of pages in the next weeks.
(note the JS interpreter mode is a very first cut, I think we have a bit to learn from how it should be done).

> Note that these pages are also set up to pose pull requests. Ehm. hint, hint @ritave.
Seriously, we are considering how to make these pages more interactive to lower the barrier for sharing code examples and having discussions.

The python (on pyodide it seems) is further away as integrating a micropip package is taking me some time to troubleshoot (a combination of not being particularly good at build systems and the situation that z3 is quite fat compared to other mixed mode packages).

bakkot:
I strongly recommend using TypeScript, which will catch errors of that form at compile time. For example, the above would have failed with `Argument of type 'string' is not assignable to parameter of type 'Bool<"main">'.`

It's possible to do this client-side, for the demo site, although it is a little bit involved. I can put a PR together at some point if I have time.

walnutwaldo:
I'm currently looking into improving the bindings to support Arrays and Quantifier expressions and have started working on this locally. @ritave you mentioned earlier that you wished to work on this as well in the future. Have you thought about how to best implement the type system with Arrays? i.e. I imagined we would want the definition to be something like `Array<DomainSort, RangeSort, Name>` but this seems like it would run into weird recursive type referencing issues

ritave:
@walnutwaldo I haven't looked into arrays yet.

Currently the whole type system is somewhat self-referencing, but if you help it by casting it into proper shape it works properly. I'd say it's worth trying to do a quick experiment if your type works.

ritave:
@walnutwaldo It's pretty hectic right now, the earliest I could take a look is 24th Oct

walnutwaldo:
Is anyone aware of any benefits of having both `IntNum` and `RealNum` both under `ArithSort` rather than having child sorts `IntSort` and `RealSort`? It feels like the type system would be more natural this way and I think more reflective of the underlying C. Unless there is a good reason to keep it the current way, I could go ahead and switch it.

ritave:
@walnutwaldo If I remember correctly, it's copying internal C++ hierarchy

PR #6101, opened 2022-06-22
Make high-level JS API more idiomatic/type-safe

bakkot:
This is a followup to @ritave's #6048. This makes the implementation and API a little more idiomatic, makes the typechecking stricter, adds a more complete example, and adds error checking. Specifically:

- the first commit makes the implementation more idiomatic JavaScript (mainly by not using classes and `bind` when a single object literal and closures will work)
- in the second commit makes the type checker able to distinguish between contexts by default (previously it used `any`, which is a special non-type sigil which tells the type checker not to check types, basically)
- the third commit changes the return types for `solve` from `Symbol` to `string` - symbols are annoying in this case because they have to be imported every time they're used (also the description for `unsat` was wrong; it said "Z3 couldn't find a solution", whereas it actually means "determined there was not a solution")
- the fourth commit adds a couple of `toString` methods useful for printing results
- the fifth commit adds the example
- the sixth commit makes `from` on integral values create an Int rather than a Real, which makes the example run ~25x faster on my machine (26 seconds, instead of 11 minutes)
- the seventh commit adds `Z3_get_error_code !== Z3_OK` checks after every nontrivial call to the low-level API
- the eighth commit adds an operator which was apparently left out of the types
- the ninth commit replaces various getters with methods, since getters should not be expensive (as calling out to wasm arguably is)

### DIFF
--- a/src/api/js/src/high-level/high-level.ts
+++ b/src/api/js/src/high-level/high-level.ts
@@ -456,7 +456,15 @@ export function createApi(Z3: Z3Core): Z3HighLevel {
     function from(value: CoercibleToExpr<Name>): AnyExpr<Name> {
       if (typeof value === 'boolean') {
         return Bool.val(value);
-      } else if (typeof value === 'number' || isCoercibleRational(value)) {
+      } else if (typeof value === 'number') {
+        if (!Number.isFinite(value)) {
+          throw new Error(`cannot represent infinity/NaN (got ${value})`);
+        }
+        if (Math.floor(value) === value) {
+          return Int.val(value);
+        }
+        return Real.val(value);
+      } else if (isCoercibleRational(value)) {
         return Real.val(value);
       } else if (typeof value === 'bigint') {
         return Int.val(value);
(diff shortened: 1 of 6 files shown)


src/api/js/src/high-level/high-level.ts
```diff
@@ -150,1724 +148,1807 @@ export function createApi(Z3: Z3Core): Z3HighLevel {
...
     }
 
-    from(primitive: boolean): Bool;
-    from(primitive: number | CoercibleRational): RatNum;
-    from(primitive: bigint): IntNum;
-    from<T extends Expr>(expr: T): T;
-    from(expr: CoercibleToExpr): AnyExpr;
-    from(value: CoercibleToExpr): AnyExpr {
+    function from(primitive: boolean): Bool<Name>;
+    function from(primitive: number | CoercibleRational): RatNum<Name>;
```
ritave, line 465: @bakkot This PR left out correct typing  for new `from` logic. `from(5)` will return `IntNum` but the overload here and in `types.ts` says that it will be `RatNum`

ritave: @bakkot No worries. Added a small comment there. Cheers!

Issue #5990, opened 2022-04-19
Deadlock in `scoped_timer`


bmourad01:
I seem to be encountering deadlock at this point in the `scoped_timer`

https://github.com/Z3Prover/z3/blob/ec57d3b15c0e0f3a7192bd3c0213a9ab3b8a7084/src/util/scoped_timer.cpp#L54

Looking through the history, I see a relevant PR (#5371). However, this issue still appears to be present in the upstream version.

For what it's worth, I'm using Z3 via the OCaml bindings.

NikolajBjorner:
@nunoplopes - the appverifier allows to enable cuzz.

nunoplopes:
@bmourad01 can you share a repro, please? Preferably without using ocaml bindings.
Otherwise I can't debug it.

NikolajBjorner:
Have you run the existing unit tests using appverifier?

nunoplopes:
I've never used appverifier. I don't have MSVC installed either. I'm a bit short on computing resources at the moment.

NikolajBjorner:
It is either pre-installed as "appverif" or can be added.
https://www.microsoft.com/en-us/research/project/cuzz-concurrency-fuzzing/

NikolajBjorner:
Maybe @bmourad01 has an easier time running this (it assumes Windows) directly from the ocaml app, and instrument some logging to the scoped_timer class to produce a trace that can help understanding what is going on.
4 decades of CAV, model checking, fuzzing, verification, concurrent separation logic, ...

bmourad01:
I'm having a hard time coming up with a minimal example that reproduces the problem. Will update if that situation changes.

nunoplopes:
Please reopen if you are able to provide more information. Thank you!

fkohlgrueber:
I ran into the same issue yesterday and was able to reproduce it with a small example program (see [here](https://github.com/fkohlgrueber/z3/commit/eac7dfa1af9fc2a477fd53dbd8fda2a64b12b3f1)):

```c++
for (int i = 0; i< 1000000; i++) {
    if (i % 10 == 0) {
        std::cout << i << std::endl;
    }
    test_scoped_eh eh;
    auto delay = rand() % 20 + 1;
    scoped_timer sc(100, &eh);
    auto end = std::chrono::steady_clock::now() + std::chrono::milliseconds(delay);
    while (std::chrono::steady_clock::now() < end) {
    }
}
```

[...]

PR #6102, opened 2022-06-22
fix data race in scoped_timer

fkohlgrueber:
The fix for #5990 introduced a data race in `scoped_timer`, see my comment in #5990 for details.

This PR fixes the issue by making sure that the `s->work` field is written last. I also took the opportunity to refactor initializing the state into a helper method, which simplifies the control flow in the constructor.


fkohlgrueber:
Great, thanks for merging! Do you have any plans on when you'll do a 4.8.18 release?

PR #6118, opened 2022-06-29
fix gc'ed callbacks in .NET propagator api

mlevatich:


src/api/dotnet/UserPropagator.cs
```diff
@@ -76,6 +76,17 @@ public class UserPropagator
         CreatedEh created_eh;
         DecideEh decide_eh;
 
+        Native.Z3_push_eh push_eh;
```
NikolajBjorner, line 79: can't these be moved as well?

PR #6139, opened 2022-07-06
fix js distributable

bakkot:
https://github.com/Z3Prover/z3/pull/6048 reorganized the files in the JS api which made the published artifact not work (so the just-published 4.9.0 doesn't work at all; attempting to import it throws). This fixes that. It also marks the package as usable on node 18, due to the inclusion of https://github.com/Z3Prover/z3/pull/6136.

### DIFF
--- a/src/api/js/package-lock.json
+++ b/src/api/js/package-lock.json
@@ -27,7 +27,7 @@
         "typescript": "^4.5.4"
       },
       "engines": {
-        "node": ">=16 <18"
+        "node": ">=16"
       }
     },
     "node_modules/@ampproject/remapping": {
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
thanks for looking into this.
Let me know if I need to crank up 4.9.1 today or we can revert the npmjs dist.

bakkot:
It's possible to `npm unpublish` the new version, so that only the old version exists, but unfortunately you can't just then re-publish the fixed thing - npm won't let you reuse a version number even after unpublishing. It will require a new version to get something which works and also is current.

NikolajBjorner:
so I will then roll a 4.9.1 now

PR #6146, opened 2022-07-08
Fix build on Mac

victorpaleologue:


victorpaleologue:
Should fix https://github.com/Z3Prover/z3/issues/6147

Issue #6160, opened 2022-07-14
proposal: reduce concurrency for ocaml bindings


c-cube:
We've long had issues with Z3 and the OCaml bindings at Imandra, and had to avoid threads entirely because of that.

cc @giltho @xvilka @ivg 

## Problem description

I finally made a small repro for the issue (see below), which can occur as soon as there are at least 2 OCaml threads, one of which uses Z3. My understanding of the problem is roughly this:

- thread A calls Z3, creates Z3 ASTs, expressions, etc. The AST nodes are wrapped in OCaml values via _custom blocks_ (a special kind of block that lives in the OCaml heap, and can contain C/C++ data).
- thread B does other stuff.
- at some point, A calls `Z3.Solver.solve solver` and releases the runtime lock (long call to C code, allows for some parallelism). At this point A's code pointer is within C++.
- meanwhile, while A is in Z3's solving code, B calls the GC for unrelated reasons. It collects dead values, some of which happen to be dead OCaml references to AST nodes. The AST nodes themselves are still alive and in use in Z3, it's just the OCaml reference that is dead.
- B calls the finalizer on an OCaml wrapper of an AST node. This finalizer decrements a refcount inside Z3.

problems:
- the refcount is not atomic, so it can become wrong
- if refcount reaches 0, then thread B is going to release the AST node. It goes through code of ast.h/ast.cpp that is *not* thread-safe, while thread A might also be doing the same to allocate other unrelated AST nodes.

both of these are grounds for race conditions, even though only thread A ever used Z3 explicitly.

## Basic fix

The simpler fix is to avoid releasing the runtime lock when calling a long-running Z3 function (simplify, solve, etc.) This means B cannot run at the same time as A.

However it's annoying because now the program can do absolutely nothing else why Z3 is being called. Even with several distinct, independent Z3 instances, only one can work at a given time.

## Harder (potential) fix

I've given a quick try at making the deallocation path of AST nodes thread-safe. Unsurprisingly it didn't work; the question is, how much effort would be required to make it work, and is it worth it (and is it worth the maintainers' sanity).

The basic approach I started is to have refcounting use `atomic<unsigned>` for ast nodes (and in parts of the OCaml stubs); coupled with a mutex used, when a refcount reaches 0, to protect the deallocation path and everything  that touches `ast_manager::m_ast_table`.

## Repro

the following archive contains a small OCaml project.

[...]

PR #6161, opened 2022-07-14
(do not merge) attempt at making parts of z3 thread-safe

c-cube:
(related to #6160)

this is a partial attempt at securing the deallocation path of AST nodes against race conditions. It's really intended for discussion, I'm not going to pretend it works.


src/ast/ast.cpp
```diff
@@ -1799,6 +1806,7 @@ static void track_id(ast_manager& m, ast* n, unsigned id) {
 #endif
 
 ast * ast_manager::register_node_core(ast * n) {
+    LOCK_AST_TABLE;
```
NikolajBjorner, line 1809: this is a very common code-path.

NikolajBjorner:
How about we enable a concurrent mode for Z3_ast_dec_ref and other dec_refs?
In the concurrent mode, it checks the thread identifier used for creating the context. 
If the thread ids don't match, the dec_ref is called from a different thread and therefore put in a queue.
The queue is flushed when there is some other call on the context in the thread it was created.
The concurrent mode is outside ast_manager. It is at the level of api_context.

You may say a context cannot be migrated safely to a different thread by the application without an explicit instruction.
Seems this is a small price to pay because such features are for power users.

NikolajBjorner:
will not take this change as it goes too deep into the core that is allowed to assume single threaded mode.

PR #6186, opened 2022-07-22
fix z3str3 re.is_range symbolic argument bug

JohnLyu2:
Currently Z3str3 doesn't support re.is_range symbolic arguments.

### DIFF
--- a/src/smt/theory_str_regex.cpp
+++ b/src/smt/theory_str_regex.cpp
@@ -725,7 +725,11 @@ namespace smt {
             return _qadd(lo, cx);
         } else if (u.re.is_range(re, sub1, sub2)) {
             SASSERT(u.str.is_string(sub1));
+            if (!u.str.is_string(sub1))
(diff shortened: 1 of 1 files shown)


src/smt/theory_str_regex.cpp
```diff
@@ -768,7 +772,11 @@ namespace smt {
             return _qmul(2, cx);
         } else if (u.re.is_range(re, sub1, sub2)) {
             SASSERT(u.str.is_string(sub1));
+            if (!u.str.is_string(sub1))
```
NikolajBjorner, line 775: Why introduce 3 copies of essentially the same code three different places?
I have added a routine now u.re.is_range(re, ch1, ch2), it returns true if re is a regex range with singleton character bounds. ch1, ch2 are unsigned integers (character codes).

JohnLyu2: Hi I have modified the code using u.re.is_range(re, ch1, ch2) in #6189 .

PR #6191, opened 2022-07-25
Improve intra-doc linking.

waywardmonkeys:


waywardmonkeys:
This is based on similar changes that I've done in the Rust documentation for the C-like low level API.

PR #6210, opened 2022-07-29
Remove usages of Z3_bool, just use bool.

waywardmonkeys:


src/api/z3_api.h
```diff
@@ -1528,7 +1528,7 @@ extern "C" {
 
        def_API('Z3_global_param_get', BOOL, (_in(STRING), _out(STRING)))
```
NikolajBjorner, line 1529: Maybe a left-over of earlier C standards.

and typedef bool Z3_bool?

waywardmonkeys: Yes, you might recall that about 4 years ago, we did the bulk of switching from `Z3_bool` to `bool` and added a requirement for C people to have a compiler with `stdbool.h` ... this was part of what prompted you to switch from 4.6 to 4.7 to indicate the possibly breaking change.

With this change, nothing is left of `Z3_bool` except for the typedef and the `#define`s for `Z3_TRUE` and `Z3_FALSE`. I didn't want to get rid of those in the same commit as you might feel those were breaking changes and would be better for when you decide to bump from 4.10 to 4.11.

PR #6211, opened 2022-07-30
Remove empty leaf destructors.

waywardmonkeys:
This removes all empty leaf destructors since the generated destructor will do the right thing and there's already the required virtual destructor above this.

This is what I believe @nunoplopes was suggesting in #6207.


NikolajBjorner:
@nunoplopes - you get the token to rebase-merge

nunoplopes:
This is great, thank you!

PR #6236, opened 2022-08-02
test/lp: Replace if linux with if not windows.

waywardmonkeys:
This is stuff that works on posix, so we can flip the check.


waywardmonkeys:
These were the only real usage of `_LINUX_`, so I wanted to get rid of them for a future patch.

PR #6247, opened 2022-08-04
mk_unix_dist.py: Fix --nopython

waywardmonkeys:
Writing to the global PYTHON_ENABLED requires that it be flagged
as a global.


waywardmonkeys:
Found this via pyflakes, not by actually running it and having a problem. (I don't know how to run it exactly.)

PR #6256, opened 2022-08-04
(cmake): fix visibility on shell z3 binary

agurfinkel:
Commit #b361226 changed symbol visibility from a global to a local option.
This creates inconsistency for shell that is compiled as an executable rather
than as z3 component.

This commit adds same local options to shell target in cmake.

Prior to the fix, clang on OSX complains with lots of warnings about symbol visibility
being different in different translation units that are linked together

### DIFF
--- a/src/shell/CMakeLists.txt
+++ b/src/shell/CMakeLists.txt
@@ -34,6 +34,14 @@ add_executable(shell
 # we don't want (I think).
   ${shell_object_files}
 )
+
+set_target_properties(shell PROPERTIES
+    # Position independent code needed in shared libraries
+    POSITION_INDEPENDENT_CODE ON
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
Even if we do this fix for now, maybe it makes sense to set the defaults and avoid the per-project config?

The defaults are just the variables with `CMAKE_` as the prefix, so setting `CMAKE_POSITION_INDEPENDENT_CODE` to `ON` prior to creating any targets would make that the default...

And sorry about this ... I haven't built on macOS in the last couple of days so I didn't notice the new warnings yet.

NikolajBjorner:
I merge this as is. 
Setting defaults sounds reasonable to me. Separate push if it makes sense.

PR #6257, opened 2022-08-04
Disable _tzcnt_u32/_tzcnt_u64 for ARM64EC

mcfi:
ARM64EC doesn't support AVX intrinsics including _tzcnt_u32/_tzcnt_u64, which are mapped to AVX2 instruction tzcnt. Therefore, neuter such intrinsics for ARM64EC compilation.

### DIFF
--- a/src/util/mpz.cpp
+++ b/src/util/mpz.cpp
@@ -50,7 +50,7 @@ Revision History:
 
 #if defined(__GNUC__)
 #define _trailing_zeros32(X) __builtin_ctz(X)
-#elif defined(_WINDOWS) && !defined(_M_ARM) && !defined(_M_ARM64) && !defined(__MINGW32__)
+#elif defined(_WINDOWS) && !defined(_M_ARM) && !defined(_M_ARM64) && !defined(_M_ARM64EC) && !defined(__MINGW32__)
 // This is needed for _tzcnt_u32 and friends.
 #include <immintrin.h>
 #define _trailing_zeros32(X) _tzcnt_u32(X)
@@ -62,7 +62,7 @@ static uint32_t _trailing_zeros32(uint32_t x) {
 }
 #endif
 
-#if (defined(__LP64__) || defined(_WIN64)) && !defined(_M_ARM) && !defined(_M_ARM64)
+#if (defined(__LP64__) || defined(_WIN64)) && !defined(_M_ARM) && !defined(_M_ARM64) && !defined(_M_ARM64EC)
  #if defined(__GNUC__)
  #define _trailing_zeros64(X) __builtin_ctzll(X)
  #else


waywardmonkeys:
Would it make sense to change the logic so that this code only happens for the right processors instead of trying to flag all of the ones where it shouldn't, which is more fragile?

If so, then we could also add sections for the ARM64 platforms with MSVC to use the right intrinsics there (https://docs.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics?view=msvc-170).

mcfi:
Arm/arm64 doesn’t have same similar intrinsics now, so it’s not possible to check arm64 cpu capability and use intrinsics.

waywardmonkeys:
ah, right, we only look for trailing zeros, not leading.

The other part is valid though, we can change:

```
#if (defined(__LP64__) || defined(_WIN64)) && !defined(_M_ARM) && !defined(_M_ARM64) && !defined(_M_ARM64EC)
```

to check for `_M_X64` instead of  `!_M_ARM && !_M_ARM64 && !_M_ARM64EC`.

And similar for the other one ...

waywardmonkeys:
But now I also see that both x86_64 and arm64 can be defined at once in https://techcommunity.microsoft.com/t5/windows-kernel-internals-blog/getting-to-know-arm64ec-defines-and-intrinsic-functions/ba-p/2957235 ...

It just seems like we can make this a bit less fragile here.

NikolajBjorner:
> Would it make sense to change the logic so that this code only happens for the right processors instead of trying to flag all of the ones where it shouldn't, which is more fragile?

Right

src/util/mpz.cpp
```diff
@@ -50,7 +50,7 @@ Revision History:
 
 #if defined(__GNUC__)
 #define _trailing_zeros32(X) __builtin_ctz(X)
-#elif defined(_WINDOWS) && !defined(_M_ARM) && !defined(_M_ARM64) && !defined(__MINGW32__)
+#elif defined(_WINDOWS) && !defined(_M_ARM) && !defined(_M_ARM64) && !defined(_M_ARM64EC) && !defined(__MINGW32__)
```
NikolajBjorner, line 53: should be defined(_WINDOWS) && (defined(_X86/64) || ...)

src/util/mpz.cpp
```diff
@@ -62,7 +62,7 @@ static uint32_t _trailing_zeros32(uint32_t x) {
 }
 #endif
 
-#if (defined(__LP64__) || defined(_WIN64)) && !defined(_M_ARM) && !defined(_M_ARM64)
+#if (defined(__LP64__) || defined(_WIN64)) && !defined(_M_ARM) && !defined(_M_ARM64) && !defined(_M_ARM64EC)
```
NikolajBjorner, line 65: also I would then set the architecture conservatively.

NikolajBjorner:
I think I should set the architecture conservatively to handle this bug/pull

mcfi:
Note that in ARM64EC compilation, both the _M_X64 and _M_ARM64EC macros are defined. To really only limit the scope of the tzcnt intrinsics, you need to check (defined(_M_X86) || defined(_M_X64) && !defined(_M_ARM64EC)) if you want a more conservative change. Thanks.

PR #6284, opened 2022-08-19
Fix Android CI CMake Java JNI error

ekilmer:
Specifying `CMAKE_ANDROID_API` is apparently more correct than specifying `CMAKE_SYSTEM_VERSION` on CMake 3.24 (and earlier versions). Reference is here https://cmake.org/cmake/help/v3.24/manual/cmake-toolchains.7.html#cross-compiling-for-android-with-the-ndk where it describes the `CMAKE_SYSTEM_VERSION` variable:
> If not specified, the value is determined as follows: If the CMAKE_ANDROID_API variable is set, its value is used as the API level."

Passing CI is here for a commit that enabled CI on my branch for these changes https://github.com/ekilmer/z3/actions/runs/2890880110

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -27,7 +27,7 @@ jobs:
       run:  |
         mkdir build
         cd build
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
Thanks for taking a look

PR #6297, opened 2022-08-23
Change to 4 digit assembly version

jfleisher:
This change allows symbols to be archived for each new build rather than overwriting the archived symbols for the 3 digit build number that is manually updated. 

For example: version will go from 4.11.1.0 to 4.11.1.4515

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -91,6 +91,7 @@ TRACE = False
 PYTHON_ENABLED=False
 DOTNET_CORE_ENABLED=False
 DOTNET_KEY_FILE=getenv("Z3_DOTNET_KEY_FILE", None)
+ASSEMBLY_VERSION=getenv("Z2_ASSEMBLY_VERSION", None)
 JAVA_ENABLED=False
 ML_ENABLED=False
 PYTHON_INSTALL_ENABLED=False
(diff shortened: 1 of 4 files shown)


scripts/mk_win_dist.py
```diff
@@ -299,9 +307,10 @@ def cp_licenses():
     cp_license(False)
 
 def init_flags():
-    global DOTNET_KEY_FILE, JAVA_ENABLED, PYTHON_ENABLED
+    global DOTNET_KEY_FILE, JAVA_ENABLED, PYTHON_ENABLED, ASSEMBLY_VERSION
```
waywardmonkeys, line 310: Since these globals aren't modified, I think that this `global ...` line is actually wrong in the first place and not needed. But that's separate from these changes.

waywardmonkeys (changes requested):
Some nits

scripts/mk_util.py
```diff
@@ -1694,10 +1716,11 @@ def mk_makefile(self, out):
             key = "<AssemblyOriginatorKeyFile>%s</AssemblyOriginatorKeyFile>" % self.key_file
             key += "\n<SignAssembly>true</SignAssembly>"
 
-        version = get_version_string(3)
+        version = get_version_string(4)
 
-        core_csproj_str = """<Project Sdk="Microsoft.NET.Sdk">
+        print(f"Version output to csproj: {version}")
```
NikolajBjorner, line 1721: sorry, but we can't use formatting with 'f' in our build scripts due to down-level python under some platforms.

PR #6318, opened 2022-09-03
Pass size to deallocation functions

nunoplopes:
WARNING: This is a proof of concept that needs a lot of testing and some fixes.

The goal of this PR is to remove the size data we store alongside every allocation. Most of the times we can get the alloc size automatically through compiler magic.
With vectors we cannot, and thus users need to pass in the number of elements.

This PR builds at least! I didn't remove the size data, but use it to assert that the passed data is equal.
Only when we are sure all is good we can move on with removing the data & reaping the perf benefits 🥳

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -757,7 +757,7 @@ public:
     }
 
     ~tmp_app() {
-        dealloc_svect(m_data);
+        dealloc_svect(m_data, app::get_obj_size(m_num_args));
     }
 
     app * get_app() {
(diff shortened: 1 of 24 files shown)


nunoplopes:
Well, it doesn't even run a single example for now.
The assertion is triggered because of dealloc on a base class ptr, but alloc was on a subclass. So we get a smaller size.

I'll check if these can be fixed or if there are just too many of them. An option would be to accept the lower size and compute an approximate memory consumption instead. Might be sufficient for most users(?).

NikolajBjorner:
From what I recall, the mimalloc interface allows to query allocation size. This would allow the code above to be simple(r) and the code below not have to perform the problematic offset adjustment?

PR #6321, opened 2022-09-04
Use glibc's malloc_usable_size when available

nunoplopes:
Glibc's malloc_usable_size() can tell us quickly the size of a malloc'ed ptr.
Implement an optional version of the allocator that uses this function instead of using an extra slot to save this information.


nunoplopes:
This also fixes a memory leak in the parallel version of allocate when the soft out-of-memory is hit.

nunoplopes:
I measured a max 0.5% memory reduction only.

The value of this patch is more on the cleanup side. Plus it fixes a mem leak when the memory limit is hit.

NikolajBjorner:
still resisting mimalloc? https://github.com/microsoft/mimalloc/blob/f2712f4a8f038a7fb4df2790f4c3b7e3ed9e219b/include/mimalloc.h#L358

nunoplopes:
> still resisting mimalloc? https://github.com/microsoft/mimalloc/blob/f2712f4a8f038a7fb4df2790f4c3b7e3ed9e219b/include/mimalloc.h#L358

For the time being yes!
Last time I tried a different alloc library, I didn't see any significant improvement.. But I never tried mimalloc.

NikolajBjorner:
mimalloc is cross platform so not specific to GLIBC. It also has support for querying allocation sizes.
So, it doesn't matter if it is an "improvement" as long as it isn't a drawback. All evidence points to the opposite: it improves over mainstream general-purpose allocators. on top of that z3 tries to be economic on allocations within time-consuming portions (search).

NikolajBjorner:
should we take your pull request? maybe yes. Still the memory layer seems out-dated in light of modern portable layers.

nunoplopes:
I think so. It fixes bugs and paves the way for follow up changes, like plugin in other allocators.

NikolajBjorner:
"paves the way" sounds like a generous characterization.

PR #6327, opened 2022-09-06
DotNet Api: Ensure native Z3 context is freed on disposing dotnet context

### LINKED ISSUE #5043: Memory leak in Z3 Context using .NET API
When multiple instances of `Context` are dynamically created and disposed inside one process memory leak is observed.

**Z3 Version:** 4.8.10 (also reproducible with previous versions)

### **Steps to reproduce (C#):** 

```
// Check allocated memory before
Console.WriteLine(Process.GetCurrentProcess().WorkingSet64);

for (var i = 0; i < 100_000; i++)
{
    var context = new Context();
    context.Dispose();
}

GC.Collect(); // trigger garbage collection (just in case)

// Check allocated memory after
Console.WriteLine(Process.GetCurrentProcess().WorkingSet64);
```

### Observed result

Memory allocated for `Context` instances is not disposed.

Console output from example above:
> 55795712 // ~55Mb
> 12888211456 // ~12Gb

### Expected result

Allocated memory should be disposed.

**Some background.** 
I noticed this issue long ago and as workaround I reused single instance of `Context` across the application. However starting from version 4.8.10 solving some predicates has started leading to non-termination. These issues are not reproducible when fresh instance of `Context` is used. It means that there are side effects from solving previous predicates. Trying to solve this by creating fresh instance of `Context` for each quaey leads to memory leak. 
By the way, what is the recommended way of using `Context` and is it safe to reuse `Context` or in other words are observed side effects in 4.8.10 results of bugs rather than expected behavior?
Thanks in advance.

ptr1120:
fixes #5043

When using the Dotnet API there is still a big amount of memory being lost over time when using the Z3 solver. I am using Z3 in a web application, where a lot of Contexts are created and solved over time. Lastly, the memory increases in a way that the application gets killed.

In my opinion, the cause for the memory loss is the custom reference counting of Dotnet objects that impedes the native context from being freed. 
The mechanism of re-scheduling the GC (`GC.ReRegisterForFinalize`) relies on the assumption that the GC will reliably clean up all Dotnet Z3Objects before the memory consumption kills the process. In my experience, it is unpredictable and too late, at which time the GC will decide to clean up these objects. 
I assume the amount of memory that is managed by the Dotnet GC is very small (most Z3 Dotnet objects contain only a pointer and some references to other objects). Whereas the memory on the native side of Z3 (that is not managed/seen by the Dotnet GC) increases with every native Z3 Context that is not being cleaned up.

I understand that reference counting is needed for the native Z3 objects. Therefore the `DecRefQueue` mechanism reliable maintains the calling of native reference-count methods for all Z3Objects that need to be reference-counted.
What I do not understand is, why there is a second reference counter for the Dotnet Z3Objects. In Dotnet, there is a garbage collector that is reliable and automatically takes care of cleaning up objects that are not referenced anymore.
When `Context.Dispose()` is called, all `DecRefQueue` are cleared -> consequently all objects that are reference-counted on the native side of Z3 got reliable dec-referenced at this point of time (except the native Z3 Context)!
So why wait on the Dotnet GC for cleaning up all these little Dotnet Z3Objects now before calling `Z3_del_context`?

I made several tests with this approach. In order to enforce the memory cleanup (by using memory pressure), I ran these tests inside a docker container where I can easily apply a memory limit (5Gb) and a memory reservation (3Gb) and disallow the container to swap the memory.
Inside the docker container, a process creates around 250 contexts and does some solving, always with an equivalent model that has around 100 mostly boolean variables and 100 constraints.
The process runs in a loop, after each cycle, the context is disposed. The loop runs in parallel, but each loop has its own context.
Without these changes, the container gets killed sometimes after 50 sometimes after 500 cycles because of hitting the 5Gb limit. The cycles after which the container gets killed are very fluctuating, maybe this shows the unpredictability of the Dotnet GC, sometime it cleans something up, sometimes not.
With these changes, the docker container does not get killed anymore. I stop the container regularly after 5000-10000 cycles (because the test takes hours). The memory reaches 3 Gb after 2-3 cycles and then nearly continuously stays around 3Gb - 3.1Gb (around the specified ram reservation for the container).
So I can see at least a significant improvement in the memory-consuming behavior and suggest discussing/integrating these changes.

Sorry for the long explanation in this pull request, due to the difficulty in reproducing/understanding this memory-related problem I felt that it is important for others to understand/follow/question which thoughts led to this change.

Here are some measurements visualized in Grafana, we can see the memory consumption (process working set) over time.
Without the changes the process gets killed after  20min (250  cycles) due to hitting the 5Gb memory limit:
Screenshot 2022-09-06 165918

With the given changes process runs 4 hours (5000 cycles) without hitting the 5Gb memory limit:

Screenshot 2022-09-06 163719

### DIFF
--- a/src/api/dotnet/Context.cs
+++ b/src/api/dotnet/Context.cs
@@ -4778,7 +4778,8 @@ namespace Microsoft.Z3
         #endregion
 
         #region Internal
-        internal readonly IntPtr m_ctx;
+        internal IntPtr m_ctx;
+        private readonly object m_ctxLock = new object();
         internal Native.Z3_error_handler m_n_err_handler = null;
         internal static Object creation_lock = new Object();
         internal IntPtr nCtx { get { return m_ctx; } }
@@ -5012,7 +5013,13 @@ namespace Microsoft.Z3
             m_n_err_handler = null;
             IntPtr ctx = m_ctx;
             if (!is_external)
-                Native.Z3_del_context(ctx);
+            {   
+                lock (m_ctxLock)
+                {
+                    Native.Z3_del_context(ctx);
+                    m_ctx = IntPtr.Zero;
+                }                                                                                                }
+            }
             GC.SuppressFinalize(this);
         }
 
--- a/src/api/dotnet/Context.cs
+++ b/src/api/dotnet/Context.cs
@@ -4778,7 +4778,7 @@ namespace Microsoft.Z3
         #endregion
 
         #region Internal
-        internal IntPtr m_ctx = IntPtr.Zero;
+        internal readonly IntPtr m_ctx;
         internal Native.Z3_error_handler m_n_err_handler = null;
         internal static Object creation_lock = new Object();
         internal IntPtr nCtx { get { return m_ctx; } }
@@ -4966,9 +4966,7 @@ namespace Microsoft.Z3
         /// Optimize DRQ
         /// </summary>
         public IDecRefQueue Optimize_DRQ { get { return m_Fixedpoint_DRQ; } }
-
-        internal long refCount = 0;
-
+        
         /// <summary>
         /// Finalizer.
         /// </summary>
@@ -5011,16 +5009,11 @@ namespace Microsoft.Z3
             m_realSort = null;
             m_stringSort = null;
             m_charSort = null;
-            if (refCount == 0 && m_ctx != IntPtr.Zero)
-            {
-                m_n_err_handler = null;
-                IntPtr ctx = m_ctx;
-                m_ctx = IntPtr.Zero;
-                if (!is_external) 
-                   Native.Z3_del_context(ctx);
-            }
-            else
-                GC.ReRegisterForFinalize(this);
+            m_n_err_handler = null;
+            IntPtr ctx = m_ctx;
+            if (!is_external)
+                Native.Z3_del_context(ctx);
+            GC.SuppressFinalize(this);
         }
 
 
--- a/src/api/dotnet/Context.cs
+++ b/src/api/dotnet/Context.cs
@@ -5013,12 +5013,12 @@ namespace Microsoft.Z3
             m_n_err_handler = null;
             IntPtr ctx = m_ctx;
             if (!is_external)
-            {   
+            {
                 lock (m_ctxLock)
                 {
                     Native.Z3_del_context(ctx);
                     m_ctx = IntPtr.Zero;
-                }                                                                                                }
+                }
             }
             GC.SuppressFinalize(this);
         }
--- a/src/api/dotnet/Z3Object.cs
+++ b/src/api/dotnet/Z3Object.cs
@@ -50,13 +50,6 @@ namespace Microsoft.Z3
                 m_n_obj = IntPtr.Zero;                
             }
 
-            if (m_ctx != null)
-            {
-                if (Interlocked.Decrement(ref m_ctx.refCount) == 0)
-                    GC.ReRegisterForFinalize(m_ctx);
-                m_ctx = null;
-            }
-
             GC.SuppressFinalize(this);
         }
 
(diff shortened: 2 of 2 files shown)


NikolajBjorner:
I am worried this change is not safe. If the GC disposes a z3object after the context is deleted it will crash.
I recently enabled functionality inside of z3 to handle reference queues. This is used for the ocaml bindings and could simplify the .net and java bindings a lot (get rid of all the queues for z3objects).

ptr1120:
I understand your concerns. But following the implementation, `DecRef` does not do what the name lets you think of. `Z3Object.DecRef()` can be called by the finalizer during `Z3Object.Dispose()`, that right. But as far as I know the implemented `DecRef` methods only **add** the object to the `DecRefQueue`. A further call of `DecRefQueue.Clear()` would be needed in order execute the native `dec_ref` method. But this further call to `DecRefQueue.Clear()` or `IncAndClear` is not a valid case, or have I overlooked something. Thus, in my understandig the finalizer will only add objects to the `DecRefQueue`.

On the other hand, getting completely rid of the `DecRefQueue` mechanism additionally to to dotnet reference counting mechanism would be an optimal solution. Is the new mechanism already working and ready to use? Currently I am on a travel for 14 days, but afterwards I could add a commit that removes the `DecRefQueue`s from the Dotnet bindings.

NikolajBjorner:
> But following the implementation, DecRef do ...

OK, I should review this again

> On the other hand, getting completely rid of the DecRefQueue mechanism a ...

```

    /**
       \brief use concurrency control for dec-ref. 
       Reference counting decrements are allowed in separate threads from the context. 
       If this setting is not invoked, reference counting decrements are not going to be thread safe.

       def_API('Z3_enable_concurrent_dec_ref', VOID, (_in(CONTEXT),))
     */
    void Z3_API Z3_enable_concurrent_dec_ref(Z3_context c);

```

We call this in the constructor where Z3_context is created.
Then the GC just calls the dec_ref method on the respective Z3Object or Ast (expression).
But it could invalidate your analysis because the Context is now invalid after a dispose (caveat: I have to re-review the code to follow your analysis).

NikolajBjorner:
OK, granted. It looks to me there is some useless code as you point out.
The native handling of dec_ref should crash, while the external queues are just dangling in the wind after the context dispose and your change should be good.

ptr1120:
I added another commit with my understanding of your your first suggestion.
The second point I did not understand. Did you mean a lock inside `Z3Object.Dispose()` around `DeRef` call (which only adds the object to the `DecRefQueue`) , or in `DecRefQueue.Clear()` which actually execute the native `dec_ref` call and is already locked.

NikolajBjorner:
Another note: The Z3Object objects own unmanaged memory that is heap allocated. For example, the Z3_solver object contains potentially a lot of allocated memory. If we just shut down the managed context and let the Z3_solver objects go uncollected there will be memory leaks, not on the context but other objects. The AST objects are different: the heap allocated for these are owned by Z3_context. The sane approach is perhaps to take a more serious dig at the interface: when setting the flag Z3_enable_concurrent_dec_ref also track in a global pool the other objects that are live. Then when shutting down a context make sure to collect these. The dec_ref functions on Z3_solver and other objects cannot be called after a Z3_context is deleted (and in fact no other operation can be performed on any references the managed code may have).

ptr1120:
Thanks for these important notes. Good to hear that the Z3 Ast objects do not be needed to be referenced counted, if I understood you correctly. These objects are the most difficult objects to keep track of (during creation of constraints). For things like solver it would be also possible to let the consumer dispose the object and therefore free the native solver. In Dotnet it is a common practice to dispose everything that implements `IDisposable`.
I think I will need to squash the commits after travel...
After that I can make a proposal regarding the DecRefQueues.

NikolajBjorner:
I am adding a different pull request

PR #6332, opened 2022-09-08
remove DecRefQueue, use Z3_enable_concurrent_dec_ref

NikolajBjorner:
The notion of reference counted contexts never worked. The reference count to a context only ends up being 0 if the GC kicks in and disposes the various z3 objects. A call to Dispose on Context should free up all resources associated with that context. In exchange none of the resources are allowed any other operation than DecRef. The invocations of DecRef are protected by a lock and test on the context that the native pointer associated with the context is non-zero. Dispose sets the native pointer to zero.

Z3_enable_concurrent_dec_ref ensures that:

- calls to decref are thread safe. Other threads can operate on the context without interference.

The Z3_context ensures that
- z3objects allocated, but not disposed during the lifetime of Z3_context are freed when Z3_context is deleted (it triggers a debug warning, but this is now benign).

### DIFF
--- a/src/api/dotnet/AST.cs
+++ b/src/api/dotnet/AST.cs
@@ -208,37 +208,22 @@ namespace Microsoft.Z3
         internal AST(Context ctx) : base(ctx) { Debug.Assert(ctx != null); }
         internal AST(Context ctx, IntPtr obj) : base(ctx, obj) { Debug.Assert(ctx != null); }
 
-        internal class DecRefQueue : IDecRefQueue
-        {
-            public DecRefQueue() : base() { }
-            public DecRefQueue(uint move_limit) : base(move_limit) { }
-            internal override void IncRef(Context ctx, IntPtr obj)
-            {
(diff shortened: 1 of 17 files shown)


NikolajBjorner:
Unit test:

```
using System;
using Microsoft.Z3;

namespace TestCSharp
{
    internal class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
            for (int i = 0; i < 1000; i++)
            {
                using (var ctx = new Context())
                {
                    Global.SetParameter("verbose", "1");
                    using var s1 = ctx.MkSolver();
                    var s2 = ctx.MkSolver();
                    var a = ctx.MkBoolConst("a");
                    using var b = ctx.MkBoolConst("b");
                    var c = ctx.MkBoolConst("c");
                    Console.WriteLine("done");
                }
            }
        }
    }
}
```

Resources for s2, a, c are purged by the context deletion. Resources for s1 and b are purged by reference counting.

PR #6358, opened 2022-09-22
Add option 'MSVC_STATIC'

gmh5225:


NikolajBjorner:
One wish: add a GitHub action that exercises this option.

CMakeLists.txt
```diff
@@ -304,6 +304,7 @@ endif()
 # Option to control what type of library we build
 ################################################################################
 option(Z3_BUILD_LIBZ3_SHARED "Build libz3 as a shared library if true, otherwise build a static library" ON)
+option(MSVC_STATIC           "Use statically-linked runtime library"                                     OFF)
```
NikolajBjorner, line 307: options that are specific to Z3 should be prefixed by Z3_
other options that work across cmake projects don't
In this way others that use z3 in a bigger cmake project don't suffer clashes.
Can MSVC_STATIC expected to be a general parameter or specific to Z3?

gmh5225:
> One wish: add a GitHub action that exercises this option.

Thank you for your reply.
This is a github action of my own. I will try to merge in later.
https://github.com/gmh5225/Z3-static-build-windows/blob/main/.github/workflows/build64.yml

CMakeLists.txt
gmh5225: I see.
I intended it to be specific to the Z3.

PR #6361, opened 2022-09-22
Dotnet Api: Fix infinite finalization of Context

ptr1120:
* suppress GC finalization of Dotnet context in favor of re-registering finalization
* enable concurrent dec-ref even if context is created without parameters


NikolajBjorner:
thanks for catching these

PR #6381, opened 2022-10-02
switch mpz to using 64-bit integers for the small case

nunoplopes:
This doesn't increase the size of the mpz structure by reusing 2 bits from the pointer.
This will hopefully reduce memory consumption by keeping more cases as "small".

Don't merge yet. Need to test it more thoroughly & benchmark.

After this, we can switch all of mp* to 64-bit integers to get a 2x speedup on this stuff.


nunoplopes:
This is not worth it.. Less than 1% speedup and there's still 1 failing test case..

PR #6393, opened 2022-10-11
JS/TS API Array support

walnutwaldo:
Would love feedback on any changes/practices to incorporate before considering combining with the main Z3 repo.

### DIFF
--- a/src/api/js/package.json
+++ b/src/api/js/package.json
@@ -50,6 +50,7 @@
     "prettier": "^2.5.1",
     "rimraf": "^3.0.2",
     "sprintf-js": "^1.1.2",
+    "ts-expect": "^1.3.0",
     "ts-jest": "^28.0.3",
     "ts-node": "^10.8.0",
     "typedoc": "^0.22.17",
@@ -57,7 +58,6 @@
   },
   "license": "MIT",
   "dependencies": {
-    "async-mutex": "^0.3.2",
-    "ts-expect": "^1.3.0"
+    "async-mutex": "^0.3.2"
   }
 }
--- a/src/api/js/package.json
+++ b/src/api/js/package.json
@@ -58,6 +58,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "async-mutex": "^0.3.2"
+    "async-mutex": "^0.3.2",
+    "ts-expect": "^1.3.0"
   }
 }
--- a/src/api/js/package.json
+++ b/src/api/js/package.json
@@ -57,6 +57,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "async-mutex": "^0.3.2"
+    "async-mutex": "^0.3.2",
+    "ts-expect": "^1.3.0"
   }
 }
--- a/src/api/js/package.json
+++ b/src/api/js/package.json
@@ -35,7 +35,8 @@
   "contributors": [
     "Kevin Gibbons <bakkot@gmail.com>",
     "Nikolaj Bjorner",
-    "Olaf Tomalka <olaf@tomalka.me>"
+    "Olaf Tomalka <olaf@tomalka.me>",
+    "Walden Yan <waldenyan20@gmail.com>"
   ],
   "devDependencies": {
     "@types/jest": "^27.5.1",
--- a/src/api/js/package.json
+++ b/src/api/js/package.json
@@ -58,7 +58,6 @@
   },
   "license": "MIT",
   "dependencies": {
-    "async-mutex": "^0.3.2",
-    "ts-expect": "^1.3.0"
+    "async-mutex": "^0.3.2"
   }
 }
--- a/src/api/js/src/high-level/high-level.test.ts
+++ b/src/api/js/src/high-level/high-level.test.ts
@@ -1,7 +1,7 @@
 import assert from 'assert';
 import asyncToArray from 'iter-tools/methods/async-to-array';
-import {init, killThreads} from '../jest';
-import {Arith, Bool, Model, Z3AssertionError, Z3HighLevel} from './types';
+import { init, killThreads } from '../jest';
+import { Arith, Bool, Model, Z3AssertionError, Z3HighLevel } from './types';
 
 /**
  * Generate all possible solutions from given assumptions.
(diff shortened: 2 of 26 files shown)


walnutwaldo:
Fixes Z3Prover/z3/#6392

src/api/js/src/high-level/types.ts
```diff
@@ -109,98 +136,148 @@ export interface Context<Name extends string = 'main'> {
...
   isBitVecSort(obj: unknown): obj is BitVecSort<number, Name>;
+
   /** @category Functions */
   isBitVec(obj: unknown): obj is BitVec<number, Name>;
+
   /** @category Functions */
   isBitVecVal(obj: unknown): obj is BitVecNum<number, Name>;
+
+  /** @category Functions */
+  isArraySort(obj: unknown): obj is SMTArraySort<any, any, Name>;
```
bakkot, line 228: Is it necessary to use `any` here, or could it be `unknown` or `AnySort<Name>[]` or something? Using `any` tends to make things typecheck which really shouldn't typecheck.

src/api/js/src/high-level/types.ts
```diff
@@ -953,11 +1197,110 @@ export interface BitVecNum<Bits extends number = number, Name extends string = '
...
+  domain(): DomainSort[0];
+
+  /**
+   * The sort of the i-th (0-indexed) dimension of the domain
+   *
+   * TODO: add example
+   *
+   * @param i index of the dimension of the domain being requested
+   */
+  domain_n(i: number): DomainSort[number];
```
bakkot, line 1235: I don't know if TypeScript will let you get away with it, but if it will,

```suggestion
  domain_n<T extends number>(i: T): DomainSort[T];
```

That way when you pass a constant value, like `.domain_n(2)`, type inference can get the actual domain.

src/api/js/src/high-level/high-level.ts
```diff
@@ -677,6 +700,44 @@ export function createApi(Z3: Z3Core): Z3HighLevel {
...
       },
     };
+    const Array = {
+      sort<DomainSort extends AnySort<Name>[], RangeSort extends AnySort<Name>>(
+        ...sig: [...DomainSort, RangeSort]
+      ): SMTArraySort<DomainSort, RangeSort, Name> {
+        const arity = sig.length - 1;
+        const r = sig[arity];
+        const d = sig[0];
+        if (arity === 1) {
```
bakkot, line 710: Probably would help to have a useful error in the case that this is invoked with only a single argument (i.e. just a range, no domain). The type signature doesn't currently prevent that, I don't think.

bakkot (commented):
Unfortunately I haven't used arrays in Z3 and so can't comment on whether this is a good wrapper for them, but nothing jumps out at me as obviously wrong. Had a few minor comments.

src/api/js/src/high-level/high-level.ts
walnutwaldo: Good suggestion! I instead opted to now update the type signature to not allow it.

src/api/js/src/high-level/types.ts
walnutwaldo: Nice!

src/api/js/src/high-level/types.ts
walnutwaldo: Nope, I changed the `any`s to be more specific now

src/api/js/package.json
```diff
@@ -57,6 +57,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "async-mutex": "^0.3.2"
+    "async-mutex": "^0.3.2",
+    "ts-expect": "^1.3.0"
```
bakkot, line 61: This should be in `devDependencies`.

walnutwaldo: good catch

src/api/js/src/high-level/high-level.test.ts
```diff
@@ -390,6 +391,83 @@ describe('high-level', () => {
     });
   });
 
+  describe('arrays', () => {
```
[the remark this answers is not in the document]
walnutwaldo: Sure, opened a PR on z3guide at https://github.com/microsoft/z3guide/pull/142 for when this is merged in.

src/api/js/src/high-level/high-level.ts
```diff
@@ -216,20 +219,22 @@ export function createApi(Z3: Z3Core): Z3HighLevel {
...
       if (kind === Z3_ast_kind.Z3_QUANTIFIER_AST) {
         if (Z3.is_quantifier_forall(contextPtr, ast))
-           return new BoolImpl(ast);
+          return new BoolImpl(ast);
         if (Z3.is_quantifier_exists(contextPtr, ast))
-           return new BoolImpl(ast);
+          return new BoolImpl(ast);
         if (Z3.is_lambda(contextPtr, ast))
-           return new ExprImpl(ast);
+          return new ExprImpl(ast);
```
NikolajBjorner, line 237: The type of a lambda is an array in z3. Note that isArray returns true only for ArrayImpl

walnutwaldo: This is fixed in some other changes that add Quantifier bindings in general. I'll make a new PR for those changes once these are merged in.

walnutwaldo:
> One comment added. I could merge if you think it is ready.

Should be good to merge in. Will keep Lambdas as `ExprImpl` for now.

NikolajBjorner:
FWIW, we are currently chasing down some regression in z3guide.
https://github.com/microsoft/z3guide/issues/147
It works with 4.11.2 but not 4.12.0. The only change to the JS API is from this commit.
Yet, I don't see how the error message (obscure) is related to the changes.

walnutwaldo:
> FWIW, we are currently chasing down some regression in z3guide.
> 
> https://github.com/microsoft/z3guide/issues/147
> 
> It works with 4.11.2 but not 4.12.0. The only change to the JS API is from this commit.
> 
> Yet, I don't see how the error message (obscure) is related to the changes.

I can try to help investigate later if it isn't resolved by then.

Issue #6408, opened 2022-10-18
`Z3_eval_smtlib2_string` overhead: can I do anything about it ?


qaristote:
[Here](https://gist.github.com/qaristote/1dc2ee7033c44a2dcd14713b47f80c54)'s an SMTLib2 file I want to evaluate with Z3 (I'm using version `4.8.15`). Using [`hyperfine`](https://github.com/sharkdp/hyperfine) (with `--runs 250` and `--warmup 5`) I benchmarked different ways of doing so:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `z3 $input >/dev/null` | 181.8 ± 5.9 | 171.7 | 208.9 | 1.00 |
| `./bindings-oneshot.c.bin $input >/dev/null` | 189.2 ± 19.8 | 183.7 | 499.5 | 1.04 ± 0.11 |
| `./bindings.c.bin $input >/dev/null` | 253.3 ± 5.9 | 248.3 | 317.9 | 1.39 ± 0.06 |

[`bindings.c.bin`](https://gist.github.com/qaristote/6910cea4473dd7bc38b0eecd4cd4ae56) uses the `Z3_eval_smtlib2_string` function from the API to evaluate the input file line by line, while [`bindings-oneshot.c.bin`](https://gist.github.com/qaristote/2bb18ab65a3d7974ca9b19fd01e4f41a) does the same thing except the whole file is passed to Z3 in one go.

As you can see, there is a significant overhead in repeatedly running `Z3_eval_smtlib2_parse` on each line instead of running it once on the whole file. Is there a reason for that ? Is there anything I can do to have the line-by-line version be faster ?

NikolajBjorner:
Mainly use the I/O pipe for parsing SMT commands.

The API function Z3_eval_smtlib2_string is not tuned for line-by-line: it re-initializes a parser every time it is invoked.
This is going to be significant overhead (or you can run a profiler to get source information). 
Tuning this code would require exposing a chunk that recycles the parser between calls.


Background.
This line is invoked from the API function:

https://github.com/Z3Prover/z3/blob/1fc77c8c004521867f1a78368f3c956c3ecfdbb8/src/parsers/smt2/smt2parser.cpp#L3209

It initializes the parser class.

[...]

Issue #6049, opened 2022-05-24
Support parametric datatypes in the API


facundominguez:
I'm looking into parametric datatypes because Liquid Haskell uses them via SMTLIB2, and using z3 via the API might offer some speed up.

z3 supports parametric datatypes via SMTLIB2, yet I can't find in the API any way to use them.

There seems to be [an implementation](https://github.com/Z3Prover/z3/blob/5a9b0dd747a2fc26513fcb15181678781e6667a7/src/cmd_context/pdecl.h), but I couldn't find any examples of using it, and for all I know it might not be intended for public consumption.

Are parametric datatypes supported in the API? If so, are there any examples or documentation? If they aren't, would it be a difficult project to have them supported?

PR #6422, opened 2022-10-24
Optimize calls to Z3_eval_smtlib2_string

facundominguez:
Loosely following the advice in #6408, this PR optimizes `Z3_eval_smtlib2_string`.

This change makes possible to have a faster SMTLIB2 interface to talk to Z3. Alternatives considered:
* Start z3 in an external process and communicate with it using IO pipes. The disadvantage of this is that it introduces IO latency and context switches between z3 and the calling application.
* Use the z3 bindings to build the queries instead of using SMTLIB2. The disadvantage of this is that it couples the application a lot with Z3, making difficult to switch to another SMT solver. It is not as simple to debug as SMTLIB2. And it also would make unavailable SMTLIB2-exclusive features, not readily available in the bindings (I know only of #6049 though).

With this PR, IO pipes is almost twice slower than `Z3_eval_smtlib2_string`.

I had to innovate in two aspects to get most of the speed up:
1. Instead of creating a new scanner for every call, this PR reuses the same scanner by resetting the input stream.
2. In addition to reusing the scanner, it also reuses the `stringstream` and the `istringstream` used to communicate with the parser.


facundominguez:
Feedback on the implementation is welcome. I think there might be a few opportunities to organize it better, but it is difficult to decide from an outsider perspective.

src/parsers/smt2/smt2parser.cpp
```diff
@@ -3204,11 +3208,16 @@ namespace smt2 {
...
+    void free_parser(parser * p) { delete p; }
 };
 
 bool parse_smt2_commands(cmd_context & ctx, std::istream & is, bool interactive, params_ref const & ps, char const * filename) {
-    smt2::parser p(ctx, is, interactive, ps, filename);
-    return p();
+    if (ctx.parser())
+        ctx.parser()->reset_input(is, interactive);
+    else
+        ctx.parser(new smt2::parser(ctx, is, interactive, ps, filename));
```
NikolajBjorner, line 3219: use alloc(smt2::parser, ctx, is, interactive, ps, filename)

src/parsers/smt2/smt2parser.cpp
```diff
@@ -3204,11 +3208,16 @@ namespace smt2 {
             }
         }
     };
+
+    void free_parser(parser * p) { delete p; }
```
NikolajBjorner, line 3212: use dealloc(p)

src/api/api_parsers.cpp
```diff
@@ -244,22 +244,24 @@ extern "C" {
         }
         scoped_ptr<cmd_context>& ctx = mk_c(c)->cmd();
         std::string s(str);
-        std::istringstream is(s);
-        ctx->set_regular_stream(ous);
-        ctx->set_diagnostic_stream(ous);
+        ctx->reset_parser_streams(s);
+
+        ous = ctx->parser_ous();
```
NikolajBjorner, line 249: not sure why ownership of ous is organized this way now.

src/api/api_parsers.cpp
```diff
@@ -244,22 +244,24 @@ extern "C" {
...
-            SET_ERROR_CODE(Z3_PARSER_ERROR, ous.str());
-            RETURN_Z3(mk_c(c)->mk_external_string(ous.str()));
+            if (ous->str().empty()) *ous << e.msg();
+            SET_ERROR_CODE(Z3_PARSER_ERROR, ous->str());
+            RETURN_Z3(mk_c(c)->mk_external_string(ous->str()));
         }
-        RETURN_Z3(mk_c(c)->mk_external_string(ous.str()));
-        Z3_CATCH_RETURN(mk_c(c)->mk_external_string(ous.str()));
+        RETURN_Z3(mk_c(c)->mk_external_string(ous->str()));
+        Z3_CATCH_RETURN(mk_c(c)->mk_external_string(ous?ous->str():""));
```
NikolajBjorner, line 265: when is ous nullptr?

src/cmd_context/cmd_context.h
```diff
@@ -295,6 +299,10 @@ class cmd_context : public progress_callback, public tactic_manager, public ast_
 
     stopwatch                    m_watch;
 
+    smt2::parser*                m_parser;
```
NikolajBjorner, line 302: seems this would be a property of the API (api_context) not cmd_context

src/cmd_context/cmd_context.h
```diff
@@ -554,6 +562,11 @@ class cmd_context : public progress_callback, public tactic_manager, public ast_
 
     void slow_progress_sample() override;
     void fast_progress_sample() override;
+    smt2::parser* parser() { return m_parser; };
```
NikolajBjorner, line 565: again, seems api_context is the place to own parsing thunks, cmd_context is also for the shell and does not reset the parser.

src/api/api_parsers.cpp
facundominguez: `ous` needs to be declared before `Z3_TRY` in order to be in scope in `Z3_CATCH_RETURN`. Doing this, however, forces the value to start initialized to `nullptr`, before it can be properly initialized after obtaining the `ctx` reference.

I'm guessing that if an exception is thrown before `ctx` is initialized, we can reach the catch block before `ous` has been initialized.

src/api/api_parsers.cpp
facundominguez: The creation of the `stringstream` seemed to be expensive in my benchmarks, which motivated me to put it in the context to reuse it across calls. But beware that C++ is not my element. I could double check this if it feels amiss.

src/cmd_context/cmd_context.h
facundominguez: Thanks for the suggestion. How do I get a reference to the `api_context` inside `Z3_eval_smtlib2_string`?

NikolajBjorner: Z3_context is the external pointer name for api_context. The function mk_c(c) coerces it to api_context.
It contains an optional cmd_context pointer (accessed using "cmd()"). It can contain other optional pointers, such as a reusable parser closure.

src/api/api_parsers.cpp
NikolajBjorner: Ah, c++ destructors are not invoked from C exceptions. This is indeed a tricky part.

src/parsers/smt2/smt2parser.cpp
facundominguez: Done.

src/parsers/smt2/smt2parser.cpp
facundominguez: Done.

src/cmd_context/cmd_context.h
facundominguez: Done.

src/cmd_context/cmd_context.h
facundominguez: Moved to api_context.

src/api/api_parsers.cpp
facundominguez: This line was reverted as a side effect of moving `ous` to the `api::context`.

src/parsers/smt2/smt2parser.cpp
```diff
@@ -3204,13 +3208,23 @@ namespace smt2 {
...
+
+    void free_parser(parser * p) { dealloc(p); }
 };
 
 bool parse_smt2_commands(cmd_context & ctx, std::istream & is, bool interactive, params_ref const & ps, char const * filename) {
     smt2::parser p(ctx, is, interactive, ps, filename);
     return p();
 }
 
+bool parse_smt2_commands_with_parser(class smt2::parser ** p, cmd_context & ctx, std::istream & is, bool interactive, params_ref const & ps, char const * filename) {
```
NikolajBjorner, line 3220: smt2::parser *& ?

NikolajBjorner:
I take this works and I can merge it after the build is confirmed?

src/parsers/smt2/smt2parser.cpp
facundominguez: I'll adopt that style. Without knowing better, I like using the '&' at the call site to signal that the parameter can be mutated (i.e. `parse_smt2_commands_with_parser(&p, ...)`).

facundominguez:
Yes, it does work for me. I might delay a bit in acknowledging the CLA because I need to get the word from my employer ([tweag.io](https://www.tweag.io/)).

If you merge early, I'll submit a second PR with tests.

src/parsers/smt2/smt2parser.cpp
facundominguez: Done.

facundominguez:
@microsoft-github-policy-service agree company="Tweag"

facundominguez:
Alright, unless there are further request, I got this PR as complete as I could.

nunoplopes:
What's the speedup you obtain? What the speedup of reusing the parser? What's the speedup of reusing the stream?

qaristote:
> What's the speedup you obtain? What the speedup of reusing the parser? What's the speedup of reusing the stream?

Following the same protocol as in #6408, using Z3 4.8.15 yields

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `z3 inputs/isunity_bug.smt2 >/dev/null` | 180.9 ± 5.0 | 172.4 | 199.8 | 1.00 |
| `./bindings-oneshot.c.bin inputs/isunity_bug.smt2 >/dev/null` | 192.7 ± 3.0 | 188.3 | 210.0 | 1.07 ± 0.03 |
| `./bindings.c.bin inputs/isunity_bug.smt2 >/dev/null` | 259.4 ± 5.9 | 252.6 | 300.7 | 1.43 ± 0.05 |

whereas using Z3 from c181b7b8d691d3646caad18c5749484979e81127 yields

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `z3 inputs/isunity_bug.smt2 >/dev/null` | 189.2 ± 7.4 | 179.1 | 212.7 | 1.00 |
| `./bindings-oneshot.c.bin inputs/isunity_bug.smt2 >/dev/null` | 191.9 ± 3.0 | 187.6 | 210.8 | 1.01 ± 0.04 |
| `./bindings.c.bin inputs/isunity_bug.smt2 >/dev/null` | 199.6 ± 2.3 | 196.1 | 208.3 | 1.06 ± 0.04 |

Notice how the overhead of `./bindings.c.bin` has basically disappeared: the difference with `./bindings-oneshot.c.bin` (~10ms) is what you can expect from the difference in the way the input file is read.

nunoplopes:
Why does Z3 get slower?
Also, are you sure you're using Z3 in the right way??? Doesn't make sense that you need to call this function thousands of times. What the heck are you doing?

qaristote:
I'm running these benchmarks in a noisy environment so I'd say the differences between the running-times of Z3 itself aren't relevant, especially as their deviations overlap. What's more relevant is the ratios between the running-times of the different ways to use Z3.

As to why we're calling this function thousands of times, we're basically using the SMTLib2 interface which we do line-by-line (and I'd say this is the simpler thus usual way to do so), except that here we're using the API instead of an external process because it's faster this way.

nunoplopes:
> As to why we're calling this function thousands of times, we're basically using the SMTLib2 interface which we do line-by-line (and I'd say this is the simpler thus usual way to do so), except that here we're using the API instead of an external process because it's faster this way.

But why can't you use the API to create the expressions by hand? Why is the input a string?

facundominguez:
> Why does Z3 get slower?
> Also, are you sure you're using Z3 in the right way??? Doesn't make sense that you need to call this function thousands of times. What the heck are you doing?

Thanks @nunoplopes for your interest. And thanks @qaristote for chiming in!

> What the heck are you doing?

We have two use cases. One is a symbolic evaluation engine that interacts with SMT solvers, and symbolic evaluation performance is notably known to be dominated by interactions with the SMT solver. The other use case is Liquid Haskell. None of these projects are eager to integrate the binding APIs of the multiple SMT solvers out there. For the modest effort of hooking the Z3 library through such a simple function, the speed up is very attractive when compared to the alternative of using IO pipes.

We'd be happy to compare benchmark results if you have any of your own.

nunoplopes:
So you're missing out on like a 2x because you don't want to use the API, but you care about a 20% speedup?

Anyway, could you please benchmark your patch without the stream part? I don't believe it's anything any speedup besides increasing memory consumption for everybody else.
Plus the out stream is just for errors, so it can't make a difference! Doesn't make sense to keep it around.

facundominguez:
> So you're missing out on like a 2x because you don't want to use the API, but you care about a 20% speedup?

If your claim is that constructing the ASTs is 2x faster than invoking `Z3_eval_smtlib2_string`, we haven't analyzed that thouroughly. But I did one experiment measuring `Z3_parse_smtlib2_string` and the overhead of parsing was marginal.

> Anyway, could you please benchmark your patch without the stream part?

We are double checking this. We might undo that bit indeed.

Thanks for your thoughts.

src/api/api_parsers.cpp
facundominguez: I've moved the streams out of the context in 12c81d7. Whatever overhead it adds, I cannot distinguish it in my latest measures.

src/parsers/smt2/smt2scanner.cpp
```diff
@@ -390,5 +390,13 @@ namespace smt2 {
         return m_cache_result.begin();
     }
 
+    void scanner::reset_input(std::istream & stream, bool interactive) {
+		m_stream = &stream;
```
nunoplopes, line 394: indentation is off here

facundominguez:
```suggestion
      m_stream = &stream;
```

src/cmd_context/cmd_context.h
```diff
@@ -41,7 +41,6 @@ Module Name:
 #include "cmd_context/tactic_manager.h"
 #include "params/context_params.h"
 
-
```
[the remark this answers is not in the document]
facundominguez: Added back.

NikolajBjorner:
> It's written in Haskell so a small overhead may be expected but it wouldn't be big enough to explain the difference we're seeing.

I don't understand what the current conclusion should be.

- Did the update improve performance? A table contains timing spanning 170min to 358min. I presume one of the lower end timings exhibit the new code and includes the parser itself.

- I haven't grokked the claim by Nuno that the change incurs a penalty for others. The parser pointer should be allocated on demand and just persist in the case it is used in the same session. It is of course liability to add features that down the road bite as complexity.

Nikolaj

nunoplopes:
They've removed the streams parts, that was not adding any speedup. The overhead now is the extra heap allocation + 8 bytes in the api context, but I guess that's ok.

qaristote:
>     * Did the update improve performance? A table contains timing spanning 170min to 358min. I presume one of the lower end timings exhibit the new code and includes the parser itself.

The tables I posted compares the running-times of different ways of using Z3 with SMTLib2. `./bindings.c.bin` uses `Z3_eval_smtlib2_string`, command by command. Its overhead can be deduced by comparing it with `./bindings-oneshot.c.bin`, which only calls the binding once on the whole input file. 
The ratio between these two programs' running-times used to be ~1.4 before this PR. After https://github.com/Z3Prover/z3/commit/c181b7b8d691d3646caad18c5749484979e81127, it went down to ~1.0, meaning the overhead had disappeared in practice. After https://github.com/Z3Prover/z3/commit/12c81d71d2a75a4c6a2bde2c55a9737de8cd23d7, which removed the streams part, this ratio went back up to ~1.1, so it seems a small overhead was added back.
The last table also contains a benchmark for `shellcmd-bytestrings.hs.bin`, which uses Z3 as an external process. I only added it to show that it was way slower to use Z3 as an external process instead of through the `Z3_eval_smtlib2_string` binding (even line by line), and thus to motivate this binding's use.

facundominguez:
> Did the update improve performance? ... I presume one of the lower end timings exhibit the new code and includes the parser itself.

Yes, it did improve performance, and your presumption is correct. Quentin is giving the details in the answer above.

I think Nuno answered well your second point, but happy to discuss it more if needed.

Quentin and I have a bit of controversy on whether there is a noticeable performance improvement when reusing streams in addition to reusing the parser and the scanner. Therefore, I'm deferring that discussion for a later PR.

PR #6435, opened 2022-11-02
Added limit to "visit" to allow detecting multiple visits

CEisenhofer:


src/util/visit_helper.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+#include "sat_literal.h"
```
NikolajBjorner, line 2: not a fan of adding things to util directory that depend on sat::literal.
This is done in a different instance, but would like to avoid it.
Can move to sat_types.h

src/sat/sat_solver.h
```diff
@@ -176,8 +177,7 @@ namespace sat {
         std::string             m_reason_unknown;
         bool                    m_trim = false;
 
-        svector<unsigned>       m_visited;
-        unsigned                m_visited_ts;
+        visit_helper            m_visited;
```
NikolajBjorner, line 180: instead of exposing this having clienets have to use s().m_visited all over the place have the solver inherit from visit_helper.

src/util/visit_helper.h
```diff
@@ -0,0 +1,39 @@
...
+    
+public:
+    
+    void init_visited(unsigned num_vars, unsigned lim = 1) {
+        init_ts(2 * num_vars, lim);
+    }
+    void mark_visited(sat::literal l) { m_visited[l.index()] = m_visited_begin + 1; }
+    void mark_visited(sat::bool_var v) { mark_visited(sat::literal(v, false)); }
+    void inc_visited(sat::literal l) {
+        m_visited[l.index()] = std::min(m_visited_end, std::max(m_visited_begin, m_visited[l.index()]) + 1);
```
NikolajBjorner, line 33: m_visited_begin + 1

src/sat/sat_solver.h
CEisenhofer: The reason I want to put this in a class is that I will need two instances of it. (I could also inherit from it in the xor-solver but then I have the problem again that I limit this datastructure to a single solver.)

NikolajBjorner:
Instead of round-tripping on this I merged and updated the code to make the visited_helper independent of literals and have the sat::solver layer handle the literal to index conversion. Note how the interface to counting marked already broke with the bool_var/literal interface. So visited_helper was not internally consistent.

NikolajBjorner:
https://github.com/Z3Prover/z3/commit/90490cb22f2600dd45826ba7070c97e091077298

Issue #6466, opened 2022-11-28
Status of Sequences and `mapi` in API


avery-laird:
Hello,

I am wondering about the status of sequences in the APIs (either python or C++). Most of the functionality documented [here](https://microsoft.github.io/z3guide/docs/theories/Sequences/) is accessible from the python API, except for functions like `seq.mapi`. Specifically, I want to use this to model the behaviour of `Store` on arrays, but using a structure (list/sequence) that is more amenable to inductive reasoning. 

Are functions like `seq.mapi` accessable from the python or c++ API? Thanks for any help.

NikolajBjorner:
not yet accessible, but if you are in a hurry, you can add a pull request with updates to the API to expose them.
These functions were added relatively recently and were so far tested with Bosque.

avery-laird:
sounds good, I have done so in #6468 .

PR #6468, opened 2022-11-28
Adding mapi to the python API.

avery-laird:
Per issue #6466.


avery-laird:

PR #6497, opened 2022-12-17
Tiny fix to qprofdiff

nikswamy:
I use qprofdiff for differential profiling for F* SMT qi profiles.
But, the most recent version fails to parse the qi profile entries correctly because of a small bug.
It is searching for a delimiter token " : " of length 3, but when it finds it, it only advances the position by 1. I've fixed that.

@wintersteiger does it makes sense? Thanks!


0946505082:
Generated 3 verification condition(s). Starting proof:
[1/3] : Precondition faculty   : OK
[2/3] : While Case True (l:8)  : OK
[3/3] : While Case False (l:8) : OK

Summary: VERIFICATION OK

PR #6526, opened 2023-01-08
Add rewrite for array selects of chain of stores of a same value

nunoplopes:
Example:
```smt
(declare-fun mem () (Array (_ BitVec 4) (_ BitVec 4)))
(declare-const x (_ BitVec 4))
(declare-const y (_ BitVec 4))
; simplifies to #x1
(simplify (select (store (store (store mem #x1 #x1) y #x1) x #x1) #x1))
```

### DIFF
--- a/src/ast/rewriter/array_rewriter.cpp
+++ b/src/ast/rewriter/array_rewriter.cpp
@@ -211,6 +211,31 @@ br_status array_rewriter::mk_select_core(unsigned num_args, expr * const * args,
     goto exit
 
     while (true) {
+        if (m_util.is_store(arg0)) {
+            SASSERT(to_app(arg0)->get_num_args() == num_args+1);
+            switch (compare_args(num_args - 1, args+1, to_app(arg0)->get_args()+1)) {
+            case l_true:
+                // select(store(a, I, v), I) --> v
+                RET(to_app(arg0)->get_arg(num_args), BR_DONE);
+
+            case l_false:
+                // select(store(a, I, v), J) --> select(a, J) if I != J
+                arg0 = to_app(arg0)->get_arg(0);
+                continue;
+
+            case l_undef:
+                // check if loading from subsequent arrays yields the same value
+                if (first) {
+                    result = to_app(arg0)->get_arg(num_args);
+                    first  = false;
+                } else if (result != to_app(arg0)->get_arg(num_args)) {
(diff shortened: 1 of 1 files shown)


nunoplopes:
Traverses a chain of stores until it finds a store whose idx is equal to the select. Rewrite works if all stores until that guaranteed store have same value.
Quicks in Alive2 benchmarks.

src/ast/rewriter/array_rewriter.cpp
```diff
@@ -198,13 +198,97 @@ bool array_rewriter::squash_store(unsigned n, expr* const* args, expr_ref& resul
...
+            switch (compare_args(num_args - 1, args+1, to_app(arg0)->get_args()+1)) {
+            case l_true:
+                // select(store(a, I, v), I) --> v
+                RET(to_app(arg0)->get_arg(num_args), BR_DONE);
+
+            case l_false:
+                // select(store(a, I, v), J) --> select(a, J) if I != J
+                do {
+                    arg0 = to_app(arg0)->get_arg(0);
+                } while (m_util.is_store(arg0) && compare_args(num_args-1, args + 1, to_app(arg0)->get_args() + 1) == l_false);
```
NikolajBjorner, line 267: inner loop optimization is similar to what is going on in outer loop.
Move m_util.is_store(arg0) up to beginning for "every instruction is holy" optimization.

src/ast/rewriter/array_rewriter.cpp
```diff
@@ -198,13 +198,97 @@ bool array_rewriter::squash_store(unsigned n, expr* const* args, expr_ref& resul
 
 br_status array_rewriter::mk_select_core(unsigned num_args, expr * const * args, expr_ref & result) {
     SASSERT(num_args >= 2);
+    expr *arg0 = args[0];
+    expr_ref tmp(m());
+    bool first = true;
+
+#define RET(x, status)            \
```
NikolajBjorner, line 205: why not a separate function that returns a status? BR_FAILED results in trying the next rewrite.
The local variable arg0 is updated in the first part of the code and should not be used in the second part because it could be unrelated to args

nunoplopes: Sorry, I didn't understand the suggestion.
arg0 is only used in this new loop; it's not used in the old code.

PR #6576, opened 2023-02-10
Added two new context::recfun overloads that use sort_vector instead of sort* and size. 

jparsert:
Previously recfun required a pointer and arity (size) to a list that contains the domain sorts of a recrusive function declaration in the c++ api.
In this pull request I added two overloads that allow the use of context::sort_vector instead.


jparsert:

PR #6596, opened 2023-02-17
Contract bit information to large unit-intervals

CEisenhofer:


src/math/polysat/op_constraint.cpp
```diff
@@ -275,7 +275,7 @@ namespace polysat {
         }
         else {
             // forward propagation
-            SASSERT(!(pv.is_val() && qv.is_val() && rv.is_val()));
```
NikolajBjorner, line 278: disabling this code is independent of the fi functionality?

src/math/polysat/forbidden_intervals.h
```diff
@@ -23,7 +23,7 @@ namespace polysat {
     struct fi_record {
         eval_interval               interval;
         vector<signed_constraint>   side_cond;
-        signed_constraint           src;
+        vector<signed_constraint>   src; // only units may have multiple src (as they can consist of contracted bit constraints)
```
NikolajBjorner, line 26: seems safe to update fi_record with src being a vector. It can be done in a self-contained push with no big risk.

src/math/polysat/viable.cpp
```diff
@@ -684,7 +720,7 @@ namespace {
             rational const b = mod(r * val + s_, mod_value);
             rational const np = mod_value - p;
             rational const nr = mod_value - r;
-            int const corr = e->src.is_negative() ? 1 : 0;
+            int const corr = e->src[0].is_negative() ? 1 : 0;
```
NikolajBjorner, line 723: assert e->src.size() == 1?

src/math/polysat/viable.cpp
```diff
@@ -742,72 +778,183 @@ namespace {
         while (e != first);
         return true;
     }
-    
-    bool viable::quick_bit_check(pvar v) {
-#if 0
-        return true;
-#endif
+
```
NikolajBjorner, line 781: must admit I am not tuned into what this should do. Possibly the trickier part of the pull request.

src/math/polysat/viable.cpp
```diff
@@ -742,72 +778,183 @@ namespace {
...
         auto add_entry = [&builder](entry* e) {
-            for (const auto& sc : e->side_cond) {
+            for (const auto& sc : e->side_cond)
                 builder.insert_eval(~sc);
-                LOG("Side cond: " << sc);
-            }
-            builder.insert_eval(~e->src);
-            LOG("Adding to core: " << e->src);
+            SASSERT(e->src.size() == 1);
```
NikolajBjorner, line 902: The assert holds because?
Generally tricky with these deep assumptions

src/math/polysat/op_constraint.cpp
CEisenhofer: Jep - In our benchmarks it performs poorly (the justification is weak)

src/math/polysat/viable.cpp
CEisenhofer: Right

src/math/polysat/viable.cpp
CEisenhofer: I dropped it. Left over from a previous state

src/math/polysat/viable.cpp
CEisenhofer: We look for the smallest x with x >= bound such that it matches the bit-pattern

PR #6612, opened 2023-02-28
Use sys.getdefaultencoding() instead of sys.stdout.encoding

twizmwazin:
On windows when using `pythonw.exe` or pyinstaller >= 5.7, `sys.stdout` is set to `None`. Instead of using stdout to infer the default encoding, use `sys.getdefaultencoding()`.


twizmwazin:
@microsoft-github-policy-service agree company="Arizona Board of Regents"

twizmwazin:
@microsoft-github-policy-service agree company="Arizona Board of Regents"

PR #6626, opened 2023-03-07
Make all methods show in java API

BramVerb:
Some methods did not show up in the java documentation, for example the `mkUnaryMinus` and `mkDiv`, making the methods final made them show up for me in the documentation. 

I do not know if this was intentional, but I do think that these methods should show up in the documentation, as they are also available in the Java API

### DIFF
--- a/src/api/java/Context.java
+++ b/src/api/java/Context.java
@@ -754,7 +754,7 @@ public class Context implements AutoCloseable {
     /**
      * Create an expression representing {@code not(a)}.
      **/
-    public BoolExpr mkNot(Expr<BoolSort> a)
+    public final BoolExpr mkNot(Expr<BoolSort> a)
     {
         checkContextMatch(a);
         return new BoolExpr(this, Native.mkNot(nCtx(), a.getNativeObject()));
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
https://z3prover.github.io/api/html/classcom_1_1microsoft_1_1z3_1_1_context.html#a4313e0284a1f0f5c31847fd7c3d14a0d

NikolajBjorner:
mkPower doesn't show up.
Perhaps all the ones with public <... are not understood by doxygen.

BramVerb:
There are way more methods that do not show up, and they all seem to include the generics syntax with `<>`.

NikolajBjorner:
so it makes better sense to update them all?

NikolajBjorner:
Bing chat gives a summary:
[According to some web sources](https://stackoverflow.com/questions/22359457/doxygen-losing-non-static-java-generic-methods)[1](https://stackoverflow.com/questions/22359457/doxygen-losing-non-static-java-generic-methods)[2](https://stackoverflow.com/questions/22359457/doxygen-losing-non-static-java-generic-methods), Doxygen may have some issues with documenting Java generic methods. [One possible workaround is to use static modifiers for your generic methods](https://stackoverflow.com/questions/22359457/doxygen-losing-non-static-java-generic-methods)[1](https://stackoverflow.com/questions/22359457/doxygen-losing-non-static-java-generic-methods), as this seems to make Doxygen accept arbitrary strings as return type modifiers. [Another possible workaround is to use a different version of Doxygen that may have fixed this bug](https://stackoverflow.com/questions/19365041/doxygen-sees-java-generic-method-as-package-protected)[3](https://stackoverflow.com/questions/19365041/doxygen-sees-java-generic-method-as-package-protected).

I hope this helps. Do you want me to show you how to use static modifiers for your generic methods or how to install a different version of Doxygen?

BramVerb:

BramVerb:
I did a search and replace for all generic methods, so it works now for the Context.java class.
I don't think this is the nicest solution, but I hope this [PR](https://github.com/doxygen/doxygen/pull/8534) for doxygen will be merged at some point to resolve the issue.

PR #6627, opened 2023-03-08
use uintptr_t instead of size_t (tptr) for portability

igcontreras:
size_t is not guaranteed to be able to contain the largest pointer in some architectures


igcontreras:

PR #6656, opened 2023-03-29
update ocaml binding to support more string apis

Hhro:


Hhro:

PR #6673, opened 2023-04-08
Some UP bugfixes in the new core

CEisenhofer:


src/sat/smt/euf_solver.cpp
```diff
@@ -349,6 +349,20 @@ namespace euf {
         si.uncache(literal(v, true));
     }
 
+    bool solver::decide(bool_var& var, lbool& phase) {
```
NikolajBjorner, line 352: This is only going to be enabled if there is a user-propagator. Otherwise, it is never used.

PR #6678, opened 2023-04-10
Bugfix relevancy propagation + UP (old core)

CEisenhofer:


src/smt/smt_context.cpp
```diff
@@ -2330,8 +2330,10 @@ namespace smt {
 
                     if (get_assignment(l1) == l_false)
                         set_conflict(b_justification(cls));
-                    else if (get_assignment(l2) == l_false)
+                    else if (get_assignment(l2) == l_false) {
+                        mark_as_relevant(l1);
```
NikolajBjorner, line 2334: do you have a test that observes the effect of the change in this line?
Would be good to clarify this. It is more disruptive than changing UP

CEisenhofer: The examples I have are probably all fixed by adding relevancy marking in UP's propagate => Not needed

Issue #6709, opened 2023-05-05
javascript tests fail with the error "TypeError: Mod._malloc is not a function"


gbagan:
Hi,

I have tried to pass javascript/wasm API tests.
So I have installed `Emscripten` and run `npm run build:wasm` and `npm run build:ts`.
Both these commands succeed but, when I ran `npm run test`, almost all test fails with the error
`TypeError: Mod._malloc is not a function`

Finally I have succeed to pass the tests by adding `_malloc` on file `scripts/build-wasm.ts` line 43:
`const extras = ['_malloc', '_set_throwy_error_handler', '_set_noop_error_handler', ...asyncFuncs.map(f => '_async_' + f)];`

I am not sure it is the best solution but at least it works.

I can submit a PR if necessary.

PR #6712, opened 2023-05-06
JS/TS:  add Optimize class

gbagan:
- add  an Optimize class in the Typescript high level API
- add tests for this class
- add _malloc to exported functions (fix https://github.com/Z3Prover/z3/issues/6709)


gbagan:

PR #6755, opened 2023-06-08
fix sort_non_basis

levnach:


baylesj:
This is the fix I had in mind. Thanks for the quick turnaround!

PR #6756, opened 2023-06-09
get cached tv value

levnach:


NikolajBjorner:
ooh, I hope it does the trick :)
I guess it requires that never ever pass a term without any coefficients to lar_solver.

PR #6771, opened 2023-06-19
use heap to track infeasible columns.

levnach:


src/math/lp/lp_core_solver_base.h
```diff
@@ -18,592 +18,595 @@ Revision History:
...
-    }
-
-    
-    unsigned get_base_column_in_row(unsigned row_index) const {
-        return m_basis[row_index];
-    }
+    return true;
+  }
+
+  bool column_has_lower_bound(unsigned j) const {
```
NikolajBjorner, line 504: this is not a real change. It is a formatting change. it should not be part of pull request.

PR #6774, opened 2023-06-20
Adding some options in support of F*

mtzguido:
In upgrading F* to use the latest Z3, these two features prove to be quite disruptive, so add options to disable them.

Not sure if the naming is adequate, any thoughts?


src/ast/rewriter/bool_rewriter.cpp
```diff
@@ -293,7 +294,8 @@ br_status bool_rewriter::mk_nflat_or_core(unsigned num_args, expr * const * args
 #endif
         if (s) {
```
NikolajBjorner, line 295: I think the code should be

```
        if (m_sort_disjunctions && s) {
            ast_lt lt;
            std::sort(buffer.begin(), buffer.end(), lt);       
            result = m().mk_or(sz, buffer.data());
            return BR_DONE;
        }
```

If nothing is sorted, then the rewrite has no effect. The rewrite engine is saved from rebuilding expressions.

src/ast/rewriter/bool_rewriter.cpp
```diff
@@ -331,7 +333,8 @@ br_status bool_rewriter::mk_flat_or_core(unsigned num_args, expr * const * args,
         if (mk_nflat_or_core(flat_args.size(), flat_args.data(), result) == BR_FAILED) {
             if (!ordered) {
```
NikolajBjorner, line 334:
```
if (!ordered && m_sort_disjunctions)
```

mtzguido:
Thanks for reviewing, good points! Just updated the PR.

PR #6791, opened 2023-07-02
User Propagator: Return if propagated lemma is redundant

CEisenhofer:
Reason for this PR.
The solver is forced to proceed in case the UP propagates new lemmas in the final callback.
However, propagation might "fail" in case the propagated expression is already true => the search may end although the user-propagator did not want it (maybe the UP just propagated some lemmas because it expected the solver to continue anyway).
We need to return something that indicates if the current propagation was really "accepted" by the core (i.e., was not considered redundant).

### DIFF
--- a/src/sat/smt/user_solver.cpp
+++ b/src/sat/smt/user_solver.cpp
@@ -43,15 +43,16 @@ namespace user_solver {
             m_prop.push_back(prop_info(explain, v, r));
     }
 
-    void solver::propagate_cb(
-        unsigned num_fixed, expr* const* fixed_ids,
-        unsigned num_eqs, expr* const* eq_lhs, expr* const* eq_rhs,
-        expr* conseq) {
+    bool solver::propagate_cb(
+            unsigned num_fixed, expr* const* fixed_ids,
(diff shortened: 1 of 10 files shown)


src/sat/smt/user_solver.cpp
```diff
@@ -43,15 +43,16 @@ namespace user_solver {
...
-    void solver::propagate_cb(
-        unsigned num_fixed, expr* const* fixed_ids,
-        unsigned num_eqs, expr* const* eq_lhs, expr* const* eq_rhs,
-        expr* conseq) {
+    bool solver::propagate_cb(
+            unsigned num_fixed, expr* const* fixed_ids,
+            unsigned num_eqs, expr* const* eq_lhs, expr* const* eq_rhs,
+            expr* conseq) {
         m_fixed_ids.reset();
         for (unsigned i = 0; i < num_fixed; ++i)
```
NikolajBjorner, line 51: Seems you want the following;

auto* n = ctx.get_enode(conseq);
if (n && s().value(enode2literal(n)) == l_true)
   return false;

PR #6803, opened 2023-07-09
Typo Fixes

THE-Spellchecker:
We have eliminated 65 typos from the codebase.

The file _src/math/simplex/model_based_opt.cpp_ had an extraneous carriage return character on line 550 which was removed by **THE-Spellchecker**. Unfortunately, as a side effect, git has mistakenly marked the entire file changed.

image


THE-Spellchecker:

PR #6814, opened 2023-07-15
code cleaning around m_touched_rows of lar_solver

levnach:


src/sat/smt/arith_solver.cpp
```diff
@@ -396,7 +396,7 @@ namespace arith {
             propagate_eqs(b.tv(), ci, k, b, value.get_rational());
 #if 0
         if (propagation_mode() != BP_NONE)
-            lp().mark_rows_for_bound_prop(b.tv().id());
+            lp().add_column_rows_to_touched_rows(b.tv().id());
```
NikolajBjorner, line 399: you managed to come up with a longer name! :-)

PR #6831, opened 2023-07-26
no lazy template pivot in lp2

levnach:
simplify the template code related to pivoting in theory_arith


NikolajBjorner:
This would require more review to merge: What ensures the pre-condition for when the is_base(s2) true?

PR #6844, opened 2023-08-04
merge theory_lra from branch arith

levnach:


src/smt/theory_lra.cpp
```diff
@@ -1641,70 +1644,129 @@ class theory_lra::imp {
...
-                    break;
-                default:
-                    break;
-                }
-                if (st == FC_CONTINUE)
-                    break;
-            }
-            return st;
+    final_check_status sat_case_of_final_check() {
+        bool giveup;
```
NikolajBjorner, line 1693: just initialize it to false;

PR #6875, opened 2023-08-24
Explain max term

levnach:
Hooking up improve_bounds()

### DIFF
--- a/src/math/lp/lar_solver.cpp
+++ b/src/math/lp/lar_solver.cpp
@@ -311,6 +311,7 @@ namespace lp {
 
     bool lar_solver::maximize_term_on_tableau(const lar_term& term,
         impq& term_max) {
+        flet f(m_mpq_lar_core_solver.m_r_solver.m_look_for_feasible_solution_only, false);    
         if (settings().simplex_strategy() == simplex_strategy_enum::undecided)
             decide_on_strategy_and_adjust_initial_state();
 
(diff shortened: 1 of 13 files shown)


src/math/lp/lar_solver.cpp
```diff
@@ -1057,6 +1034,110 @@ namespace lp {
...
+            if (s.m_basis_heading[j] >= 0) {
+                SASSERT(s.m_d[j].is_zero());
+                continue;
+            }
+            const mpq& d_j = s.m_d[j];
+            if (d_j.is_zero()) continue;
+            TRACE("lar_solver", tout << "d[" << j << "] = " << d_j << "\n";);
+            TRACE("lar_solver", s.print_column_info(j, tout););
+            const ul_pair& ul = m_columns_to_ul_pairs[j];
+            svector<constraint_index> deps;
```
NikolajBjorner, line 1054: unused?

src/math/lp/lar_solver.cpp
```diff
@@ -1057,6 +1034,110 @@ namespace lp {
...
+                bound_dep = ul.lower_bound_witness();               
+            }
+            dep = m_dependencies.mk_join(dep, bound_dep);
+        }
+        return dep;
+    }
+
+    
+
+    void lar_solver::get_explanation_of_maximum(const lar_term& term, explanation& exp) {
```
NikolajBjorner, line 1072: seems you want the explain functionality to be there even in spite it is in a broken state with nested dependencies. 
There really is a dependency tree now. An explanation is a set of inequalities with coefficients. Each is justified by either a literal or, recursively, an explanation for the negation. 
When searching we just need the flattened set of literals, when justifying the inference it can be split into portions. You could also multiply the coefficients. For example, suppose the bound x <= k has coefficient c1, but is explained by the core x > k * c2 + bound2 * c3 + bound3 * c4.
you could include bound2 with coefficient c3*c1/c2, and bound3 with coefficient c4*c1/c2.

src/math/lp/lar_solver.cpp
levnach: svector<constraint_index> deps;     - removed

src/math/lp/lar_solver.cpp
levnach: I removed this method and the test method using explanation.

PR #6883, opened 2023-08-30
[draft] Use half SipHash-1-3 for AST hashing

nunoplopes:
I've been hitting a bunch of cases where I see billions of AST hash collisions and then the run time goes through the roof.
We've partially offset this by increasing the initial AST hash table size, but I still many bad collisions.

This patch changes the code a bit to use a more modern API, where we have a hasher collecting data, and then there's a final mixing step.
I've used the half SipHash-1-3 (https://en.wikipedia.org/wiki/SipHash) to get started.

I'll benchmarks multiple algorithms. This one works well (in terms of number of collisions), but it's a bit slow.

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -93,18 +93,17 @@ bool parameter::operator==(parameter const & p) const {
     }
 }
 
-unsigned parameter::hash() const {
-    unsigned b = 0;
+void parameter::addHash(GenHash &hash) const {
+    hash.add(get_kind());
     switch (get_kind()) {
-    case PARAM_INT:      b = get_int(); break;
-    case PARAM_AST:      b = get_ast()->hash(); break;
-    case PARAM_SYMBOL:   b = get_symbol().hash(); break;
-    case PARAM_RATIONAL: b = get_rational().hash(); break;
-    case PARAM_DOUBLE:   b = static_cast<unsigned>(get_double()); break;
-    case PARAM_ZSTRING:  b = get_zstring().hash(); break;
-    case PARAM_EXTERNAL: b = get_ext_id(); break;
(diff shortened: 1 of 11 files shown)


nunoplopes:
Other usual suspects to try out:
 - https://en.wikipedia.org/wiki/MurmurHash
 - https://github.com/google/cityhash

NikolajBjorner:
C:\nz3\src\util/hash.h(80): warning C4267: '+=': conversion from 'size_t' to 'uint32_t', possible loss of data

nunoplopes:
Some numbers.

Alive2 running on LLVM unit tests:
image
The sys time goes down due to fewer hash tables re-hashings and thus fewer memory allocations.
Option 2 is the one that finds more bugs in LLVM. Note that Alive2 uses a timeout.

---------------------

Now let's look at compilation of a few C programs (sqlite3, bzip2, gzip, etc):
image

Again, 2nd option finds the most bugs and has fewer OOMs, so it can prove more stuff.

Option 2 is the one in this PR.

NikolajBjorner:
It appears to be a consistent 5% slowdown with the new hash for QF_BV.
I ran master with two different random seeds and pull request with 3 different random seeds over 41K benchmarks.
There is a 0.5% variation between random seeds for the same code and 5% between the different hash functions.

NikolajBjorner:
Another issue could be the hash-tables themselves. It uses low-order bits for bucket fitting. 
The entropy of these low order bits better be high. If the hash is given as a low entropy integer it may not be the case. Better hash table implementations may be less prone to collisions. You already "fixed" some of this by pre-assigning a larger hash-table to avoid wasting time in doubling.

nunoplopes:
ok, let's close it then. I realize Alive2 is not the most common benchmark 😅

Issue #6880, opened 2023-08-29
Configure a quantifier so it cannot be instantiated by e-matching


utaal:
I am interested in configuring a quantifier so it cannot be instantiated by e-matching (but it can be instantiated by mbqi), in the context of program verification. Here's a synthetic example:

```smt2
(set-option :auto_config false)
(set-option :smt.mbqi false)

(declare-sort Foo)
(declare-fun one (Foo) Bool)
(declare-fun two (Foo) Bool)

(declare-fun dummy_pattern (Foo) Bool)

(assert
  (forall ((f Foo)) (!
    (=>
      (one f)
      (two f)
    )
    :pattern (dummy_pattern f)
    :qid forall_f_one_two
  ))
)

(declare-const f1 Foo)
(assert (not
  (=>
    (one f1)
    (two f1)
  )
))

(check-sat)
```

This results in `unknown` (`(:reason-unknown "smt tactic failed to show goal to be sat/unsat (incomplete quantifiers)")`), as expected.
* commenting out the line `:pattern (dummy_pattern f)` results in `unsat`, as expected;
* setting `smt.mbqi` to true results in `unsat`, as expected.

I'm looking for a way to accomplish the same (preventing e-matching instantiation) without having to provide a dummy pattern. Removing all the patterns lets z3 pick a pattern for the quantifier, and something like `:pattern ()` is just ignored.

NikolajBjorner:
there isn't a way at present to disable pattern inference (by setting parameters).
Is this a feature request?

utaal:
> there isn't a way at present to disable pattern inference (by setting parameters).
> Is this a feature request?

Now that I know that there is in fact a way to disable pattern inference, I think it's indeed a feature request. Though it would only be enabling some experimentation for research work at this stage (in the context of program verification within EPR), so it's not critical for us at this stage.

Would you be open to adding this feature? We may be able to contribute the implementation. If so, should this be an additional attribute on the quantifier? (`:no-pattern-qi`? other suggestions?)

NikolajBjorner:
I don't see what value  `:no-pattern-qi` has over using a  `:pattern (dummy_pattern f)`. They seem to achieve the same.
The parameter addition would disable pattern inference across all quantifiers. Maybe that isn't desired either?

utaal:
> I don't see what value  :no-pattern-qi has over using a `:pattern (dummy_pattern f)`.

[...]

PR #6884, opened 2023-09-01
add the option :pi.enabled to disable pattern inference for quantifiers

utaal:
As suggested by https://github.com/Z3Prover/z3/issues/6880#issuecomment-1701531126. Is this the right way to approach this?

Should I add a test, and if so what's the best place to do so? (I tried it and it seems to work when setting the option in an smtlib2 input file)

Let me know if you prefer a different name for the option.


src/params/pattern_inference_params.cpp
```diff
@@ -21,6 +21,7 @@ Revision History:
 
 void pattern_inference_params::updt_params(params_ref const & _p) {
     pattern_inference_params_helper p(_p);
+    m_pi_enabled                 = p.enabled();
```
NikolajBjorner, line 24: good

PR #6932, opened 2023-10-04
Add RCF functions to OCaml API

wintersteiger:
This adds the RCF functions from `z3_rcf.h` to the OCaml API.


wintersteiger:
@microsoft-github-policy-service agree company="Imandra"

PR #6949, opened 2023-10-17
Update script to use importlib_resources

rsetaluri:
To avoid a deprecation warning, this change updates scripts/update_api.py to use 'importlib_resources' instead of 'pkg_resources'.

See https://setuptools.pypa.io/en/latest/pkg_resources.html and https://importlib-resources.readthedocs.io/en/latest/migration.html for more information.

### DIFF
--- a/scripts/update_api.py
+++ b/scripts/update_api.py
@@ -1827,17 +1827,25 @@ def write_core_py_preamble(core_py):
   core_py.write(
 """
 # Automatically generated file
+import atexit
 import sys, os
+import contextlib
 import ctypes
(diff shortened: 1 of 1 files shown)


rsetaluri:
This is a starting point for this change to get rid of deprecation warnings. I'm not sure what versions of python are supported for build/install, and whether or not `importlib_resources` exists for all of those. Furthermore, I'm unsure where the dependency on `importlib_resources` needs to be declared.

NikolajBjorner:
Thanks!!

PR #6973, opened 2023-10-30
Fixed parsing of | and \

CEisenhofer:
The \ does not only escape the | but also counts itself as a character of the name. Also, \\ does not escape \

e.g.,
In SMT-LIB input |\|| becomes |\\\|| 
|\\| becomes |\\\\|
|\\|| becomes |\\\\\|
...


NikolajBjorner:
- separate scanner changes from user-propagator
- esp. for a change to old code paths add unit test that establishes bug vs fix.

CEisenhofer:
This is not related to user-propagation 🤔

CEisenhofer:
Where should parsing unit-tests go to?
test/smt2print_parse.cpp ?

NikolajBjorner:
either in a new or old file in the test directory yes.

PR #6992, opened 2023-11-14
Fix memory and concurrency issues in OCaml API

wintersteiger:
Fixes multiple issues with the OCaml API:

- Global hash table for constructor objects meant that all contexts that use datatypes were never garbage collected
  - Adds `Z3_constructor_num_fields` to remove the need for the hash table
- Flush API solver objects more eagerly
- Fixes a latent memory leak in `Z3.Goal.simplify`
- Makes the object count of OCaml context trackers `_Atomic` where available
- Fixes a latent race on `api::context::m_allocated_objects`

### DIFF
--- a/src/api/api_context.cpp
+++ b/src/api/api_context.cpp
@@ -38,10 +38,9 @@ namespace api {
     void object::inc_ref() { ++m_ref_count; }
 
     void object::dec_ref() { SASSERT(m_ref_count > 0); if (--m_ref_count == 0) m_context.del_object(this); }
-
+    
     unsigned context::add_object(api::object* o) {
         flush_objects();
-        lock_guard lock(m_mux);
         unsigned id = m_allocated_objects.size();
         if (!m_free_object_ids.empty()) {
             id = m_free_object_ids.back();
@@ -87,12 +86,14 @@ namespace api {
     void context::flush_objects() {
 #ifndef SINGLE_THREAD
         if (!m_concurrent_dec_ref)
-            return;
-        lock_guard lock(m_mux);
-        if (m_asts_to_flush.empty() && m_objects_to_flush.empty())
-            return;
-        m_asts_to_flush2.swap(m_asts_to_flush);
-        m_objects_to_flush2.swap(m_objects_to_flush);
+            return;        
+        {
+            lock_guard lock(m_mux);
+            if (m_asts_to_flush.empty() && m_objects_to_flush.empty())
+                return;
+            m_asts_to_flush2.swap(m_asts_to_flush);
+            m_objects_to_flush2.swap(m_objects_to_flush);
+        }
         for (ast* a : m_asts_to_flush2)
             m().dec_ref(a);
         for (auto* o : m_objects_to_flush2) {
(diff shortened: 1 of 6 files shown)


src/api/api_context.cpp
```diff
@@ -81,14 +82,12 @@ namespace api {
...
-            return;        
-        {
-            lock_guard lock(m_mux);
-            if (m_asts_to_flush.empty() && m_objects_to_flush.empty())
-                return;
-            m_asts_to_flush2.swap(m_asts_to_flush);
-            m_objects_to_flush2.swap(m_objects_to_flush);
-        }
+            return;
+        lock_guard lock(m_mux);
```
NikolajBjorner, line 86: global lock guard allows contention during dec_ref

src/api/api_context.cpp
```diff
@@ -38,9 +38,10 @@ namespace api {
     void object::inc_ref() { ++m_ref_count; }
 
     void object::dec_ref() { SASSERT(m_ref_count > 0); if (--m_ref_count == 0) m_context.del_object(this); }
-    
+
     unsigned context::add_object(api::object* o) {
         flush_objects();
+        lock_guard lock(m_mux);
```
NikolajBjorner, line 44: what is the race that caused adding a lock guard here?
add_object on a context should only be called within the same thread.

src/api/api_solver.cpp
```diff
@@ -468,8 +468,10 @@ extern "C" {
     void Z3_API Z3_solver_dec_ref(Z3_context c, Z3_solver s) {
         Z3_TRY;
         LOG_Z3_solver_dec_ref(c, s);
-        if (s) 
-            to_solver(s)->dec_ref();        
+        if (s) {
+            to_solver(s)->dec_ref();
+            mk_c(c)->flush_objects();
```
NikolajBjorner, line 473: flush_objects should only be invoked when new objects are created or the context is deleted.
The assumption is that new objects are only created in the main thread of the context.

src/api/api_context.cpp
wintersteiger: Yes, I much prefer that to random segfaults though!

src/api/api_solver.cpp
wintersteiger: Yeah, this was mainly useful for debugging, it can go.

src/api/api_context.cpp
wintersteiger: `flush_objects` can be called concurrently by the GC, which also modifies `m_allocated_objects`. It's the same reason for having locks in .NET/Java and DecRefQueues to help with contention.

I believe that in OCaml 5 we could possibly get away with it, if we remove all the `*_dec_ref` calls in `z3.ml`, because the OCaml 5 GC doesn't parallelize internally, but I'm sure someone's already working on that...

Actually, looking at it again, I think I may have tricked myself with the call to `flush_objects` in `Z3_solver_dec_ref`. I'll try it without.

wintersteiger: Yeah, I think it's fine without those changes, although I will have to get rid of the `*_dec_ref`s in `z3.ml`. I do see a significant increase in non-deterministic unknowns, but that's probably a side-effect.

PR #6999, opened 2023-11-16
Qel fixes

hgvk94:


src/qe/qe_mbp.cpp
```diff
@@ -405,8 +406,18 @@ class mbproj::impl {
         return true;
     }
 
+    bool has_unsupported_th(const expr_ref_vector fmls) {
```
NikolajBjorner, line 409: shouldn't you use allow-lists of supported theories instead of blocked lists of unsupported theories? Someone will then start fuzzing this using the special relations theory or user-propagator operators?

NikolajBjorner:
On closer inspection there appear to be further systematic issues. Arrays are not necessarily 1-dimensional but the code made some unchecked assumptions. I have updated the code some places.

PR #7014, opened 2023-11-23
z3.h: Don't include `stdio.h`

waywardmonkeys:
This doesn't seem to actually be used or needed here.


waywardmonkeys:
Thank you!

PR #7015, opened 2023-11-25
cmake: Require cmake 3.16 or later.

waywardmonkeys:
Support for requiring cmake < 3.4 may go away soon (according to a deprecation notice when building).

Ubuntu 20.04 provides cmake 3.16 and current is 3.27, so that seems like a reasonable version to require.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Enforce some CMake policies
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.16)
 
 set(CMAKE_USER_MAKE_RULES_OVERRIDE_CXX "${CMAKE_CURRENT_SOURCE_DIR}/cmake/cxx_compiler_flags_overrides.cmake")
 project(Z3 VERSION 4.12.3.0 LANGUAGES CXX)


waywardmonkeys:
I need this for another change that will requires cmake 3.12 or later, but wanted to introduce this separately to ensure no issues with it.

aytey:
Can I suggest 3.12 rather than 3.16 then? It seems like this would exclude some users “pointlessly” if you only need 3.12 features.

waywardmonkeys:
Well, I don’t think this is actually tested, so not really sure that it even works on any given version. Is there someone that is actually stuck on something as old as 3.12?

waywardmonkeys:
FWIW, 3.12 was released in July 2018 and 3.16 in November 2020.

Along similar lines, Z3 requires C++17 and suggests or needs at least Visual Studio 2019 on Windows.

NikolajBjorner:
The way I see it: It is a non-goal for z3 builds to be compatible with stale software pipelines, so 3.16 (instead of 3.12) makes better sense. Downlevel compatibility is a well-known ask for users not upgrading their pipelines, but it is a trap when it comes to flexiblity.

PR #7030, opened 2023-12-02
Clarify optimizer guarantees

hexagonrecursion:
To me as a layperson from outside your theoretical field the phrasing such as "model meets the objectives", "solves objectives" and "check satisfiability while optimizing objective functions" sounds ambiguous. To me this phrasing sounds like it could mean either of the two options:
1. Best effort - try to find an optimal solution, but it's ok if you find a sub-optimal solution or if you can't prove that there are no better solutions
2. All or nothing - either find an optimal solution and prove that it is optimal or fail

From the C API documentation and from several bug reports about non-optimal solutions I have inferred that the second interpretation is correct.

Here is my attempt to document the guarantees given by the optimizer interface

The word "optimal" sounds unambiguous to me. If something is sub-optimal than it is *not* optimal. The words "optimize", "optimization" and "objective" on the other hand often do not imply that the result is perfect.

### DIFF
--- a/src/api/dotnet/Optimize.cs
+++ b/src/api/dotnet/Optimize.cs
@@ -220,7 +220,7 @@ namespace Microsoft.Z3
         /// <summary>
         /// Check satisfiability of asserted constraints.
         /// Produce a model that (when the objectives are bounded and 
-        /// don't use strict inequalities) meets the objectives.
+        /// don't use strict inequalities) is optimal.
         /// </summary>
         ///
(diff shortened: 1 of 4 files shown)


hexagonrecursion:
I'll think about this. Especially about the "Governing Law/Jurisdiction" section.

hexagonrecursion:
Authorship clarification "Check consistency and produce optimal values." is copy-paste from your C API docs, "is optimal" is my own addition.

PR #7039, opened 2023-12-05
simple scheme of choosing a column for gomory cut

levnach:
The idea is to choose columns with small non-integral part, and, of course, randomize the choice. Break the equality by comparing usage_in_terms.

### DIFF
--- a/src/math/lp/int_solver.cpp
+++ b/src/math/lp/int_solver.cpp
@@ -198,15 +198,9 @@ namespace lp {
         if (r == lia_move::undef) lra.move_non_basic_columns_to_bounds();
         if (r == lia_move::undef && should_hnf_cut()) r = hnf_cut();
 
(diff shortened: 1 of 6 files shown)


src/math/lp/int_solver.cpp
```diff
@@ -849,19 +829,71 @@ namespace lp {
...
+            return lra.usage_in_terms(j) > lra.usage_in_terms(k);
+        });
+        unsigned_vector ret;
+        unsigned n = static_cast<unsigned>(sorted_vars.size());
+
+        while (num_cuts-- && n > 0) {
+            unsigned k = random() % n;
+           
+            double k_ratio = k / (double) n;
+            k_ratio *= k_ratio*k_ratio;  // square k_ratio to make it smaller
```
NikolajBjorner, line 869: this is a cube, not a square

PR #7042, opened 2023-12-05
Use built-in `importlib.resources` on Python 3.9+

mgorny:
Use built-in `importlib.resources` module rather than the external `importlib_resources` package on Python 3.9 and newer.  The latter is only intended as a backport for old Python versions, and since modern Linux distributions may no longer support such old Python versions, they also no longer provide importlib_resources (this is the case on Gentoo).

CC @rsetaluri


mgorny:
Thank you.

PR #7058, opened 2023-12-14
Use `noexcept` more.

waywardmonkeys:
This applies it to swap functions as well as move constructors, move assignment, etc.

### DIFF
--- a/src/ast/recfun_decl_plugin.cpp
+++ b/src/ast/recfun_decl_plugin.cpp
@@ -606,7 +606,7 @@ namespace recfun {
         : m_lhs(from.m_lhs),
           m_def(from.m_def),
           m_args(from.m_args) {}
-    case_expansion::case_expansion(case_expansion && from)
+    case_expansion::case_expansion(case_expansion && from) noexcept
         : m_lhs(from.m_lhs),
           m_def(from.m_def),
           m_args(std::move(from.m_args)) {}
--- a/src/ast/recfun_decl_plugin.cpp
+++ b/src/ast/recfun_decl_plugin.cpp
@@ -602,15 +602,6 @@ namespace recfun {
         m_args.append(n->get_num_args(), n->get_args());
     }
 
-    case_expansion::case_expansion(case_expansion const & from)
-        : m_lhs(from.m_lhs),
-          m_def(from.m_def),
-          m_args(from.m_args) {}
-    case_expansion::case_expansion(case_expansion && from) noexcept
-        : m_lhs(from.m_lhs),
-          m_def(from.m_def),
-          m_args(std::move(from.m_args)) {}
-
     std::ostream& case_expansion::display(std::ostream & out) const {
         return out << "case_exp(" << m_lhs << ")";
     }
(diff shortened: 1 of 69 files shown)


src/ast/recfun_decl_plugin.cpp
```diff
@@ -606,7 +606,7 @@ namespace recfun {
         : m_lhs(from.m_lhs),
           m_def(from.m_def),
           m_args(from.m_args) {}
-    case_expansion::case_expansion(case_expansion && from)
+    case_expansion::case_expansion(case_expansion && from) noexcept
```
nunoplopes, line 609: these copy/move constructors should just be deleted

waywardmonkeys: @nunoplopes Just this one? Or more?

nunoplopes: I only spotted this one, but there might be others. The ones that do reference counting can't go; everything else should go.

NikolajBjorner: Note that case-expansion is not on a hot path so seems overtuned to add a move constructor here in the first place.

waywardmonkeys: In the same file, there was a move constructor for `body_expansion` as well (which was already `noexcept`), so I removed that one as well.

src/ast/recfun_decl_plugin.h
```diff
@@ -302,7 +302,6 @@ namespace recfun {
         expr_ref_vector     m_args;
         case_expansion(recfun::util& u, app * n);
         case_expansion(case_expansion const & from);
```
nunoplopes, line 304: the copy constructor must go too. You can't have just one of them. Either all or nothing.

src/ast/recfun_decl_plugin.h
```diff
@@ -325,8 +324,6 @@ namespace recfun {
             m_pred(pred), m_cdef(&d), m_args(args) {}
         body_expansion(body_expansion const & from):
```
nunoplopes, line 325: idem

src/util/f2n.h
```diff
@@ -45,7 +45,7 @@ class f2n {
         m_manager.set(m_one, ebits, sbits, 1);
     }
 
-    f2n(f2n && other) : m_manager(other.m_manager), m_mode(other.m_mode), m_ebits(other.m_ebits), m_sbits(other.m_sbits),
+    f2n(f2n && other) noexcept : m_manager(other.m_manager), m_mode(other.m_mode), m_ebits(other.m_ebits), m_sbits(other.m_sbits),
```
nunoplopes, line 48: this one can go too.

waywardmonkeys:
Updated, rebased to fix a merge conflict, and force pushed.

src/util/f2n.h
waywardmonkeys: When this one goes away, there's a build error on Ubuntu in CI. I've added it back for now.

PR #7069, opened 2023-12-24
Faster gc

levnach:
Use polarity of the non-basic variables to create bounds on basic variables in Gomory cuts


levnach:
z3 in this version is not much faster than the master, but it produces less timeouts: around 560 in QF_LIA, and around 62 in benchmark-submission.  The corresponding numbers in master are 676, and 88.

PR #7099, opened 2024-01-25
Expose forall and exists to Julia

remysucre:
A new release of Z3 is also needed to trigger the Julia binary wrapper update: https://github.com/Z3Prover/z3/blob/master/src/api/julia/README.md#julia-part

### DIFF
--- a/src/api/julia/z3jl.cpp
+++ b/src/api/julia/z3jl.cpp
@@ -303,6 +303,8 @@ JLCXX_MODULE define_julia_module(jlcxx::Module &m)
     m.method("xnor", &xnor);
     m.method("min", &min);
     m.method("max", &max);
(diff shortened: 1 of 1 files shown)


remysucre:
Thanks for merging this! When should I expect the next release? (Just so that I can plan, not trying to rush you)

NikolajBjorner:
Given that Julia bindings are not created with releases, what is the dependency on a release?

remysucre:
`Z3.jl` [downloads](https://github.com/JuliaPackaging/Yggdrasil/blob/46738f5f664a402fe1a04ba1756aab693b06d18f/Z/z3/build_tarballs.jl#L16) the released z3 binaries to use the bindings. The [README](https://github.com/Z3Prover/z3/blob/master/src/api/julia/README.md#julia-part) is inaccurate and I'll change it.

PR #7115, opened 2024-02-07
Fix some typos.

waywardmonkeys:


src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3850,12 +3850,12 @@ bool seq_rewriter::pred_implies(expr* a, expr* b) {
     Utility function to decide if two BDDs (nested if-then-else terms)
     have exactly the same structure and conditions.
 */
-bool seq_rewriter::ite_bdds_compatabile(expr* a, expr* b) {
```
NikolajBjorner, line 3853: need also to fix header file

waywardmonkeys: Fixed and pushed.

PR #7116, opened 2024-02-08
conditionally depend on importlib_resources

cj81499:
In Python >= 3.9, `importlib.resources` is available as part of the stdlib.

This change makes it so that the backport is only installed for Python versions where it is actually needed and used.

Relevant docs: https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers

Originally noticed here: https://github.com/cj81499/advent-of-code/pull/330#discussion_r1482498607


cj81499:
@NikolajBjorner thanks for merging! :)

Any idea when this will be released?

Thanks again!

PR #7136, opened 2024-02-22
ci: Stop using deprecated `::set-output`.

waywardmonkeys:
This has been deprecated and the replacement functionality is described here:

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


waywardmonkeys:
Sorry, I left the old line in ...

It should have removed the `run: echo "::set-output name=date::$(date +'%Y-%m-%d')"` line

PR #7200, opened 2024-04-02
Update README.md

angelica-moreira:
Minor change suggestion.
It seems that it is needed to add opt=-x for running the make script in a x64 machine.


NikolajBjorner:
Let me check. Seems like a regression.

NikolajBjorner:
it works for me with -x on Windows 11, python 3.12.

angelica-moreira:
Updated the python version, and it worked, sorry.

PR #7235, opened 2024-05-27
Fix compilation error in column_info

waywardmonkeys:
Reference to `m_low_bound` should be `m_lower_bound`.


waywardmonkeys:
This got picked up by the WebAssembly build of the Rust bindings because an emscripten update got a new version of LLVM / Clang, so I guess that was the canary in the coal mine. This code was last modified in 2017 / 2018 in commit 6202b2f2e486c7800c8b21c9912e17e74de5328c.

PR #7244, opened 2024-06-04
Clear up inefficient code

CoolThi:
Multiple `if` would confuse compilers to produce unnecessary branches. This patch clear it up.


NikolajBjorner:
this is not used in a performance critical phase.

PR #7297, opened 2024-07-18
add new hashtable unit tests

ChuyueSun:


src/test/hashtable.cpp
```diff
@@ -20,7 +21,7 @@ Revision History:
 #include<iostream>
 #include<unordered_set>
 #include<stdlib.h>
-
+#include <cassert>
```
NikolajBjorner, line 24: use VERIFY or SASSERT in z3 code base

ChuyueSun: Got it.

PR #7298, opened 2024-07-19
new heap invariants

ChuyueSun:
Note that the second new invariant (`All indices in m_value2indices that are not used in m_values should be 0`) is expensive to check. So should not be turned on during runtime.

### DIFF
--- a/src/test/heap.cpp
+++ b/src/test/heap.cpp
@@ -38,7 +38,7 @@ static void tst1() {
     for (int i = 0; i < N * 3; i++) {
         int val = heap_rand() % N;
         if (!h.contains(val)) {
-            SASSERT(!t.contains(val));
+            VERIFY(!t.contains(val));
             h.insert(val);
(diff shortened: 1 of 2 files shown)


NikolajBjorner (commented):
ENSURE is actually a macro for VERIFY. If you replace it, then use VERIFY.
This allows running the heap unit test in release mode to quicker cover more cases.

ChuyueSun:
> ENSURE is actually a macro for VERIFY. If you replace it, then use VERIFY. This allows running the heap unit test in release mode to quicker cover more cases.

Sounds good.

ChuyueSun:
> ENSURE is actually a macro for VERIFY. If you replace it, then use VERIFY. This allows running the heap unit test in release mode to quicker cover more cases.

Actually, I don't understand the difference between these two. Seems to me that `ENSURE` is an alias for `VERIFY`.

PR #7299, opened 2024-07-19
add new permutation class invariant

ChuyueSun:


src/util/permutation.cpp
```diff
@@ -66,11 +67,19 @@ void permutation::display(std::ostream & out) const {
 bool permutation::check_invariant() const {
     SASSERT(m_p.size() == m_inv_p.size());
     unsigned n = m_p.size();
+    std::vector<bool> check_vector(n, false); // To check for duplicate and out-of-range values
```
NikolajBjorner, line 70: use bool_vector for uniformity within z3 code base

ChuyueSun: > use bool_vector for uniformity within z3 code base

Got it.

Issue #7305, opened 2024-07-22
Add SA_ONSTACK flag to SIGINT signal handlers (required for CGO)


tiborschneider:
When including z3 from golang, we noticed that z3 registers `SIGINT` handlers without setting the `SA_ONSTACK` flag. This is required for CGO: https://pkg.go.dev/os/signal#hdr-Go_programs_that_use_cgo_or_SWIG. 

There seems to be lots of places in the code base where signal handlers are used. Fortunately, most are in `src/shell`, `src/api`, and `src/test`, all of which are not used in `z3.h`. The only file that would need changes is `src/util/scoped_ctrl_c.cpp`. Are there any plans to flag use `sigaction` with `SA_ONSTACK` instead of `signal` in that file?

NikolajBjorner:
sounds like updating to using sigaction instead of signal is the way to go.
z3.h uses src/api, which depends on scoped_ctrl. This is to allow killing console applications built on top of z3's API.

NikolajBjorner:
The issue is a good test case for github workspaces. Trying it out. I am not totally sure about whether the change preserves semantics, and it may be better to replace signal everywhere by sigaction. ctrl-c is used also from the shell.

nunoplopes:
Is this a real problem or just theoretical?
Because the linked page from Go says that programs have limited stack. Which is fine.. Z3's signal handlers don't do much. The ctr-C one just terminates the program or sets some terminating flag. The amount of stack usage is minimal.

I suggest that unless there is proof that something doesn't work, we don't try to fix a problem that doesn't exist. These things are super tricky to get right..

PR #7312, opened 2024-07-27
Fix SIGINT handler to use sigaction with SA_ONSTACK flag

NikolajBjorner:
Related to #7305

Update SIGINT signal handler to use `sigaction` with `SA_ONSTACK` flag.

* Replace `signal` function with `sigaction` function in `src/util/scoped_ctrl_c.cpp` to register the SIGINT handler.
* Set the `SA_ONSTACK` flag for the SIGINT signal handler in `src/util/scoped_ctrl_c.cpp`.
* Update `on_ctrl_c` function to use `sigaction` in `src/util/scoped_ctrl_c.cpp`.
* Modify `scoped_ctrl_c` struct in `src/util/scoped_ctrl_c.h` to include `struct sigaction m_old_action` instead of `void (STD_CALL *m_old_handler)(int)`.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/Z3Prover/z3/issues/7305?shareId=e00543fa-583e-4b92-affa-e8a73d553547).

### DIFF
--- a/src/util/scoped_ctrl_c.cpp
+++ b/src/util/scoped_ctrl_c.cpp
@@ -26,11 +26,10 @@ static void on_ctrl_c(int) {
         g_obj->m_cancel_eh(CTRL_C_EH_CALLER);
         if (g_obj->m_once) {
             g_obj->m_first = false;
-            signal(SIGINT, on_ctrl_c); // re-install the handler
         }
     }
     else {
-        signal(SIGINT, g_obj->m_old_handler); 
+        sigaction(SIGINT, &g_obj->m_old_action, nullptr);
         raise(SIGINT);
     }
 }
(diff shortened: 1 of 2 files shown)


src/util/scoped_ctrl_c.cpp
```diff
@@ -43,15 +42,17 @@ scoped_ctrl_c::scoped_ctrl_c(event_handler & eh, bool once, bool enabled):
     m_old_scoped_ctrl_c(g_obj) {
     if (m_enabled) {
         g_obj = this;
-        m_old_handler = signal(SIGINT, on_ctrl_c); 
+        struct sigaction sa;
+        sa.sa_handler = on_ctrl_c;
+        sa.sa_flags = SA_ONSTACK;
```
nunoplopes, line 47: but without calling `sigaltstack`, isn't this useless?

My man page says:

> Call the signal handler on an alternate signal stack
              provided by [sigaltstack(2)](https://www.man7.org/linux/man-pages/man2/sigaltstack.2.html).  If an alternate stack is not
              available, the default stack will be used.  This flag is
              meaningful only when establishing a signal handler.

NikolajBjorner: It is what Github WorkSpaces provided in the first shot. 
Besides looking suspicious, it also isn't cross-platform. For Windows, we would not use this.
Then also, I am not sure if the code in 'shell' is broken because it only considers sigint.
Altogether more than a quick change, but something that you as the highest esteemed and glorious expert in operating sophisticated machinery may find a spare moment to consider when not occupied near sunny grass lawns or ocean side camps (it is August, isn't it?).

PR #7313, opened 2024-07-27
Gcc 15 two phase

SoapGentoo:
Fix for GCC 15 two-phase lookup
    
* GCC 15 is more aggressive about checking dependent names:
      https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=r15-2117-g313afcfdabeab3
    
Bug: https://bugs.gentoo.org/936634


SoapGentoo:

PR #7337, opened 2024-08-12
Fix building with Windows SDK and Clang-CL

### LINKED ISSUE #7318: __GNUC__ isn't always defined
We're cross-compiling Z3 for Windows using MSVC Sysroot and clang-cl on Linux.
The compile steps are fine, but linking would fail due to ``_tzcnt_u64`` symbol not defined.

I'm clueless with how MSVC works, so here are some educated assumptions:
From https://github.com/llvm/llvm-project/blob/07d2709a17860a202d91781769a88837e4fb5f2a/clang/lib/Frontend/InitPreprocessor.cpp#L868 we could see that while ``__clang__`` and  ``__llvm__`` are always defined, but ``__GNUC__`` is only defined when building with GNU C Version specified, which is not the case when using Clang-CL.

This results in the following Z3 code:

https://github.com/Z3Prover/z3/blob/0c16d34eb0eb9eb2627606431c631d896d547f6f/src/util/mpz.cpp#L50 and https://github.com/Z3Prover/z3/blob/0c16d34eb0eb9eb2627606431c631d896d547f6f/src/util/mpz.cpp#L65

falls back to using ``#define _trailing_zeros**(X) _tzcnt_u**(X)`` 

Grepping through Windows SDK:

``` 
WinSDK/VC/Tools/MSVC/14.26.28801/include/immintrin.h:extern unsigned int     _tzcnt_u32(unsigned int);
WinSDK/VC/Tools/MSVC/14.26.28801/include/ammintrin.h:unsigned int _tzcnt_u32(unsigned int);
WinSDK/VC/Tools/MSVC/14.26.28801/bin/Hostx64/x86/c1xx.dll: binary file matches
WinSDK/VC/Tools/MSVC/14.26.28801/bin/Hostx64/x86/c1.dll: binary file matches
```

We can see that these two builtins are listed as an external function, which makes compilation success. 
However without c1xx and c1.dll to lower those intrinsics calls, clang generates them to actual lib call, which fails in linking due to those functions only exists as intrinsics

Naville:
This maybe fixes #7318

Definitely need a second pair of eyes to review, and maybe CI for testing

### DIFF
--- a/src/util/mpz.cpp
+++ b/src/util/mpz.cpp
@@ -50,7 +50,7 @@ Revision History:
     #define HAS_BUILTIN(X) __has_builtin(X)
 #else
     #define HAS_BUILTIN(X) 0
-#end
+#endif
 #if HAS_BUILTIN(__builtin_ctz)
 #define _trailing_zeros32(X) __builtin_ctz(X)
 #elif defined(_WINDOWS) && (defined(_M_X86) || (defined(_M_X64) && !defined(_M_ARM64EC))) && !defined(__clang__)
--- a/src/util/mpz.cpp
+++ b/src/util/mpz.cpp
@@ -46,10 +46,14 @@ Revision History:
 #define LEHMER_GCD
 #endif
 
-
-#if defined(__GNUC__)
+#ifdef __has_builtin
+    #define HAS_BUILTIN(X) __has_builtin(X)
+#else
+    #define HAS_BUILTIN(X) 0
+#end
+#if HAS_BUILTIN(__builtin_ctz)
 #define _trailing_zeros32(X) __builtin_ctz(X)
-#elif defined(_WINDOWS) && (defined(_M_X86) || (defined(_M_X64) && !defined(_M_ARM64EC)))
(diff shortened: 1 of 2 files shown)


src/util/mpz.cpp
```diff
@@ -46,10 +46,14 @@ Revision History:
 #define LEHMER_GCD
 #endif
 
-
-#if defined(__GNUC__)
+#ifdef __has_builtin
+    #define HAS_BUILTIN(X) __has_builtin(X)
+#else
+    #define HAS_BUILTIN(X) 0
+#end
```
NikolajBjorner, line 53: #endif

Naville: Oops, fixed

Naville:
I don't really understand why https://github.com/Z3Prover/z3/blob/83f47bd84bbd2f22b1394cac03a024f1e49c5183/cmake/cxx_compiler_flags_overrides.cmake#L9 is required, and it hasn't been updated for 7 years

PR #7351, opened 2024-08-23
Migrate string_buffer to use C++ smart pointer.

saikat107:
1. `string_buffer` class is now migrated to use smart pointers instead of manually allocated and deallocated memory.
The `buffer` function decides and returns if the buffer is currently using initial buffer or the allocated buffer.
2. Added `tst_small_buffer_expand` to test memory expansion of the small buffer, i.e., the test forces the buffer to go from `m_initial_buffer` to dynamically allocated `m_buffer`.

### DIFF
--- a/src/tactic/aig/aig_tactic.cpp
+++ b/src/tactic/aig/aig_tactic.cpp
@@ -27,24 +27,21 @@ class aig_manager;
 class aig_tactic : public tactic {
     unsigned long long m_max_memory;
     bool               m_aig_gate_encoding;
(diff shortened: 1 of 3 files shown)


NikolajBjorner:
alloc/dealloc use memory_manager. STL allows associating custom allocators with classes that call new/delete. 
Better to have a uniqe_ptr wrapper that ensures all memory uses memory_manager.

nunoplopes:
Plus the changes to `string_buffer` remove several of optimizations that it has. I don't see the point of this PR.

PR #7353, opened 2024-08-25
Avoid broken stack at few places

catap:
Here a similar fixes which I had discovered in the codebase. All of them follow the pattern: an object is allocated on the stack, but stored as pointer. As result an access to it may lead to crash, or write into it brokes the stack.


catap:

Issue #6902, opened 2023-09-14
Segfault on OpenBSD


bentley:
The following testcase from Symbiyosys (executed with `z3 -smt2 -in`) crashes on OpenBSD -current.

```
; SMT-LIBv2 description generated by Yosys 0.32 (git sha1 fbab08acf14, c++ 13.0.0 -O2 -fPIC -Os)
; yosys-smt2-stdt
; yosys-smt2-module top
(set-option :produce-models true)
(set-logic ALL)
(declare-datatype |top_s| ((|top_mk|
(|top_is| Bool)
(|top#0| (_ BitVec 8)) ; \counter
(|top#1| Bool) ; \clk
(|top#2| (_ BitVec 1)) ; \_witness_.anyinit_procdff_12
(|top#3| (_ BitVec 1)) ; $formal$autotune_div.sv:4$1_EN
)))
; yosys-smt2-witness {"offset": 0, "path": ["\\counter"], "smtname": 0, "smtoffset": 0, "type": "reg", "width": 8}
; yosys-smt2-register counter 8
; yosys-smt2-wire counter 8
(define-fun |top_n counter| ((state |top_s|)) (_ BitVec 8) (|top#0| state))
; yosys-smt2-input clk 1
; yosys-smt2-wire clk 1
; yosys-smt2-clock clk posedge
; yosys-smt2-witness {"offset": 0, "path": ["\\clk"], "smtname": "clk", "smtoffset": 0, "type": "posedge", "width": 1}
; yosys-smt2-witness {"offset": 0, "path": ["\\clk"], "smtname": "clk", "smtoffset": 0, "type": "input", "width": 1}
(define-fun |top_n clk| ((state |top_s|)) Bool (|top#1| state))
; yosys-smt2-anyinit top#2 1 autotune_div.sv:3.5-6.8
; yosys-smt2-witness {"offset": 0, "path": ["\\_witness_", "\\anyinit_procdff_12"], "smtname": 2, "smtoffset": 0, "type": "init", "width": 1}
; yosys-smt2-register _witness_.anyinit_procdff_12 1
; yosys-smt2-wire _witness_.anyinit_procdff_12 1
(define-fun |top_n _witness_.anyinit_procdff_12| ((state |top_s|)) Bool (= ((_ extract 0 0) (|top#2| state)) #b1))
[...]

PR #7354, opened 2024-08-26
Avoid broken stack by mbp_arrays

catap:


src/model/model_evaluator.h
```diff
@@ -32,6 +32,7 @@ class model_evaluator {
     struct imp;
     imp *  m_imp;
 public:
+    model_evaluator(model_evaluator & mev);
```
NikolajBjorner, line 35: Not sure about why you add this constructor. 
It breaks some assumptions with m_imp being unique.

src/qe/mbp/mbp_arrays.cpp
```diff
@@ -726,7 +726,7 @@ namespace mbp {
             model_evaluator mev(mdl);
             mev.set_model_completion(true);
             M = &mdl;
-            m_mev = &mev;
+            flet<model_evaluator*> _mev(m_mev, &mev);
```
NikolajBjorner, line 729: It may be more dilligent but it doesn't address any bug.

catap: `mev` is allocated on stack inside this function. So, keep it as pointer inside `m_mev` lead to broken stack. If you wish I may go back to dynamic allocation of this, but using flet seems cleaner.

src/model/model_evaluator.h
catap: indeed it doesn't required

src/qe/mbp/mbp_arrays.cpp
NikolajBjorner: Note that there is only one public method, m_mev doesn't get used from other methods.
This is why I say it doesn't address any existing bug. So it is more about cosmetics and better quality code. Any new code should have used the flet pattern, for old code there is a tradeoff between the overhead of churn.

catap: Well, `m_mev` is used inside this file outside of function on which stack it is allocated and from which stack a pointer has been get. If I understand code corectly, it keep pointer to `mev` allocated on the stack of `operator()`. When it's called who knows that value is here.

Have I missed something?

NikolajBjorner: There is no bug.
The only place the private class array_project_eqs_util is used is here:

```
        array_project_eqs_util pe (m);
        pe (mdl, arr_vars, fml, aux_vars);
        
```

There could be a future bug. But then there can be future bugs in many things.
There is risk even when adding fixes for future bugs.

Since we are spending time on this code, go ahead and remove the other two files that should not be changed.
The change is for the better, but the overhead in reviewing it and taking risk isn't the best use of time.

catap: anyway I can reproduce https://github.com/Z3Prover/z3/issues/6902 quite easy and I simple hunting to cause of this issue.

PR #7356, opened 2024-08-27
New python packaging and tests

rhelmot:
The new best-practices way of building a python package looks like this:

- use `python -m build` to generate wheels
- use `auditwheel repair` to verify that those wheels are conformant to symbol versioning standards and are tagged correctly

To this end, I was able to simplify the python bindings build substantially. This involved making some upstream changes to auditwheel, which is generally not very happy about auditing cross builds, but it was done and my PR is pending. The "most best-practices" way of doing cross builds is to use a full container image for the target architecture and using binfmt_misc + qemu-user to perform the build and audit, but I don't think this is possible on azure.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,8 +43,20 @@ jobs:
     - ${{if eq(variables['runRegressions'], 'True')}}:
       - template: scripts/test-regressions.yml
 
-- job: LinuxBuildsArm64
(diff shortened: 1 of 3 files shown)


NikolajBjorner:
I removed a line that includes uploading to an audit wheel and copied part into nightly.yml where I can create local artifacts

Issue #6890, opened 2023-09-07
ThirdParty Memory Allocators still causes crash


Naville:
From our internal project that uses Z3 and [snmalloc](https://github.com/microsoft/snmalloc/blob/7b597335aedc2c60b0174f6e67a7f7560f841760/src/snmalloc/override/malloc.cc#L6), we see some weird crash at [this line](https://github.com/Z3Prover/z3/blob/3aea4ebf42dd2b26261657e9a4823f91f9b3e736/src/util/memory_manager.cpp#L307)

Going to the top of the file, one can see that:

```
#elif defined(_WINDOWS)
# include <malloc.h>
# define HAS_MALLOC_USABLE_SIZE
# define malloc_usable_size _msize
#endif
```

While I havn't take deeper look into it, but apparently snmalloc doesn't provide overrides for ``_msize``. So in this case, the program would try to malloc from snm's heap, and uses WinCRT to query the usable size.

# Proposed fix
Make ``HAS_MALLOC_USABLE_SIZE`` an user configurable parameter

nunoplopes:
They just need to provide an override for _msize.

BTW, I would be very curious to see perf results of using Z3 with snmalloc, including replacing all of Z3's hand-made allocators.

Naville:
I'm not a memallocators expert (or expert in general). But is malloc() + ``msize()`` sort of like mixing usage of different APIs?

nunoplopes:
> I'm not a memallocators expert (or expert in general). But is `malloc()` + `msize()` sort of like mixing usage of different APIs?

No: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/msize?view=msvc-170#remarks

NikolajBjorner:
How about 

```
#elif defined(_WINDOWS) && defined(_msize)
# include <malloc.h>
# define HAS_MALLOC_USABLE_SIZE
# define malloc_usable_size _msize
#endif
```

nunoplopes:
_msize is not a macro on windows. At least the documentation doesn't say so. So that `#if` trick won't work.

Naville:
Any reason we can't revert #6321 and disable related code on Windows?
Though, I agree that wait for snmalloc to fix the issue would be a equally good if not better idea :)

nunoplopes:
> Any reason we can't revert #6321 and disable related code on Windows? Though, I agree that wait for snmalloc to fix the issue would be a equally good if not better idea :)

[...]

PR #7384, opened 2024-09-18
Add cmake option for malloc_usable_size

remysucre:
Add a cmake option `USE_MALLOC_USABLE_SIZE` (defaults to `TRUE`) as [suggested](https://github.com/Z3Prover/z3/issues/6890#issuecomment-1713347718) to address known issues with third-party allocators (#6890). 

I'm tempted to set this to `FALSE` by default. Is there evidence that `malloc_usable_size` actually improves performance? The only thing I could find seems [speculative](https://github.com/Z3Prover/z3/pull/6321#issuecomment-1237439487), and lots of people are patching to remove it (see the mentions at the end [here](https://github.com/Z3Prover/z3/pull/6321)).

### DIFF
--- a/src/util/memory_manager.cpp
+++ b/src/util/memory_manager.cpp
@@ -13,19 +13,17 @@ Copyright (c) 2015 Microsoft Corporation
 #include "util/error_codes.h"
 #include "util/debug.h"
 #include "util/scoped_timer.h"
-#ifdef __GLIBC__
+#ifndef HAS_MALLOC_USABLE_SIZE
+// drop calls to malloc_usable_size
+#elif defined(__GLIBC__)
 # include <malloc.h>
-# define HAS_MALLOC_USABLE_SIZE
 #elif defined(__APPLE__)
 # include <malloc/malloc.h>
-# define HAS_MALLOC_USABLE_SIZE
 # define malloc_usable_size malloc_size
 #elif defined(__FreeBSD__)
 # include <malloc_np.h>
-# define HAS_MALLOC_USABLE_SIZE
 #elif defined(_WINDOWS)
 # include <malloc.h>
-# define HAS_MALLOC_USABLE_SIZE
 # define malloc_usable_size _msize
 #endif
 
--- a/src/util/memory_manager.cpp
+++ b/src/util/memory_manager.cpp
@@ -13,17 +13,21 @@ Copyright (c) 2015 Microsoft Corporation
 #include "util/error_codes.h"
 #include "util/debug.h"
 #include "util/scoped_timer.h"
-#ifndef HAS_MALLOC_USABLE_SIZE
+#ifndef USE_MALLOC_USABLE_SIZE
 // drop calls to malloc_usable_size
 #elif defined(__GLIBC__)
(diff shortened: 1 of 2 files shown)


remysucre:

src/util/memory_manager.cpp
```diff
@@ -13,19 +13,17 @@ Copyright (c) 2015 Microsoft Corporation
 #include "util/error_codes.h"
 #include "util/debug.h"
 #include "util/scoped_timer.h"
-#ifdef __GLIBC__
+#ifndef HAS_MALLOC_USABLE_SIZE
+// drop calls to malloc_usable_size
+#elif defined(__GLIBC__)
 # include <malloc.h>
-# define HAS_MALLOC_USABLE_SIZE
```
NikolajBjorner, line 18: Not sure about this: the assumption of what has to go into the build configuration changes materially.

remysucre: Not exactly sure what you mean, but I pushed another change. Does that look better?

remysucre: Do you mean "whether to call `malloc_usable_size()` or not" is too low-level to be added to the build configuration?

NikolajBjorner: You are changing the build behavior. 
The behavior before: user doesn't configure anything. For selected configurations it uses the usable size.
New behavior: the user has to explicitly configure, with the CMAKE build only, the HAS_MALLOC_USABLE_SIZE. Otherwise it reverts to not using the usable size. 

Even this new code:
```
#ifndef USE_MALLOC_USABLE_SIZE
// drop calls to malloc_usable_size
#elif defined(__GLIBC__)
# include <malloc.h>
# define HAS_MALLOC_USABLE_SIZE
#elif defined(__APPLE__)
# include <malloc/malloc.h>
# define HAS_MALLOC_USABLE_SIZE
# define malloc_usable_size malloc_size
#elif defined(__FreeBSD__)
# include <malloc_np.h>
# define HAS_MALLOC_USABLE_SIZE
#elif defined(_WINDOWS)
# include <malloc.h>
# define HAS_MALLOC_USABLE_SIZE
# define malloc_usable_size _msize
#endif
```

doesn't make sense to me:

#ifndef USE_MALLOC_USABLE_SIZE
 // use malloc usable size is not defined
#else
 // use malloc usable size is defined. The uses of #define HAS_MALLOC_USABLE_SIZE are redundant.
 
 Is this really what you intend? Do you intend #ifdef instead of #ifndef?
 I just can't take pull requests that don't make sense. In the end I have to fumble with regressions.

remysucre: I see. I pushed [another change](https://github.com/Z3Prover/z3/compare/master...remysucre:z3:master) to keep `malloc_usable_size` by default, and the user can pass an `Z3_AVOID_MALLOC_USABLE_SIZE` option to drop the calls. I believe this implements exactly the recommendation [here](https://github.com/Z3Prover/z3/issues/6890#issuecomment-1723508107) which got a 👍 from you some time ago. Should I open another PR?

PR #7388, opened 2024-09-20
Julia now uses the C API.

PallHaraldsson:


PallHaraldsson:
See #7387 to probably merge first, or together. I didn't find a way to add this commit there.

Note, possibly the README shouldn't explain the C++ since legacy for Julia's Z3.jl?

PR #7396, opened 2024-09-23
Refresh gcm script

pelikhan:
Simplified gcm script and supporting files for intellisense.
- inquirer dependency is buitin.


genaisrc/genaiscript.d.ts
```diff
@@ -0,0 +1,2631 @@
+type OptionsOrString<TOptions extends string> = (string & {}) | TOptions
```
NikolajBjorner, line 1: should this be part of pull request?

Issue #7399, opened 2024-09-25
WASM bindings are broken


NikolajBjorner:
Z3 now uses c++20. #7379
However, there is something amiss for more than a month now. The continuous build/test of WASM bindings is broken.
@pelikhan and I can confirm it remains broken when setting it up over codespaces (following the steps in the README.md and then running npm test).
Some observations:

- emsdk uses node 18, the current default is 20. Switching to node 18 mode doesn't seem to matter.
- it runs out of memory. 
- Commenting out the unit test (number 17) does not change the memory issue, it just runs out of memory later.
- Adding an early return before number 17 causes an exception.
- Adding console.log to calls like

```
    constructor(ptr: Z3_solver | string = Z3.mk_solver(contextPtr)) {
        this.ctx = ctx;
        let myPtr: Z3_solver;
        if (typeof ptr === 'string') {
          myPtr = check(Z3.mk_solver_for_logic(contextPtr, _toSymbol(ptr)));
        } else {
          myPtr = ptr;
        }
        this.ptr = myPtr;
        console.log("create solver");
        Z3.solver_inc_ref(contextPtr, myPtr);
        cleanup.register(this, () => { console.log("dec-ref-solver"); Z3.solver_dec_ref(contextPtr, myPtr);});
      }
```
suggests that the cleanup actions are never run.
I can see that cleanup is an instance of FinalizationRegistry, that is connected to GC. 

- Should the gc be forced? Global.gc, std.global.gc, std.gc?

@bakkot

bakkot:
I can confirm the build worked at 3e518b9e8b387f1bb664a619e0aec45f4b62dfba and fails at bf34600f08d732680f7a9b3d85a257caf435418b. In fact it fails even when asking for 2GB rather than 1GB on [this line](https://github.com/Z3Prover/z3/blob/82eb18674b446f3a90992fc57cbb690ccd4cfa08/src/api/js/scripts/build-wasm.ts#L72).

I notice a new warning when building at the later commit, not present on the former:

```
../src/nlsat/nlsat_solver.cpp:2840:23: warning: binding reference member 'm_lit_num' to stack allocated parameter 'ln' [-Wdangling-field]
 2840 |             m_lit_num(ln) {
      |                       ^~
../src/nlsat/nlsat_solver.cpp:2837:31: note: reference member declared here
 2837 |             unsigned_vector & m_lit_num;
      |                               ^
```
[...]

PR #7400, opened 2024-09-25
wasm: increase memory and attempt to GC in tests

bakkot:
Attempt to fix https://github.com/Z3Prover/z3/issues/7399.

Just forcing GC or increasing memory size to 2GB alone are each insufficient, but in combination tests pass again at least on my machine. This is not an elegant solution but hopefully at least un-breaks the build. I can try to experiment with newer versions of emscripten or a different test runner if I have time later.

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -48,7 +48,7 @@ jobs:
           source $(dirname $(which emsdk))/emsdk_env.sh
           which node
           which clang++
-          npm run build:wasm -- -sINITIAL_MEMORY=128MB -sALLOW_MEMORY_GROWTH
+          npm run build:wasm
 
       - name: Test
         run: npm test
--- a/src/api/js/package.json
+++ b/src/api/js/package.json
@@ -29,7 +29,7 @@
     "clean": "rimraf build 'src/**/*.__GENERATED__.*'",
     "lint": "prettier -c '{./,src/,scripts/,examples/}**/*.{js,ts}'",
     "format": "prettier --write '{./,src/,scripts/}**/*.{js,ts}'",
-    "test": "jest",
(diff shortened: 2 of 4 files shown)


NikolajBjorner:
Something different is happening. It times out in "can do simple proofs". 
I would merge this PR even if it doesn't run to completion because it looks easier to fix than the original code.

NikolajBjorner:
I get the following now in code spaces:

● Test suite failed to run

    RuntimeError: Aborted('PThread' was not exported. add it to EXPORTED_RUNTIME_METHODS (see the Emscripten FAQ))

      887 |   // TODO(https://github.com/google/closure-compiler/pull/3913): Remove if/when upstream closure gets fixed.
      888 |   /** @suppress {checkTypes} */
    > 889 |   var e = new WebAssembly.RuntimeError(what);
          |           ^
      890 |
      891 |   readyPromiseReject(e);
      892 |   // Throw the error whether or not MODULARIZE is set because abort is used

      at abort (build/z3-built.js:889:11)
      at Object.get (build/z3-built.js:1200:9)
      at killThreads (src/jest.ts:51:6)
      at Object.<anonymous> (src/high-level/high-level.test.ts:88:22)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       39 passed, 39 total
Snapshots:   0 total
Time:        46.001 s
Ran all test suites.
@NikolajBjorner ➜ .../z3/src/api/js (master) $

bakkot:
For code spaces, my guess is that the version of emscripten in code spaces is newer than the version in CI. If you do `emsdk install 3.1.15 && emsdk activate 3.1.15` does that fix it?

I should update the emscripten version used in CI; it's way behind. But your experience suggests there have been some breaking changes so this might not happen right away.

PR #7428, opened 2024-10-19
js: Adding manual release methods

### LINKED ISSUE #7427: JS High-level api - Memory cleanup
I've noticed that in the js high level api, memory of some object will never be freed from the wasm memory until the api shuts down and garbage collects. This works fine in short-lived applications, but if the library is used in a long living application where new variables / solvers / functions whatnot are created dynamically, there is to my knowledge no way to free up the memory it allocates. 

Currently all cleanup callbacks are registered on the `createApi` instance https://github.com/Z3Prover/z3/blob/master/src/api/js/src/high-level/high-level.ts#L118, and i'm trying to understand why its not either registered on the individual context the object is created within (so you can garbage collect an entire context and cleanup its associated memory), or individually on the objects that are allocating the memory. 
For example, why not register `FinalizationRegistry` on `Solver` class itself, so if nothing points to an instance of `Solver`, it will allow the memory to be freed from the wasm memory. 

Am I missing something that wont make that possible? 

Tagging @bakkot  / @ritave since this design mostly comes from you?

HalfdanJ:
Adding methods allowing manual reference cleanup of object created by high level api. 

I was considering if it instead should be called decRef like the native api, and java api is called. https://github.com/Z3Prover/z3/blob/master/src/api/java/Solver.java#L418, or if it should be called something more intuitive to a javascript developer, for whom the concept of references is abstracted away. My suggestion in this pr is the latter. 

Also considering if we want to add runtime checks that asserts no methods can be called on objects that are released. 

Fixes #7427

@bakkot

### DIFF
--- a/src/api/js/src/high-level/high-level.ts
+++ b/src/api/js/src/high-level/high-level.ts
@@ -178,6 +178,10 @@ export function createApi(Z3: Z3Core): Z3HighLevel {
       ctxs.forEach(other => assert('ctx' in other ? ctx === other.ctx : ctx === other, 'Context mismatch'));
     }
 
+    function _assertPtr<T extends Number>(ptr: T | null): asserts ptr is T {
+      if (ptr == null) throw new TypeError('Expected non-null pointer');
+    }
+
     // call this after every nontrivial call to the underlying API
     function throwIfError() {
       if (Z3.get_error_code(contextPtr) !== Z3_error_code.Z3_OK) {
@@ -1235,18 +1239,17 @@ export function createApi(Z3: Z3Core): Z3HighLevel {
       toString() {
         return this.sexpr();
       }
-
-      release() {
-        Z3.dec_ref(contextPtr, this.ast);
-        cleanup.unregister(this);
-      }
     }
 
     class SolverImpl implements Solver<Name> {
       declare readonly __typename: Solver['__typename'];
 
-      readonly ptr: Z3_solver;
       readonly ctx: Context<Name>;
+      private _ptr: Z3_solver | null;
+      get ptr(): Z3_solver {
+        _assertPtr(this._ptr);
+        return this._ptr;
+      }
 
       constructor(ptr: Z3_solver | string = Z3.mk_solver(contextPtr)) {
         this.ctx = ctx;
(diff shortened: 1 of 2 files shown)


bakkot:
You'll need to pass `this` as the third argument to `register` for the `unregister` call to work.

HalfdanJ:
Good catch.

bakkot:
Looks good. It would kind of be nice to be defensive against the user calling `release` multiple times, or using the other methods after calling it, but that requires runtime checks and so may not be worthwhile.

HalfdanJ:
Yeah was considering that, could be done storing a Boolean on the objects. I think what would make sense design wise is make a super class that provides the registration and release of the reference, that super class can then be providing that Boolean flag, and assertion function without it having to be implemented on all class implementations. 

The child classes just have to provide the increase and decrease functions to the super constructor.

bakkot:
> could be done storing a Boolean on the objects

That would work, but the standard practice here is to set `ptr` to `null` once it's been released. That also means you don't need a second field.

> I think what would make sense design wise is make a super class that provides the registration and release of the reference

Eh... you still need to _call_ the assertion function in every method on every class, so having the actual implementation be in a superclass doesn't buy you much (vs just having a top-level `function checkNonNull(ptr: number | null) { if (ptr == null) throw new TypeError }` which you call at the top of every method). And adding a superclass adds complexity to some already quite complex types. I would prefer avoid the superclass.

HalfdanJ:
I'm going to limit the scope to just adding `release` to `Solver` `Optimize` and `Model` in this case i think. Making ptr possibly null on all objects means assertions need to be added a lot of places. 

One big downside of doing this is that its a breaking change type wise to the library. Any usage of the library that references `.ptr` will have a new type.

bakkot:
Sound fine.

I wouldn't worry too much about making `ptr` nullable changing the types: that change reflects a change in the underlying reality, i.e., that now the `ptr` might no longer point to valid memory, where it always did before (unless you were reaching into the low-level API). Also I doubt many people are relying on it anyway. But that can be changed later, if we are inclined to.

HalfdanJ:
Ok, assertion added, i think this solution with an internal `_ptr` field is best since it avoids any issues with typescript not allowing setting ptr to null, and it adds assertion to any place that requests the `ptr` field from any of the objects.

bakkot (approved):

PR #7479, opened 2024-12-16
Add 2 new API functions to get depth & groundness of exprs

nunoplopes:
Add:
 - Z3_is_ground
 - Z3_get_depth

These cannot be implemented efficiently from outside of Z3. These functions are O(1) when run internally.

That's my Xmas gift for myself this year 🎅🎅


nunoplopes:
SantaClauseHohohoGIF

PR #7481, opened 2024-12-17
Optimize expr_safe_replace for quantifiers when all source patterns are vars

nunoplopes:
I get > 40% speedups in some benchmarks with this patch.

Alive2 calls var_subst to replace var:0 with something else. But if this falls in the `has_quantifier(n)` case, it calls expr_safe_replace.
This patch adds a fast path for when all the source exprs are variables by using var_subst, which is much faster than expr_sage_replace (it has multiple fast paths).

@NikolajBjorner could you please check the sanity of this patch? 🙏


nunoplopes:
> z3-test var_subst fails. Can you check what is going on?

fixed! (it was using the inverse var order)

PR #7514, opened 2025-01-18
use cmake from PyPI only when system executable is not available

mgorny:
Rather than pulling `cmake` from PyPI unconditionally, add it to build dependencies only if the system `cmake` executable cannot be found. This eliminates an unnecessary dependency on systems featuring CMake, and ensures that whenever possible, a downstream patched CMake version is used that is more compatible with the system in question.


mgorny:
Thanks!

PR #7519, opened 2025-01-21
Performance improvements for seq-sls

CEisenhofer:
+ bugfix for stoi


src/ast/sls/sls_seq_plugin.cpp
```diff
@@ -196,8 +197,27 @@ namespace sls {
                     update(e, rational(sx.indexofu(sy, val_z.get_unsigned())));
                 return false;
             }
-            // last-index-of
-            // str-to-int
+            if (seq.str.is_last_index(e, x, y) && seq.is_string(x->get_sort())) {
+                // TODO
+                SASSERT(false);
```
NikolajBjorner, line 202: Use UNREACHABLE() or NOT_IMPLEMENTED_YET()

PR #7529, opened 2025-01-26
Set default behaviour to exit on error in CLI mode

### LINKED ISSUE #7528: Z3 ignores errors
Hi, Z3 is commonly used for formal verification where you generally want to be really really sure you haven't made mistakes and everything should be failsafe. So I find it really weird that by default Z3 ignores errors and will produce `sat` or `unsat` even if there are obvious errors in the model.

This was noted [here](https://github.com/Z3Prover/z3/issues/1445) and I ran into it [here](https://github.com/rems-project/sail/pull/747#issuecomment-2574872676) but not changed.

I find this quite surprising - you really don't want your formal verification tool to say something is proven because it parses the Z3 output incorrectly (perhaps because it made the reasonable assumption that errors wouldn't be ignored).

Wouldn't the safer behaviour be to stop on errors. If there are any situations where you actually do want to continue (@alasdair suggested competitions) then you can opt in with a `--dangerous-ignore-errors` flag or similar).

Timmmm:
This changes the default `error-behaviour` to `immediate-exit` for smtlib2 when running in CLI mode. This makes it fail-safe. There's no reason to ignore errors in CLI mode and it may lead to accidentally "verified" proofs if the user makes an easy mistake.

This does not change the default when used via an API because `immediate-exit` calls `_Exit(1)` and that isn't a good thing for a library to do.

Fixes #7528

### DIFF
--- a/src/shell/smtlib_frontend.cpp
+++ b/src/shell/smtlib_frontend.cpp
@@ -160,6 +160,11 @@ unsigned read_smtlib2_commands(char const * file_name) {
     signal(SIGINT, on_ctrl_c);
     cmd_context ctx;
 
+    // Fail safe when used from the command line - if there's an error exit(1).
+    // We don't want to do this when used as a library because it would kill
+    // the entire process.
+    ctx.set_exit_on_error(true);
+
     ctx.set_solver_factory(mk_smt_strategic_solver_factory());
     install_dl_cmds(ctx);
     install_dbg_cmds(ctx);


NikolajBjorner:
Sorry if I was unclear.
There is already an option to force z3 to exit on error by setting a configuration.
Use this for your application.

I am going to get support calls to revert your change.

Timmmm:
Sorry I didn't quite get what you meant about interactive text mode. I'll try again.

PR #7530, opened 2025-01-26
Add -repl flag and set default behaviour to exit on error

Timmmm:
This changes the default `error-behaviour` to `immediate-exit` for smtlib2 when running in CLI mode. This makes it fail-safe. There's no reason to ignore errors in CLI mode and it may lead to accidentally "verified" proofs if the user makes an easy mistake.

This does not change the default when used via an API because `immediate-exit` calls `_Exit(1)` and that isn't a good thing for a library to do.

When passing commands via standard input there are two use cases:

* Running Z3 from another program (e.g. a compiler), in which case you don't want to ignore errors.
* Using it as a basic REPL, in which case you don't.

I added a `-repl` flag for the second use case. This also opens the door for potentially adding a "proper" REPL in future (with history, etc.).

### DIFF
--- a/src/shell/main.cpp
+++ b/src/shell/main.cpp
@@ -48,6 +48,7 @@ typedef enum { IN_UNSPECIFIED, IN_SMTLIB_2, IN_DATALOG, IN_DIMACS, IN_WCNF, IN_O
 static char const * g_input_file          = nullptr;
 static char const * g_drat_input_file     = nullptr;
 static bool         g_standard_input      = false;
+static bool         g_repl                = false;
 static input_kind   g_input_kind          = IN_UNSPECIFIED;
 bool                g_display_statistics  = false;
 bool                g_display_model       = false;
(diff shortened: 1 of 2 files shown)


Timmmm:
I had a brief look for documentation to update but didn't see any - let me know if I missed it.

NikolajBjorner:
(set-option :error-behavior immediate-exit)
(declare-const x Int)
(assert x)
(assert (> x 0))
(check-sat)

The other place set_exit_on_error is called is based on this command. I was assuming you would know it since you already found set_exit_on_error. The documentation is probably bad.

Timmmm:
The documentation probably is bad (this option is entirely undocumented as far as Google knows)... but that's missing the point anyway. This PR is intended to make the behaviour *fail-safe*. Documentation can't achieve that because one of the failures that this makes safe is not reading the documentation! (Let's not pretend that any of us always read all of the documentation.)

The classic illustration of fail-safe API design is Python's `yaml.load()` and `yaml.load_safe()`. It is very easy to not read the documentation and just use `yaml.load()` and assume that it has sensible behaviour, when in fact it does crazy unsafe things. It's not fail-safe. To put it another way, dangerous behaviours should be opt-in, not opt-out. Systems should be safe by default. 

The point of this PR is to make ignoring errors (the dangerous behaviour) opt-in, not opt-out. The existing option is an opt-out mechanism.

I was going to try to point to some article hopefully explaining this better than me, but it seems like [there isn't really a term for this concept](https://www.reddit.com/r/rust/comments/esqhr2/everything_is_safe_by_default_approach/) (someone suggested "common sense"!). I should write a blog post...

Anyway please would you reconsider this from that perspective?

PR #7534, opened 2025-01-27
Preserve fingerprint in trace

JonasAlaif:
Commits https://github.com/Z3Prover/z3/commit/554191885eca4a97a2e5ff2af52a4d16f513ba61 and https://github.com/Z3Prover/z3/commit/23c4728d684141b53b25d4236ef47870786e331a break the trace file since it requires linking a `[new-match]` to a later `[instance]` (with potentially other `[new-match]` between).

This PR tries to achieve the same thing as those commits (removing the platform dependent value due to the pointer cast) while preserving a unique link between the two lines. @NikolajBjorner is `get_data_hash` the correct thing to use to uniquely identify a fingerprint/new match?

### DIFF
--- a/src/smt/qi_queue.cpp
+++ b/src/smt/qi_queue.cpp
@@ -188,7 +188,7 @@ namespace smt {
 
     void qi_queue::display_instance_profile(fingerprint * f, quantifier * q, unsigned num_bindings, enode * const * bindings, unsigned proof_id, unsigned generation) {
         if (m.has_trace_stream()) {
-            m.trace_stream() << "[instance] 0x0";
+            m.trace_stream() << "[instance] " << f->get_data_hash();
             if (m.proofs_enabled())
                 m.trace_stream() << " #" << proof_id;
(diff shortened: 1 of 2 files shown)


JonasAlaif:

src/smt/qi_queue.cpp
```diff
@@ -188,7 +188,7 @@ namespace smt {
 
     void qi_queue::display_instance_profile(fingerprint * f, quantifier * q, unsigned num_bindings, enode * const * bindings, unsigned proof_id, unsigned generation) {
         if (m.has_trace_stream()) {
-            m.trace_stream() << "[instance] 0x0";
+            m.trace_stream() << "[instance] " << f->get_data_hash();
```
NikolajBjorner, line 191: f can be null.
I removed references to f because the format of printing the pointers was platform specific. 
Can you make this safer by using get_data_hash(f) instead, and 

unsigned get_data_hash(fingerprint* f) {
      return f ? f->get_data_hash() : 0;
}

PR #7537, opened 2025-01-29
Removed minor divergence case for seq-sls

CEisenhofer:


NikolajBjorner:
actually looks like the place updates is being processed it should skip identity updates. This lessens burden on all places where updates are added. Alternative is to have a function that performs the push_back operation and checks identity. Either way is more resilient coding approach.

PR #7553, opened 2025-02-13
Fix build when Z3_API macro is non-empty

pclayton:
Z3 fails to build with VS 2022 when the macro `Z3_API` is defined as `__stdcall` due to differences between API function declarations and definitions.  This PR updates function definitions to match declarations by
  - adding the `Z3_API` macro where missing;
  - changing `Z3_API char const *` to `Z3_string Z3_API`.


pclayton:

PR #7557, opened 2025-02-16
Build fix for eee96ec: create mbp::def w/ brace initialization

jevinskie:
eee96ec removed the explicit constructor for mbp_plugin.h's struct def. Change qe_mbp.cpp and q_mbi.cpp to use brace initialization.


jevinskie:

jevinskie:
@NikolajBjorner looks like you beat me to it by a minute or so. :)

NikolajBjorner:
it built locally so had to round-trip over build bots and left it in partial state yesterday

Issue #7575, opened 2025-03-07
License of DLLs in Windows Wheels


hyandell:
Hi Z3 folk,

Would it be possible to include licensing information in the Windows Python wheels for the various DLLs? I assume these are freely distributable standard Microsoft files?

Specifically I mean these files:

```
ls lib/
libz3.dll			msvcp140_1.dll			msvcp140_atomic_wait.dll	vcomp140.dll			vcruntime140_1.dll
msvcp140.dll			msvcp140_2.dll			msvcp140_codecvt_ids.dll	vcruntime140.dll		vcruntime140_threads.dll
```

NikolajBjorner:
Z3 is OSS. You are invited to add a pull request to fix the issue.

Otherwise, the license information is clearly indicated on main distribution sites for z3 and VS redist.

PR #7576, opened 2025-03-07
Reference VS licensing in README

hyandell:
Related to #7575. I think I've identified the right licensing file per the recommendation to use at least VS 2019. If VS 2022 is used however, it seems that the link should be to https://visualstudio.microsoft.com/license-terms/vs2022-addons-and-extensions/


NikolajBjorner:
added link in README to c-runtime following the pattern of this PR

PR #7597, opened 2025-03-27
Remove vestiges of old ml api

jberdine:
I noticed this old stuff and it seems better to clean it up. It dates from the ocaml api I worked on for the Z3 v3 to v4 transition.


jberdine:

PR #7613, opened 2025-04-06
Fix various race conditions, mostly signal-based

mikulas-patocka:
Hi

I reviewed the source code and found out that there are some (mostly signal-based) race conditions.

The first 3 patches fix race conditions that happen when Ctrl-C is pressed in a multithreaded program (#7603).
The 4th patch fixes a race condition when a multithreaded program calls fork while there are some threads doing calculations inside libz3.
The 5th patch adds a boolean option "ctrl_c" that can be used to disable SIGINT signal handling - it is useful when I want to use the z3 library in a program that does signal handling on its own.

Mikulas

### DIFF
--- a/src/util/cancel_eh.h
+++ b/src/util/cancel_eh.h
@@ -34,15 +34,13 @@ public:
     cancel_eh(T & o): m_obj(o) {}
     ~cancel_eh() override { if (m_canceled) m_obj.dec_cancel(); if (m_auto_cancel) m_obj.auto_cancel(); }
     void operator()(event_handler_caller_t caller_id) override {
-        if (caller_id != CTRL_C_EH_CALLER)
-            signal_lock();
+        signal_lock();
         if (!m_canceled) {
             m_caller_id = caller_id;
             m_canceled = true;
             m_obj.inc_cancel(); 
         }
-        if (caller_id != CTRL_C_EH_CALLER)
-            signal_unlock();
+        signal_unlock();
     }
     bool canceled() {
         bool ret;
--- a/src/util/cancel_eh.h
+++ b/src/util/cancel_eh.h
@@ -18,28 +18,46 @@ Revision History:
 --*/
 #pragma once
 
+#include <atomic>
 #include "util/event_handler.h"
+#include "util/scoped_ctrl_c.h"
 
 /**
    \brief Generic event handler for invoking cancel method.
 */
 template<typename T>
 class cancel_eh : public event_handler {
-    bool m_canceled = false;
+    std::atomic<bool> m_canceled = false;
     bool m_auto_cancel = false;
     T & m_obj;
 public:
     cancel_eh(T & o): m_obj(o) {}
     ~cancel_eh() override { if (m_canceled) m_obj.dec_cancel(); if (m_auto_cancel) m_obj.auto_cancel(); }
     void operator()(event_handler_caller_t caller_id) override {
+        if (caller_id != CTRL_C_EH_CALLER)
+            signal_lock();
         if (!m_canceled) {
             m_caller_id = caller_id;
             m_canceled = true;
             m_obj.inc_cancel(); 
         }
+        if (caller_id != CTRL_C_EH_CALLER)
+            signal_unlock();
     }
-    bool canceled() const { return m_canceled; }
-    void reset() { m_canceled = false; }
+    bool canceled() {
+        bool ret;
+        if (!m_canceled)
+            return false;
+        signal_lock();
+        ret = m_canceled;
+        signal_unlock();
+        return ret;
+    }
+    void reset() {
+        signal_lock();
+        m_canceled = false;
+        signal_unlock();
+    }
     T& t() { return m_obj; }
     void set_auto_cancel() { m_auto_cancel = true; }
 };
--- a/src/util/scoped_ctrl_c.cpp
+++ b/src/util/scoped_ctrl_c.cpp
@@ -20,6 +20,7 @@ Revision History:
 #include <signal.h>
 #include <cstring>
 #include <mutex>
+#include "util/gparams.h"
 #include "util/scoped_ctrl_c.h"
 
 #ifdef _WINDOWS
@@ -106,6 +107,8 @@ scoped_ctrl_c::scoped_ctrl_c(event_handler & eh, bool once, bool enabled):
     m_first(true),
     m_once(once),
     m_enabled(enabled) {
+    if (gparams::get_value("ctrl_c") == "false")
+        m_enabled = false;
     if (m_enabled) {
         signal_lock();
         active_contexts.push_back(this);
--- a/src/util/scoped_ctrl_c.cpp
+++ b/src/util/scoped_ctrl_c.cpp
@@ -36,7 +36,7 @@ static struct sigaction old_sigaction;
 #endif
 static bool signal_handled = false;
 
-static void signal_lock(void) {
+void signal_lock(void) {
 #ifdef USE_SIGNAL
     context_lock.lock();
 #else
@@ -50,7 +50,7 @@ static void signal_lock(void) {
 #endif
 }
 
-static void signal_unlock(void) {
+void signal_unlock(void) {
 #ifdef USE_SIGNAL
     context_lock.unlock();
 #else
--- a/src/util/scoped_ctrl_c.cpp
+++ b/src/util/scoped_ctrl_c.cpp
@@ -12,46 +12,125 @@ Abstract:
 Author:
 
     Leonardo de Moura (leonardo) 2011-04-27.
+    Mikulas Patocka 2025-04-05. (rewritten to be thread safe)
 
 Revision History:
 
 --*/
-#include<signal.h>
+#include <signal.h>
+#include <cstring>
+#include <mutex>
 #include "util/scoped_ctrl_c.h"
 
-static scoped_ctrl_c * g_obj = nullptr;
+#ifdef _WINDOWS
+#define USE_SIGNAL
+#endif
 
-static void on_ctrl_c(int) {
-    if (g_obj->m_first) {
-        g_obj->m_cancel_eh(CTRL_C_EH_CALLER);
-        if (g_obj->m_once) {
-            g_obj->m_first = false;
-            signal(SIGINT, on_ctrl_c); // re-install the handler
-        }
+static std::mutex context_lock;
+static std::vector<scoped_ctrl_c *> active_contexts;
+#ifdef USE_SIGNAL
+static void (*old_handler)(int);
+#else
+static sigset_t context_old_set;
+static struct sigaction old_sigaction;
+#endif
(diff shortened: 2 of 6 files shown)


NikolajBjorner:
thanks, running workflows to check build

NikolajBjorner:
FYI @nunoplopes

nunoplopes:
It's not a very good practice to stuff 5 patches in one PR. Even more so with complicated stuff like this.

mikulas-patocka:
The patches 1, 2, 3 and 5 are dependent, so it's not possible to send them as separate pull requests.

NikolajBjorner:
> 5 patches in one PR.

Generally, correct. Patches with many independent changes end up getting lost because something doesn't work and then all the effort is wasted.

> The patches 1, 2, 3 and 5 are dependent, so it's not possible to send them as separate pull requests.

First impression: correct. signals influence rlimit. If there isn't a way to mutex independently we would get into the direction indicated by the patch.

Overall, updating these files is risky and can affect everything. 
Everything depends on scoped_timer (and ctr-c/ signals).
Mikulas is the first to note how multiple ctr-c from threads would stomp on each-other.

src/util/scoped_timer.cpp
```diff
@@ -122,34 +120,27 @@ scoped_timer::~scoped_timer() {
 
 void scoped_timer::initialize() {
 #ifndef _WINDOWS
-    static bool pthread_atfork_set = false;
-    if (!pthread_atfork_set) {
+    static std::atomic<bool> pthread_atfork_set = false;
```
nunoplopes, line 123: there is no concurrency in initialize()

src/util/scoped_timer.cpp
```diff
@@ -122,34 +120,27 @@ scoped_timer::~scoped_timer() {
...
-        workers.lock();
-        for (auto w : available_workers) {
-            w->work = EXITING;
-            w->cv.notify_one();
-        }
-        decltype(available_workers) cleanup_workers;
-        std::swap(available_workers, cleanup_workers);
-        workers.unlock();
+    workers.lock();
+    for (auto w : available_workers) {
```
nunoplopes, line 132: this simplification seems wrong. The worker thread also locks workers. You need a loop to let the worker thread get the lock and exit.

Also, what is the change that makes finalize() be called once while before it was called many times?

nunoplopes:
Regarding the ctrl-C patches, there can only be more than one active signaling if you are using multiple Z3 contexts in different threads solving multiple formulas at once.
Note that this is not a common usage, hence the patches must not degrade performance for everybody else.

src/util/scoped_ctrl_c.cpp
```diff
@@ -12,46 +12,125 @@ Module Name:
...
+    signal_handled = false;
+}
+
+static void on_sigint(int) {
+    signal_lock();
+#ifdef USE_SIGNAL
+    if (signal_handled)
+        signal(SIGINT, on_sigint);
+#endif
+    for (auto a : active_contexts) {
```
nunoplopes, line 95: this is racy. what if a thread starts a new ctrl-c handler?

src/util/cancel_eh.h
```diff
@@ -18,28 +18,46 @@ Revision History:
...
 #include "util/event_handler.h"
+#include "util/scoped_ctrl_c.h"
 
 /**
    \brief Generic event handler for invoking cancel method.
 */
 template<typename T>
 class cancel_eh : public event_handler {
-    bool m_canceled = false;
+    std::atomic<bool> m_canceled = false;
```
nunoplopes, line 30: should use Z3's atomic<>

src/util/scoped_ctrl_c.cpp
```diff
@@ -12,46 +12,136 @@ Module Name:
...
+static struct sigaction old_sigaction;
+static unsigned signal_lock_depth = 0;
+#endif
+static bool signal_handled = false;
+
+void signal_lock(void) {
+#ifdef USE_SIGNAL
+    context_lock.lock();
+#else
+    sigset_t set, old_set;
```
nunoplopes, line 45: why is all this stuff needed?  Isn't the mutex lock sufficient?

NikolajBjorner:
> Regarding the ctrl-C patches, there can only be more than one active signaling if you are using multiple Z3 contexts in different threads solving multiple formulas at once. Note that this is not a common usage, hence the patches must not degrade performance for everybody else.

Some context: One of the rewrites (by Nuno) for scoped_timer (to avoid spawning fresh threads) was to deal with a big perf bottleneck in the Alive2 tool.

src/util/scoped_timer.cpp
mikulas-patocka: ok, this part of the patch may be dropped

src/util/scoped_ctrl_c.cpp
```diff
@@ -12,46 +12,136 @@ Module Name:
...
+#endif
 
-static void on_ctrl_c(int) {
-    if (g_obj->m_first) {
-        g_obj->m_cancel_eh(CTRL_C_EH_CALLER);
-        if (g_obj->m_once) {
-            g_obj->m_first = false;
-            signal(SIGINT, on_ctrl_c); // re-install the handler
-        }
+static std::recursive_mutex context_lock;
```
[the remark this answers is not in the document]
mikulas-patocka: You can't use z3's mutex here. If z3 is compiled as single-threaded, the z3's mutex is no-operation.

On Windows, signal handlers are ran in a separate thread (even if the program is single-threaded), so you need a real mutex here, even in single-threaded build.

It is recursive because of this stack trace:
        SIGINT signal
        on_sigint - calls signal_lock()
        a->m_cancel_eh()
        cancel_eh::operator()
        m_obj.inc_cancel
        reslimit::inc_cancel - also calls signal_lock()

src/util/cancel_eh.h
mikulas-patocka: You can't use z3's atomic here - because you need atomic even in single-threaded build, because Windows runs signal handlers in a separate thread - see the answer below).

src/util/scoped_ctrl_c.cpp
mikulas-patocka: Suppose that you grab a mutex, receive a signal, and attempt to grab the same mutex again from the signal handler - the result is deadlock (or corruption of the pthread_mutex_t structure).

So you need to block the signal before you attempt to grab the mutex.

On Windows, the SIGINT signal handler is always ran in a separate thread, so a simple mutex is sufficient there.

src/util/scoped_timer.cpp
mikulas-patocka: But the spin-waiting is worse problem than leaving busy workers behind.

This piece of code is being called if any thread does fork() - and in my use-case, it is possible that some thread does fork()+exec() and another thread does some z3 calculation at the same time.

The question is - what should we do here with busy workers? The current code spin-waits, which is incorrect - the result is that fork() is spin-waiting.

So, I removed the spin-wait loop from this function.

nunoplopes: No, we can't leave stuff behind. intialize can be called again after finalize. Alive2 does that several times to cleanup the memory and start fresh.
For long-running processes, we can't have memory leaks or accumulate dead-locked threads.

src/util/scoped_ctrl_c.cpp
nunoplopes: If a 2nd signal is delivered to a separate thread, it just gets blocked on the mutex. It doesn't get deadlocked; eventually the 1st handler unblocks and the 2nd gets a chance to run.

src/util/scoped_timer.cpp
mikulas-patocka: I thought about it and I think it is safe. The idle thread is added to "available_workers" in scoped_timer::~scoped_timer destructor.

If we call the "finalize" method as a part of total memory cleanup, all the scoped_timers' destructors were already executed and all the worker threads are already on "available_workers" vector. So, we don't need to loop; the first loop iteration will clean all the threads.

If the "finalize" method is called from single-threaded program's fork(), then all the scoped timers' destructors are already called and the case is analogous to the previous case.

If the "finalize" method is called from multi-threaded program's fork(), then it breaks down - the "num_workers" variable is the total amount of workers (both sleeping and busy), and we loop until we terminated "num_workers" threads - that means that if the number of sleeping workers is less than "num_workers", the function just spins.

Then, there is unlocked assignment to "num_workers = 0" and "available_workers.clear()" that can race with other threads doing z3 work and corrupt memory. available_workers.clear() is not needed, because it was already cleared by std::swap(available_workers, cleanup_workers) (and that was correctly locked).

If you think that the loop in scoped_timer::finalize is needed, describe some case where it is needed.

I think that the problem came from the fact that the function was originally written to clean up memory when libz3 was cleaned up (it works in this case) and later, it was added to pthread_atfork, causing havoc in multithreaded programs.

nunoplopes: Your argument sounds good, including that the call to 'pthread_atfork' is still present. Hence finalize may get called when a thread is running the solver.
We need to close the timer threads on fork() since otherwise they get deadlocked.

mikulas-patocka:
> > Regarding the ctrl-C patches, there can only be more than one active signaling if you are using multiple Z3 contexts in different threads solving multiple formulas at once. Note that this is not a common usage, hence the patches must not degrade performance for everybody else.
> 
> Some context: One of the rewrites (by Nuno) for scoped_timer (to avoid spawning fresh threads) was to deal with a big perf bottleneck in the Alive2 tool.

I think that the patches won't degrade performance on Windows because they just replace one mutex with another. But they may degrade performance on Posix, because we need to call the syscall to block and unblock the signal when we take and drop the mutex. And a syscall is slower than a mutex.

Do you have some benchmark that I could run to test the performance impact of the patches?

I've got another idea that could be implemented - the Ctrl-C signal handler could just increment a global atomic variable (without taking any locks), and the global atomic variable would be tested by the compute threads. Do you think that this could work? - or is there any problem with it?

Anyway, if it is not possible to accept the Ctrl-C fixes, could you accept at least the patch 4 (a fork race) and patch 5 (make Ctrl-C configurable). If Ctrl-C were configurable, then I would just disable it and the races wouldn't affect me.

nunoplopes:
Can you split the patches into multiple PRs so I can benchmark them?
Locks and atomic variables are expensive when you have dozens of cores. That why we have a single-core version. But since that's not what distros compile, we also have to offer decent performance for the multi-threaded version.

mikulas-patocka:
> Can you split the patches into multiple PRs so I can benchmark them? Locks and atomic variables are expensive when you have dozens of cores. That why we have a single-core version. But since that's not what distros compile, we also have to offer decent performance for the multi-threaded version.

OK, I have created 3 pull requests - #7617, #7618, #7619

NikolajBjorner:
this PR is subsumed by 2-3 other PRs

PR #7614, opened 2025-04-07
include LICENSE.txt in wheels

### LINKED ISSUE #7604: Python wheels do not contain the LICENSE.txt file
To reproduce, simply download a wheel, e.g., z3_solver-4.14.1.0-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl, from https://pypi.org/project/z3-solver/#files and unzip it.  There's no license file.

The license file is however present in the corresponding source distribution, in the core folder.

The [license](https://github.com/Z3Prover/z3?tab=License-1-ov-file#readme) says "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software." so it really should be included in the wheels.

markdryan:
Update setup.py so that we copy LICENSE.TXT to src/api/python before creating the sdist.  Any wheels built from this sdist will now contain the LICENSE.txt file.

Fixes #7604

### DIFF
--- a/src/api/python/setup.py
+++ b/src/api/python/setup.py
@@ -245,6 +245,7 @@ def _copy_sources():
     shutil.rmtree(SRC_DIR_LOCAL, ignore_errors=True)
     os.mkdir(SRC_DIR_LOCAL)
 
+    shutil.copy(os.path.join(SRC_DIR_REPO, 'LICENSE.txt'), ROOT_DIR)
     shutil.copy(os.path.join(SRC_DIR_REPO, 'LICENSE.txt'), SRC_DIR_LOCAL)
     shutil.copy(os.path.join(SRC_DIR_REPO, 'z3.pc.cmake.in'), SRC_DIR_LOCAL)
     shutil.copy(os.path.join(SRC_DIR_REPO, 'CMakeLists.txt'), SRC_DIR_LOCAL)


markdryan:
Note that LICENSE.txt is already include in the sdist, but just not in the right place to get picked up when building the wheels.  setup.py already copies it to the core folder.  It's not clear to me whether it's supposed to be in core or not so I just left the original code as is and added a separate copy to core's parent.

NikolajBjorner:
Thanks, 

I am merging and checking with nightly build. In case it does the job it is good otherwise, I will revert. 
I don't understand what reasoning is behind ROOT_DIR vs LOCAL_SRC_DIR, as the update only uses ROOT_DIR for LICENSE.txt. @rhelmot

rhelmot:
the difference between ROOT_DIR and LOCAL_SRC_DIR is that ROOT_DIR is the folder that setup.py lives and LOCAL_SRC_DIR is the folder that CMakeLists.txt lives in (specifically, LOCAL_SRC_DIR is the subset of the whole z3 src tree that is copied into the python bindings dir for the purpose of isolated builds).

This is probably a correct change, but I'm not sure if it completes the job of making sure that the license shows up in the binary distributions. Do we still use the configuration where we assemble a wheel from the canonical z3 build? iirc no, and this should solve the problem in the case I remember building which builds z3 in the manylinux docker container, but just covering my bases.

NikolajBjorner:
I checked, the license does show up.

PR #7617, opened 2025-04-09
Make Ctrl-C handling thread-safe (#7603)

mikulas-patocka:
The Ctrl-C handling is not thread safe, there's a global variable g_obj that is being accessed without any locking. The signal handlers are per-process, not per-thread, so that different threads step over each other's handlers. It is unpredictable in which thread the signal handler runs, so the handler may race with the scoped_ctrl_c destructor.

Fix this by introducing the functions signal_lock and signal_unlock. signal_lock blocks the SIGINT signal and then takes a mutex (so that the signal handler can't be called while the mutex is held). signal_unlock drops the mutex and restores the signal mask.

We protect all the global variables with signal_lock and signal_unlock.

Note that on Windows, the SIGINT handler is being run in a separate thread (and there is no way how to block it), so we can use a simple mutex to synchronize the signal handler with the other threads.

In class cancel_eh, the operator () may be called concurrently by the timer code and the Ctrl-C code, but the operator () accesses class' members without any locking. Fix this race condition by using the functions signal_lock() and signal_unlock().

There is this possible call trace:
        SIGINT signal
        on_sigint
        a->m_cancel_eh()
        cancel_eh::operator()
        m_obj.inc_cancel
        reslimit::inc_cancel
        lock_guard lock(*g_rlimit_mux);

Here we take a mutex from a signal - this is subject to deadlock (if the signal interrupted another piece of code where the mutex is already held).

To fix this race, we remove g_rlimit_mux and replace it with signal_lock() and signal_unlock(). signal_lock and signal_unlock block the signal before grabbing the mutex, so the signal can't interrupt a piece of code where the mutex is held and the deadlock won't happen.


NikolajBjorner:
I am not following:
> (if the signal interrupted another piece of code where the mutex is already held).
The g_rlimit_mux is used exclusively inside the rlimit class and for touching the m_cancel flag + updating children.
What action within rlimit would cause a signal? The main unsafe operation I see is memory allocation which could page fault and even fail.

mikulas-patocka:
> I am not following:
> 
> > (if the signal interrupted another piece of code where the mutex is already held).
> 
> The g_rlimit_mux is used exclusively inside the rlimit class and for touching the m_cancel flag + updating children. What action within rlimit would cause a signal? The main unsafe operation I see is memory allocation which could page fault and even fail.

The SIGINT signal could interrupt it. According to the POSIX specification, when the user presses Ctrl+C, the SIGINT signal may be delivered to any thread in the process (the only exception is that the signal couldn't be delivered to a thread that has the signal blocked).

So, we grab a mutex, execute some code in the rlimit class with the mutex held, the user presses Ctrl+C, the signal handler is delivered to the thread that already holds the mutex, the signal handler attempts to grab the mutex again and it deadlocks.

In order to avoid this deadlock possibility, we need to block the SIGINT signal before we grab the mutex.

It's a similar pattern to spinlocks and interrupt masking inside the kernel. If we grab a spinlock and the spinlock may be grabbed by an interrupt, there is deadlock risk. In order to avoid this deadlock, we must block interrupts before grabbing the spinlock and re-enable interrupts after releasing the spinlock.

src/util/scoped_ctrl_c.cpp
```diff
@@ -12,49 +12,137 @@ Module Name:
...
 
-static void on_ctrl_c(int) {
-    if (g_obj->m_first) {
-        g_obj->m_cancel_eh(CTRL_C_EH_CALLER);
-        if (g_obj->m_once) {
-            g_obj->m_first = false;
-            signal(SIGINT, on_ctrl_c); // re-install the handler
-        }
+static std::recursive_mutex context_lock;
+static std::vector<scoped_ctrl_c *> active_contexts;
```
NikolajBjorner, line 31: Note: 
ptr_vector will not work across resets.
We would need to initialize / finalize memory if any of this global state goes over z3's memory_manager. 
In principle, I prefer to ensure all heap traffic goes over this. 
In practice, anything std::string will use the native allocator.

src/util/cancel_eh.h
```diff
@@ -18,28 +18,44 @@ Revision History:
...
     cancel_eh(T & o): m_obj(o) {}
     ~cancel_eh() override { if (m_canceled) m_obj.dec_cancel(); if (m_auto_cancel) m_obj.auto_cancel(); }
     void operator()(event_handler_caller_t caller_id) override {
+        signal_lock();
         if (!m_canceled) {
             m_caller_id = caller_id;
             m_canceled = true;
             m_obj.inc_cancel(); 
         }
+        signal_unlock();
```
NikolajBjorner, line 43: this is a bit peculiar. When m_canceled is true you force an interrupt by using signal_lock, but I don't see how it constrains behaviors.

NikolajBjorner:
Nuno, I am inclined to merge pending an understanding to my question on why there is a lock being taken in cancel_eh.
Perf-wise, I take this code isn't on any critical path so doesn't affect Alive2 or other use cases (other than fixing thread safety).

nunoplopes:
Please give me a couple of days to measure perf & read the patch. I'm under a bit under water, sorry..

nunoplopes:
Doesn't compile:
```
../src/util/scoped_ctrl_c.cpp: In constructor ‘scoped_ctrl_c::scoped_ctrl_c(event_handler&, bool, bool)’:
../src/util/scoped_ctrl_c.cpp:110:5: error: class ‘scoped_ctrl_c’ does not have any field named ‘m_old_scoped_ctrl_c’
  110 |     m_old_scoped_ctrl_c(g_obj) {
      |     ^~~~~~~~~~~~~~~~~~~
src/math/polynomial/rpolynomial.cpp
../src/util/scoped_ctrl_c.cpp:110:25: error: ‘g_obj’ was not declared in this scope
  110 |     m_old_scoped_ctrl_c(g_obj) {
      |                         ^~~~~
```

NikolajBjorner:
@nunoplopes - this would just be a stale field from a merge. Would you be able to check it?

src/util/cancel_eh.h
NikolajBjorner: @mikulas-patocka - I still have this question

NikolajBjorner:
see https://github.com/Z3Prover/z3/pull/7755

PR #7657, opened 2025-05-21
Centralize and document TRACE tags using X-macros

IamYJLee:
- Introduce X-macro-based trace tag definition
- Add script to generate Markdown documentation from trace_tags.def

### DIFF
--- a/scripts/generate_trace_tag_docs.py
+++ b/scripts/generate_trace_tag_docs.py
@@ -1,0 +1,35 @@
+# generate_trace_tag_docs.py
+# This script generates a Markdown file from the trace_tags.def file.
+# It extracts the trace tags and their descriptions and formats them into a table.
+# The generated Markdown file can be used for documentation purposes.
+
+# Trace Tags Documentation
+
+# | Tag | Description |
+# |-----|-------------|
+# | `mam_int` | mam initialization |
+# | `mam_test` | mam test |
+# | `propagate` | propagation |
+
+import re
+
+def_file = "./src/util/trace_tags.def"
+output_md = "./trace_tags.md"
+
+with open(def_file, "r") as f:
+    lines = f.readlines()
+
+entries = []
+for line in lines:
+    match = re.match(r'X\(\s*(\w+)\s*,\s*"([^"]+)"\s*\)', line)
+    if match:
+        tag, desc = match.groups()
+        entries.append((tag, desc))
+
+with open(output_md, "w") as f:
+    f.write("# Trace Tags Documentation\n\n")
+    f.write("| Tag | Description |\n")
+    f.write("|-----|-------------|\n")
+    for tag, desc in entries:
+        f.write(f"| `{tag}` | {desc} |\n")
+
--- a/src/smt/smt_context.cpp
+++ b/src/smt/smt_context.cpp
@@ -1702,6 +1702,7 @@ namespace smt {
      */
     bool context::propagate() {
         TRACE("propagate", tout << "propagating... " << m_qhead << ":" << m_assigned_literals.size() << "\n";);
+        TRACE_NEW(TraceTag::propagate, tout << "propagating... " << m_qhead << ":" << m_assigned_literals.size() << "\n";);
         while (true) {
             if (inconsistent())
                 return false;
--- a/src/util/trace.h
+++ b/src/util/trace.h
@@ -20,6 +20,7 @@ Revision History:
 #pragma once
 
 #include<fstream>
+#include "util/trace_tags.h"
 
 #ifdef SINGLE_THREAD
 # define is_threaded() false
@@ -81,6 +82,7 @@ static inline void finalize_trace() {}
 #define STRACEBODY(CODE) CODE; tout.flush()
 
 #define TRACE(TAG, CODE) TRACE_CODE(if (is_trace_enabled(TAG)) { THREAD_LOCK(TRACEBODY(TAG, CODE)); })
+#define TRACE_NEW(TAG, CODE) TRACE_CODE(if (is_trace_enabled(to_string(TAG))) { THREAD_LOCK(TRACEBODY(to_string(TAG), CODE)); })
 
 #define STRACE(TAG, CODE) TRACE_CODE(if (is_trace_enabled(TAG)) { THREAD_LOCK(STRACEBODY(CODE)); })
 
--- a/src/util/trace_tags.def
+++ b/src/util/trace_tags.def
@@ -1,0 +1,3 @@
+X(mam_int, "mam initialization")
+X(mam_test, "mam test")
+X(propagate, "propagation")
--- a/src/util/trace_tags.h
+++ b/src/util/trace_tags.h
@@ -1,0 +1,43 @@
+#pragma once
+
+enum class TraceTag {
+#define X(tag, desc) tag,
+#include "trace_tags.def"
+#undef X
+    Count
+};
+
+// Convert TraceTag to string
+inline const char* to_string(TraceTag tag) {
+    switch (tag) {
+#define X(tag, desc) case TraceTag::tag: return #tag;
+#include "trace_tags.def"
+#undef X
+    default: return "Unknown";
+    }
+}
+
+// Return description of TraceTag
+inline const char* get_trace_tag_doc(TraceTag tag) {
+    switch (tag) {
+#define X(tag, desc) case TraceTag::tag: return desc;
+#include "trace_tags.def"
+#undef X
+    default: return "Unknown tag";
+    }
+}
+
+// Return the number of TraceTags
+inline constexpr int trace_tag_count() {
+    return static_cast<int>(TraceTag::Count);
+}
+
+// Return all TraceTags as an array
+inline const TraceTag* all_trace_tags() {
+    static TraceTag tags[] = {
+#define X(tag, desc) TraceTag::tag,
+#include "trace_tags.def"
+#undef X
+    };
+    return tags;
+}


src/util/trace_tags.h
```diff
@@ -0,0 +1,43 @@
...
+#include "trace_tags.def"
+#undef X
+    default: return "Unknown";
+    }
+}
+
+// Return description of TraceTag
+inline const char* get_trace_tag_doc(TraceTag tag) {
+    switch (tag) {
+#define X(tag, desc) case TraceTag::tag: return desc;
```
NikolajBjorner, line 23: you can use #tag instead of desc to retrieve the string version of the enum.

src/util/trace_tags.def
```diff
@@ -0,0 +1,3 @@
+X(mam_int, "mam initialization")
+X(mam_test, "mam test")
+X(propagate, "propagation")
```
NikolajBjorner, line 3: for simplicity: we should just have one field.

scripts/generate_trace_tag_docs.py
```diff
@@ -0,0 +1,35 @@
+# generate_trace_tag_docs.py
```
NikolajBjorner, line 1: something calls this script, e.g., called from doc/mk_api_doc.py to include in api doc package?

src/smt/smt_context.cpp
```diff
@@ -1702,6 +1702,7 @@ namespace smt {
      */
     bool context::propagate() {
         TRACE("propagate", tout << "propagating... " << m_qhead << ":" << m_assigned_literals.size() << "\n";);
+        TRACE_NEW(TraceTag::propagate, tout << "propagating... " << m_qhead << ":" << m_assigned_literals.size() << "\n";);
```
NikolajBjorner, line 1705: Replacing TRACE by TRACE_NEW and populating trace_tags.def is definitely scriptable. 
Based on the original github issue is there any "module" information you wanted to add?

src/util/trace.h
```diff
@@ -81,6 +82,7 @@ static inline void finalize_trace() {}
 #define STRACEBODY(CODE) CODE; tout.flush()
 
 #define TRACE(TAG, CODE) TRACE_CODE(if (is_trace_enabled(TAG)) { THREAD_LOCK(TRACEBODY(TAG, CODE)); })
+#define TRACE_NEW(TAG, CODE) TRACE_CODE(if (is_trace_enabled(to_string(TAG))) { THREAD_LOCK(TRACEBODY(to_string(TAG), CODE)); })
```
NikolajBjorner, line 85: Can't the TRACE macro itself pre-pend the class name TraceTag to eliminate common clutter from uses?

src/smt/smt_context.cpp
IamYJLee: Yes — I’d like to categorize all TRACE tags, starting with:

- SMT Internalization: tracking e-graph construction
- Congruence Closure Core: focusing on quantifier instantiation

The goal is to build a structured view of all trace points across the solver.

NikolajBjorner: So what is then the plan to achieve this based on C macros?
Have a third column with a class (which is also defined in the enum column) and create a map from class to members such that when the class is traced and TRACE is hit for a member, the trace is invoked?

scripts/generate_trace_tag_docs.py
IamYJLee: Currently, the script is intended to be run manually and is not yet hooked into doc/mk_api_doc.py.
Integration into the API doc pipeline can be considered in the future if that makes sense.

NikolajBjorner: right, first this script has to be pushed :-)

src/util/trace.h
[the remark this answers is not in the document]
IamYJLee: I've updated the code as suggested — `is_trace_enabled` accepts the enum directly.

The changes are included in commit cac9782.  
Please take a look and let me know if anything else needs refinement!

IamYJLee: I've updated the code as suggested — `TRACE_NEW` now prepends TraceTag.

The changes are included in commit cac9782.  
Please take a look and let me know if anything else needs refinement!

nunoplopes:
Stepping back a little, who needs to know about the tracing mechanism besides the Z3 developers?  Probably no one.

Also, a simple grep (e.g.`grep -RoP --no-filename 'TRACE\s*\("\K[^"]+' src`) is sufficient to gather all trace tags from the code automatically.

src/smt/smt_context.cpp
NikolajBjorner: I believe the simpler and also usable approach would be if you use a single name-space for tag_name and tag-class-name and introduce a hierarchy by making X be of the form
X(tag_name, tag_class_name, "description")
X(tag_class_name, tag_class_name, "description")
So all classes are also tags and we can either take the 1step closure or transitive closure if that turns out to be more useful.
There would be no extra -tr_class option either.

I would merge this PR first and implement tags everywhere so rename TRACE_NEW to TRACE.
Porting all other tags is scriptable by assigning every tag to its own class by default.
It may be an AI query hack to refactor into classes, except the code base is not a friend of small context windows.

IamYJLee:
> Stepping back a little, who needs to know about the tracing mechanism besides the Z3 developers? Probably no one.
> 
> Also, a simple grep (e.g.`grep -RoP --no-filename 'TRACE\s*\("\K[^"]+' src`) is sufficient to gather all trace tags from the code automatically.

While it's true that we can extract trace tags using `grep`, I believe having a structured and documented list — especially with class-based organization — can be meaningful not just for Z3 developers, but also for users or researchers like myself who aim to deeply understand how input conditions are processed internally.

Even though TRACE is primarily a debugging mechanism, exposing it in a more systematic and categorized way could help advanced users trace how specific components like `E-matching`, `internalization`, or `congruence closure` are triggered, which is valuable for learning, debugging, and contributing effectively.

NikolajBjorner:
FWIW, a simple AI prompt created basically this.
Modify/reuse for creating file with X.
Extra credit: sort entries.

```
import os
import re

# Initialize an empty set to store unique tags
unique_tags = set()

# Define the directory to be walked
directory = 'src'

# Walk through all files in the directory recursively
for root, dirs, files in os.walk(directory):
    files = [f for f in files if f.endswith(".cpp") or f.endswith(".h")]
    for file in files:
        file_path = os.path.join(root, file)
        print(file_path)
        
        # Read the content of the file
        with open(file_path, 'r') as f:
            content = f.read()
        
        # Find all occurrences of TRACE("sometag"
        matches = re.findall(r'TRACE\("([^"]+)"', content)
        
        # Replace occurrences of TRACE("sometag" with TRACE(sometag)
        new_content = re.sub(r'TRACE\("([^"]+)"', r'TRACE(\1', content)
        
        # Write the modified content back to the file
        with open(file_path, 'w') as f:
            f.write(new_content)
        
        # Add found tags to the set
        unique_tags.update(matches)

# Print out the accumulated tags in the specified format
for i, tag in enumerate(unique_tags):
    print(f'X({tag}, "{tag}")')
```

IamYJLee:
@NikolajBjorner 
Please review the overall changes in this PR.
`trace_tags.def` includes all current tags using the X(tag, tag_class, description) format, but class names and descriptions are still placeholders and will be refined later.

src/util/trace_tags.h
```diff
@@ -0,0 +1,81 @@
...
+    
+    return class_tags;
+}
+
+// Find TraceTag by string
+inline TraceTag find_trace_tag_by_string(const char* tag_str) {
+    #define X(tag, class, desc) if (strncmp(#tag, tag_str, strlen(#tag)) == 0) return TraceTag::tag;
+    #include "trace_tags.def"
+    #undef X
+    return TraceTag::Count;  // 매칭되는 태그가 없는 경우
```
NikolajBjorner, line 80: remove unicode comment

src/util/trace_tags.h
```diff
@@ -0,0 +1,81 @@
+#pragma once
+
+#include <cstring>
+
+enum class TraceTag {
+#define X(tag, class, desc) tag,
```
NikolajBjorner, line 6: to be safer, avoid identifier "class". I run into compiler incompatibilities a lot.

src/util/trace_tags.h
```diff
@@ -0,0 +1,81 @@
...
+    #include "trace_tags.def"
+    #undef X
+    return count;
+}
+
+// Return all tags that belong to the given class
+inline const TraceTag* get_tags_by_class(TraceTag cls, int& count) {
+    count = count_tags_in_class(cls);
+    static TraceTag* class_tags = nullptr;
+    if (class_tags) delete[] class_tags;
```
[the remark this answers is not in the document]
nunoplopes: Or maybe just get rid of this class altogether?

If this is to keep, you can just define an iterator that generates the result on-the-fly.

NikolajBjorner:
Suggestuin is to defer implementing this feature in first pr. Then merging will be quicker. Milestone 1 is using enums.

src/util/trace_tags.h
IamYJLee: Fixed. PTAL.

src/util/trace_tags.h
IamYJLee: Fixed. PTAL.

src/util/trace_tags.h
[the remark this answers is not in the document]
IamYJLee: @NikolajBjorner @nunoplopes 
I've created a follow-up issue to track the implementation. (https://github.com/Z3Prover/z3/issues/7663)
I'd appreciate any feedback or suggestions on the proposed implementation direction in that issue!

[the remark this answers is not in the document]
NikolajBjorner: could you disable or remove this for the current PR?
Then I can merge what there is. With the code below exposed, someone will just start filing a gotcha fuzz bug that there is a memory leak.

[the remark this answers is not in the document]
IamYJLee: Disabled the code per your comment. PTAL.

PR #7660, opened 2025-05-26
Bazel: expose `z3_static` target for working macOS, win32 builds

tsnl:
The current Bazel build files introduced in #7646 do not work on macOS or Windows. The current implementation expects `libz3.so` to be produced by the CMake build. However, on macOS and Windows, we produce `libz3.dylib` and `z3.dll` respectively.

Updating the Bazel build file to output `libz3.dylib` on macOS and `z3.dll` on Windows reveals more issues. On macOS, exposing `libz3.dylib` is insufficient because running executables linked against this dylib causes a runtime error, complaining that `libz3<version>.dylib` is not found. (**TODO:** report test results on Windows)

One solution is to specify both `libz3.dylib` and `libz3<version>.dylib` in the list of output shared libraries exposed by the `cmake` rule in the Bazel build. However, Bazel [cannot depend on the contents of a VERSION file to determine CMake's outputs](https://stackoverflow.com/questions/56007307/read-local-file-content-in-bazel-build-system), meaning we'd have to hard-code the current Z3 version in the Bazel file and keep it in-sync with the rest of the project. Brittle.

A simpler workaround is to also expose the statically-linked libraries also produced by Z3. This is what I've done in this PR.

---

This PR adds the `z3_static` Bazel target to the `BUILD.bazel` file. This allows consumption of Z3 as a static library using the Bazel build system. It also unblocks users on macOS and Windows until the default `z3_dynamic` target can be fixed on those platforms.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,6 +2,13 @@
 common --enable_bzlmod
 common --noenable_workspace
 
+# Specifies...
+#   --config=windows on Windows hosts
+#   --config=linux on Linux hosts
+#   --config=macos on macOS hosts
+# NOTE: We assume our host and target platforms are identical.
+common --enable_platform_specific_config
+
 # Use C++20.
 build --cxxopt=-std=c++20
 build --host_cxxopt=-std=c++20
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -17,10 +17,42 @@ filegroup(
 )
 
 cmake(
-    name = "z3",
-    generate_args = ["-G Ninja"],
+    name = "z3_dynamic",
+    generate_args = [
+        "-G Ninja", 
+        "-D Z3_EXPORTED_TARGETS=",  # prevents installation, leaving symlinks between dylibs intact on copy
+    ],
     lib_source = ":all_files",
     out_binaries = ["z3"],
-    out_shared_libs = ["libz3.so"],
+    out_shared_libs = select({
+        "@platforms//os:linux": ["libz3.so"],
+        # "@platforms//os:osx": ["libz3.dylib"],    # FIXME: this is not working, libz3<version>.dylib is not copied
+        # "@platforms//os:windows": ["z3.dll"],     # TODO: test this
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+    visibility = ["//visibility:public"],
+)
+
+cmake(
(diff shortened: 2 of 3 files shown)


NikolajBjorner:
Is MODULE.bazel.lock added as intended or by accident?

NikolajBjorner:
> However, Bazel [cannot depend on the contents of a VERSION file to determine CMake's outputs](https://stackoverflow.com/questions/56007307/read-local-file-content-in-bazel-build-system), meaning we'd have to hard-code the current Z3 version in the Bazel file and keep it in-sync with the rest of the project. Brittle.

Indeed. It already has been bit-rotting to 4.14.0 while current version is 4.15.1.
I update z3's version manually in four different files at the moment. (release.yml, nightly.yml, cmakelists.txt and mk_project.py)
This step can/should be scripted and/or there is something better to do with the scripts that could pick it up from CMakeLists.txt.

tsnl:
I've removed the MODULE.bazel.lock file. This file can checked in once the Bazel build files stabilize.

@NikolajBjorner, if you can tolerate updating a fifth file when there's a version change, I can update the PR to fix the dynamically-linked targets on macOS. The symmetry between BUILD.bazel and CMakeLists.txt should reduce the cognitive load of remembering to update these files.

Are there any other changes you'd like me to make? I'm happy to take a stab at adding some CI targets for the Bazel toolchain to avoid similar issues cropping up again.

One possible way to automate keeping the versions in sync would be to specify the version when configuring the build. CMake could read the `Z3_VERSION` variable as a configuration option when generating build files (e.g. `cmake .. -GNinja -DZ3_VERSION=4.15.1`). We'd need to generate the BUILD.bazel file from a template or by patching it before building, but this shouldn't be too bad, especially with proper tests.

NikolajBjorner:
Thanks, taking the current improvements.
For Bazel, decisions on what to support has to be made by the contributors. 
There is no dependency on this for building releases.

smolkaj:
Nice!! Thanks for adding support for Mac and Windows, @tsnl.

Sorry for the silence, I just returned from parental leave.

smolkaj:
On a related note -- I'd like to get Z3 added to the [Bazel Central Registry](https://registry.bazel.build/) eventually. Though I currently don't have cycles to take up this work myself.

PR #7672, opened 2025-06-04
Fix library header padding for `install_name_tool` to be able to rewrite paths on OSX

### LINKED ISSUE #7623: Installation with opam fails on MacOS if switch name is too long
I tried installing z3 (version 4.14.1) via the opam package manager. I am on MacOS 15.4 + arm64.
 The opam switch that I am in has a name that is 13 characters long. Installation fails due to issues with `install-name-tool`. 
```
# install_name_tool -id $(ocamlfind printconf destdir)/stublibs/libz3.dylib libz3.dylib
# error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: changing install names or rpaths can't be redone for: libz3.dylib (for architecture arm64) because larger updated load commands do not fit (the program must be relinked, and you may need to use -headerpad or -headerpad_max_install_names)
```
If I try installing z3 with a shorter switch name then it succeeds. Probably the issue is caused due to some changes in the new MacOS releases.
On the earlier MacOS versions, z3 got installed via opam with no issues even on switches with longer names.

wintersteiger:
Fixes #7623

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -2077,6 +2077,9 @@ class MLComponent(Component):
                 # Some ocamlmklib's don't like -g; observed on cygwin, but may be others as well.
                 OCAMLMKLIB += ' -g'
 
+            if IS_OSX:
+                OCAMLMKLIB += ' -ldopt -Wl,-headerpad_max_install_names'
+
             z3mls = os.path.join(self.sub_dir, 'z3ml')
 
             LIBZ3ML = ''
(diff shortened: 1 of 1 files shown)


giltho:
Hello, any chance this could warrant a new release (e.g. 4.15.1), so it can then make its way into the opam repository? :)

wintersteiger:
https://github.com/ocaml/opam-repository/pull/28015

PR #7711, opened 2025-07-01
collect statistics from theories on a timeout, or ctrl-C

levnach:


NikolajBjorner:
97193b4a1d623a4c6cd57c0fde779a77b2061ae1

NikolajBjorner:
There is a race condition in smt_tactic.cpp: m_ctx can be deleted during collect_statistics. 
This is fixable with a mutex which should only be activated if m_ctx is tested as non-null.

PR #7713, opened 2025-07-02
trace : Classify tag_names unique to smt_internalize.cpp

IamYJLee:
I’ve grouped the `tag_names` used only in `smt_internalize.cpp` under the `smt_internalize` tag_class.
For `tag_names` that appear across multiple files such as `smt_context.cpp`, `smt_clause_proof.cpp`, and `smt_internalize.cpp`, further classification is needed.
I’d appreciate your advice on whether we should consider renaming `tag_names` that are used in multiple places.

For example, 
https://github.com/Z3Prover/z3/blob/8de80e666b1dd3b6e55cf3ff26e71073a7899685/src/smt/smt_internalizer.cpp#L620

https://github.com/Z3Prover/z3/blob/8de80e666b1dd3b6e55cf3ff26e71073a7899685/src/smt/smt_clause_proof.cpp#L240

https://github.com/Z3Prover/z3/blob/8de80e666b1dd3b6e55cf3ff26e71073a7899685/src/smt/smt_consequences.cpp#L55

https://github.com/Z3Prover/z3/blob/8de80e666b1dd3b6e55cf3ff26e71073a7899685/src/smt/smt_context.cpp#L2433


IamYJLee:
@microsoft-github-policy-service agree company="NSHC"

src/util/trace_tags.def
```diff
@@ -63,8 +63,32 @@ X(get_uninterp_proc, get_uninterp_proc, "get uninterp proc")
 X(berlekamp_matrix, berlekamp_matrix, "berlekamp matrix")
 X(pob, pob, "pob")
 X(theory_str, theory_str, "theory str")
+X(smt_internalize, smt_internalize, "smt internalize")
```
NikolajBjorner, line 66: we should keep these alphabetical. 
Otherwise, small updates become chaotic.
The python script that generates trace_tags_def can sort.

NikolajBjorner: maybe change the convention for X to use the group first, then the tag.
Then alphabetical order is stable relative to group.

IamYJLee: what do you think about sorting them like this:
```
X(A_tag_class, A_tag_class ...)
X(A_tag_name, A_tag_class ...)
X(B_tag_name, A_tag_class ...)

X(B_tag_class, B_tag_class ...)
X(A_tag_name, B_tag_class ...)
X(B_tag_name, B_tag_class ...)
```

NikolajBjorner:
> I’ve grouped the `tag_names` used only in `smt_internalize.cpp` under the `smt_internalize` tag_class. For `tag_names` that appear across multiple files such as `smt_context.cpp`, `smt_clause_proof.cpp`, and `smt_internalize.cpp`, further classification is needed. I’d appreciate your advice on whether we should consider renaming `tag_names` that are used in multiple places.
> 
> For example,
> 
> https://github.com/Z3Prover/z3/blob/8de80e666b1dd3b6e55cf3ff26e71073a7899685/src/smt/smt_internalizer.cpp#L620
> 
> https://github.com/Z3Prover/z3/blob/8de80e666b1dd3b6e55cf3ff26e71073a7899685/src/smt/smt_clause_proof.cpp#L240
> 
> https://github.com/Z3Prover/z3/blob/8de80e666b1dd3b6e55cf3ff26e71073a7899685/src/smt/smt_consequences.cpp#L55
> 
> https://github.com/Z3Prover/z3/blob/8de80e666b1dd3b6e55cf3ff26e71073a7899685/src/smt/smt_context.cpp#L2433

It could be a group but then having too many specialized trace tags doesnt produce really clear value to me.
So fallback to trace on operations relevant for context can cover good scenarios. 

That said, sometimes when debugging conflict resolution I end up shuffling tag names.

src/util/trace_tags.def
NikolajBjorner: Here is what i described:
```
X(A_tag_class, A_tag_class ...)
X(A_tag_class, A1_tag_name ...)
X(A_tag_class, A2_tag_name ...)

X(B_tag_class, B_tag_class ...)
X(B_tag_class, B1_tag_name ...)
X(B_tag_cass, B2_tag_name ...)

PR #7717, opened 2025-07-02
make sure that we do no access a null m_ctx

levnach:
@NikolajBjorner , does it make sense? just checking with an std::atomic flag that m_ctx is not being destroyed before accessing it.


NikolajBjorner:
this will not work

PR #7721, opened 2025-07-03
another try to avoid the race with atomic swap

levnach:


src/smt/tactic/smt_tactic_core.cpp
```diff
@@ -101,10 +103,26 @@ class smt_tactic : public tactic {
...
+            // If successful, we own the context and must deallocate it
+            if (m_ctx.compare_exchange_strong(ctx, nullptr)) {
+                try {
+                    ctx->collect_statistics(st);
+                } catch (...) {
+                    // If exception occurs, fall back to cached stats
+                    st.copy(m_stats);
+                }
+                // We won the race, so we're responsible for cleanup
+                dealloc(ctx);
```
NikolajBjorner, line 120: this is wrong. The context should only be deallocated after search has completed.

levnach: maybe just restore the behavior of collect_statistics that worked on m_ctx directly and catch an exception if thrown?

NikolajBjorner:
this is a bit far from a good fix, so rather close this PR

PR #7755, opened 2025-08-03
fix #7603: race condition in Ctrl-C handling

### LINKED ISSUE #7603: Crashes in a multithreaded program on Ctrl-C

nunoplopes:
This implementation is much simpler than #7617.
When receiving a signal, it cancels all ongoing scoped handlers. The cancel EH must become atomic so there are no races between the signal thread and the canceled thread.

### DIFF
--- a/src/api/api_ast.cpp
+++ b/src/api/api_ast.cpp
@@ -805,7 +805,7 @@ extern "C" {
         cancel_eh<reslimit> eh(m.limit());
         api::context::set_interruptable si(*(mk_c(c)), eh);
         {
-            scoped_ctrl_c ctrlc(eh, false, use_ctrl_c);
+            scoped_ctrl_c ctrlc(eh, use_ctrl_c);
             scoped_timer timer(timeout, &eh);
             try {
                 m_rw(a, result);
--- a/src/api/api_datalog.cpp
+++ b/src/api/api_datalog.cpp
@@ -287,7 +287,7 @@ extern "C" {
             cancel_eh<reslimit> eh(mk_c(c)->m().limit());
             api::context::set_interruptable si(*(mk_c(c)), eh);        
             scoped_timer timer(timeout, &eh);
-            scoped_ctrl_c ctrlc(eh, false, use_ctrl_c);
+            scoped_ctrl_c ctrlc(eh, use_ctrl_c);
             try {
                 r = to_fixedpoint_ref(d)->ctx().query(to_expr(q));
             }
--- a/src/api/api_opt.cpp
+++ b/src/api/api_opt.cpp
@@ -154,7 +154,7 @@ extern "C" {
         bool     use_ctrl_c  = to_optimize_ptr(o)->get_params().get_bool("ctrl_c", true);
         api::context::set_interruptable si(*(mk_c(c)), eh);        
         {
-            scoped_ctrl_c ctrlc(eh, false, use_ctrl_c);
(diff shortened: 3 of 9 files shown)


NikolajBjorner:
was "once" always false or not used?

NikolajBjorner:
these appear to be the occurrences with once semantics:

cmd_context\cmd_context.cpp:#include "util/scoped_ctrl_c.h"
cmd_context\cmd_context.cpp:    scoped_ctrl_c ctrlc(eh);
cmd_context\cmd_context.cpp:        scoped_ctrl_c ctrlc(eh);
cmd_context\cmd_context.cpp:        scoped_ctrl_c ctrlc(eh);
cmd_context\cmd_context.cpp:    scoped_ctrl_c ctrlc(eh);
cmd_context\cmd_context.cpp:    scoped_ctrl_c ctrlc(eh);
cmd_context\cmd_context.cpp:        scoped_ctrl_c ctrlc(eh);
cmd_context\eval_cmd.cpp:#include "util/scoped_ctrl_c.h"
cmd_context\eval_cmd.cpp:            scoped_ctrl_c ctrlc(eh);
cmd_context\extra_cmds\polynomial_cmds.cpp:#include "util/scoped_ctrl_c.h"
cmd_context\simplify_cmd.cpp:#include "util/scoped_ctrl_c.h"
cmd_context\simplify_cmd.cpp:            scoped_ctrl_c ctrlc(eh);
cmd_context\tactic_cmds.cpp:#include "util/scoped_ctrl_c.h"
cmd_context\tactic_cmds.cpp:                scoped_ctrl_c ctrlc(eh);
cmd_context\tactic_cmds.cpp:                scoped_ctrl_c ctrlc(eh);
muz\fp\dl_cmds.cpp:#include "util/scoped_ctrl_c.h"
muz\fp\dl_cmds.cpp:            scoped_ctrl_c ctrlc(eh);


So basically from the command-line.
It has the effect that if the first ctrl-c doesn't finish the job the second ctrl-c will invoke a heavier handed interrupt.
If you remove the "once" argument, what is the new semantics?

nunoplopes:
I tested with some long running files and the behavior is the same as before: you press ctrl-c once and you get a soft resource limit. You press again and it quits straight away.
The new code is simpler because it sets the handler just once and records the previous handler (which is the default exit). Once our handler is called, it restores the previous (default) one.

I also did a perf test and it's neutral.

PR #7761, opened 2025-08-07
4015

levnach:
I fixed some of the ambiguous calls with the pattern foo(bar(...), ..., bor(...)) causing different parameter evaluation order depending on the compiler. I have not fixed all such places. 
Now z3 timeouts on 4015.smt2 on all platforms, I believe :-) At least it is a consistent behavior.


levnach:
I added an option to regressions/smt2/4015.smt2 to use a random seed. The example diverges on many seeds - it seems an investigation is in order.

PR #7768, opened 2025-08-08
Finest factors

levnach:
Using only irreducible polynomials in the default projection of lsat.

### DIFF
--- a/src/nlsat/nlsat_explain.cpp
+++ b/src/nlsat/nlsat_explain.cpp
@@ -593,7 +593,7 @@ namespace nlsat {
         /**
            \brief Add factors of p to todo
         */
-        void add_factors(polynomial_ref & p) {
+        void insert_fresh_factors_in_todo(polynomial_ref & p) {
             if (is_const(p))
                 return;
             elim_vanishing(p);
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
you added a unit test that appears to show that the change breaks your square free fix.

levnach:
The change in the pull request was to replace each polynomial participating in project() by the set of its irreducible factors - these factors define the signs of the original polynomial on the cell. The square free check became obsolete. As the bonus, after the change, z3 started processing the benchmark from issue 7489 under 20 seconds, so I added the benchmark to z3test.

NikolajBjorner:
It breaks the regression tests?

Z3  build/z3) produced unexpected output for z3test/regressions/smt2/7489.smt2:
EXPECTED
sat
PRODUCED
unsat

levnach:
I see : the release produced an "unsat".
I was checking the debug version that was giving a "sat".

[881/881] Linking CXX executable z3
levnach@Levs-MacBook-Pro release % ./z3 ~/dev/z3test/regressions/smt2/7489.smt2 
unsat
levnach@Levs-MacBook-Pro release % cd ../debug 
levnach@Levs-MacBook-Pro debug % ./z3 ~/dev/z3test/regressions/smt2/7489.smt2
sat

It would be nice to understand what is going on.

levnach:
The problem goes to the factorization functionality. There is a polynomial p that has a squared factor (x0 + 8) but z3 cannot factor p, so the factorization just returns the whole p. As a result the finest_factors heuristic does not work.
 
 p = x0^3 x1^4 + 24 x0^2 x1^4 + 192 x0 x1^4 + 512 x1^4 + 2 x0^4 x1^3 - 48 x0^3 x1^3 - 1632 x0^2 x1^3 - 12800 x0 x1^3 - 30720 x1^3 + x0^5 x1^2 - 56 x0^4 x1^2 - 320 x0^3 x1^2 + 5120 x0^2 x1^2 + 28928 x0 x1^2 + 2048 x1^2 + 16 x0^5 x1 - 992 x0^4 x1 - 5376 x0^3 x1 + 98816 x0^2 x1 + 585728 x0 x1 + 196608 x1 + 64 x0^5 - 3328 x0^4 - 30464 x0^3 + 26624 x0^2 - 524288 x0 - 5767168.

PR #7807, opened 2025-08-23
Attempt at adding the README to the NuGet package

SolalPirelli:
**This is untested because I don't see how to run `scripts/release.yml` myself**, but I think I know what the problem is, and I figured a PR would be easier to look at than an issue comment with suggestions.

Locally, _using the latest `nuget.exe`_, doing what [this commented line](https://github.com/Z3Prover/z3/blob/master/scripts/mk_nuget_task.py#L106) does and adding `<readme>content/README.md</readme>` to the nuspec, without needing any explicit "file" nodes, works. It leads to a package with a README that Nuget.org recognizes.

However:
- there was a bug with using readmes when also using symbols up until .NET 6 RC 2 (= NuGet 6.0): NuGet/Home#10791
- `scripts/release.yml` passes `versionSpec: 5.x` to `NuGetToolInstaller@0`, so the bug will trigger

This PR changes the version of NuGet used to 6.x, which should contain the fix and thus work.

:warning: There is still a warning printed on the command line when packing that the _symbols_ package doesn't have a readme, but this seems like a NuGet bug, symbols packages aren't supposed to have a readme anyway.

### DIFF
--- a/scripts/release.yml
+++ b/scripts/release.yml
@@ -471,7 +471,7 @@ stages:
 
      
   - job: NuGetPublish
-    condition: eq(1,0)
+    condition: eq(1,1)
     displayName: "Publish to NuGet.org"
     steps:
     - task: DownloadPipelineArtifact@2


SolalPirelli:

NikolajBjorner:
@copilot can you make similar updates to nightly.yaml that were made to release.yml?

NikolajBjorner:
we try to see effect on Nightly first

NikolajBjorner:
https://github.com/Z3Prover/z3/releases/tag/Nightly 
The nuget package contains README.txt in the content directory.

@SolalPirelli 

@levnach - this was a trigger to examine what was going on with the nightly build. It is fairly often broken for random reasons so it needs to be considered.

SolalPirelli:
Nice! Does this mean the next (non-nightly) release will be pushed to NuGet.org as well, or is there something else to do?

NikolajBjorner:
The packages would not be signed. I removed signing pipeline which uses an internal service that is very laborious to set up and the process changed last year. If I get internal help to resurrect this, then the packages may be signed. Otherwise, they are unsigned. I presume nuget.org doesn't require signed packages.

NikolajBjorner:
We still have something to fix. Uploading nightly fails.
The package manifest contains an invalid Version: '4.15.4.5884-5884'
The minor version gets appended by a script somewhere.

SolalPirelli:
I believe a lack of signature is fine, NuGet.org doesn't require it. AFAIK the only people who ""need"" signed packages are folks in companies with heavy bureaucracy around dependencies.

SolalPirelli:
The nightly version is set [here](https://github.com/Z3Prover/z3/blob/master/scripts/nightly.yaml#L8), presumably removing the -$BuildId part would make it work. I don't see any non-NuGet uses of `NightlyVersion`.

But for release it's [hardcoded](https://github.com/Z3Prover/z3/blob/master/scripts/release.yml) so it should be fine.

NikolajBjorner:
See also https://github.com/Z3Prover/z3/pull/7820

NikolajBjorner:
I am unable to update nuget.org because the package isn't signed and my organization requires signed packages. 
I have neither the skillsets nor bandwidth to re-add certificates or understanding of if there is a simpler process than the internal approach that I can follow.

For now, you can now probably publish your own clones (called something else than Microsoft.Z3) on nuget if that works for you, or install directly from the builds.

PR #7831, opened 2025-08-29
Parallel solving

ilanashapiro:
Iterative deepening experiment (no PQ yet): this builds on the depth-splitting experiment, and only applies to the frugal strategy. Each cube that's worked on by a thread returns some statistics (conflicts, propagations, decisions restarts). Based on these statistics, I developed a naive metric of "hardness" for that cube, and the batch manager maintains a running average of the hardness of all the cubes solved by the workers. If the cube hardness > overall avg hardness, we cube (frugally, and depth-first). Otherwise, we don't cube. This means the max_depth cutoff is now _adaptive_ and per-cube.


NikolajBjorner:
have to PR into ilana branch not master

Issue #7953, opened 2025-09-29
Quantifier.getBody() again throws Cast Error in Java API (Issue #2460 is still not fixed)


ruijiefang:
TLDR: It seems like the fix in https://github.com/Z3Prover/z3/issues/2460 has been reverted, and the bug has resurfaced again.

Array values in the Z3 model sometimes show up as lambda terms, but in Z3 Java API, they get represented as a `Quantifier` instead --- but `Quantifier.getBody` is no longer guaranteed to be a `BoolExpr`. Thus, if I call `Quantifier.getBody` on a lambda term, sometimes Java throws a cast error, and there is no good way to overcome this. In particular, it seems like the fix in the above issue has been reverted in the commit https://github.com/Z3Prover/z3/commit/3bca1fbcd82d0dd8afec741d8759a72f8490f1e2

From a user point of view, it would be great if lambda values in the Z3 model actually get represented as a `z3.Lambda` AST node in the Java API (or alternatively, not have `Quantifier` be a subclass of `BoolExpr`) --- but at a minimum, it would be good if `Quantifier.getBody` can be fixed again to not perform a forced cast.

ruijiefang:
I suggested a fix in https://github.com/Z3Prover/z3/pull/7955 --- Not sure if it completely addresses the issue, and many thanks in advance for looking at this! (For background, I'm interacting with Z3's Java API via my own Scala wrapper, and my Z3 usage centers on array and datatype theories, so encountering this has been a real pain).

PR #7955, opened 2025-09-30
Correctly distinguish between `Lambda` and `Quantifier` in Z3 Java API

ruijiefang:
This is my attempt at patching https://github.com/Z3Prover/z3/issues/7953. I'm not sure if this is up to standard for a PR, but I will put this here in case it helps. Thank you!


ruijiefang:

PR #7959, opened 2025-10-03
fix: add .gitmodules

lixitrixi:
This PR adds a gitmodules entry for the "z3test" submodule, pointing to the correct repository.

Currently when using this repo as a submodule and running `git submodule update --init --recursive`, I get the following error:
```
fatal: No url found for submodule path '<path>/z3/z3test' in .gitmodules
```

It's worth saying I'm using the [prove-rs/z3.rs](https://github.com/prove-rs/z3.rs) Rust library as a submodule, which itself uses Z3Prover/z3 as a submodule. Because of this I need to pass the `--recursive` flag and I run into this error. Is there a reason why `.gitmodules` isn't present?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "z3test"]
+	path = z3test
+	url = https://github.com/Z3Prover/z3test.git


NikolajBjorner:
Something on your build chain assumes a reference to z3test?
Our CI just references z3test as a separate repository. It isn't a submodule of z3.

lixitrixi:
That's strange... I don't reference z3test anywhere.

I can see from the main repository page on desktop that there's a folder called `z3test`. It has the icon of a subdirectory but clicking it doesn't lead anywhere. This file resolves that issue on my fork and it now leads to the test repo.

agurfinkel:
Sees like a friendly AI added an undesired connection to z3test. @NikolajBjorner, there is z3test file now in repo root. Was introduced by this [link](https://github.com/Z3Prover/z3/commit/ee083a2e6c09bfe824d0abf57d99be4ecf0e0380#diff-56e8f8179991f4e605a2448620acbbb7fc13cfd3bbd77256a89c83fbacc6607a)

PR #7960, opened 2025-10-04
Add cube tree optimization about resolving cores recursively up the path, to prune

ilanashapiro:
Solving this issue: "If the core is {A, C} in in the A->B->C branch and the core is {A, !C} in the A->B->!C branch we currently still explore A->!B. This is not necessary because the resolvent of {A, C} and {A, !C} is A which means that the branch A is closed. You can go directly to !A. "

Benchmarks: 
run-0728-smt_qfnia-threads-4-recursiveresolve3-stats.md
run-0727-smt_qflia-threads-4-recursiveresolve3-stats.md

Baselines:
run-0723-smt_qfnia-threads-4-stats.md
run-0729-smt_qflia-threads-4-stats.md

QF_NIA_small: solves 2 more than baseline
QF_LIA: solves 9 more than baseline

I got 1 segfault (`terminated by signal[31mramon: signal 11 (SIGSEGV Segmentation fault)[0m`) on the QF_LIA benchmark (z3-poly-testing/inputs/QF_LIA/c10_problem__003.smt2.slack.smt2) that I'm unable to repro locally. all the code I added is under lock when called in the batch manager.

### DIFF
--- a/src/util/search_tree.h
+++ b/src/util/search_tree.h
@@ -163,9 +163,6 @@ namespace search_tree {
         }
 
         // Given complementary sibling nodes for literals x and ¬x, sibling resolvent = (core_left ∪ core_right) \ {x, ¬x}
-        // if there are other complements in the union of the cores, they are not removed, but that is ok, because the core is
-        // strengthened, and these complements came from a higher-up branch, which will be resolved away recursively as we move up the tree
-        // in conclusion: the sibling resolvent only promises to eliminate the split variable’s literal/complement.Other complements may still appear, which is ok since they just reflect conflicts from higher up the path.
         vector<literal> compute_sibling_resolvent(node<Config>* left, node<Config>* right) {
           vector<literal> res;
 
@@ -216,7 +213,7 @@ namespace search_tree {
 
               // empty resolvent → subtree is UNSAT
               if (resolvent.empty()) {
-                  p->set_core(resolvent);  // empty core
+                  p->set_core(resolvent); // empty core
                   close_node(p);
                   p = p->parent();
                   continue;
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
use https://github.com/Z3Prover/z3/tree/parallel

src/util/search_tree.h
```diff
@@ -154,8 +162,109 @@ namespace search_tree {
...
+          while (p) {
+              auto left = p->left();
+              auto right = p->right();
+              if (!left || !right) return;
+
+              // only attempt when both children are closed and each has a core
+              if (left->get_status() != status::closed || right->get_status() != status::closed) return;
+              if (!left->has_core() || !right->has_core()) return;
+
+              auto resolvent = compute_sibling_resolvent(left, right);
```
NikolajBjorner, line 223: not only subtree, but all is unsat

ilanashapiro:
there is a bug, I am working on it

src/util/search_tree.h
```diff
@@ -131,31 +139,139 @@ namespace search_tree {
...
+      // Cores labeling nodes are subsets of the literals on the path to the node and the (external) assumption literals.
+      // If a parent is open, then the one of the children is open.
+      void close_with_core(node<Config>* n, vector<literal> const &C, bool allow_resolve = true) {
+          if (!n || n->get_status() == status::closed)
+              return;
+
+          n->set_core(C);
+          n->set_status(status::closed);
+
+          close_with_core(n->left(), C, false);
```
ilanashapiro, line 152: is it ok to pass the same core down? we're closing the children regardless but there is some semantic inconsistency because these nodes didn't directly produce this core

PR #7977, opened 2025-10-15
Fixed tests by making methods public for finite sets

kper:
The tests fail because they use the methods from `finite_set_rewriter` which are private. To fix the tests, I made them public.

### DIFF
--- a/src/ast/rewriter/finite_set_rewriter.h
+++ b/src/ast/rewriter/finite_set_rewriter.h
@@ -34,12 +34,7 @@ where the signature is defined in finite_set_decl_plugin.h.
 */
 class finite_set_rewriter {
     finite_set_util  m_util;
-    // Rewrite rules for set operations
-    br_status mk_union(unsigned num_args, expr * const * args, expr_ref & result);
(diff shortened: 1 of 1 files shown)


src/ast/rewriter/finite_set_rewriter.h
```diff
@@ -34,12 +34,7 @@ where the signature is defined in finite_set_decl_plugin.h.
 */
 class finite_set_rewriter {
     finite_set_util  m_util;
-    // Rewrite rules for set operations
-    br_status mk_union(unsigned num_args, expr * const * args, expr_ref & result);
-    br_status mk_intersect(unsigned num_args, expr * const * args, expr_ref & result);
-    br_status mk_difference(expr * arg1, expr * arg2, expr_ref & result);
-    br_status mk_subset(expr * arg1, expr * arg2, expr_ref & result);
-public:
+    public:
```
NikolajBjorner, line 37: oh, that's why copilot made these methods public.
I am going to make the tests use a class test_finite_sets and declare it as a friend.

PR #7978, opened 2025-10-15
Implement more cases for `finite_set_decl_plugin::are_distinct`

kper:
This PR adds more cases for `finite_set_decl_plugin::are_distinct` to determine whether `e1` and `e2` are distinct.

### DIFF
--- a/src/ast/finite_set_decl_plugin.cpp
+++ b/src/ast/finite_set_decl_plugin.cpp
@@ -204,12 +204,8 @@ bool finite_set_decl_plugin::are_distinct(app* e1, app* e2) const {
         return true;
     if (r.is_singleton(e1) && r.is_empty(e2))
         return true;
-    if(r.is_union(e1) && r.is_empty(e2))
-        return true;
-    if(r.is_empty(e1) && r.is_union(e2))
-        return true;
     if(r.is_singleton(e1) && r.is_singleton(e2))
-        return e1 != e2;
+        return m_manager->are_distinct(e1, e2);
     
     // TODO: could be extended to cases where we can prove the sets are different by containing one element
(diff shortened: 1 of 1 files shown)


src/ast/finite_set_decl_plugin.cpp
```diff
@@ -204,6 +204,13 @@ bool finite_set_decl_plugin::are_distinct(app* e1, app* e2) const {
         return true;
     if (r.is_singleton(e1) && r.is_empty(e2))
         return true;
+    if(r.is_union(e1) && r.is_empty(e2))
```
NikolajBjorner: this isn't right.
For example set.empty is not distinct from set.empty union set.empty.

src/ast/finite_set_decl_plugin.cpp
```diff
@@ -204,6 +204,13 @@ bool finite_set_decl_plugin::are_distinct(app* e1, app* e2) const {
...
+        return true;
+    if(r.is_empty(e1) && r.is_union(e2))
+        return true;
+    if(r.is_singleton(e1) && r.is_singleton(e2))
```
NikolajBjorner, line 207: This is close, you want to recursively call m_manager->is_distinct(e1, e2)

src/ast/finite_set_decl_plugin.cpp
kper: Thanks, I removed them

src/ast/finite_set_decl_plugin.cpp
kper: I use the `are_distinct` function now because `is_distinct` only allows one argument.

PR #7986, opened 2025-10-16
Create fresh values for finite set using bitvector

lorenzwinkler1:
This approach looks at the `next_index` as a bitvector, and creates a set $R$, where $x_i\in R \Leftrightarrow$ i-th bit in `next_index` is $1$, when $x_i$ is the i-th element of the sort of the set (credits to @CEisenhofer who proposed this during class).

I would be happy to discuss whether this approach is suitable.

Advantages
- handles infinite-domain element-sorts well (I think the implementation which was hinted at would only return singleton sets in that case)

Disadvantages
- can at most create as many singleton sets/use as many values of the element sort as the number of bits of `unsigned int`.

### DIFF
--- a/src/model/finite_set_value_factory.cpp
+++ b/src/model/finite_set_value_factory.cpp
@@ -38,7 +38,7 @@ expr * finite_set_value_factory::get_fresh_value(sort * s) {
 
     // take the bitmask of next_index to determine which elements to include
     // i.e. if next_index = 13 = 1101_2, then we include elements values_e[0], values_e[2], and values_e[3]
-    // new elements for sort s are created on the fly
+    // new elements for element sort are created on the fly
     int num_shifts = 0;
     auto r = u.mk_empty(s);
     // check the rightmost bit of next_index


lorenzwinkler1:

src/model/finite_set_value_factory.cpp
```diff
@@ -33,31 +33,37 @@ expr * finite_set_value_factory::get_fresh_value(sort * s) {
...
+    auto r = u.mk_empty(s);
+    // check the rightmost bit of next_index
+    while ( next_index > 0)
+    {
+        if (next_index & 1) {
+            // the element values_e[next_index] is in the set
+            expr* element = nullptr;
+            if(num_shifts > values_e.size()) {
+                // create a fresh element for the element sort
+                element = m_model.get_fresh_value(elem_sort);
```
NikolajBjorner, line 52: this will not work for FiniteSet(Bool).
get_fresh_value(Bool) will return true, false, and then null.

NikolajBjorner:
Using bit-vectors is going to be possible, but requires a fix. 
A fix could be of the form of using [set_e, values_e] for the element set, if the index of the bit-vector is below the size of values_e just access a value there.

However, there is a going to be a simpler approach that uses a different iteration over the value space of sets.
It is as follows:

```
if values.size() == 0: 
    let r = u.mk_empty(set_sort);
    register_value(r)
    return r
    
if values.size() is a power of 2:
   let e = create-fresh-element(element_sort)
   if (!e)
      return nullptr;
   register_value(e)
   let r = u.mk_singleton(r)
   register_value(r)
   return r
 
 let N be the greatest number N < values.size() such that N is a power of 2.
 let r = values[values.size() - N]
 r = u.mk_union(e_values.back(), r);
 register_value(r)
 return r;
 ```
 
 It should enumerate the sets in order of:
 
 ```
 {}, {v1} , {v2}, {v1, v2}, {v3}, {v1, v3}, {v2, v3}, {v1, v2, v3}, {v4}, {v1, v4}, {v2, v4}, {v1, v2, v4}, ...
 ```

PR #7988, opened 2025-10-16
Enabling Control Flow Guard (CFG) by default for MSVC on Windows, with options to disable CFG.

hwisungi:
Updated Python build script and CMakeList.txt to enable Control Flow Guard (CFG) by default for MSVC on Windows.
When CFG is enabled for MSVC, `/guard:cf` option is added to cl.exe, and `/GUARD:CF` and `/DYNAMICBASE` (required by `/GUARD:CF` option) options are added to link.exe.
When CFG is disabled, `/guard:cf-` option is added to cl.exe, and `/GUARD:NO` option is added to link.exe. `/DYNAMICBASE` is not changed, if set explicitly. CFG can be enabled for any flavors

Below examples show how to build z3 with CFG enabled. 

# Building z3 as an amd64 retail DLL with CFG enabled:
 
## CMake
Run these commands, starting at the root of the z3 repo enlistment:
``` 
md build-cmake
cmake .. -G "Visual Studio 17 2022" -A x64 -DCMAKE_BUILD_TYPE=Release -DZ3_BUILD_LIBZ3_SHARED=ON -DZ3_ENABLE_CFG=ON
cd build-cmake
cmake --build . --config Release
```
- `Z3_ENABLE_CFG` (enable CFG) is ON by default. So, it is actually redundant. Setting it to OFF will disable CFG.

## Python
Run these commands, starting at the root of the z3 repo enlistment:
 
```
python scripts/mk_make.py -x
cd build
nmake
```
- `--guardcf` option enables it CFG is enabled by default
- To disable CFG, add `--no-guardcf` option to the script.

# Verifying the built binary:
Run
```
dumpbin /headers /loadconfig libz3.dll
```
and look for the GUARD CF properties
 
## When CFG is enabled - tables for CFG will have entries:
```
    0000000180DB3730 Guard CF address of check-function pointer
    0000000180DB3740 Guard CF address of dispatch-function pointer
    0000000180DB3870 Guard CF function table
                276A Guard CF function count
            10017500 Guard Flags
                       CF instrumented
                       FID table present
                       Protect delayload IAT
                       Delayload IAT in its own section
                       Export suppression info present
                       Long jump target table present
```

## When CFG is disabled - tables for CFG will be empty: 
```
0000000180C4D730 Guard CF address of check-function pointer
0000000180C4D740 Guard CF address of dispatch-function pointer
0000000000000000 Guard CF function table
               0 Guard CF function count
00000100 Guard Flags
                 CF instrumented
```

### DIFF
--- a/scripts/mk_util.py
+++ b/scripts/mk_util.py
@@ -645,7 +645,8 @@ if os.name == 'nt':
     IS_WINDOWS=True
     # Visual Studio already displays the files being compiled
     SHOW_CPPS=False
-    # Enable Control Flow Guard by default on Windows
+    # Enable Control Flow Guard by default on Windows with MSVC
+    # Note: Python build system on Windows assumes MSVC (cl.exe) compiler
     GUARD_CF = True
 elif os.name == 'posix':
     if os.uname()[0] == 'Darwin':
@@ -2538,6 +2539,8 @@ def mk_config():
     config = open(os.path.join(BUILD_DIR, 'config.mk'), 'w')
     global CXX, CC, GMP, GUARD_CF, STATIC_BIN, GIT_HASH, CPPFLAGS, CXXFLAGS, LDFLAGS, EXAMP_DEBUG_FLAG, FPMATH_FLAGS, LOG_SYNC, SINGLE_THREADED, IS_ARCH_ARM64
     if IS_WINDOWS:
+        # On Windows, Python build system assumes MSVC (cl.exe) compiler
+        # GUARD_CF is only supported with MSVC, which is the default on Windows
         CXXFLAGS = '/nologo /Zi /D WIN32 /D _WINDOWS /EHsc /GS /Gd /std:c++20 -D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR'
         config.write(
             'CC=cl\n'


CMakeLists.txt
```diff
@@ -362,34 +362,69 @@ endif()
...
+  message(STATUS "Enabling Control Flow Integrity (CFI) for Clang")
+  z3_add_cxx_flag("-fsanitize=cfi" REQUIRED)
+  z3_add_cxx_flag("-fsanitize-cfi-cross-dso" REQUIRED)
+endif()
+# End CFI section
+
+################################################################################
+# Control Flow Guard (MSVC only)
+################################################################################
+option(Z3_ENABLE_CFG "Enable Control Flow Guard security checks" ON)
```
NikolajBjorner: So this flag is ON by default. 
But if the compiler is different from MSVC it raises a fatal error.
This seems incorrect: a user may intend to use a different compiler.
The user should not have to configure an unrelated flag to avoid a fatal error.

hwisungi: I will set it to OFF by default for other compilers and turn it ON for MSVC only.

NikolajBjorner: Thanks, and it also appears to have affected the existing CI with the OCaml Binding CI failing

NikolajBjorner: LMK about when it is off

PR #7993, opened 2025-10-20
Setting up param tuning python prototyping experiment

ilanashapiro:
For some reason this PR is showing a lot of files changed (potentially due to new commits on the master branch)? The experiment is in param-tuning-experiment.py. I also changed some other files to modify the OnClause hook

This file sets up the param tuning Python prototyping experiment outlined here: https://github.com/NikolajBjorner/z3papers/blob/main/threads/notes/parameter_sampling.md

However, there are 2 parts still missing for the experiment to be ready (the Z3 surgery discussed at the bottom of the linked document):

(1) right now, the clauses selected from each iteration of the proof prefix thread are chosen naively via the Python OnClause hook. we are just choosing the first n clauses encountered, which is naive

(2) I was also not entirely sure where to modify Z3 to be able to dynamically update the params during solving, for the main solver that's running in parallel mode. right now, the main thread's solver doesn't receive param updates from the proof prefix thread due to this.

### DIFF
--- a/src/smt/smt_parallel.cpp
+++ b/src/smt/smt_parallel.cpp
@@ -412,17 +412,7 @@ namespace smt {
         switch (m_state) {
         case state::is_running:  // batch manager is still running, but all threads have processed their cubes, which
                                  // means all cubes were unsat
-            if (!m_search_tree.is_closed())
-                throw default_exception("inconsistent end state");
-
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
It has resolve conflicts

PR #8006, opened 2025-10-30
Setting up param tuning infrastructure in C++

ilanashapiro:
The code compiles but I haven't tested it yet
Also, I can't figure out how to access the relevant metrics from the solver statistics object, so the best param state isn't getting updated at all right now

### DIFF
--- a/src/smt/smt_parallel.cpp
+++ b/src/smt/smt_parallel.cpp
@@ -155,7 +155,8 @@ namespace smt {
                 unsigned best_param_state_idx = replay_proof_prefixes(candidate_param_states);
 
                 if (best_param_state_idx != 0) {
-                    best_param_state = candidate_param_states[best_param_state_idx];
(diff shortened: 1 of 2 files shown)


src/smt/smt_parallel.cpp
```diff
@@ -64,6 +64,122 @@ namespace smt {
...
+                }
+
+                // Replay the negated clause
+                lbool r = probe_ctx->check(negated_lits.size(), negated_lits.data());
+
+                ::statistics st;
+                probe_ctx->collect_statistics(st);
+                unsigned conflicts = 0, decisions = 0, rlimit = 0;
+
+                // I can't figure out how to access the statistics fields, I only see an update method
```
ilanashapiro, line 117: I tried to access the statistics here but couldn't figure out which function to call on the solver statistics object (this is wrong/commented out now)

PR #8007, opened 2025-10-30
expose a getter for the solver statistics

ilanashapiro:
it compiles but is still untested

closing since we're using the existing exposed fields in smt_params


src/smt/smt_parallel.cpp
```diff
@@ -141,7 +139,7 @@ namespace smt {
         m_my_param_state.insert(symbol("smt.arith.nl.cross_nested"), smtp.arith_nl_cross_nested());
         m_my_param_state.insert(symbol("smt.arith.nl.delay"), smtp.arith_nl_delay());
         m_my_param_state.insert(symbol("smt.arith.nl.expensive_patching"), smtp.arith_nl_expensive_patching());
-        m_my_param_state.insert(symbol("smt.arith.nl.gb"), smtp.arith_nl_gb());
+        // m_my_param_state.insert(symbol("smt.arith.nl.gb"), smtp.arith_nl_gb());
```
ilanashapiro, line 142: commenting this out for now since i was getting a field not found compiler error

PR #8008, opened 2025-10-31
Updates to param tuning

ilanashapiro:
debug some things

instead of saving many contexts in candidate_param_states, create a throwaway context for each tester on the fly

also create the mutation on the fly for each tester

in smt_internalizer, change m_recorded_clauses to m_recorded_cubes. m_recorded_cubes is of type vector<expr_ref_vector> (i.e. a vector of cubes, where each cube is an expr_ref_vector) since i couldn't figure out how to pass in an expr_ref to ctx->check

### DIFF
--- a/src/util/statistics.cpp
+++ b/src/util/statistics.cpp
@@ -81,21 +81,6 @@ struct str_lt {
 typedef map<char const *, unsigned, str_hash_proc, str_eq_proc> key2val;
 typedef map<char const *, double, str_hash_proc, str_eq_proc> key2dval;
 
-double statistics::get_val(char const * key) const {
-    key2val m_u;
-    key2dval m_d;
-    mk_map(m_stats, m_u);
-    mk_map(m_d_stats, m_d);
-
-    unsigned val = 0;
-    double dval = 0.0;
-    if (m_u.find(key, val))
-        return static_cast<double>(val);
-    if (m_d.find(key, dval))
-        return dval;
-    return 0.0;
-}
-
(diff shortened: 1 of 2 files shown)


src/smt/smt_parallel.cpp
```diff
@@ -168,41 +168,23 @@ namespace smt {
...
-                unsigned best_param_state_idx = replay_proof_prefixes(candidate_param_states);
-
-                if (best_param_state_idx != 0) {
-                    m_param_state = candidate_param_states[best_param_state_idx];
+                auto [best_param_state, found_better_params] = replay_proof_prefixes();
+
+                // NOTE: we either need to return a pair from replay_proof_prefixes so we can return a boolean flag indicating whether better params were found.
+                // or, we have to implement a comparison operator for param_values
+                // or, we update the param state every single time even if it hasn't changed
+                // for now, I went with option 1
```
ilanashapiro, line 182: see comment (I know this is likely not ideal and will need to be changed)

src/smt/smt_parallel.cpp
```diff
@@ -83,50 +83,51 @@ namespace smt {
         return r;
     }
 
-    unsigned parallel::param_generator::replay_proof_prefixes(vector<param_values> const& candidate_param_states, unsigned max_conflicts_epsilon=200) {
+    std::pair<parallel::param_generator::param_values, bool> parallel::param_generator::replay_proof_prefixes(unsigned max_conflicts_epsilon=200) {
```
ilanashapiro, line 86: see the other comment I left below about this

PR #8012, opened 2025-11-04
Debugging param tuning (I think it runs now)

ilanashapiro:
Debug several things:

"warming up" the param thread context's manager that's default-constructed by running it through an ast_translation (still not sure why this is necessary but it crashes otherwise upon context alloc)

make sure the param probe thread is killed along with the worker when the batch manager sets unsat/sat/exception (need to check the manager's limit at several places inside param thread)

make sure we don't create multiple nested param tuners within child parallel objects that are created upon ctx->check calls (this required modification to the check function in smt_context.cpp)

finally, make sure the param probe contexts are run in single-threaded mode so we don't spawn a bunch of new workers from the param probe's separate process (this results in a segfault)

### DIFF
--- a/src/smt/smt_parallel.cpp
+++ b/src/smt/smt_parallel.cpp
@@ -70,6 +70,8 @@ namespace smt {
             return l_undef;
         IF_VERBOSE(1, verbose_stream() << " PARAM TUNER running prefix step\n");
         ctx->get_fparams().m_max_conflicts = m_max_prefix_conflicts;
+        ctx->get_fparams().m_threads = 1;
+
         m_recorded_cubes.reset();
         ctx->m_recorded_cubes = &m_recorded_cubes;
         lbool r = l_undef;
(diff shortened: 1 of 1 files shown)


src/smt/smt_parallel.cpp
```diff
@@ -320,7 +339,11 @@ namespace smt {
     }
 
     parallel::param_generator::param_generator(parallel& p)
-        : p(p), b(p.m_batch_manager), m_p(p.ctx.get_params()), m_l2g(m, p.ctx.m) {
+        : b(p.m_batch_manager), m_p(p.ctx.get_params()), m_l2g(m, p.ctx.m) {
+        // patch fix so that ctx = alloc(context, m, p.ctx.get_fparams(), m_p); doesn't crash due to some issue with default construction of m
+        ast_translation m_g2l(p.ctx.m, m);
+        m_g2l(p.ctx.m.mk_true());
```
ilanashapiro, line 345: i am not sure why this is necessary, but if i don't do this step and I just pass in the default-constructed manager, it crashes upon ctx allocation

PR #8021, opened 2025-11-14
remove ocamlfind destdir directory check

PerfectLaugh:
This check causes compilation error when using directly in dune.
Somehow, $DESTDIR doesn't created automatically with ocamlfind in dune. Removing this check should make dune dependency works again.


PerfectLaugh:
Draw off PR. Linking doesn't work out

PR #8031, opened 2025-11-16
Small bugfix in compute_sibling_resolvent

ilanashapiro:
I was making my slides for my talk this week and going through the search_tree.h code I think I may have found a small bug in compute_sibling_resolvent


NikolajBjorner:
somehow search_tree in this branch is based on an older version than search_tree.h in the master branch.
I didn't find the bug in the master branch.
Perhaps update search_tree to be more current?

ilanashapiro:
I believe we added the optimization here about pruning the tree via resolving cores upward? I don't think that exists on the master branch yet (started with this commit https://github.com/Z3Prover/z3/commit/52fd59df1b0d628b16515725adc90f790f9921cd and ended with this commit https://github.com/Z3Prover/z3/commit/c5d65cdedddfb23484175c9234f904988fa6b3fd)

the master branch just has close_node, vs parallel now has close_with_core

PR #8044, opened 2025-11-24
Reorder at correct spot

levnach:
- **parameter correct order experiment**
- **try reordering before analyzing bounds**


levnach:
It is a useful change: the performance is slightly better and 20+ files solved on QF_NIA_small, tested with two random seeds.

PR #8046, opened 2025-11-26
Return bool instead of int from Z3_rcf_interval

jberdine:
In the underlying realclosure implementation, the interval operations for
{`lower`,`upper`}`_is_`{`inf`,`open`} return `bool` results. Currently these
are cast to `int` when surfacing them to the API. This patch keeps them at
type `bool` through to `Z3_rcf_interval`.

No worries if you don't want to take API-breaking changes like this, it's
not hard to work around.

### DIFF
--- a/src/api/api_rcf.cpp
+++ b/src/api/api_rcf.cpp
@@ -385,7 +385,7 @@ extern "C" {
         Z3_CATCH_RETURN(nullptr);
     }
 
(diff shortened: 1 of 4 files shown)


jberdine:
draft because this depends on https://github.com/Z3Prover/z3/pull/8047

NikolajBjorner:
There was an ancient time when bool was an int. 
This code was written before even committing to the C99 standard ;-).

PR #8048, opened 2025-11-26
Return bool instead of int in extra_API for Z3_open_log

jberdine:
The C declaration returns `bool`.

The last of the small bool vs int things I hit.

### DIFF
--- a/src/api/z3_api.h
+++ b/src/api/z3_api.h
@@ -5877,7 +5877,7 @@ extern "C" {
        \sa Z3_append_log
        \sa Z3_close_log
 
-       extra_API('Z3_open_log', INT, (_in(STRING),))
+       extra_API('Z3_open_log', BOOL, (_in(STRING),))
     */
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
self-help suggestion: add your name to RELEASE_NOTES with these kind of PRs because I will just slop around it.

Actually, this is a use for an agentic workflow, @pelikhan. I bet you have already been using one for a couple of years at this point.

NikolajBjorner:
hmm, I should have checked the builds.
The Ocaml, Java and probably more are broken.
Good if you get a chance to chime in to fix the straight-forward regressions.

jberdine:
Oh, I'm very sorry, I didn't intend to spend your time on this sort of thing. I didn't have enough enabled in my cmake command. I've checked now, but it looks like you have already fixed everything. I'm not sure just what gets run on PRs, and don't seem to be able to see the jobs or results.

PR #8050, opened 2025-11-26
Fix _in vs _out def_API param for Z3_solver_get_levels

jberdine:

### DIFF
--- a/src/api/z3_api.h
+++ b/src/api/z3_api.h
@@ -7132,7 +7132,7 @@ extern "C" {
        \brief retrieve the decision depth of Boolean literals (variables or their negations).
        Assumes a check-sat call and no other calls (to extract models) have been invoked.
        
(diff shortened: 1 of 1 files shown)


jberdine:
I checked the new code generated in build/src/api/ml/z3native_stubs.c and the corresponding .ml and .mli. The code in z3.py looks to not be generated using this declaration. I looked at the code generated in build/src/api/java/Native.cpp but TBH I'm lost in the JNI details.

NikolajBjorner:
you are adding this PR not for your _needs_ but because you have a _need_ for order and harmony?

jberdine:
> you are adding this PR not for your needs but because you have a need for order and harmony?
> 

I won't deny that there is a bit of that. :-)

But I am generating binding code from the `def_API` lines, and that code is just wrong with `_in` instead of `_out`. I figured to send this to save the next consumer of that `def_API` a little debugging.

As for `Z3_solver_get_levels` itself, currently my usage is just testing and exploration, I can't claim to have a new interesting use of that API.

PR #8061, opened 2025-12-06
Finite Sets: Simplify expressions before range local check

stromberger:
While implementing the metamorphic fuzzer for finite sets I stumbled upon performance issues when set ranges intersected. For example:

```
(declare-const s (FiniteSet Int))

(assert (not (set.subset (set.range 1 2) s)))
(assert (not (set.subset (set.range 2 3) s)))

(assert (not (set.in 1 s)))

(check-sat)
```

This does not terminate within reasonable time (>1 min; appears to not terminate). However, replacing `(set.range 1 2)` with `(set.range 1 1)` results in `SAT` immediately.

I am not sure what causes the bug exactly but I (and Claude) think there is some amplification of unsimplified terms when testing range membership. This patch simplifies the expression before checking containment and simplifies the `lo`/`hi` boundary expressions afterwards. The program above now results in `SAT` in <1s.

If you need more details and/or test cases let me know!

### DIFF
--- a/src/smt/theory_finite_set.cpp
+++ b/src/smt/theory_finite_set.cpp
@@ -27,7 +27,7 @@ namespace smt {
     theory_finite_set::theory_finite_set(context& ctx):
         theory(ctx, ctx.get_manager().mk_family_id("finite_set")),
         u(m),
-        m_axioms(m), m_find(*this),
+        m_axioms(m), m_rw(m), m_find(*this),
         m_cardinality_solver(*this)
     {
         // Setup the add_clause callback for axioms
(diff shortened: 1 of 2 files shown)


stromberger:

NikolajBjorner:
This is a good find.

The bug is that the range axioms are created using a rewriter that normalizes lo and hi. So the filter that checks if a lo/hi pair were already added in the E-graph is going to fail.

Your fix appears to address this based on the behavior of how axioms are created.

We could consider a more principled approach that caches generated range 
axioms more directly within theory_finite_sets without having to make brittle assumptions about behavior of rewriting.

PR #8066, opened 2025-12-11
Search tree core resolution optimization

ilanashapiro:
for merging into master branch


src/smt/smt_parallel.cpp
```diff
@@ -115,10 +115,6 @@ namespace smt {
                     b.set_unsat(m_l2g, unsat_core);
                     return;
                 }
-                // report assumptions used in unsat core, so they can be used in final core
```
NikolajBjorner: I don't think the assumption tracking should be removed?

NikolajBjorner:
something is missing in the backtrack function. 
When setting the state to is_unsat (the root is closed). 
Then extract the core (that contains assumption literals) from the root.

PR #8088, opened 2025-12-16
BLD: Add CMake option to build Python bindings without rebuilding libz3 (redux)

### LINKED ISSUE #8016: BLD: provide CMake option to install only python bindings
Hi!

I help run [conda-forge](https://conda-forge.org), a big packaging ecosystem for python and many more languages. For almost a year, we've [providing](https://github.com/conda-forge/z3prover-feedstock) `z3` as a package that can be installed also through `conda install -c conda-forge z3-solver`.

Conda as a paradigm is strictly more powerful than what PyPI and wheels can offer, namely, we can package native C/C++/Fortran libraries and binaries for linux/osx/win just as well as python libraries.

Unsurprisingly, the C++ [library](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Flibz3-4.15.4-h171cf75_0.conda) underlying `z3-solver` is _independent_ of the python version being used. This is a very common pattern, and we like to package things in a way that reduces vendoring and other redundancies; in other words, there's a python-independent `libz3`, which `z3-solver` can then depend on, and everything works fine.

However, there's a limitation in building the package at the moment; due to the very bespoke `python scripts/mk_make.py` setup, I've preferred to go with the CMake-based builds; the problem with this is just that we're very inefficiently building the library portion 6 times, rather than once, namely (indentation indicates dependence):

* `libz3` - necessary ✅ 
  * `z3-solver` (py3.10) - rebuilds library, discards it because it matches what `libz3` provides ❌ 
  * `z3-solver` (py3.11) - rebuilds library, discards it because it matches what `libz3` provides ❌ 
  * `z3-solver` (py3.12) - rebuilds library, discards it because it matches what `libz3` provides ❌ 
  * `z3-solver` (py3.13) - rebuilds library, discards it because it matches what `libz3` provides ❌ 
  * `z3-solver` (py3.14) - rebuilds library, discards it because it matches what `libz3` provides ❌ 

If there was an easy way to install just the python bindings on top of an already-installed `libz3` (everything for us is happens in one central but relocatable `$PREFIX`), that would allow us to cut this down to building the library once, and then do only very quick installs for the python bindings. This would cut the build time from 3-5h down to ~1h.

I have enough experience with this that I could _probably_ hack something together (and we do test our packages before publishing), but then it would be me who has to maintain it, and I was hoping that this is a reasonable request that the project would be willing to accommodate. 😅🙃

h-vetinari:
Built on top of #8028 (rebased on master), tested successfully in https://github.com/conda-forge/z3prover-feedstock/pull/15 (where those 7 commits are backported to 4.15.4).

One of the key points was that we need to avoid `add_subdirectory(src)`, otherwise the whole library will be rebuilt. However, putting that step behind `if (Z3_BUILD_LIBZ3_CORE)` means that some quantities need to move out to the main CMakeLists.txt, otherwise they are missing.

I'm not "married" to the exact approach I took, I tried to keep changes minimal; at least it's enough for now to achieve the goal of not rebuilding the library N times pointlessly.

Fixes #8016

Closes #8028

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -548,6 +548,32 @@ set(Z3_GENERATED_FILE_EXTRA_DEPENDENCIES
 )
 
 ################################################################################
+# API header files
+################################################################################
+# This lists the API header files that are scanned by
+# some of the build rules to generate some files needed
+# by the build; needs to come before add_subdirectory(src)
+set(Z3_API_HEADER_FILES_TO_SCAN
+  z3_api.h
+  z3_ast_containers.h
+  z3_algebraic.h
+  z3_polynomial.h
+  z3_rcf.h
+  z3_fixedpoint.h
+  z3_optimization.h
+  z3_fpa.h
+  z3_spacer.h
+)
+set(Z3_FULL_PATH_API_HEADER_FILES_TO_SCAN "")
+foreach (header_file ${Z3_API_HEADER_FILES_TO_SCAN})
+  set(full_path_api_header_file "${CMAKE_CURRENT_SOURCE_DIR}/src/api/${header_file}")
+  list(APPEND Z3_FULL_PATH_API_HEADER_FILES_TO_SCAN "${full_path_api_header_file}")
+  if (NOT EXISTS "${full_path_api_header_file}")
+    message(FATAL_ERROR "API header file \"${full_path_api_header_file}\" does not exist")
+  endif()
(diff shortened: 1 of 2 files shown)


h-vetinari:

h-vetinari:
Thanks for merging this! 🙏

PR #8112, opened 2026-01-03
set source & target version java compile flags in cmake build to match the python/make build for consistent bytecode generation

### LINKED ISSUE #8111: CMake build system compiles Java bindings with inconsistent Java source & target version
### Problem
The CMake build system compiles the Java bindings with the JDK version of the build environment as target instead of a consistent, controlled Java source & target version.
This causes compatibility issues when building with newer JDKs.

### Current behavior
When building the java bindings with CMake using JDK 25 for example, the resulting com.microsoft.z3.jar contains Java 25 bytecode.

### Expected behavior
Java bindings should compile with consistent source/target compatibility regardless of build JDK version and build system

### Root Cause
No -source or -target compile flags are specified in the CMake build, while the python script generating the Makefile [specifies](https://github.com/Z3Prover/z3/blob/5dc812728e50b003b5dfdf1481bd18bb13f5d24a/scripts/mk_util.py#L1932) -source 1.8 -target 1.8.

bu99ed:
fixes #8111

### DIFF
--- a/src/api/java/CMakeLists.txt
+++ b/src/api/java/CMakeLists.txt
@@ -187,6 +187,8 @@ add_custom_target(build_z3_java_bindings
 # TODO: Should we set ``CMAKE_JNI_TARGET`` to ``TRUE``?
 # REMARK: removed VERSION to fix issue with using this to create installations.
 
+set(CMAKE_JAVA_COMPILE_FLAGS -source 1.8 -target 1.8)
+
 add_jar(z3JavaJar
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
this hardwires to 1.8. Is that really the best approach?

NikolajBjorner:
given that the damage was already done in the python build, it can be a separate fix to update versions. 1.8 appears to be from 2014.

PR #8211, opened 2026-01-16
Lattice based refutation

lorenzwinkler1:
This Pull request implements a prototype of a lattice-based-refutation mechanism for the theory of finite sets, as proposed in [the projects page of the TU Wien guest lectures](https://z3prover.github.io/slides/guest-lectures-wien-2025/projects.md).

In particular, this implementation takes into account partial order constraints, and leverages transitivity rules to identify unsatisfiable assignments, based on those rules.

- The implementation keeps track of subset relations $x_1\subseteq x_2$, as well as non-relations $x_3\not\subseteq x_4$.
- When a direct conflict is identified, i.e. $x_1\subseteq x_2$ and $x_1\not\subseteq x_2$ is asserted, a conflict is triggered using `ctx.set_conflict(...)` with a justification for all the subsetrelations
- When a cycle is identified, i.e. $x_1\subseteq x_2 \subseteq ... \subseteq x_n \subseteq x_1$, then equality between all pairs in the chain is asserted through `ctx.add_eq(x_i, x_{i+1},...)`.

### Technical Details

The refutation system is called through two callbacks in the functions `new_diseq_eh` and `new_eq_eh` of the `theory_finite_set`-class.
Since subset relations are expressed through intersections, they are obtained by matching the pattern `x_1 = intersect(x1, x2)`, in which case $x_1\subseteq x_2$. If this pattern can not be matched, then for arbitrary expressions `expr1 = expr2`, we set $expr1\subseteq expr2$ and $expr2 \subseteq expr1$.

Reachability information is stored in the following variables:
- `reachable`: bit-matrix, where `reachable[i,j]` (or to be precise `reachable[i*NUM_WORDS+j/64]&(1<<j%64)`) indicates that the theory variable `j` is a subset of the theory variable `i`. This matrix is updated through row-wise or-operations whenever a new subset relation is added.
- `links`: stores the equalities justifying a direct link (used for building justifications)
- `non_links`: bit-matrix, where `non_links[i,j]` (or to be precise ...) indicates, that `i` is not a subset of `j`.
- `non_link_justifications`: stores the enodes involved in the disequality justifying the non-subset relation
- `link_dls`: stores the decision levels where the links were asserted (see limitations for a reason why this is stored).

#### Large Conflict Justifications
In some cases, the conflict clauses of the partial-orders become very large, and span multiple decision levels. This is for instance the case in the `example-assignment-large.smt2` example. If the refutation system is turned on, then z3 becomes way slower, since the refutation system basically creates a conflict clause for every combination of chosen or-literals. This results in about 65000 refutations, and results in a runtime of over 1 second (instead of ~40ms). Note that the increased runtime does not stem from some kind of overhead due to tracking the relations - if just the line of code that sets the conflict is commented out, the runtime is nearly unchanged.
To overcome this issue, a conflict is not triggered when the number of conflict literals with non-zero decision level is greater than 10.

#### Hardcoded number of relations
Currently reachability (=partial order relation) is saved only for a maximum number of theory variables, hardcoded to 5*64. The reason for that is, that for larger instances the overhead could become too large, and also resizing of the reachability matrix would be quite expensive. Implementing some resizing functionality would be potential future work, although I doubt that it would be very useful.

### Small Experiments
For the added examples you can find below some experiments (that are to be taken with a grain of salt).
The runs were performed with Release configuration, and recorded using hyperfine with 100 runs per entry.

| name | without refutation | with refutation | comments |
| :---         |     ---:      |          ---: | :--- |
| `example-chain.smt2`   | 25.1 ms ±   3.1 ms     | 24.2 ms ±   2.7 m    |  |
| `example-assignment.smt2`     | 26.6 ms ±   3.0 ms    | 26.0 ms ±   2.9 ms      |  |
| `example-assignment-large.smt2`     | 42.1 ms ±   5.5 ms  | 46.5 ms ±   5.3 ms  | here no conflicts were created - the additional time thus approximately captures the overhead for this example |
| `cycle.smt2`     | 29.0 ms ±   3.4 ms       | 24.4 ms ±   3.3 ms      |  |


src/smt/theory_finite_set_lattice_refutation.cpp
```diff
@@ -0,0 +1,307 @@
...
+            return;
+        }
+        if (subset_th == null_theory_var || superset_th == null_theory_var){
+            return;
+        }
+        reachability.set_reachability(subset_th, superset_th, justifying_equality);
+        SASSERT(reachability.is_reachable(subset_th, superset_th));
+        if(reachability.is_reachable(superset_th, subset_th)){
+            TRACE(finite_set, tout << "cycle_detected: " << subset_th << " <--> " << superset_th);
+            vector<enode_pair> path;
```
lorenzwinkler1, line 267: should space for this vector be allocated in a specific place?

src/smt/theory_finite_set_lattice_refutation.cpp
```diff
@@ -0,0 +1,307 @@
...
+            }
+            SASSERT(success);
+        }while(source != dest);
+        TRACE(finite_set, tout << "get_path_num_decisions: "<<num_decisions);
+    }
+
+    bool reachability_matrix::check_reachability_conflict(theory_var source, theory_var dest){
+        if(is_reachable(source,dest) && is_reachability_forbidden(source, dest)){
+            TRACE(finite_set, tout << "found_conflict1: "<<source<<" -> "<<dest);
+            vector<enode_pair> path;
```
lorenzwinkler1, line 187: should space for this vector be allocated in a specific place?

PR #8264, opened 2026-01-21
preserve the initial state of the solver with push/pop for multiple objectives

levnach:
This pull request, it seems, fixes three bugs. One algorithmic in opt_solver.cpp.cpp, and two memory corruption bugs.


numinit:
This fixed z3 in nixpkgs, and we're applying it as a patch, thanks. Could use a newly tagged version soon?

Issue #8036, opened 2025-11-20
Example Missing for Z3_parser_context


Toby-Shi-cloud:
I like the incremental paring future (#6123).
However, there seems to be no example for any languages about `Z3_parser_context`.

I took much time looking into the commit 815518d to find how to use `Z3_parser_context` in c++.

It will be really helpful if z3 add incremental paring examples.

NikolajBjorner:
I closed a PR generated by copilot with incremental parsing examples.
Copilot didn't generate proper c++ calls so the example is somewhat low quality, but may be enough to get an idea.
Also, the python bindings let you use incremental parsing for a minimal mcp server, https://github.com/Z3Prover/z3/blob/master/src/api/mcp/z3mcp.py

PR #8722, opened 2026-02-21
Add `parser_context` in C++ API

Toby-Shi-cloud:
#8036 

I add a wrapper class for `Z3_parser_context` in C++ api.

I think this would be useful for users who want an incremental parser.

Moreover, I found no `explicit` keyword in C++ api, especially for boolean conversation. Sometimes z3 objects are accidentally converted to int or other integer types. `explicit` is introduced in C++11, which is not a new feature at 2026. If z3 wants to support C++03, maybe we can use a macro to solve this issue.

### DIFF
--- a/src/api/c++/z3++.h
+++ b/src/api/c++/z3++.h
@@ -868,6 +868,44 @@ namespace z3 {
 
     };
 
+    class parser_context : public object {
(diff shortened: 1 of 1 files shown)


Toby-Shi-cloud:

PR #8752, opened 2026-02-24
Nl2lin

ValentinPromies:
Merge master into nl2lin and move linear cell construction into levelwise

### DIFF
--- a/src/nlsat/nlsat_explain.cpp
+++ b/src/nlsat/nlsat_explain.cpp
@@ -1733,7 +1733,7 @@ namespace nlsat {
             
             // We call levelwise directly without normalize, simplify, elim_vanishing to preserve the original polynomials
             var max_x = max_var(m_ps);
-            bool levelwise_ok = levelwise_single_cell(m_ps, max_x, m_cache, true);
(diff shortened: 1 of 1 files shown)


levnach:
There were some issues.
Z3 has a test environment at https://github.com/Z3Prover/z3test. It is a good idea to clone it ,for example, to directory “tz3” and run “python tz3/scripts/test_benchmarks.py ./z3 tz3/regressions/smt2”. This test currently gives me several incorrect results and a few timeouts: and it should work even on the debug version of z3.
I noticed that the new code produces empty explanations, which cause the unsound “unsat” results.
There was also exponential growth of the coefficients on , for example, problem tz3/regressions/smt2/8099.smt2.
Some more useful debug tricks:
Run z3 with nlsat.check_lemmas=true: I believe this option works now, even in the release build.
Run z3 with nlsat.log_lemmas=true, capture the output, including stderr, and then run z3 again on the output. There should not be lemmas which are “sat”.
I have not tried the last two options on the PR.

NikolajBjorner:
you can just run the CI/CD checks. They will run z3test regression tests.

NikolajBjorner:
Z3  build/z3) produced unexpected output for z3test/regressions/smt2/3689.smt2:
EXPECTED
sat
PRODUCED
unsat

PR #8776, opened 2026-02-26
fix: replace 33 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


NikolajBjorner:
why?

PR #8785, opened 2026-02-26
Expose `mk_re_allchar` in OCaml API

filipeom:
Noticed it was missing.


filipeom:

PR #8792, opened 2026-02-27
fix: replace 33 bare except clauses with except Exception

haosenwang1018:
Replace bare `except:` clauses with `except Exception:` for PEP 8 compliance.


haosenwang1018:
CLA signing required, cannot proceed with automation. Closing per project policy.

PR #8856, opened 2026-03-04
Add ASan/UBSan memory safety CI workflow

angelica-moreira:
Adds a CI workflow that checks for memory safety issues on every push using two runtime sanitizers:

   - ASan: Builds and tests Z3 with AddressSanitizer to catch runtime memory errors
   - UBSan: Builds and tests Z3 with UndefinedBehaviorSanitizer to catch undefined behavior

 Also includes a Copilot agentic workflow that generates a summary report as a GitHub Discussion when findings are detected. The workflow runs on push (path-filtered to src/) and can also be triggered manually with a full-scan option. No existing files are modified, all changes are additive under .github/workflows/.

### DIFF
--- a/.github/workflows/memory-safety-report.md
+++ b/.github/workflows/memory-safety-report.md
@@ -1,0 +1,205 @@
+---
+description: >
+  Generates a detailed Memory Safety report for Z3 by analyzing ASan/UBSan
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
is the md file to be compiled using aw?

PR #8977, opened 2026-03-13
 Add missing AST query methods to Java API

angelica-moreira:
**Summary**

This PR exposes several small utility functions from the Z3 C API that were not available in the Java API.

**Added methods**

- `Expr.isGround()`:  returns true if the expression has no free variables.
- `Expr.isLambda()`:  returns true if the expression is a lambda quantifier.
- `AST.getDepth()`: returns the maximum depth of the AST (longest root-to-leaf path).
- `ArraySort.getArity()`: returns the number of dimensions of the array sort.
- `DatatypeSort.isRecursive()`: returns true if the datatype refers to itself.
- `FPExpr.isNumeral()`: returns true if the floating-point expression is a concrete value.

**Implementation Details**

Each method simply calls the existing auto-generated Native binding.

**Notes**

- `Constructor.getNumFields()` was considered but already exists in the Java API and returns a cached value.
- Added examples to JavaExample.java covering all new methods.
- Inspired by #8762, which addressed a similar gap where C API functions were available through `Native.*` but lacked typed Java wrappers. Without wrappers, users must call `Native.*` directly, which bypasses type safety and is less convenient.

### DIFF
--- a/examples/java/JavaExample.java
+++ b/examples/java/JavaExample.java
@@ -2392,6 +2392,28 @@ class JavaExample
         System.out.println("FpNumeralExample passed.");
     }
 
+    @SuppressWarnings("unchecked")
+    void isLambdaExample(Context ctx) throws TestFailedException
+    {
+        System.out.println("IsLambdaExample");
+        Log.append("IsLambdaExample");
+
+        // build lambda x : Int . x + 1
+        IntExpr x = (IntExpr) ctx.mkBound(0, ctx.getIntSort());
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
@angelica-moreira - I am baby sitting an agentic workflow for API coherence. It works in parts. The process isn't optimal: a lot of this can be automated with a script that greps specifically for z3_api C functions in the other front ends (for Java it converts the names into Caml-back form). It should catch these things, but it also priortizes changes.

PR #9038, opened 2026-03-18
fix #8185: add FPA rewriter simplifications for fma with zero multipl…

### LINKED ISSUE #8185: Incorrect model in mixed FP/Real logic triggered by extra string constraint
Z3 reports sat but generates an invalid model `(fails model_validate=true)`.

Crucially, this behavior is triggered by the inclusion of a seemingly irrelevant String constraint: `(assert (and x (= (str.from_int 0) "0")))`. If I remove this constraint or remove `(set-option :produce-proofs true)`, Z3 returns unknown instead of producing the incorrect model.

Input:
```smt2
(set-option :produce-proofs true)
(declare-const x8 Int)
(declare-const x Bool)
(declare-fun s (Real) Real)
(assert (> 
    (+ (s 0.0) 
        (ite (fp.isNormal 
            (fp.fma RNE 
                (fp (_ bv0 1) (_ bv0 8) (_ bv0 23)) 
                ((_ to_fp 8 24) RNE (to_real x8)) 
                ((_ to_fp 8 24) RNE 0.5)))
            x8 
            1)) 
    0.0)
)
(assert (and x (= (str.from_int 0) "0")))
(check-sat)
(get-model)
```

Output:
```smt2
z3 test.smt2 model_validate=true
sat
(error "line 17 column 10: an invalid model was generated")
(
  (define-fun x () Bool
    true)
  (define-fun x8 () Int
    0)
  (define-fun /0 ((x!0 Real) (x!1 Real)) Real
    (/ 1.0 126.0))
  (define-fun s ((x!0 Real)) Real
    0.0)
)
```
For reference, cvc5 solves this instance correctly and produces a valid model.
```smt2
sat
(
(define-fun x8 () Int (- 2))
(define-fun x () Bool true)
(define-fun s ((_arg_1 Real)) Real 3.0)
)
```

Commit: eca8e192312fface752121b427f22061b6db0db3

levnach:
…icand and classification of int-to-float conversions

Three optimizations that eliminate expensive FP bit-blasting:

1. fma(rm, ±0, y, z): decompose into simpler operations instead of building the full FMA bit-blast circuit. When z is a concrete non-zero value, simplifies to ite(isFinite(y), z, NaN).

2. isNaN/isInf/isNormal applied to to_fp(rm, to_real(int_expr)): convert to pure integer constraints. For integers, to_fp never produces NaN or subnormals, so classification reduces to overflow threshold checks based on the rounding mode.

3. Push classification functions (isNaN, isInf, isNormal) through ite when both branches are concrete FP numerals.

Together these transform the original formula from requiring full FP bit-blasting (timeout) to pure integer/real reasoning (instant).

### DIFF
--- a/src/ast/rewriter/fpa_rewriter.cpp
+++ b/src/ast/rewriter/fpa_rewriter.cpp
@@ -428,6 +428,86 @@ br_status fpa_rewriter::mk_fma(expr * arg1, expr * arg2, expr * arg3, expr * arg
             result = m_util.mk_value(t);
             return BR_DONE;
         }
+
+        // fma(rm, ±0, y, z) or fma(rm, x, ±0, z): when one multiplicand is
+        // a concrete zero the full FMA bit-blast circuit is unnecessary.
+        // IEEE 754: 0 * finite = ±0, so fma reduces to fp.add(rm, ±0, z).
+        //           0 * NaN = NaN, 0 * inf = NaN.
+        expr *other_mul = nullptr;
+        bool zero_is_neg = false;
+        scoped_mpf vzero(m_fm);
+        if (m_util.is_numeral(arg2, vzero) && m_fm.is_zero(vzero)) {
+            other_mul = arg3;
+            zero_is_neg = m_fm.is_neg(vzero);
+        }
+        else {
+            scoped_mpf v3tmp(m_fm);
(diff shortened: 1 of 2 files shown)


levnach:
This is a big change created by Claude Opus 4.6.  It needs a review by people familiar with operations on floats.

PR #9115, opened 2026-03-24
Mfact

levnach:
Improve polynomial.cpp by adding factorization of multivariate polynomials with Hensel lifting.


levnach:
In all tests this branch slightly outperforms master.

PR #9249, opened 2026-04-08
Horner emon bounds

1arie1:
An attempt to harden agains solve_eqs that splits monomials.

First commit seems like a genuine bug.

Second is my attempt to use Horner to recover monomials that are split. 
Less sure about that one


1arie1:
Seems like unrelated tests are failing. I'll wait until they are fixed on master and rebase

1arie1:
@NikolajBjorner this should be good for unit tests. Please look at the commit message and the change.

NikolajBjorner:
I am not sure about the AI diagnostics.

- Should we really use fold-unfold instead of solve-eqs?
- The diagnostics for lar_term::== doesn't address whether returning false is _correct_, rather how it is used in historical context.

1arie1:
The problem is with this map: https://github.com/Z3Prover/z3/blob/9c81571eb88b6714161a4c1ae028f6dee4c6d95e/src/math/lp/lar_solver.cpp#L70

Every lookup fails because `term_comparer` forwards to `lar_term::==`, and that is always false. 

My expectation is that `lar_term` is correct, but `term_comparer` is broken.

1arie1:
I forgot about fold-unfold. Is it in `master`? 

For now, I'm trying various ways to limit solve_eqs or power up Horner. 
What I have so far helps a bit (i.e., with minimized benchmarks), but not completely (i.e., full benchmarks).

Mostly I'm looking for simple non-invasive things to see how far I get with my helpers.

1arie1:
found fold-unfold...

1arie1:
can I/you bring fold-unfold into master?

NikolajBjorner:
> can I/you bring fold-unfold into master?

It is there now. 
Optionally we allow configuring with something like solve_eqs.conservative (bool) to set fold-unfold instead of solve-eqs?

1arie1:
@NikolajBjorner I will try it out and get back to you soon. 

This, however, does not invalidate the PR. Please see my answer above about `term_comparer`

NikolajBjorner:
> The problem is with this map:
> 
> https://github.com/Z3Prover/z3/blob/9c81571eb88b6714161a4c1ae028f6dee4c6d95e/src/math/lp/lar_solver.cpp#L70
> 
> Every lookup fails because `term_comparer` forwards to `lar_term::==`, and that is always false.
> 
> My expectation is that `lar_term` is correct, but `term_comparer` is broken.

@levnach - see code review. this is my understanding too. Why is == not performing structural equaltiy test?

levnach:
I believe initially lar_term::operator== was required by some template and I gave the simplest implementation, that was not invoked anywhere up to my knowledge.
Later the map in question was introduced, and I did not catch this behavior.
term_hasher  does not behave correctly as well: term_hasher should not depend on the order elements in lar_term::m_coeffs. 
Claude is saying that the only usage of lar_term::operator()== is in this map. 
I would suggest we fix this operator, and make term_hasher order invariant.

1arie1:
My understanding was that `lar_term` was unique, so structural equality is not expected, with the exception of a few cases like looking up a key in a map. For that reason, disabling equality (for example, replacing it by assert false or unreachable) is desirable. If it is really not used anywhere, this should be easy to test. In this case, my definition of `term_comparer` is the correct solution because it introduces equality specifically for key lookups, preventing accidental misuse.

For the `term_hasher` -- I am expecting `lar_term` to be normalized so that the variables are sorted. If that is the case, the current hash is not dependent on order of the elements. If my assumption is wrong and terms are not normalized -- then why? is it hard? If terms are not normalized, there are likely other places in the code that are affected. 

Note that current hasher is optimized for long terms. It uses at most 10 coefficients. I don't see how to preserve this without normalizing terms.

I suggest separating comparer and hasher changes. The comparer change is very local and easy to evaluate. hasher change might have other effect on performance and should be evaluated separately.

1arie1:
It seems like the issue with `lar_term` is deeper because it affects equality of `term_o` that is a subclass of `lar_term`. Also the equality in debug build is defined differently from equality in release mode. I'm going to create a new PR focusing on just `==` and `!=` operators of `lar_term` so that we can discuss that without other changes in the PR.

NikolajBjorner:
@1arie1 - you should merge with master. Apparently I cannot push to this fork

1arie1:
This PR has a mix of lar_term fixes and new functionality for Horner. I will re-create it against master

PR #9259, opened 2026-04-09
Fix scaled_min and max_reg test failures

levnach:
## Summary

Fixes two failing tests in `test-z3 /a`: **scaled_min** and **max_reg**.

### max_reg — wrong assertion count
Commit bb48e3a disabled 5 weighted-sum tests with `#if 0`, reducing active optimizations from 7 to 2, but changed the assertion to expect 6 instead of 2.

**Fix:** Update assertion to `ENSURE(num_sat == 2)`.

### scaled_min — optimizer regression from #9235
The mod-factor-propagation commit (#9235) added `ensure_nla()` + `add_bounded_division` in the `is_mod` handler. For problems containing `mod` but no nonlinear multiplication (like the scaled_min test), this unnecessarily created the NLA solver, whose final-check lemmas caused the optimizer to converge to a suboptimal finite value instead of `-infinity`.

**Fix:**
- Introduce a separate `m_mod_divisions` vector (`(x, y, r)` tuples) used exclusively by `check_mod_mult()`, keeping `m_bounded_divisions` clean.
- Remove unnecessary idiv term creation from the `is_mod` handler.
- Defer mod-division registration: buffer in `m_pending_mod_divisions` when NLA is not yet active; flush when `ensure_nla()` first creates the NLA solver.

All 89 tests pass after these changes.

### DIFF
--- a/src/math/lp/nla_core.h
+++ b/src/math/lp/nla_core.h
@@ -218,6 +218,7 @@ public:
     void add_idivision(lpvar q, lpvar x, lpvar y, lpvar r) { m_divisions.add_idivision(q, x, y, r); }
     void add_rdivision(lpvar q, lpvar x, lpvar y, lpvar r) { m_divisions.add_rdivision(q, x, y, r); }
     void add_bounded_division(lpvar q, lpvar x, lpvar y, lpvar r) { m_divisions.add_bounded_division(q, x, y, r); }
+    void add_mod_division(lpvar x, lpvar y, lpvar r) { m_divisions.add_mod_division(x, y, r); }
     void set_add_mul_def_hook(std::function<lpvar(unsigned, lpvar const*)> const& f) { m_add_mul_def_hook = f; }
     lpvar add_mul_def(unsigned sz, lpvar const* vs) { SASSERT(m_add_mul_def_hook); lpvar v = m_add_mul_def_hook(sz, vs); add_monic(v, sz, vs); return v; }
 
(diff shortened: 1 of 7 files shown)


NikolajBjorner:
@1arie1 - your PR got redone.

agurfinkel:
This does not look good. From the description, I see that the core issue is forcing nla when it should not be forced. 

However, the changes seem a bit chaotic. I will try to address the main concern first...

NikolajBjorner:
the only thing that was required to fix the unit test was to change ENSURE(6 == blah) to ENSURE(2 == blah).
Why are we now adding more changes?
Arie's PR also "looked chaotic" to me. It included a new argument that was tracked. I initially trusted Arie to have checked this to be necessary, but with unit tests failing I don't know.

levnach:
Two tests were failing. The change in theory_lra fixing the second failing test.

agurfinkel:
:)

In theory_lra during linearize, when certain division terms are detected, extra axioms are injected, and, divisions are registered in nla solver. The handling of different choices (idiv term first versus mod term first) are not handled consistently. Some branches checks that nla is already available, while others force nla by a call to `ensure_nla()`.

My changes followed the convention of nearby code to call `ensure_nla`. I will fix it to be less intrusive.
I will leave the logic as it is on all other branches.

The proposed fixed delayed mod registration with nla. This might be a better long term solution, but it was applied narrowly to mod only. The table registering div and mod was split. The fix did not preserve all prior behavior either.

agurfinkel:
Proposed fix: https://github.com/Z3Prover/z3/pull/9260

PR #9260, opened 2026-04-10
Fix scaled_min test failure from #9235 mod-factor-propagation

1arie1:
The is_mod handler in theory_lra called ensure_nla(), which unnecessarily created the NLA solver for pure linear problems, causing the optimizer to return a finite value instead of -infinity.

Fix: check `m_nla` instead of calling `ensure_nla()`, matching the pattern used by the is_idiv handler. The mod division is only registered when NLA is already active due to nonlinear terms.

Update mod_factor tests to use QF_NIA logic and assert the mul term before the mod term so that internalize_mul triggers ensure_nla() before mod internalization.


1arie1:
I will add a fix to the other failing test as a separate commit on this PR

1arie1:
rebased against existing fix, should be good now

agurfinkel:
@NikolajBjorner this seems good now

PR #9284, opened 2026-04-12
Fixes for lar_term== operator

1arie1:
The main changes are

1. rewrite `term_comparator` to do structural equality
2. delete `lar_term::operator==` and `lar_term::operator!=`
3. change `lar_term==` to reference equality in one place it is used

The code compiles in Debug mode but not in Release mode because `term_o`, a sub-class of `lar_term`, overrides equality in Debug mode only.

### DIFF
--- a/src/math/lp/lar_solver.cpp
+++ b/src/math/lp/lar_solver.cpp
@@ -36,7 +36,12 @@ namespace lp {
 
         struct term_comparer {
             bool operator()(const lar_term& a, const lar_term& b) const {
-                return a == b;
+                if (a.size() != b.size()) return false;
+                for (const auto& p : a) {
+                    auto const* e = b.coeffs().find_core(p.j());
+                    if (!e || e->get_data().m_value != p.coeff()) return false;
+                }
+                return true;
             }
         };
 
--- a/src/math/lp/lar_term.h
+++ b/src/math/lp/lar_term.h
@@ -129,8 +129,8 @@ public:
         add_monomial(a, v1);
         add_monomial(b, v2);
     }
-    bool operator==(const lar_term & a) const {  return false; } // take care not to create identical terms
-    bool operator!=(const lar_term & a) const {  return ! (*this == a);}
+    bool operator==(const lar_term & a) const = delete; //  take care not to create identical terms
+    bool operator!=(const lar_term & a) const = delete;  
     // some terms get used in add constraint
     // it is the same as the offset in the m_constraints
 
--- a/src/math/lp/nla_types.h
+++ b/src/math/lp/nla_types.h
@@ -42,7 +42,7 @@ namespace nla {
         ineq(lpvar v, lp::lconstraint_kind cmp, int i): m_cmp(cmp), m_term(v), m_rs(rational(i)) {}
         ineq(lpvar v, lp::lconstraint_kind cmp, rational const& r): m_cmp(cmp), m_term(v), m_rs(r) {}
         bool operator==(const ineq& a) const {
-            return m_cmp == a.m_cmp && m_term == a.m_term && m_rs == a.m_rs;
+            return m_cmp == a.m_cmp && &m_term == &a.m_term && m_rs == a.m_rs;
         }
         const lp::lar_term& term() const { return m_term; };
         lp::lconstraint_kind cmp() const { return m_cmp;  };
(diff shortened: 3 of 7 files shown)


1arie1:
@NikolajBjorner @levnach This isolates the issue with `lar_term` equality. I will add comments with questions to this PR.

Note that the code compiles in Debug mode but not Release. I will add comments for that as well.

src/math/lp/lar_solver.cpp
```diff
@@ -36,7 +36,12 @@ namespace lp {
 
         struct term_comparer {
             bool operator()(const lar_term& a, const lar_term& b) const {
-                return a == b;
+                if (a.size() != b.size()) return false;
+                for (const auto& p : a) {
```
1arie1, line 40: if `lar_term` are normalized, this can be done more efficiently instead of linear scan through `b`

src/math/lp/lar_term.h
```diff
@@ -129,8 +129,8 @@ class lar_term {
         add_monomial(a, v1);
         add_monomial(b, v2);
     }
-    bool operator==(const lar_term & a) const {  return false; } // take care not to create identical terms
-    bool operator!=(const lar_term & a) const {  return ! (*this == a);}
+    bool operator==(const lar_term & a) const = delete; //  take care not to create identical terms
```
1arie1, line 132: if these operators are not to be used, it is best to delete them rather than provide a trivial implementation.

src/math/lp/nla_types.h
```diff
@@ -42,7 +42,7 @@ namespace nla {
         ineq(lpvar v, lp::lconstraint_kind cmp, int i): m_cmp(cmp), m_term(v), m_rs(rational(i)) {}
         ineq(lpvar v, lp::lconstraint_kind cmp, rational const& r): m_cmp(cmp), m_term(v), m_rs(r) {}
         bool operator==(const ineq& a) const {
-            return m_cmp == a.m_cmp && m_term == a.m_term && m_rs == a.m_rs;
+            return m_cmp == a.m_cmp && &m_term == &a.m_term && m_rs == a.m_rs;
```
1arie1: @levnach what is the intention here? I've changed comparison to reference compare. Note that this comparison was always false in the current code. Not sure what else this will affect.

src/test/nla_intervals.cpp
```diff
@@ -207,13 +207,82 @@ void test_nla_intervals_fractional() {
     VERIFY(true); // Placeholder
 }
 
+void test_fetch_normalized_term_column() {
```
1arie1, line 210: this is a test for `term_comparer` implementation.

1arie1:
The problem with Debug vs Release is tis block

https://github.com/Z3Prover/z3/blob/1544462f47efc1bb267d30407aef711eee515e0b/src/math/lp/dioph_eq.cpp#L241-L252

It overrides `==` and `!=` for debug, but is not present in release. I suggest this code should always be enabled.

src/math/lp/nla_types.h
levnach: Sorry for the mess, it was never meant to be used in release mode. This should be deleted if nobody uses this map.

1arie1: this is not a map. this is part of `ineq` class

1arie1: if `lar_term` is canonicalized, my fix here is correct. Two `ineq` are equal if they have exact same term (i.e., identity vs equality). This also depends on whether `ineq::operator==` is on some performance critical path. Doing structural comparison in that case might be too expensive.

1arie1:
`Ineq::operator==` is not used anywhere. So I've deleted it and `ineq::operator!=` to go with it.

Everything is in a good state in this PR. 
1. `lar_term` has no `==` and no uses
2. map that uses `lar_term` as a key provides structural equality
3. `o_term` does not inhering `operator==` and uses specifically defined `eq()` friend function.

If there are no performance regressions, this should be good.

Q: Is `lar_term` sorted/normalized? If yes, structural equality can be improved. If not, I'll need to review what else might be affected.

1arie1:
This is another `lar_term` function that confuses me

https://github.com/Z3Prover/z3/blob/1566d3cc413a22458fec020cde8961d5e5d7dd94/src/math/lp/lar_term.h#L299-L316

The name says that normalization is by min var, but the implementation is minimization by first var. 
I assumed that terms are ordered by variables, but maybe that is not a correct assumption.

1arie1:
Once again forgot to check what happens in unit tests :(

The `ineq::operator==` is used in tests. I have to investigate what is expected there. Interestingly, that equality was always false in prior builds, so what was the test checking...

levnach:
> Once again forgot to check what happens in unit tests :(
> 
> The `ineq::operator==` is used in tests. I have to investigate what is expected there. Interestingly, that equality was always false in prior builds, so what was the test checking...

It is dead code, should be removed.

1arie1:
I've commented out the offending code. If that is the correct course of action, I'm happy to remove it completely.

Waiting to see whether CI is happy.

PR #9376, opened 2026-04-24
Core Minimization Thread + some bugfixes to backtracking on backbones

ilanashapiro:
There are strong signals on LIA and NIA for including the core min thread:
run-1411-Z3-threads-4-qfnia-core_min-stats.md
run-1410-Z3-threads-4-qflia-core_min-stats.md
run-1409-Z3-threads-4-qfnia-stats.md
run-1408-Z3-threads-4-qflia-stats.md
http://mtzguido.tplinkdns.com:8081/z3/compare_stats.html

This PR implements the core minimization algorithm on a dedicated thread, and re-prunes the search tree with the minimized core if it's smaller than the original node's current core (i.e. if the original node's core wasn't changed/made smaller from UNSAT propagation from another thread)

Also, there are some bugfixes: each time we close a node for the first time (i.e. not re-backtracking with a smaller core), we also close valid external targets, which are other nodes in the search tree that have the same literal and whose path contains the current core. this is especially useful for closing the negation of proven backbones (since their core is just the negated backbone). Finally, I removed the epoch field from node leases, since this is now obselete, and fixed a small bug where we need to look at the shallowest timed-out leaf, not the shallowest activated leaf, in the tree expansion algorithm (per-SMTS).

Also, the conflict budget for each flipped literal check in core min is currently set to 5000. for the hard SMT COMP problems we probably need to increase this.


ilanashapiro:
making more changes, closing for now

PR #9391, opened 2026-04-25
Improvements to NLA lemmas

1arie1:
Monomial sandwich

Monomial bounds

McCormick tangents

see individual commit messages for details


agurfinkel:
@levnach @NikolajBjorner I went carefully over this, but would appreciate a second look.

everything is gated and off by default. there is no change to default behavior

1arie1:
I would like to trigger copilot review again if that is possible

PR #9405, opened 2026-04-27
Use the minimum generation number among matching enodes

CanCebeci:
This patch is meant to improve stability. It ensures that the generation number corresponds to the shortest path of instantiations for the term to be created. This essentially prevents cases where the generation number is determined by some high-generation term that sticks around after backtracking.


CanCebeci:
@microsoft-github-policy-service agree company="EPFL IC IINFCOM SYSTEMF"

PR #9411, opened 2026-04-28
Build Z3 through Bazel natively

EdSchouten:
Right now Z3 ships with a BUILD.bazel file. However, by the looks of it it's slightly pointless. It merely invokes CMake using rules_foreign_cc. This takes away most of the benefits of using Bazel, such as fine-grained caching and being able to parallelize builds across multiple machines.

In my case I've observed the existing rules_foreign_cc based build to be very slow and also not reproducible. Let's solve this by replacing BUILD.bazel with one that simply declares a cc_library() target. We name this one "z3", so that it's possible to depend on it using "@z3".

For compatibility we let "z3_dynamic" and "z3_static" be aliases pointing to the new "z3" target.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake")
+load("@rules_cc//cc:defs.bzl", "cc_library")
 load("@rules_license//rules:license.bzl", "license")
 
 package(default_applicable_licenses = [":license"])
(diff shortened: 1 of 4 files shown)


NikolajBjorner:
@smolkaj

EdSchouten:
Before submitting this change I was not aware that contributing required signing a CLA, as it wasn't mentioned in the project's README.

As I am not in a position where I can sign this CLA, I have no choice but to close this PR. My apologies.

PR #9498, opened 2026-05-11
fix edge case in algebraic number comparison

ValentinPromies:
So far, `algebraic_numbers compare_core ` handles an edge case incorrectly: 
- If the two compared numbers (`a`, `b`) are different,
- the intervals still overlap after refinements, and
- both a and b are a root of the second polynomial (`cell_b->m_p`), e.g. they are the first and second root

then the method would return `sign_zero` (i.e. "equal"). This behavior can be replicated with the provided test case (before the fix). This requires `algebraic.factor=false`, though i first encountered it during solver runs on QF_NRA instances with the default `algebraic.factor=true`, which apparently means that the polynomials for anums are still not always factored.

The fix is to compare the interval bounds of b to a and vice versa. Then the Sturm-Tarski check is only run if `a` and `b` both lie in the intersection of the intervals, because only then is it guaranteed to be correct.

### DIFF
--- a/src/math/polynomial/algebraic_numbers.cpp
+++ b/src/math/polynomial/algebraic_numbers.cpp
@@ -2028,6 +2028,20 @@ namespace algebraic_numbers {
             }
             IF_VERBOSE(9, verbose_stream() << "sturm 1\n");
 
+
+            // Check whether a can be separated from b's interval and vice versa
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
@levnach

PR #9500, opened 2026-05-11
Fix FPA/BV model-validation soundness bug for Array + Datatype theories (Fixes #9488)

### LINKED ISSUE #9488: Invalid model was generated
# Invalid model returned for SAT result on FP + datatypes + arrays

## Summary

Z3 returns `sat` on a formula whose final assertion evaluates to `false`
in the returned model. Running with `model_validate=true` confirms Z3's own
model checker rejects the model:

```
$ z3 model_validate=true repro.smt2
sat
(error "line N column M: an invalid model was generated")
```

## Version

[Master branch](https://github.com/Z3Prover/z3/commit/2c7b256db21f364cf37752e91cf750bf0b570f2c)

## Trigger

The bug is triggered by the combination of

1. a `(fp.add x ((_ to_fp 11 53) #x0000000000000000))` term, and
2. a `(fp.sub ((_ to_fp 11 53) RTZ ((_ sign_extend 32) ((_ extract 31 0)
   #x0000000000000000))) (fp.neg x))` term

asserted to feed equivalent post-store memory cells in a `MemCell`
datatype wrapped in nested `Array` selects/stores.

Replacing the two `#x0000000000000000` constants with
`#x3ff0000000000000` (+1.0) yields `unsat` deterministically. Z3's
`model_validate=true` accepts the unsat case.

### Original

https://pastebin.com/RejGjimP

### Minimized (witth [ddSMT](https://github.com/ddsmt/ddSMT))

```
(declare-const x Bool)
(declare-datatypes ((M 0)) (((i (g2 (_ BitVec 4))) (f (e Float64)) x4 (t (g Int)) x5)))
(define-sort MemPtr () (Array Int (Array Int (Array M M))))
(define-const zero_pointer_i_1d (Array Int Int) ((as const (Array Int Int)) 0))
(define-fun-rec str_reverse () String "")
(declare-fun s () MemPtr)
(declare-fun p () (Array Int M))
(assert (= (p 2) (f (fp.sub RNE (e (p 0)) (e (p 0))))))
(assert (= (p 1) (t (g (p 0)))))
(declare-fun m () (Array Int (Array Int Int)))
(assert (= zero_pointer_i_1d (m 0)))
(declare-fun o () (Array Int M))
(assert (= (o 0) (f (e (select (select (s 1) (g (p 2))) (p 0))))))
(check-sat)
```

## Expected

`unsat` (the formula is in fact equivalence-checking
`0 - (-x)` against `x + 0`, which holds in IEEE 754 under SMT-LIB FP
semantics for all `x` including ±0 and NaN).

Note, that changing 0 to 1 makes formula unsat.

## Actual

```
$ z3 repro.smt2
sat

$ z3 model_validate=true repro.smt2
sat
(error "line N column M: an invalid model was generated")
```

tryh4rd-26:
This PR resolves a soundness bug where Z3 generates self-contradictory models on combinations of Floating-Points, Datatypes, and multi-dimensional Arrays, leading to model validation failures under model_validate=true 

Fixes https://github.com/Z3Prover/z3/issues/9488

The issue was tha when an FPA term is wrapped in a non-FPA theory (like a Datatype accessor such as `(e (select p 0)))` , the Datatype theory correctly gives it a value in the model (like `+zero`). But, its bit-blasted Bit-Vector (BV) representative was being left unconstrained. Because of this, the BV solver ends up assigning it a default value (like a NaN bitpattern). This creates a `split-brain` situation where the exact same logical term gets two conflicting values, which is why Z3's model validator was rejecting its own models.

I aligned the model generation logic to ensure that Floating-Point values and their bit-blasted Bit-Vector representatives are perfectly synchronized:

1. Registered `fpa2bv_value_proc` to bind the BV representative's value to the evaluated FPA term's value.
2.  I removed formal dependencies for `fpa2bv_value_proc` to prevent topological sorting cycles in the model generator. Instead, I implemented safe on-the-fly evaluation inside `mk_value` when matching constructors and accessors are encountered.
3. Updated the accessor fallback path to only read a constructor's field if the accessor actually belongs to the active constructor. This prevents split-brain representation mismatches for multi-constructor datatypes.
4. Updated `register_fpa2bv_processors` to accept `roots`. This guarantees that newly registered wrappers are correctly scheduled and evaluated, even if they were not present in the initial relevance scan.
5. Avoid redundant allocations of `fpa2bv_value_proc` when a valid processor is already registered for the same node.

I also verified with:

* Verified that the original reproducer (`final.smt` from the issue) and the minimized reproducer compile, run, and resolve cleanly under `model_validate=true` with no errors, warnings, or crashes.
*  Ran all internal Python unit tests and doctests via `z3test.py` and `z3test.py z3num`. All tests passed successfully with exit code 0.

*Note: Minor usage of github copilot was used to triage for issues and solutions*

### DIFF
--- a/src/params/smt_parallel_params.pyg
+++ b/src/params/smt_parallel_params.pyg
@@ -4,6 +4,9 @@ def_module_params('smt_parallel',
                   params=(
                         ('inprocessing', BOOL, False, 'integrate in-processing as a heuristic simplification'),
                         ('sls', BOOL, False, 'add sls-tactic as a separate worker thread outside the search tree parallelism'),
-			('failed_literal_backbones', BOOL, False, 'use failed literal backbone test'),
-                        ('num_global_bb_threads', UINT, 0, 'default is 0 (off), can configure to 1 (negative mode only) or 2 (negative and positive mode) separate worker thread(s) outside the search tree parallelism'),
-              ))
+	                  ('num_global_bb_fl_threads', UINT, 0, 'run failed-literal backbone worker threads; default is 0 (off), supported values are 1 (negative mode only) or 2 (negative and positive mode)'),
+                        ('num_global_bb_batch_threads', UINT, 0, 'run Janota-style chunking backbone worker threads; default is 0 (off), supported values are 1 (negative mode only) or 2 (negative and positive mode)'),
+                        ('local_backbones', BOOL, False, 'enable local backbones experiment within the search tree parallelism'),
+                        ('core_minimize', BOOL, True, 'minimize unsat cores used for parallel cube backtracking'),
+                        ('ablate_backtracking', BOOL, False, 'ablation: pass entire cube as core instead of unsat core during backtracking'),
+                  ))
--- a/src/smt/smt_model_generator.cpp
+++ b/src/smt/smt_model_generator.cpp
@@ -26,7 +26,6 @@ Revision History:
 #include "smt/smt_model_generator.h"
 #include "smt/proto_model/proto_model.h"
 #include "model/model_v2_pp.h"
-#include "smt/theory_fpa.h"
 
 namespace smt {
 
@@ -132,11 +131,6 @@ namespace smt {
                 procs.push_back(proc);
                 root2proc.insert(r, proc);
             }
-        }
-
-        theory * fpa_th = m_context->get_theory(m.mk_family_id("fpa"));
-        if (fpa_th) {
-            static_cast<theory_fpa*>(fpa_th)->register_fpa2bv_processors(*this, root2proc, roots, procs);
         }
     }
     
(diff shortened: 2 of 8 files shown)


aleksisch:
Original input gives `unknown` after applying this fix.

tryh4rd-26:
> Original input gives `unknown` after applying this fix.

Hi @aleksisch,

For now, I’ve mainly focused on fixing the false-positive `sat` case and ensuring model generation remains sound. With this patch, Z3 no longer produces the invalid split-brain FP/BV models and correctly returns `unknown` instead of an unsound `sat`.

I’ll continue investigating the underlying theory-combination/model-construction issue to see if this case can be strengthened to a deterministic `unsat` result in the future.

wintersteiger:
Great effort, thanks for looking into this!

I think `smt_model_generator` should not depend on `theory_fpa`; FPA doesn't do anything special enough to handle on this level. If the connection between a theory variable and it's BV encoding is lost, the bug is much more likely in `theory_fpa.cpp` and it should be possible to fix it there.

wintersteiger:
(Also, `theory_fpa` should not depend on datatypes - there communication is via theory variables, no special handling required here)

tryh4rd-26:
Hi @wintersteiger and @aleksisch,

I’ve refactored and cleaned stuff a bit for the fix based on your feedback, @wintersteiger. I ended up rolling back all the model generator and datatype changes, so now it’s just a ~16 lines fix in `theory_fpa.cpp`.

The issue was basically a communication gap. When a term had an FP sort but wasn't from any FPA family (suppose say a `get-fp` accessor pulling from an Array), Z3 was creating two separate bitvec reps. Since they weren't really linked by any constraints, the BV solver would give them inconsistent values, leading to those weird "spurious sat" results that failed model validation.

To fix it, I updated `theory_fpa::relevant_eh` to make sure these terms are actually tied to their converted representations. It now asserts that the `wrap` expression has to match the concat'd BV fields, which is the same logic already used for the FP numerals.

The reproducer smt faile on aleksich's pastebin now correctly hits `unsat` and passes with `model_validate=true` just fine. I also ran the full `test-z3` suite and all 89 tests are green.

Could you guys pull the latest and let me know what you think?

wintersteiger (approved):
Awesome, that looks a lot better! I think this change makes sense - @NikolajBjorner what do you think?

NikolajBjorner:
Thanks a lot @wintersteiger for following up on it.
I got a chance to look at the PR in current form and from what I can tell it is a good change.

Issue #7990, opened 2025-10-18
Wrong verdict with `(set-option :sat.smt true)`


ThomasHaas:
On the following file [dglm.txt](https://github.com/user-attachments/files/22984268/dglm.txt), Z3 returns a wrong verdict when configured with `(set-option :sat.smt true)`. More precisely, it returns the wrong result on the *very final* call to `(check-sat)`.

**Result:** SAT, after ~18secs on my machine.
**Expected:** UNSAT. Confirmed by Z3 with `(set-option :sat.smt false)` in around 30secs. Other SMT solvers also confirm this result.

Interestingly, when adding `(set-option :model_validate true)`, Z3 reports `(error "line ... column 10: an invalid model was generated")` also for several intermediate `(check-sat)` calls, even though the SAT verdict is correct for those.

It seems the incremental inprocessing unsoundly deletes some irredundant clauses or fails to reconstruct them correctly.

**Background:** The task was generated by [Dartagnan](https://github.com/hernanponcedeleon/Dat3M) in the context of verifying a concurrent data structure.

NikolajBjorner:
it returns unsat in current z3

PR #9529, opened 2026-05-13
Fix soundness bug in sat.smt incremental mode 

tryh4rd-26:
Attempt at resolving #7990.

issue was a soundness bug where Z3 returns a wrong `sat` on the final `check-sat` of an incremental session w/ `sat.smt=true` linked in the og thread and I'm pretty confident I've got the root cause nailed.

It's in `elim_unconstrained` preproc. In the sat-smt path the simplifier was being too aggressive and pruning vars it considered "unconstrained" from a purely bool view, but those vars are still live for the theory solvers and model recon bridge. Axing them corrupts enough state that the solver loses track of theory bounds and ends up spitting out a bogus `sat` instead of correctly falling back to `unsat`.

I fixed it in `src/solver/solver_preprocess.cpp` by adding a guard to skip `elim_unconstrained` when `sat.smt=true`. It's not the most surgical fix and I'd love to revisit it once the simplifier can be made theory-aware, but I think it's the right safe call for now. I verified w/ the repro from the issue (`dglm.txt` in the original issue thread) and the final verdict is now `unsat`, matching the non-smt engine which is exactly what I'd expect. I also confirmed `model_validate=true` no longer throws "invalid model" on any intermediate or final call. Ran the full unit suite w/ `build/test-z3 /a /j:4` and got 90/90, and `api_bug` + `deep_api_bugs` are clean too so I don't think I've broken anything.


NikolajBjorner:
this is an AI slop fix. I closed a similar "fix" by copilot.

Issue #9139, opened 2026-03-26
Soundness bug(s) in nonlinear solver


wintered:
https://github.com/Z3Prover/z3/commit/456b99ced2719b22eec7bb25d1ab25bd2357bf4c
````
$ cvc5 -q --check-models --check-proofs --strings-exp bug1.smt2
sat
$ z3 model_validate=true bug1.smt2
unsat
$ cat bug1.smt2
(declare-const b Real) 
(assert (is_int (- (/ (/ (+ 3.0 b) (+ (- (- 3.0)) 2.0)) (/ (+ 3.0 b) (+ (- (- 3.0)) 2.0))) b))) 
(assert (< (- 2.0) b)) 
(check-sat)    
````

wintered:
https://github.com/Z3Prover/z3/commit/456b99ced2719b22eec7bb25d1ab25bd2357bf4c
````
$ cvc5 -q -i --check-models --check-proofs --strings-exp bug2.smt2
sat
sat
$ z3 model_validate=true bug2.smt2
sat
unsat
$ cat bug2.smt2
(declare-fun x0 () Real)
(assert (<= (* (+ x0 (- 13.0)) (- (/ (- (* x0 (- 1.0)) 13.0) (* x0 (- (/ (- 1.0) 13.0)))))) (- 1.0)))
(assert (> (+ 1.0 x0) (/ (- 13.0) (- 1.0))))
(check-sat)
(push)
(assert (or (> 0.0 (- 13.0)) (> 0.0 (- 13.0))))
(check-sat)
````

levnach:
Fixed by 43791ebf2abeb91fbe78656dd122c0f28b61098c. An arithmetic error in translation LRA->NLA was causing the bug.

PR #9563, opened 2026-05-18
Fix NLSAT check_assignment soundness bug (issue #9139)

levnach:
There was an algebraic error in deleted now mk_definition_assignment: 
Fixed by Claude.
setup_assignment_solver inlined term/monic definitions via mk_definition_assignment, scaling each term polynomial by den=lcm(denominators) to keep coefficients integral but then dropping `den`. This made definitions[v] = den_v * v for some hidden den_v, which propagated multiplicatively through monic factors. The constraint encoder then substituted definitions[v] into LRA constraints as if it equaled v, silently multiplying coefficients by the hidden factor. Combined with monic/term equalities never being asserted as NLSAT clauses (they were only baked in via the broken substitution), NLSAT's linear projection produced unsound bounds like `5*j8 <= 1` instead of `j8 <= 1`, which were returned as LRA lemmas and refuted satisfiable formulas.

Rewrite setup_assignment_solver in the same three-phase shape used by setup_solver_terms:

  1. Declare one NLSAT polyvar per LRA variable (identity map), seed m_values from the LRA model, and populate m_vars2mon for the inverse mapping in process_polynomial_check_assignment.
  2. Encode each LRA constraint as a single flat linear polynomial over the leaf polyvars (local lcm rescaling cancels on both sides of the relation).
  3. Assert monic equalities (j_mon - v1*v2*... = 0) via add_monic_eq and term equalities (den*j_term - sum(coeff_i * v_i) = 0) via add_term.

Remove the now-unused mk_definition_assignment helper and an unused polynomial::manager& binding in add_lemma, and update a stale comment on mk_definition.

Verified bug1.smt2 and bug2.smt2 from the issue now return sat (with model_validate=true) in both release and debug builds, and all 90 test-z3 /a unit tests still pass.

### DIFF
--- a/src/math/lp/nra_solver.cpp
+++ b/src/math/lp/nra_solver.cpp
@@ -64,8 +64,10 @@ struct solver::imp {
         m_lp2nl.reset();
     }
 
-    // Create polynomial definition for variable v used in setup_assignment_solver.
-    // Side-effects: updates m_vars2mon when v is a monic variable.
+    // Create polynomial definition for variable v used in setup_solver_poly.
+    // The definition recursively expands monic and term variables into
+    // polynomials in leaf variables, scaled by an integer denominator
+    // tracked in `denominators` to keep the coefficients integral.
     void mk_definition(unsigned v, polynomial_ref_vector &definitions, vector<rational>& denominators) {
         auto &pm = m_nlsat->pm();
         polynomial::polynomial_ref p(pm);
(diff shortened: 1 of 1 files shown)


levnach:
@ValentinPromies, it seems a bug was fixed related to your method.

src/math/lp/nra_solver.cpp
```diff
@@ -64,8 +64,10 @@ struct solver::imp {
         m_lp2nl.reset();
     }
 
-    // Create polynomial definition for variable v used in setup_assignment_solver.
-    // Side-effects: updates m_vars2mon when v is a monic variable.
+    // Create polynomial definition for variable v used in setup_solver_poly.
```
NikolajBjorner, line 67: this comment is obsolete.
You removed the code that removes the intermediary variables.
Maybe this is a good thing also for perf?

levnach: I am going to run a perf test, and also a test to find unsound lemmas.

levnach: @ValentinPromies wrote: 
While reading, i realized that the proposed bug fix (your PR from Monday) might not be the right fit with the linear cells (regarding their intended effect). The problem is that this way, the cell construction would always pick a multiplication constraint v - v_1*...*v_k = 0 for which i don't think it can derive more interesting results than e.g. the tangent lemmas.
We discussed at some point, that it would be more helpful to construct the cell for the "de-linearized" constraint, as the resulting lemma might affect several multiplications at once. This is why we introduced the substitution with mk_definition_assignment in the first place.

I think i do understand the bug, and I'll try to come up with something to salvage the substitution. Then we can check whether that works better or not.

PR #9591, opened 2026-05-21
Fix #9171: add expand-records preprocessing to enable constant folding

### LINKED ISSUE #9171: Regression in a constant-folding example
@levnach This is probably within your purview.

The following benchmark was quickly answered `unsat` with z3 compiled on March 2nd, 2026. However, it seems to loop forever with a z3 compiled just this morning.

Note that the benchmark has nonlinear operations, but it is essentially constant-folding. i.e., if the solver simply constant-folds the operations, it should be able to establish `unsat`.

```
(set-logic ALL)

(declare-datatype Rat ((mk-rat (num Int) (den Int))))

(define-fun rat-eq ((x Rat) (y Rat)) Bool
  (= (* (num x) (den y)) (* (den x) (num y))))

(declare-fun x () Rat)
(assert (> (den x) 0))
(declare-fun y () Rat)
(assert (> (den y) 0))

(assert (rat-eq x (mk-rat (- 1) 1)))
(assert (rat-eq y (mk-rat 1999327 1000000)))

(assert (not (rat-eq
  (mk-rat (- 2999327) 1000000)
  (mk-rat (- (* (num x) (den y)) (* (num y) (den x)))
          (* (den y) (den x))))))

(check-sat)
```

levnach:
Fixes #9171.

## Problem

Issue #9171 ships an SMT-LIB2 benchmark that declares a single-constructor `Rat` algebraic datatype and asserts a pile of non-linear Int constraints over the selectors `(num x)`/`(den x)`. The problem is unsatisfiable just by literal-by-literal arithmetic constant folding, but Z3 was diverging — and for some random seeds it diverged even on builds from early March.

Root causes:

1. **`solve-eqs` only substitutes uninterpreted constants.** In `extract_eqs.cpp` / `solve_eqs.cpp` the predicate `can_be_var(e)` requires `is_uninterp_const(e)`. Selector applications like `(num x)` are not uninterpreted constants, so equalities like `(= (num x) 7)` were never propagated. The variables of interest were hidden behind datatype selectors and the existing constant folding machinery never saw them as raw integer variables.
2. **Tactic dispatch saw the un-simplified syntax.** `default_tactic.cpp` picks a sub-tactic based on probes over the original formula. Because the formula mentions a datatype (`Rat`), the dispatcher chose the heavyweight `smt` sub-tactic (theory_arith + nlsat), which loops on this nonlinear input. The trivial `qfnia-tactic` (bit-blast) would have solved it instantly, but the dispatcher never saw the QF_NIA-shaped formula.

## Fix

Introduces an `expand-records` preprocessing step that rewrites every free constant whose sort is a **non-recursive single-constructor** algebraic datatype into the constructor applied to fresh field constants, e.g.

```
x : Rat   ->   x  :=  (mk-rat x!num x!den)
```

Subsequent simplification rewrites `(num (mk-rat x!num x!den))` to `x!num`, etc. After the rewrite, the formula no longer mentions the datatype at all — it is pure non-linear integer arithmetic over the fresh field constants, and the existing arithmetic rewriter and `solve-eqs` finish the job immediately.

Guards: only non-recursive single-constructor datatypes are expanded (prevents infinite expansion on lists, trees, etc.).

Model preservation: the substitution is pushed onto the model trail so models for the original datatype-typed constant can be reconstructed.

## Wiring

The transformation is wired into all three preprocessing entry points so it applies regardless of which path the user takes:

- **`init_preprocess`** (`src/solver/solver_preprocess.cpp`) — modern `simplifier_solver` path (e.g. when `(set-simplifier …)` is used).
- **`asserted_formulas::reduce`** (`src/solver/assertions/asserted_formulas.{h,cpp}`) — legacy SMT context path.
- **`default_tactic`** (`src/tactic/portfolio/default_tactic.cpp`) — inserted into the `and_then(...)` chain *before* the logic-class `cond` probes, so the dispatcher sees the post-expansion formula and routes to `qfnia-tactic`. **This is the critical piece** for the user-visible `(check-sat)` path.

## Files

New:

- `src/ast/simplifiers/expand_records.{h,cpp}` — the `dependent_expr_simplifier`.
- `src/tactic/core/expand_records_tactic.h` — thin `dependent_expr_state_tactic` wrapper registered as `expand-records`.

Modified:

- `src/ast/simplifiers/CMakeLists.txt`, `src/tactic/core/CMakeLists.txt` — register new sources/headers.
- `src/solver/solver_preprocess.cpp` — hook into `init_preprocess`.
- `src/solver/assertions/asserted_formulas.{h,cpp}` — `expand_records()` step at the start of `reduce()`.
- `src/tactic/portfolio/default_tactic.cpp` — insert `expand-records` + `simplify` before the logic-class probes.
- `src/util/trace_tags.def` — register `expand_records` trace tag (needed for the debug build's `TRACE(...)` macro check).

## Verification

- Original benchmark from #9171: now solves as **`unsat` in ~0.14s** on the release build (previously diverged).
- Unit tests: **89/90 pass on both release and debug builds**. The single failing test, `tptp`, fails identically on unmodified `master` — pre-existing, unrelated to this change.

cc @NikolajBjorner — could you review please?

### DIFF
--- a/src/ast/simplifiers/expand_records.cpp
+++ b/src/ast/simplifiers/expand_records.cpp
@@ -1,0 +1,101 @@
+/*++
+Copyright (c) 2026 Microsoft Corporation
+
+Module Name:
+
+    expand_records.cpp
+
+Abstract:
+
+    See expand_records.h.
+
+Author:
+
+    Lev Nachmanson 2026-05
+
+--*/
+
+#include "ast/ast_pp.h"
+#include "ast/ast_util.h"
+#include "ast/expr_substitution.h"
+#include "ast/for_each_expr.h"
+#include "ast/rewriter/expr_replacer.h"
+#include "ast/simplifiers/expand_records.h"
+
+
+void expand_records_simplifier::reduce() {
+    if (!m_enabled)
+        return;
+
+    m_fmls.freeze_suffix();
+
+    // Collect candidate free constants.
+    obj_hashtable<expr> seen_vars;
+    ptr_vector<app>     vars;
+    expr_mark           visited;
+
+    auto visit = [&](expr* e) {
+        if (!is_uninterp_const(e))
+            return;
+        if (seen_vars.contains(e))
+            return;
+        if (!is_expandable(e->get_sort()))
+            return;
+        // Skip frozen symbols (e.g., introduced by other simplifiers, or
+        // symbols that other modules rely on being preserved verbatim).
+        if (m_fmls.frozen(to_app(e)->get_decl()))
+            return;
+        seen_vars.insert(e);
+        vars.push_back(to_app(e));
+    };
+
+    for (unsigned i : indices()) {
+        for (expr* t : subterms::all(expr_ref(m_fmls[i].fml(), m), nullptr, &visited))
+            visit(t);
+    }
+
+    if (vars.empty())
+        return;
+
+    // Build the substitution v -> (c v!f1 ... v!fn).
+    scoped_ptr<expr_substitution> subst = alloc(expr_substitution, m);
+    expr_ref_vector pinned(m);
+
+    for (app* v : vars) {
+        sort* s = v->get_sort();
+        func_decl* ctor = m_dt.get_datatype_constructors(s)->get(0);
+        auto const* accs = m_dt.get_constructor_accessors(ctor);
+        expr_ref_vector args(m);
+        std::string base = v->get_decl()->get_name().str();
+        for (func_decl* acc : *accs) {
+            std::string nm = base + "!" + acc->get_name().str();
+            app* fresh = m.mk_fresh_const(nm.c_str(), acc->get_range());
+            args.push_back(fresh);
+        }
+        expr_ref new_val(m.mk_app(ctor, args.size(), args.data()), m);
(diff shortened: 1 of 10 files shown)


src/solver/assertions/asserted_formulas.cpp
```diff
@@ -535,6 +538,70 @@ void asserted_formulas::commit(unsigned new_qhead) {
     m_qhead = new_qhead;
 }
 
+void asserted_formulas::expand_records() {
```
NikolajBjorner, line 541: remove this from asserted formulas

src/ast/simplifiers/expand_records.cpp
```diff
@@ -0,0 +1,101 @@
+/*++
+Copyright (c) 2026 Microsoft Corporation
+
+Module Name:
+
+    expand_records.cpp
+
```
NikolajBjorner, line 7: what is the solve_eqs step that doesn't work?
Is it (first x) where x is a pair? 
By rewriting x to pair(x1, x2) then (first (pair(x1,x2)) is x1
and (first x) = t with x not in t can be solved by setting x1 = t

PR #9704, opened 2026-06-03
Remove duplicate is_nullable_rec from seq_rewriter, delegate to derive::is_nullable

NikolajBjorner:
`seq_rewriter` contained `is_nullable_rec` and `is_nullable_symbolic_regex` that duplicated logic already present in `seq::derive::is_nullable`. This removes the duplication and consolidates nullability computation in `derive`.

## Changes

- **`seq_derive.cpp`**: Fix `is_to_re` case in `derive::is_nullable` to recurse into the inner sequence expression via `is_nullable(r1)` rather than emitting a raw `mk_eq(empty, r1)`. This makes it fully equivalent to the old `seq_rewriter::is_nullable_rec` — concrete cases like `to_re("") → true` and `to_re("abc") → false` now simplify correctly.
- **`seq_derive.h`**: Expose a public `nullable(expr* r)` method on `derive` that forwards to the private `is_nullable`.
- **`seq_rewriter.cpp`**: `seq_rewriter::is_nullable` now delegates to `m_derive.nullable(r)`. Removes `is_nullable_rec` and `is_nullable_symbolic_regex` bodies.
- **`seq_rewriter.h`**: Remove the two now-deleted method declarations.


src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,679 @@
...
+        // δ(r1 · r2) = δ(r1) · r2 ∪ (if nullable(r1) then δ(r2) else ∅)
+        if (re().is_concat(r, r1, r2)) {
+            expr_ref d1 = derive_rec(r1);
+            expr_ref d1_r2 = mk_deriv_concat(d1, r2);
+            expr_ref nullable_r1 = is_nullable(r1);
+            if (m.is_true(nullable_r1))
+                return mk_union(d1_r2, derive_rec(r2));
+            if (m.is_false(nullable_r1))
+                return d1_r2;
+            // Conditional: nullable is a Boolean expression
```
NikolajBjorner, line 196: I think we should just return mk_derivative expression here instead of a Boolean. Then also remove the code that computes a symbolic condition for nullability because it is also bogus for regexes that are symbolic.
For symblic regexes nullability should be a predicate and not constant true.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,679 @@
...
+            expr_ref n2 = is_nullable(r2);
+            expr_ref result(m);
+            m_br.mk_ite(cond, n1, n2, result);
+            return result;
+        }
+        // Unknown: use membership test
+        if (m_util.is_re(r, s))
+            return expr_ref(re().mk_in_re(u().str.mk_empty(s), r), m);
+
+        return expr_ref(m.mk_true(), m);
```
NikolajBjorner: This is shady and suggests we should not have this function

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,679 @@
...
+        if (m_cache.find(r, cached))
+            return expr_ref(cached, m);
+
+        // Depth check
+        if (m_depth >= m_max_depth) {
+            // Return stuck derivative (the derivative operator applied symbolically)
+            return expr_ref(re().mk_derivative(m_ele, r), m);
+        }
+
+        ++m_depth;
```
NikolajBjorner: we should be able to use flet<unsigned> _scoped_depth(m_depth, m_depth++);

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,679 @@
...
+            // All elements were identity/absorbed - should not happen in practice
+            // but handle gracefully
+            UNREACHABLE();
+            return expr_ref(m.mk_true(), m);
+        }
+        if (args.size() == 1)
+            return expr_ref(args.get(0), m);
+        // Build right-associated union
+        expr_ref result(args.back(), m);
+        for (unsigned i = args.size() - 1; i > 0; ) {
```
NikolajBjorner: use simpler pattern for (unsigned i = args.size(); i-- > 0; ) {
       // don't decrement i inside of loop
}

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+    // -------------------------------------------------------
+    // ITE-tree combinators (analogous to REsharp mk_binary/mk_unary)
+    // -------------------------------------------------------
+
+    expr_ref derive::ite_combine_binary(expr* d1, expr* d2,
+        std::function<expr_ref(expr*, expr*)> const& op) {
+        expr *c1, *t1, *e1, *c2, *t2, *e2;
+
+        // Both are leaves (non-ITE)
```
NikolajBjorner: use this approach instead:

bool is_ite1 = m.is_ite(d1, c1, t1, e1);
bool is_ite2 = m.is_ite(d2, c2, t2, e2);

avoid other calls to m.is_ite

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,757 @@
...
+        m_util(m),
+        m_autil(m),
+        m_br(m),
+        m_trail(m),
+        m_ele(m) {
+        m_br.set_flat_and_or(false);
+    }
+
+    void derive::reset() {
+        m_cache.reset();
```
NikolajBjorner, line 48: The cache should index over both m_ele and r.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,757 @@
...
+    expr_ref derive::operator()(expr* r) {
+        SASSERT(m_util.is_re(r));
+        sort* seq_sort = nullptr, * ele_sort = nullptr;
+        VERIFY(m_util.is_re(r, seq_sort));
+        VERIFY(m_util.is_seq(seq_sort, ele_sort));
+        expr_ref v(m.mk_var(0, ele_sort), m);
+        return (*this)(v, r);
+    }
+
+    expr_ref derive::eval(expr* ele, expr* d) {
```
NikolajBjorner: don't introduce this function. rely on operator() to enforce simplifcation of the ite level

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,757 @@
...
+                }
+            }
+            result = false;
+            return true;
+        }
+
+        return false;
+    }
+
+    expr_ref derive::simplify_ite(expr* d) {
```
NikolajBjorner: rewrite simplify_ite_rec to take as argument a vector of signed conditions.
The signed conditions are extracted from the ite condition.
For ite(a & b, t, e) the path used for simplifying the term t is based on the vector [(a, false), (b, false)].
The term used for simplifying e is based on the vector [(a & b, true)]

Use simplification rules of the form:
simplify_ite_rec(path, ite(x = v, t, e)) = t if ((x = v), false) in path.
simplify_ite_rec(path, ite(x = v, t, e)) = e if ((x = v), true) in path.
simplify_ite_rec(path, ite(x = v, t, e)) = e if ((x = w), false) in path, v and w are values and they are distinct (m.are_distinct(v, w)).
Include also simplification rules for range constraints, that is a condition can be a range of the form (lo <= x & x <= hi) where lo and hi are character constants. then 
simplify_ite_rec([(lo <= x, false), (x <= hi, false)], ite(x = v, t, e)) = e if v < lo or v > hi.
simplify_ite_rec([(lo <= x, false), (x <= hi, false)], ite(x = v, t, e)) = simplify_ite_rec([(lo <= x, false), (x <= hi, false), (x = v, false)], t) if lo <= v <= hi

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2914,358 +2914,15 @@ bool seq_rewriter::check_deriv_normal_form(expr* r, int level) {
...
-            }
-            m_br.mk_and(path, c1, c2);
-            if (m().is_false(c2))
-                result = nothing();
-            else
-                // observe that the precondition |r1|>0 is implied by c1 for use of mk_seq_rest
-                result = m().mk_ite(c1, re().mk_to_re(mk_seq_rest(r1)), nothing());
-        }
-    }
-    else if (re().is_reverse(r, r2))
```
NikolajBjorner, line 2990: Port elements from handling reverse into the new code in derive.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3909,221 +3543,6 @@ expr_ref seq_rewriter::mk_der_cond(expr* cond, expr* ele, sort* seq_sort) {
...
-            expr_ref r1(re().mk_to_re(tl), m());
-            result = mk_der_concat(result, r1);
-            return result;
-        }
-        else if (str().is_empty(r1)) {
-            //observe: str().is_empty(r1) checks that r = () = epsilon
-            //while mk_empty() = [], because deriv(epsilon) = [] = nothing
-            return mk_empty();
-        }
-        else if (str().is_itos(r1)) {
```
NikolajBjorner, line 4022: port relevant portions of he is_itos case into derive.

src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -3909,221 +3543,6 @@ expr_ref seq_rewriter::mk_der_cond(expr* cond, expr* ele, sort* seq_sort) {
...
-            // recall: [] denotes the empty language (nothing) regex, () denotes epsilon or empty sequence
-            // construct the term (if (r1 != () and (ele = (first r1)) then (to_re (rest r1)) else []))
-            hd = mk_seq_first(r1);
-            m_br.mk_and(m().mk_not(m().mk_eq(r1, str().mk_empty(seq_sort))), m().mk_eq(hd, ele), result);
-            tl = re().mk_to_re(mk_seq_rest(r1));
-            return re_and(result, tl);
-        }
-    }
-    else if (re().is_reverse(r, r1)) {
-        if (re().is_to_re(r1, r2)) {
```
NikolajBjorner, line 4051: port relevant portions of the is_reverse case into derive

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,950 @@
...
+        for (auto const& [cond, sign] : path) {
+            // Direct match: c == cond
+            if (c == cond)
+                return sign ? simplify_ite_rec(path, e) : simplify_ite_rec(path, t);
+
+            // c is (x = v), cond is (x = w) with sign=false (cond is true, so x=w)
+            // If v != w, then c is false → take else branch
+            expr* lhs1 = nullptr, * rhs1 = nullptr, * lhs2 = nullptr, * rhs2 = nullptr;
+            if (!sign && m.is_eq(cond, lhs1, rhs1) && m.is_eq(c, lhs2, rhs2)) {
+                if (m_util.is_const_char(lhs1)) std::swap(lhs1, rhs1);
```
NikolajBjorner: use m.is_value instead of m_util.is_const_char

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,950 @@
...
+
+        // reverse(to_re(s1 ++ s2)) = reverse(to_re(s2)) · reverse(to_re(s1))
+        if (re().is_to_re(r, s) && u().str.is_concat(s, r1, r2)) {
+            expr_ref a(re().mk_reverse(re().mk_to_re(r2)), m);
+            expr_ref b(re().mk_reverse(re().mk_to_re(r1)), m);
+            return expr_ref(re().mk_concat(a, b), m);
+        }
+
+        // Stuck — cannot normalize further
+        return expr_ref(nullptr, m);
```
NikolajBjorner: this should be return expr_ref(re().mk_reverse(r), m)

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,950 @@
...
+                        }
+                    }
+                }
+                if (has_lo && has_hi && lo_bound <= v_val && v_val <= hi_bound) {
+                    // v is in range [lo, hi], so c is satisfiable
+                    // Add (x = v, false) to path and simplify t
+                    path.push_back({ c, false });
+                    expr_ref st = simplify_ite_rec(path, t);
+                    path.pop_back();
+                    expr_ref se = simplify_ite_rec(path, e);
```
NikolajBjorner: you have to path.push_back({c, true}); before simplify_ite_rec(path, e) for the else case of c

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,950 @@
...
+                    return mk_ite(c, st, se);
+                }
+            }
+        }
+
+        // Try to evaluate c directly
```
NikolajBjorner: move simplification to the beginning of the function before checking if it is determined by the path

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,950 @@
...
+        bool cond_val;
+        if (eval_cond(c, cond_val))
+            return simplify_ite_rec(path, cond_val ? t : e);
+
+        // Cannot simplify c: recurse into branches with extended paths
+        // True branch: add conjuncts of c
+        path_t extended_path(path);
```
NikolajBjorner: you can just record 

auto sz = path.size()
append to path instead of extended_path
and after the call to simplify_ite_rec use path.shrink(sz);

src/ast/rewriter/seq_derive.h
```diff
@@ -0,0 +1,141 @@
+/*++
+Copyright (c) 2025 Microsoft Corporation
```
NikolajBjorner: it is 2026

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,950 @@
...
+        }
+
+        return mk_concat(d, tail);
+    }
+
+    // -------------------------------------------------------
+    // Post-processing: simplify ITE conditions w.r.t. m_ele
+    // -------------------------------------------------------
+
+    bool derive::eval_cond(expr* cond, bool& result) {
```
NikolajBjorner: can you change the signature to be 

lbool eval_cond(expr* cond)

so that l_undef corresponds to returning false, and l_true corresponds to result being set to true.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,950 @@
+/*++
+Copyright (c) 2025 Microsoft Corporation
```
NikolajBjorner: it is 2026

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,950 @@
...
+    Produces an ITE-tree (transition regex) representation following
+    the approach of RE# (Varatalu, Veanes, Ernits - POPL 2025).
+
+    The symbolic derivative δ(r) maps each character to the resulting
+    derivative state via an ITE-tree. The free variable (:var 0) represents
+    the input character.
+
+Authors:
+
+    Nikolaj Bjorner (nbjorner) 2025-06-03
```
NikolajBjorner: it is 2026

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,999 @@
...
+        }
+
+        return mk_concat(d, tail);
+    }
+
+    // -------------------------------------------------------
+    // Post-processing: simplify ITE conditions w.r.t. m_ele
+    // -------------------------------------------------------
+
+    bool derive::eval_cond(expr* cond, bool& result) {
```
NikolajBjorner: can you change the signature to be

lbool eval_cond(expr* cond)

so that l_undef corresponds to returning false, and l_true corresponds to result being set to true.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,999 @@
...
+                    path.pop_back();
+                    expr_ref se = simplify_ite_rec(path, e);
+                    return mk_ite(c, st, se);
+                }
+            }
+        }
+
+        // Cannot simplify c: recurse into branches with extended paths
+        // True branch: add conjuncts of c
+        auto sz = path.size();
```
NikolajBjorner: hoist this to an auxiliary routine 

void push_path(path_t& path, expr* c, bool sign);
So that the return value can be used for path.shrink(..) after a recursive call to simplify_ite_rec.
if sign is false and m.is_and(c) then recursively push_path for each conjunct. 
If sign is true and m.is_or(c), then recursively push_path for each disjunction.
Otherwise path.push({c, sign}); 

Furthermore create an auxiliary routine simplify_ite_rec(path, c, t, e)

auto sz = path.size();
push_path(path, c, false);
auto st = simplify_ite_rec(path, t);
path.shrink(sz);
push_path(path, c, true);
auto se = simplify_ite_rec(path, e);
return {st, se };

Refactor other code to use this.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1115 @@
...
+                        unsigned bound = 0;
+                        if (le_rhs == lhs_c && m_util.is_const_char(le_lhs, bound)) {
+                            lo_bound = bound; has_lo = true;
+                        }
+                        if (le_lhs == lhs_c && m_util.is_const_char(le_rhs, bound)) {
+                            hi_bound = bound; has_hi = true;
+                        }
+                    }
+                }
+                if (has_lo && has_hi && lo_bound <= v_val && v_val <= hi_bound) {
```
NikolajBjorner: we don't need this code. It can be removed. The same code occurs below.
So break out of the loop.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1115 @@
...
+
+        // not(e1)
+        if (m.is_not(cond, e1)) {
+            lbool inner = eval_cond(e1);
+            if (inner != l_undef)
+                return inner == l_true ? l_false : l_true;
+        }
+
+        // and(...)
+        if (m.is_and(cond)) {
```
NikolajBjorner, line 1197: better style:
INitialize lbool r = l_true;
Set r = l_undef inside the loop if v == l_undef;
return r

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1115 @@
...
+                lbool v = eval_cond(arg);
+                if (v == l_false) return l_false;
+                if (v == l_undef) return l_undef;
+            }
+            return l_true;
+        }
+
+        // or(...)
+        if (m.is_or(cond)) {
+            for (expr* arg : *to_app(cond)) {
```
NikolajBjorner, line 1210: better style: 
Initialize r = l_false
set r = l_undef inside the loop if v == l_undef;
return r

src/ast/rewriter/seq_derive.h
```diff
@@ -0,0 +1,156 @@
...
+            std::function<expr_ref(expr*, expr*)> const& op);
+
+        // ITE-tree unary combinator (analogous to REsharp mk_unary)
+        expr_ref ite_combine_unary(expr* d, std::function<expr_ref(expr*)> const& op);
+
+        // Distribute concatenation through ITE/union in derivative
+        expr_ref mk_deriv_concat(expr* d, expr* tail);
+
+        // Extract head character and tail from a sequence expression
+        bool get_head_tail(expr* s1, expr* s2, expr_ref& hd, expr_ref& tl);
```
NikolajBjorner, line 150: is this function repeated elsewhere? Can you just use a single definition?

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1188 @@
...
+        if (args.empty()) {
+            UNREACHABLE();
+            return expr_ref(m.mk_true(), m);
+        }
+        if (args.size() == 1)
+            return expr_ref(args.get(0), m);
+        // Build right-associated union
+        expr_ref result(args.back(), m);
+        for (unsigned i = args.size() - 1; i-- > 0; )
+            result = expr_ref(re().mk_union(args.get(i), result), m);
```
NikolajBjorner: use mk_union method on this class instead to allow using subsumption

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1188 @@
...
+        if (args.empty()) {
+            UNREACHABLE();
+            return expr_ref(m.mk_true(), m);
+        }
+        if (args.size() == 1)
+            return expr_ref(args.get(0), m);
+        // Build right-associated intersection
+        expr_ref result(args.back(), m);
+        for (unsigned i = args.size() - 1; i-- > 0; )
+            result = expr_ref(re().mk_inter(args.get(i), result), m);
```
NikolajBjorner: use mk_inter/mk_intersect rom this class to support subsumption.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1188 @@
...
+        VERIFY(m_util.is_re(r, s));
+        return expr_ref(re().mk_in_re(u().str.mk_empty(s), r), m);
+    }
+
+    // -------------------------------------------------------
+    // Smart constructors with simplification
+    // -------------------------------------------------------
+
+    // Lightweight structural subsumption: checks if L(a) ⊆ L(b)
+    // Returns true only when subsumption can be determined structurally.
```
NikolajBjorner: resharer approach: should this be a complete check modulo a list of regexes?

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1188 @@
...
+    // -------------------------------------------------------
+
+    lbool derive::eval_cond(expr* cond) {
+        expr* lhs = nullptr, * rhs = nullptr, * e1 = nullptr;
+        unsigned ch1 = 0, ch2 = 0;
+
+        if (m.is_true(cond)) return l_true;
+        if (m.is_false(cond)) return l_false;
+
+        // elem = char or char = elem
```
NikolajBjorner: use instead is_char_const_range

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1188 @@
...
+        if (m.is_eq(cond, lhs, rhs)) {
+            if (rhs == m_ele) std::swap(lhs, rhs);
+            if (lhs == m_ele && u().is_const_char(rhs, ch1) && u().is_const_char(m_ele, ch2))
+                return ch1 == ch2 ? l_true : l_false;
+            if (lhs == rhs) return l_true;
+        }
+
+        // char_le(lhs, rhs)
```
NikolajBjorner: use result of is_char_const range

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1188 @@
...
+                return l_true;
+            // char_le(x, max_char) is always true
+            if (u().is_const_char(rhs, vr) && vr == u().max_char())
+                return l_true;
+        }
+
+        // not(e1)
+        if (m.is_not(cond, e1)) {
+            lbool inner = eval_cond(e1);
+            if (inner != l_undef)
```
NikolajBjorner: does the ~ operator on l_bool work for this?

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1206 @@
...
+            // c asserts x ∉ [lo, hi]
+            if (all_contained) return l_false; // all values are in [lo,hi], so ¬(x∈[lo,hi]) is false
+            if (!any_overlap) return l_true;   // no values are in [lo,hi], so ¬(x∈[lo,hi]) is true
+        }
+        return l_undef;
+    }
+
+    std::pair<expr_ref, expr_ref> derive::simplify_ite_rec(path_t& path, intervals_t& intervals, expr* c, expr* t, expr* e) {
+        auto sz = path.size();
+        auto saved_intervals = intervals;
```
NikolajBjorner: rewrite push_path to return an lbool.
Use the code from simplify_ite_rec that checks with path in this function.

        for (auto const& [cond, sign] : path) {
            if (c == cond)
                return sign ? simplify_ite_rec(path, intervals, e) : simplify_ite_rec(path, intervals, t);

            expr* lhs1 = nullptr, * rhs1 = nullptr, * lhs2 = nullptr, * rhs2 = nullptr;
            if (!sign && m.is_eq(cond, lhs1, rhs1) && m.is_eq(c, lhs2, rhs2)) {
                if (m.is_value(lhs1)) std::swap(lhs1, rhs1);
                if (m.is_value(lhs2)) std::swap(lhs2, rhs2);
                if (lhs1 == lhs2 && m.are_distinct(rhs1, rhs2))

Return l_true if c, sign is subsumed. In this case make sure path is not changed and return simplify_ite_rec(path, intervals, sign ? e : t)
Return l_false if c is !sign is subsumed. In this case make sure path is not changed and return simplify_ite_rec(path, intervals, sign ? t : e)

rewrite push_intervals to append only on intervals. This avoid copying intervals and restoring it. Instead shrink intervals.shrink(interval_size) where interval_size is captured along with path.size()

also rewrite push_intervals to return an lbool.
Return l_true if c, sign is subsumed. Make sure intervals is not changed in this case. In this case shrink path to sz and return simplify_ite_rec(path, intervals, sign ? e : t)
Return l_false if c is !sign is subsumed. Make sure intervals is not changed in this case. In this case shrink path to sz and return simplify_ite_rec(path, intervals, sign ? t : e)

return { st, se } if both pushes return l_undef.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1529 @@
...
+                std::swap(c1, c2);
+                std::swap(t1, t2);
+                std::swap(e1, e2);
+            }
+            expr_ref r1(m), r2(m);
+            bool have_r1 = false, have_r2 = false;
+            // Under c1: what do we know about c2?
+            if (pred_implies(c1, c2)) {
+                r1 = mk_inter(t1, t2); have_r1 = true;
+            } else if (pred_implies(c1, m.mk_not(c2))) {
```
NikolajBjorner: this leads to memory leak

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1529 @@
...
+            }
+        }
+
+        return false;
+    }
+
+    // Predicate implication for character range conditions.
+    // Returns true if: whenever cond_a is true, cond_b must also be true.
+    // Used for BDD-merge of derivative ITE trees.
+    bool derive::pred_implies(expr* a, expr* b) {
```
NikolajBjorner, line 653: you have to pin a, b in the beginning in case caller has created these expressions directly when calling the function.
Add the line:
expr_ref _a(a, m), _b(b, m);

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1529 @@
...
+        // ~∅ → Σ*
+        if (re().is_empty(a))
+            return expr_ref(re().mk_full_seq(a->get_sort()), m);
+        // ~Σ* → ∅
+        if (re().is_full_seq(a))
+            return expr_ref(re().mk_empty(a->get_sort()), m);
+
+        // Push through ITE: ~(ite(c, t, e)) → ite(c, ~t, ~e)
+        expr* c, * t, * e;
+        if (m.is_ite(a, c, t, e)) {
```
NikolajBjorner, line 836: only distribute if either t or e is empty, full or a complement.

src/ast/rewriter/seq_derive.cpp
```diff
@@ -0,0 +1,1529 @@
...
+            }
+            if (have_r1 || have_r2) {
+                if (!have_r1) r1 = mk_inter(t1, b);
+                if (!have_r2) r2 = mk_inter(e1, b);
+                return mk_ite(c1, r1, r2);
+            }
+        }
+
+        // ITE hoisting with depth bound (fallback when pred_implies doesn't fire)
+        if (m_inter_hoist_depth < m_max_inter_hoist_depth) {
```
NikolajBjorner: remove this code, m_max_inter_hoist_depth is effectively 0

NikolajBjorner:
@copilot 
revert the changes to datatype_factory
revert the changes to seq_regex
move expr_ref mk_regex_reverse(expr* r); functionality from seq_rewriter to seq_derive. It is currently duplicated.
move the expr_ref is_nullable_rec(expr* r); functionality from seq_rewriter to seq_derive. Currently seq_derive uses a weaker version to produce nullability Boolean expressions. 
if expr_ref mk_regex_concat(expr* r1, expr* r2); is not used, then delete it from seq_rewriter. Ensure that seq_derive handling of concatentation is feature compatible.

PR #9768, opened 2026-06-08
Fixes necessary to compile z3 included in clang-tidy via FetchContents.

davedets:
The page https://github.com/Z3Prover/z3/blob/master/README-CMake.md#adding-z3-as-a-dependency-to-a-cmake-project advises using the CMake FetchContent feature to include z3 as source into other CMake project. I'm trying to do this to use Z3 within a ClangTidy checker. This is one of a series of PR's aimed at getting Z3 to compile cleanly when included this way.

This initial PR fixes all the errors, allowing the compilation to succeed.  Subsequent diffs will address warnings.

I tested only the CMake compilation, on a Mac.

*Missing Z3_THROWs*

Update z3++.h to use Z3_THROW in a couple of places. Clang compiles with exceptions disabled so we get messages like:

```
￼/Users/daviddetlefs/llvm-project/build_dbg/_deps/z3-src/src/api/c++/z3++.h:4928:17: error: cannot use 'throw' with exceptions disabled￼4928 |                 throw exception("rcf_num objects from different contexts");￼
```

NOTE TO REVIEWERS: I'm not complete clear on the usage conventions for Z3_THROW. With exception disabled, it seems like the throwing function will just continue. If there's somethign else that should be done, like setting some error state, please let me know.

*CMake component name collision*

There was an error at the CMake level, a name collision (on "opt"). Apparently CMake components are named using a flat namespace, so it's easy to see how this could occur. It seems to me that the right global way to fix this would be to encourage people to use some form of "qualified name" convention in naming their component. The fix I chose was a local version of this, changing the Z3 component name to z3_opt. (It didn't seem feasible to make the change in clang.)

NOTE TO REVIEWERS: If you think this is OK, please let me know if

a) You'd like me to also change the name of the opt directory, to keep thecomponent-name == directory-name invariant, and

b) You'd like me to make this z3_ change more globally, to future-proof (somewhat) against similar component name collisions.


davedets:
-  I have sole ownership of intellectual property rights to my
   Submissions and I am not making Submissions in the course of work for my
   employer.

davedets:

PR #9800, opened 2026-06-09
Towards clean z3 build in clang-tidy: Eliminate zero-length-array warnings.

davedets:

This one concerns "-Wzero-length-array".  Many classes in Z3 use the "trailing array" idiom: the last field of the class C is a zero-length array of some type T, and the allocation of an instance adds extra space for some number of T's (usually the number is held in some other field of C).

When the flag -Wzero-length-array is used, this gives warnings like:
```
warning: zero size arrays are an extension [-Wzero-length-array]
```
This PR first adds -Wzero-length-array to the CLANG_ONLY_WARNINGS defined in compiler_warnings.cmake.  This causes the warnings to occur in a normal clang Z3 build.  We then make a trailing_array.h, that defines a TRAILING_ARRAY(type, field) macro.  This is defined, on per-comnpiler basis, to disable such warnings if they're enabled.  I chose this route because the use of the idiom is baked deeply into Z3, so it didn't seem feasable to consider alternative patterns (like, for example, having the trailing field be a pointer to a separately allocated array).

The rest of the changes in the PR are uses of this include file and macro in place of "raw" zero-length arrays.

With these changes, all instances of this warning are removed.


nunoplopes:
Seems a bit overkill. You can just change CXXFLAGS to add `-Wno-zero-length-array`. There's no point in this gigantic patch.

NikolajBjorner:
@davedets - Nuno has weighed in. I suspected it had some flavor of overkill given the invasiveness of the update for a well-known idiom.

davedets:
> You can just change CXXFLAGS to add -Wno-zero-length-array

I'm not sure I understand the suggestion.  Again, I'm tryng to get Z3 to compile cleanly when included via FetchContents as a library used in Clang-Tidy.  Are you suggesting that I change the Clang-Tidy build?  Or are you suggesting I do this locally via the CXXFLAGS environment variable when I do builds, overriding the choices Clang-Tidy makes?

If the former, well, I don't think I'm going to be able to convince Clang-Tidy to change their (aggressive) build rules.

If the latter, if I eventually succeed in getting the clang-tidy checker I'm building (the one that uses Z3) into clang-tidy, wouldn't we then have the problem for anyone building clang-tidy?

But maybe I've misunderstood the suggestion, and there's a third interpretation.  Perhaps you're suggesting doing something in the Z3 CMake build.  I'll see if adding -Wno-zero-length-array in CLANG_ONLY_WARNINGS in compiler_warnings.cmake has the desired effect when it's included in the clang-tidy build.  I agree that would satisfy my objections if it works.

nunoplopes:
> But maybe I've misunderstood the suggestion, and there's a third interpretation. Perhaps you're suggesting doing something in the Z3 CMake build. I'll see if adding -Wno-zero-length-array in CLANG_ONLY_WARNINGS in compiler_warnings.cmake has the desired effect when it's included in the clang-tidy build. I agree that would satisfy my objections if it works.

This one.

davedets:
>> <change compiler_warnings.cmake>

>This one.

I verified that the recommended change eliminated the zero-length array warnings for Z3 compiled as part of clang-tidy.

I've updated davedets/z3/master to eliminate the changes except for the recommended change.

Thanks for pushing me to get a simpler solution for this!

NikolajBjorner:
Nobody expects the Nuno nudge

PR #9825, opened 2026-06-11
Parallel tactic (#9824)

NikolajBjorner:
Add new parallel algorithm as a tactic (parallel_tactical2.cpp)
Don't port over old experiments from smt_parallel that we aren't using (sls, inprocessing, failed_literal_mode for bb detection)
Fix bugs: lease cancellation/reslimit race condition, involves changing lease epoch to simple boolean flag
Also, now there is a single shared set of params for the tactic and smt_parallel

**Test runs for the parallel_tactical2 vs old smt_parallel version:**
run-2747-Z3-threads-4-qflia-30s-stats.md
run-2746-Z3-threads-4-qflia-30s-parallel_tactic-stats.md
run-2745-Z3-threads-1-qfbv-30s-stats.md
run-3013-Z3-threads-4-qfbv-30s-parallel_tactic-stats.md  --> note this is indeed run-3013, I reran after a bugfix in inc_sat_solver
run-2743-Z3-threads-4-qfnia-30s-stats.md
run-2742-Z3-threads-4-qfnia-30s-parallel_tactic-stats.md

**Test runs for the new smt_parallel with bugfixes:**
run-2801-Z3-threads-4-qflia-30s-smtparallel-bugfixes-stats.md, run-2800-Z3-threads-4-qflia-30s-smtparallel-bugfixes-stats.md
run-2797-Z3-threads-4-qfnia-30s-smtparallel-bugfixes-stats.md
compare to old smt_parallel:
run-2747-Z3-threads-4-qflia-30s-stats.md
run-2743-Z3-threads-4-qfnia-30s-stats.md

Note that there is a slight regression on lia in run-2800. The source of this appears to be the new new LP largest-cube LIA heuristic param, which is enabled by default. disabling this param in run-2801 restored performance (I didn't change this in this PR though, just something to note)

http://mtzguido.tplinkdns.com:8081/z3/compare_stats.html


NikolajBjorner:
@copilot 

Can you avoid the const_cast in the functions in smt_solver.cpp by creating const versions of relevant functions or otherwise?

expr_ref_vector get_assigned_literals() override {
            expr_ref_vector result(m);
            auto& ctx = const_cast<smt::kernel&>(m_context).get_context();
            for (auto lit : ctx.assigned_literals()) {
                expr* atom = ctx.bool_var2expr(lit.var());
                if (!atom)
                    continue;
                result.push_back(lit.sign() ? m.mk_not(atom) : atom);
            }
            return result;
        }

        unsigned get_assign_level(expr* e) const override {
            auto& ctx = const_cast<smt::kernel&>(m_context).get_context();
            expr* atom = e;
            get_manager().is_not(e, atom);
            if (!ctx.b_internalized(atom))
                return UINT_MAX;
            return ctx.get_assign_level(ctx.get_bool_var(atom));
        }

        bool is_relevant(expr* e) const override {
            auto& ctx = const_cast<smt::kernel&>(m_context).get_context();
            expr* atom = e;
            get_manager().is_not(e, atom);
            return ctx.b_internalized(atom) && ctx.is_relevant(atom);
        }

        unsigned get_num_bool_vars() const override {
            return const_cast<smt::kernel&>(m_context).get_context().get_num_bool_vars();
        }

        sat::bool_var get_bool_var(expr* e) const override {
            auto& ctx = const_cast<smt::kernel&>(m_context).get_context();
            expr* atom = e;
            get_manager().is_not(e, atom);
            return ctx.b_internalized(atom) ? ctx.get_bool_var(atom) : sat::null_bool_var;
        }

NikolajBjorner:
@copilot also fix the build errors

../src/sat/smt/atom2bool_var.cpp:1:1: error: unknown type name 'add'
    1 | add a smart constructor to seq_util to create the simplified range given two characters. It would create either the empty regex, the singleton string or a range expression./*++
      | ^
../src/sat/smt/atom2bool_var.cpp:1:6: error: expected ';' after top level declarator
    1 | add a smart constructor to seq_util to create the simplified range given two characters. It would create either the empty regex, the singleton string or a range expression./*++
      |      ^
      |      ;

NikolajBjorner:
@copilot 

lits that is returned should be a sorted version of candidates instead of just the highest scoring literal.
It should sort candidates by highest score first.
It can do something similar to the sat_solver by shuffling candidates with the same scores.

This is the relevant code in smt_solver that should be updated:

                    if (!vars.empty() && !selected_vars.is_marked(e))
                        continue;
                    candidates.push_back(e);
                    double new_score = ctx.get_activity(v);
                    // Match smt_parallel: cube split tie-breaking is independent
                    // of the per-worker SMT search seed.
                    if (new_score > score || !result || (new_score == score && m_rand(++n) == 0)) {
                        score = new_score;
                        result = e;
                    }
                }

                vars.reset();
                vars.append(candidates);

                expr_ref_vector lits(m);
                if (result)
                    lits.push_back(result);
                return lits;

PR #9838, opened 2026-06-12
Add versioned shared object names in Bazel rules

shantanu-gontia:
Adds versioned shared library names (`libz3.so.4.17`, `libz3.4.17.dylib`, etc.) alongside the unversioned ones to enable Bazel rules to locate the correct library binaries.


shantanu-gontia:

PR #9840, opened 2026-06-12
Porting seq_split to master

CEisenhofer:


CEisenhofer:
@NikolajBjorner Currently just the code. Still needs to be used within the legacy solver

CEisenhofer:
Update: Added it to the seq solver (probably quite buggy)

src/ast/rewriter/seq_split.h
```diff
@@ -0,0 +1,118 @@
...
+struct split_pair {
+    expr_ref m_d;
+    expr_ref m_n;
+    split_pair(expr* d, expr* n, ast_manager& m) : m_d(d, m), m_n(n, m) {
+        SASSERT(d && n);
+    }
+};
+
+// A split-set is a union of individual splits.
+typedef vector<split_pair> split_set;
```
NikolajBjorner, line 45: what about making this API more abstract.
The service it provides is to expose an iterator of split_sets.

You can have the iterator return a std::option<split_pair> to communicate failure case. Or you can have the iterator expose a method if it provided only a partial split enumeration.

namespace seq {
    class split {
        struct imp;
        imp * m_imp;
    public;
        split(...);
        class iterator { // see term_enumerator branch for term_enumerator.h

        };

        class set {
        public:
             iterator begin();
             iterator end();
        };

        set operator(expr* r);

    };
}

src/ast/rewriter/seq_split.h
```diff
@@ -0,0 +1,118 @@
...
+    ast_manager&   m() const;
+    seq_util&      seq() const;
+    seq_util::rex& re() const;
+
+    // Push <d, n> onto `out`, unless `oracle` rejects it.
+    void push(split_set& out, split_oracle const& oracle, expr* d, expr* n) const;
+
+    // S1 cap S2 = { <D1 cap D2, N1 cap N2> } dropping any pair with a bottom
+    // component (and any rejected by `oracle`).  Returns false on threshold overrun.
+    bool intersect(split_set const& s1, split_set const& s2, split_set& result,
```
NikolajBjorner, line 139: how about lazy expansion of split sets?
Considering the notes in z3paper/resplit/prompts it outlines an algebraic datatype corresponding to suspended computations. 
You can use the expr class with declarations that are local to this class to maintain these suspended computations instead of coming up with a separate type.

Then considering "threshold". Could it be controlled outside of this class given that expansion would be lazy. Even lazy evaluation could end up bloating space so you may eventually keep the threshold parameter

NikolajBjorner:
NB @veanes

NikolajBjorner:
there are changes to seq_rewriter.
Can these go into a separate PR?
Some changes are cosmetic such as replacing "else if" by "if". 
What are the substantial ones?
Note that the changes are in code that the derive branch is replacing

src/smt/seq_regex.cpp
```diff
@@ -128,6 +128,31 @@ namespace smt {
             return;
         }
 
+        if (th.get_fparams().m_seq_regex_factorization_enabled) {
```
NikolajBjorner, line 131: this is for legacy solver.
Shouldn't we focus on c3 branch?
With updates like this we also have to test and fine tune parameters. Preferrably not guard functionality under parameter settings.

NikolajBjorner:
@CEisenhofer - can you also add unit tests for this pr?

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+    seq_split.cpp
+
+Abstract:
+
+    Regex split decomposition (the split function sigma).  See seq_split.h.
+
+Author:
+
+    Nikolaj Bjorner (nbjorner) 2026-6-10
```
NikolajBjorner: ?

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+        kept.push_back(row_r);
+    }
+
+    split_set result;
+    for (row const& k : kept)
+        result.push_back(pairs[k.idx]);
+    pairs.swap(result);
+}
+
+std::pair<expr*, expr*> seq_split::split_membership(expr* str, expr* regex, unsigned threshold, split_set& result) const {
```
NikolajBjorner: reference counts on returned expressions appear not accounted for in a self-contained way. Can you return a pair of expr_ref?

src/ast/rewriter/seq_split.h
```diff
@@ -0,0 +1,118 @@
...
+// Optional lookahead oracle.  Called for each candidate split <D, N> as it is
+// generated; returns true to keep it, false to prune it.  An empty oracle (the
+// default) keeps everything, so sigma is unchanged.  See seq_split::compute.
+typedef std::function<bool(expr* D, expr* N)> split_oracle;
+
+class seq_split {
+    seq_rewriter& m_rw;       // for mk_re_append + manager / seq_util access
+    seq_subset    m_subset;   // language-subset checks for subsumption
+
+    ast_manager&   m() const;
```
NikolajBjorner: just use the attribute "m" for ast_manager and don't introduce this one.
It was an old convention that turned out to be pure overhead.

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+    SASSERT(head && tail);
+
+    // Build the constant lookahead c and (if non-empty) an oracle that
+    // prunes splits whose postfix cannot match c.
+    zstring c;
+    for (i = 0; i < run_len; ++i) {
+        unsigned cv;
+        VERIFY(seq().str.is_unit(tokens.get(run_start + i), ch));
+        VERIFY(seq().is_const_char(ch, cv));
+        c = c + zstring(cv);
```
NikolajBjorner, line 775: first build a vector, then create a string from that vector. Concatenating non-mutual strings is a perf trap.

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+    }
+    split_oracle oracle;
+    if (!c.empty())
+        oracle = [this, &c](expr*, expr* n) { return split_lookahead_viable(n, c); };
+
+    // Decompose the regex into a split-set via the shared seq_split engine
+    if (!m_rw.split(regex, result, threshold, split_mode::strong, oracle)) {
```
NikolajBjorner, line 782: is this roundabout?
seq_rewriter calls compute or split again, right?

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+        return { nullptr, nullptr };
+    }
+
+    m_rw.simplify_split(result);
+
+    // Eagerly consume the constant run c from the tail by taking the c-derivative
+    // of each postfix
+    if (!c.empty()) {
+        unsigned w = 0;
+        for (i = 0; i < result.size(); ++i) {
```
NikolajBjorner, line 793: can you use for comprehension whenever it applies?

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+}
+
+std::pair<expr*, expr*> seq_split::split_membership(expr* str, expr* regex, unsigned threshold, split_set& result) const {
+    ast_manager& m = this->m();
+
+    expr_ref_vector tokens(m);
+    vector<expr*> stack;
+    stack.push_back(str);
+
+    while (!stack.empty()) {
```
NikolajBjorner, line 688: this is a utility function somewhere

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+            }
+            ch = seq().mk_char(s[0]);
+            tokens[i] = seq().str.mk_string(s.extract(1, s.length() - 1));
+        }
+        else
+            i++;
+        regex = m_rw.mk_derivative(ch, regex);
+    }
+
+    if (i > 0) {
```
NikolajBjorner, line 718: there may be a self-contained function for that somewhere. If not, why not add it? You are shifting a suffix into a prefix.

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+    if (!c.empty())
+        oracle = [this, &c](expr*, expr* n) { return split_lookahead_viable(n, c); };
+
+    // Decompose the regex into a split-set via the shared seq_split engine
+    if (!m_rw.split(regex, result, threshold, split_mode::strong, oracle)) {
+        result.clear();
+        return { nullptr, nullptr };
+    }
+
+    m_rw.simplify_split(result);
```
NikolajBjorner: isn't this just calling into seq_split?

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+        }
+    }
+    pairs.swap(out);
+}
+
+void seq_split::simplify(split_set& pairs) const {
+    seq_util::rex& r = re();
+
+    // 1. drop pairs with a bottom (empty-language) component.
+    unsigned w = 0;
```
NikolajBjorner, line 625: this is a "filter" function.

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+    expr* a = nullptr, *b = nullptr;
+
+    // to_re(s): split the literal word s at every position.
+    expr* s = nullptr;
+    if (rex.is_to_re(r, s)) {
+        zstring str;
+        vector<expr*> stack;
+        stack.push_back(s);
+
+        while (!stack.empty()) {
```
NikolajBjorner, line 277: The code optimizes but for fail fast but could be constructed from basic pieces and probably without a blimp on the perf radar:
first get concats.
Then create string,
return false if there is a concat that isn't a string.

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+    if (rex.is_full_seq(r)) {
+        const expr_ref ex(r, mm);
+        push(result, oracle, ex, ex);
+        return true;
+    }
+
+    // union: sigma(r0 | ... | r_{n-1}) = U sigma(ri)   (re.union may be n-ary)
+    if (rex.is_union(r)) {
+        app* ap = to_app(r);
+        for (unsigned i = 0; i < ap->get_num_args(); ++i) {
```
NikolajBjorner: you can iterate direcly:
for (expr* arg : *ap) {

}

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+        }
+        return true;
+    }
+
+    // concat: sigma(r0...r_{n-1}) = U_i  (r0...r_{i-1}) . sigma(ri) . (r_{i+1}...r_{n-1})
+    // (re.++ may be n-ary)
+    if (rex.is_concat(r)) {
+        app* ap = to_app(r);
+        const unsigned n = ap->get_num_args();
+        for (unsigned i = 0; i < n; ++i) {
```
NikolajBjorner, line 340: you can iterate directly:
for (auto arg : *ap) {

}

src/ast/rewriter/seq_split.cpp
```diff
@@ -0,0 +1,545 @@
...
+                for (unsigned j = 0; j < i; ++j) {
+                    expr* arg = ap->get_arg(j);
+                    left = left ? expr_ref(rex.mk_concat(left, arg), mm) : expr_ref(arg, mm);
+                }
+            }
+            if (i == n - 1)
+                right = rex.mk_epsilon(seq_sort);
+            else {
+                right = ap->get_arg(i + 1);
+                for (unsigned j = i + 2; j < n; ++j) {
```
NikolajBjorner, line 354: use iterator for (auto arg : *ap) {

}

src/ast/rewriter/seq_split.cpp
CEisenhofer: Yep; it is. Residue of the porting

src/ast/rewriter/seq_split.cpp
CEisenhofer: Sure; will do

src/ast/rewriter/seq_split.cpp
CEisenhofer: Sure

src/ast/rewriter/seq_split.cpp
CEisenhofer: Why not 🙃?

src/smt/seq_regex.cpp
CEisenhofer: But it might affect performance quite a bit, so a feature flag seems the way to go. Esp., since this feature is actually for the nseq solver.
So just having it enabled always?

src/ast/rewriter/seq_split.h
CEisenhofer: Have it locally. WIP. However, currently quite buggy, so comitted this "eager" approach for now to collect some feedback before leaving. (Will apply the changed once I am back on the 22nd.)

PR #9856, opened 2026-06-14
Prevent unsound lia2card elimination across recfun definitions

### LINKED ISSUE #9855: lia2card drops a bounded equality that still flows through a recfun definition
lia2card drops a bounded equality that still flows through a recfun definition

I hit a soundness bug on current master where `lia2card` removes a bounded integer equality even though the constant still appears in the body of a reachable recursive function.

Minimal repro:

```smt2
(set-option :model_validate true)
(declare-fun fixedAdd () Int)
(declare-fun variableAdd () Int)
(define-fun-rec $$add$$ ((a Int) (b Int)) Int
  (ite (= 0 b) 2 (- a (+ 0 (- fixedAdd b)))))
(assert (= fixedAdd 0))
(assert (= 1 ($$add$$ 1 3)))
(check-sat-using (then lia2card smt))
```

Current result on master:

```text
sat
(error "line 8 column 36: an invalid model was generated")
```

Expected result:

```text
unsat
```

`lia2card` looks like the source of the problem:

- `check-sat-using smt` returns `unsat`
- `apply lia2card` rewrites the goal to

```text
true
(= 1 ($$add$$ 1 3))
```

That rewrite drops `(= fixedAdd 0)` even though `fixedAdd` still constrains `$$add$$` through the recfun body. The weakened goal is satisfiable if the solver picks `fixedAdd = 3`, and model reconstruction later forces `fixedAdd = 0`, so model validation fails.

I also reproduced the same shape from a larger current-master regression:

```smt2
(set-option :model_validate true)
(declare-fun fixedAdd () Int)
(declare-fun variableAdd () Int)
(define-fun-rec $$add$$ ((a Int) (b Int)) Int
  (ite (= 0 b) 2 (- a (+ 0 (- fixedAdd b)))))
(assert (= fixedAdd (* 9 fixedAdd)))
(assert (= 1 ($$add$$ 1 3)))
(check-sat)
```

That query currently reports:

```text
sat
(error "line 8 column 10: an invalid model was generated")
```

I have a local fix that makes `lia2card` skip bounded replacement for constants that occur in recursive-function RHS definitions reachable from the current goal. That fix also makes `./test-z3 /seq fpa` pass with an added regression.

peter941221:
Prevent lia2card from erasing bounded constants that still flow through recfun definitions

Problem

`lia2card` can replace a bounded integer constant even when that constant still appears in the RHS of a recursive function that is reachable from the current goal.

On current master, this turns an unsatisfiable query into `sat` plus model-validation failure:

```smt2
(set-option :model_validate true)
(declare-fun fixedAdd () Int)
(declare-fun variableAdd () Int)
(define-fun-rec $$add$$ ((a Int) (b Int)) Int
  (ite (= 0 b) 2 (- a (+ 0 (- fixedAdd b)))))
(assert (= fixedAdd 0))
(assert (= 1 ($$add$$ 1 3)))
(check-sat-using (then lia2card smt))
```

Current result:

```text
sat
(error "line 8 column 36: an invalid model was generated")
```

Root cause

`lia2card` sees `fixedAdd` as a bounded uninterpreted integer constant and rewrites `(= fixedAdd 0)` away.

`apply lia2card` on the repro currently produces:

```text
true
(= 1 ($$add$$ 1 3))
```

That rewrite is not sound here because `fixedAdd` still constrains `$$add$$` through the recfun body. The weakened goal becomes satisfiable, and model reconstruction later rebuilds an inconsistent model.

Change

- collect recfun declarations reachable from the current goal
- traverse their RHS definitions
- mark uninterpreted constants from those definitions as unsafe for bounded replacement
- skip `lia2card` replacement for those constants

Tests

Added a focused regression in `src/test/fpa.cpp`:

```smt2
(set-option :model_validate true)
(declare-fun fixedAdd () Int)
(declare-fun variableAdd () Int)
(define-fun-rec $$add$$ ((a Int) (b Int)) Int
  (ite (= 0 b) 2 (- a (+ 0 (- fixedAdd b)))))
(assert (= fixedAdd 0))
(assert (= 1 ($$add$$ 1 3)))
(check-sat-using (then lia2card smt))
```

Validation

- `./test-z3 /seq fpa`
- the direct minimal `lia2card` repro now returns `unsat`
- `apply lia2card` on the minimal repro now preserves `(= fixedAdd 0)` instead of rewriting it to `true`

Non-regression smoke

- `x = 0` without recfuns still rewrites to `true`
- bounded 0/1 integer constraints still compile to cardinality form

Closes #9855

### DIFF
--- a/src/tactic/arith/lia2card_tactic.cpp
+++ b/src/tactic/arith/lia2card_tactic.cpp
@@ -24,6 +24,8 @@ Notes:
 #include "ast/rewriter/expr_safe_replace.h"
 #include "ast/ast_util.h"
 #include "ast/ast_pp_util.h"
+#include "ast/decl_collector.h"
+#include "ast/recfun_decl_plugin.h"
 #include "tactic/tactical.h"
 #include "params/tactic_params.hpp"
 #include "ast/simplifiers/bound_manager.h"
@@ -178,13 +180,29 @@ public:
         }
     }
 
+    void mark_recfun_dependent_consts(goal const& g, expr_mark& unsafe_vars) {
+        decl_collector dc(m);
+        for (unsigned i = 0; i < g.size(); ++i)
+            dc.visit(g.form(i));
+        recfun::util rec(m);
+        for (func_decl* f : dc.get_rec_decls()) {
+            expr_ref rhs(rec.get_def(f).get_rhs(), m);
+            for (expr* term : subterms::all(rhs)) {
+                if (is_uninterp_const(term))
+                    unsafe_vars.mark(term);
+            }
+        }
+    }
+
     void operator()(goal_ref const & g, goal_ref_buffer & result) override {
         m_bounds.reset();
         m_mc.reset();
         expr_ref_vector axioms(m);
         expr_safe_replace rep(m);
+        expr_mark unsafe_vars;
 
         tactic_report report("lia2card", *g);
+        mark_recfun_dependent_consts(*g, unsafe_vars);
 
         bound_manager bounds(m);
         for (unsigned i = 0; i < g->size(); ++i)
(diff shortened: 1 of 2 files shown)


peter941221:

NikolajBjorner:
superseeded by global check

PR #9891, opened 2026-06-17
python: build a PyPI-publishable Pyodide (PEP 783) wheel

alcides:
## Summary

z3 already builds under Pyodide (there is a `pyodide.yml` workflow and an
`IS_PYODIDE` path in `setup.py`), but that path uses `pyodide build` and produces
a wheel tagged `emscripten_<pyodide-version>_wasm32`, which is pinned to a single
Pyodide release and rejected by PyPI — so today it's only usable as a CI artifact.

[PEP 783](https://peps.python.org/pep-0783/) introduced the portable
`pyemscripten_<date>_wasm32` platform tag that **PyPI accepts** and `micropip` can
install at runtime. This makes `z3-solver` build that wheel via
`cibuildwheel --platform pyodide`.

## Changes

- **`setup.py`** — for the emscripten target, use the wheel platform tag that
  pyodide-build provides verbatim via `_PYTHON_HOST_PLATFORM` (e.g.
  `pyemscripten_2026_0_wasm32`) instead of reconstructing an `emscripten_*` tag.
  Falls back to the previous behaviour when that env var is absent.
- **`setup.py` / `CMakeLists.txt` / `pyproject.toml`** — switch the Pyodide build
  from JS-based exceptions (`-fexceptions`, `-sDISABLE_EXCEPTION_CATCHING=0`) to
  **native wasm exception handling** (`-fwasm-exceptions -sSUPPORT_LONGJMP=wasm`),
  matching the ABI of the Pyodide 314 / emscripten 5 main module. With the old
  flags `libz3.so` imports `invoke_*` trampolines the runtime no longer provides,
  so the wheel builds but the first `Z3_mk_config` call fails at runtime with
  `Dynamic linking error: cannot resolve symbol invoke_vi`.
- **`pyproject.toml`** — add `[tool.cibuildwheel]` / `[tool.pyodide.build]` so
  `cibuildwheel --platform pyodide` builds and tests the wheel (`cp314`).
- **`.github/workflows/pyodide-pypi.yml`** (new) — build with cibuildwheel and
  publish to PyPI (trusted publishing) on tags. Existing `pyodide.yml` unchanged.

## Verification

Built with `cibuildwheel 4.1.0` / `pyodide-build 0.35.0` / `emscripten 5.0.3`,
CPython 3.14 / Pyodide 314:

- Produces `z3_solver-4.17.0.0-py3-none-pyemscripten_2026_0_wasm32.whl`.
- `z3test.py` passes in the Pyodide runtime (node + wasm32).
- Installed via `micropip` and solves SMT problems both under node and in a
  browser (`sat` with a model, `unsat` for a contradiction).

🕵️‍♂️ Reviewed by [Alcides Fonseca](https://wiki.alcidesfonseca.com)

### DIFF
--- a/src/api/python/pyproject.toml
+++ b/src/api/python/pyproject.toml
@@ -1,3 +1,29 @@
 [build-system]
 requires = ["setuptools>=70"]
 build-backend = "setuptools.build_meta"
+
+# --- Pyodide / WebAssembly (PEP 783) build configuration ---------------------
+# Consumed by pyodide-build (invoked directly or via `cibuildwheel --platform
+# pyodide`). These flags are forwarded to the emscripten toolchain that compiles
+# libz3 to wasm32. setup.py's IS_PYODIDE branch appends the same -fexceptions
+# flags defensively, but declaring them here is what cibuildwheel relies on.
+# Pyodide 314 / emscripten 5 builds its main module with *native wasm*
+# exception handling and wasm longjmp (see Pyodide's Makefile.envs). Side
+# modules like libz3.so MUST match that ABI: building with the legacy
+# `-fexceptions` (JS-based EH) makes libz3 import `invoke_*` trampolines that the
+# Pyodide runtime no longer provides -> "cannot resolve symbol invoke_vi" at the
+# first Z3 call. WASM_BIGINT is required because the Z3 C API passes 64-bit ints
+# across the ctypes/JS boundary.
+[tool.pyodide.build]
+cflags = "-fwasm-exceptions -sSUPPORT_LONGJMP=wasm"
+cxxflags = "-fwasm-exceptions -sSUPPORT_LONGJMP=wasm"
+ldflags = "-fwasm-exceptions -sSUPPORT_LONGJMP=wasm -sWASM_BIGINT"
+
+# --- cibuildwheel: produce a PyPI-publishable pyemscripten wheel -------------
+[tool.cibuildwheel]
+# Pyodide 314 ships CPython 3.14; match the single ABI it targets.
+build = "cp314-*"
+
+[tool.cibuildwheel.pyodide]
+# z3test.py is the upstream smoke test; run it inside the Pyodide test venv.
+test-command = "python {project}/z3test.py z3"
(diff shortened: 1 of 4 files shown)


alcides:
@microsoft-github-policy-service agree company="University of Lisbon"

src/api/python/pyproject.toml
```diff
@@ -1,3 +1,29 @@
...
+# across the ctypes/JS boundary.
+[tool.pyodide.build]
+cflags = "-fwasm-exceptions -sSUPPORT_LONGJMP=wasm"
+cxxflags = "-fwasm-exceptions -sSUPPORT_LONGJMP=wasm"
+ldflags = "-fwasm-exceptions -sSUPPORT_LONGJMP=wasm -sWASM_BIGINT"
+
+# --- cibuildwheel: produce a PyPI-publishable pyemscripten wheel -------------
+[tool.cibuildwheel]
+# Pyodide 314 ships CPython 3.14; match the single ABI it targets.
+build = "cp314-*"
```
NikolajBjorner, line 25: we have to change this with new versions of Python?

NikolajBjorner:
@alcides - do you know what happened to the nightly build?

src/api/python/pyproject.toml
alcides: Unfortunately, it's static. But my guess is that you can keep the oldest version supported (or the latest, probably in this case).

alcides:
The build used an older version of pyodide, which is incompatible with these changes. I create #9931 to fix this.

Pyodide and emscripten do not support all versions of Python. My recommendation would be for z3 to support the latest stable version of Python (3.14 right now) that matches the latest stable version of pyodide. Keeping all versions working looks like a little hell, until this python-in-wasm thing stabilizes.

My experience tells me that the browser/javascript communities moves fast with new releases, while Python can have multiple older applications running successfully (I know several python2 scripts and apps running in production today). Even [if that comes at the cost of vulnerabilities](https://safedep.io/mini-shai-hulud-strikes-again-314-npm-packages-compromised/).

PR #9892, opened 2026-06-17
Use "override" keyword where needed.

davedets:

The PR adds
```
  "-Wsuggest-override"
  "-Winconsistent-missing-override"
 ```
 to the CLANG_ONLY_WARNINGS.  This exposes a relatively small number of places where method overrides did not use the "override" keyword.  The PR fixes those.
 
(In cmd_util.h, I also made the *_CMD macros be uniform in not ending the class they define with a semicolon; the invocation of the macro can add the semicolon.)


davedets:
Oops: I had only done a debug build; StaticMatrix<T> is a subclass of Matrix<T> only in debug builds, so the "override" has to be dependent on Z3DEBUG.  FIxed that.

PR #9904, opened 2026-06-19
Fix missing field initializers.

davedets:

This one fixes warnings about "missing field initializers" -- struct initializers that leave some field uninitialized.


src/util/obj_hashtable.h
```diff
@@ -62,6 +62,11 @@ class obj_map {
     struct key_data {
         Key * m_key = nullptr;
         Value m_value;
+
```
NikolajBjorner, line 65: did @nunoplopes remove these methods earlier this week?

nunoplopes:
This patch is the wrong approach.

First of all, the warning is a false positive. It's more a symptom of a bad design of the find/remove functions that take a key and a value, but they don't touch the value.

If you really want to get rid of the waning, there's the easy way that degrades performance (`Value m_value{};`) and the proper way (fix find, remove, and friends).

davedets:
I'd be happy to attempt the fix that Nuno proposes; I was trying not to be too intrusive.  Alternatively, I could just disable the warning.  (My "fix" in this PR is not avoiding any actual errors, it's just silencing the warnings.).  I'll work on the first approach, but let me know if people would prefer the second.

davedets:
I went down the path I understood Nuno to be suggesting for a couple of hours -- trying to make the Entry type of core_hashtable allow access to the key of the entry, so that the signatures of find, etc., could take just the key type, not the whole entry.  This got pretty complicated, and the changes, if I were able to succeed, were going to be extensive.  I  gave up.

Has the question of whether the custom hashtables provide a significant perf boost compared to the standard library implementations been settled?  If not, is it worth investigating?

PR #9908, opened 2026-06-20
Term enumeration

NikolajBjorner:


NikolajBjorner:
@coplilot 
- move term_enumeration.cpp/h from the ast directory to the ast/rewriter directory.
- make sure tests build too (test-z3) with updated include paths in test/term_enumeration.cpp
- add a observational equivalence filter to term_enumeration based on th_rewriter. It applies rewriting to generated terms and returns the rewritten term as a result. If a previous term that rewrites to the same was already produced, then don't return it. For example, the generator should not produce both (+ 0 1) and (+ 1 0) and 1. They all simplify to the same

PR #9931, opened 2026-06-23
python/pyodide: pin cibuildwheel pyodide-version to 314.0.0

alcides:
Follow-up to #9891 (merged), which added the PEP 783 `pyemscripten_*_wasm32` Pyodide wheel build.

## What

Pin the Pyodide runtime cibuildwheel builds against:

```toml
[tool.cibuildwheel.pyodide]
pyodide-version = "314.0.0"
```

## Why

The wheel's emscripten ABI — and therefore its `pyemscripten_YYYY_P_wasm32` platform tag — is determined by the Pyodide version cibuildwheel resolves. Today that's left implicit, so a future cibuildwheel default bump would silently change the tag and the runtime the wheel is installable on.

`314.0.0` is CPython 3.14 / emscripten 5.0.3 → `pyemscripten_2026_0_wasm32`, which is exactly what `cibuildwheel==4.1.0` (already pinned in `pyodide-pypi.yml`) resolves by default for `cp314`. So this changes nothing about the produced wheel today — it just makes the choice explicit and reproducible, and documents that the in-browser Pyodide must be loaded in lockstep when this is bumped.

Concretely motivated by the current Nightly Pyodide wheel being tagged `pyemscripten_2024_0` (emscripten 3.1.58 / Pyodide 0.26 ABI): that binary only matches Pyodide 0.26, whose micropip predates the PEP 783 `pyemscripten_` prefix and rejects the tag — so it isn't `micropip`-installable on any released Pyodide. Pinning the version prevents that kind of tag/runtime mismatch in this build path.

No functional change to the wheel; CI build/test path is unchanged.

### DIFF
--- a/.github/workflows/pyodide-pypi.yml
+++ b/.github/workflows/pyodide-pypi.yml
@@ -1,0 +1,57 @@
+name: Pyodide Wheel (PyPI)
+
+# Builds a PEP 783 `pyemscripten_*_wasm32` wheel for z3-solver using cibuildwheel
+# and publishes it to PyPI on tag pushes. Unlike the legacy pyodide.yml (which
+# uses `pyodide build` and produces a Pyodide-version-locked emscripten_* wheel
+# uploaded only as a CI artifact), this wheel is installable at runtime with
+# micropip from PyPI. See src/api/python/pyproject.toml [tool.cibuildwheel].
+
+on:
+  release:
+    types: [created]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build-pyodide:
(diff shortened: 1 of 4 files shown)


NikolajBjorner:
the PR has some conflicts.
the old pyodide builds got broken. (nightly, pyodide build - trying to throw copilot at it to patch whatever shifted)

NikolajBjorner:
resolve conflicts and add PR again

NikolajBjorner:
@alcides 
https://github.com/Z3Prover/z3/actions/workflows/pyodide-pypi.yml
what is intent wrt publishing?

PR #9957, opened 2026-06-25
Remove unnecessary semicolons.

davedets:

This one adds:
```
  "-Wextra-semi"
```

which exposes a large number of extraneous semicolons.  Three main categories:

* after the closing brace of namespace decl.
* after the closing brace of an extern "C" decl.
* after a function definition.

There are some macros that define functions, so "DEFN_MACRO(...);" also produced warnings, and I deleted these semis.

This PR is very large, but it consists entirely of deletions of semicolons.  (So while the number of files and lines touched is large, the number of actual characters changed is relatively small :-)).   I propose it, rather than just disabling the warning, because I think these changes really do improve and modernize the code base.  The extra semicolons in these places are (if Google's summary of the standard is to be believed) illegal by the letter of the law, and allowing them to be warnings relies on the mercy of the compiler.  The use of semis after function defs was inconsistent, sometimes in the same file.

If there's a policy on large PRs like this, like doing them incrementally, please let me know.


src/api/api_arith.cpp
```diff
@@ -76,11 +76,11 @@ extern "C" {
         Z3_CATCH_RETURN(nullptr);
     }
```
NikolajBjorner, line 78: these semi-colons help with some editors

NikolajBjorner:
I don't think it makes sense to remove semi-colons after the macros. There are many instances here.
Editors start aligning text oddly if they don't see the semi-colons. They don't look into the macros.
If the macros can be changed to avoid linters to complain, sure. Otherwise, it is an overall loss for code maintainance due to editor behavior.

NikolajBjorner:
- new conflicts in PR
- I don't understand the value of removing the ; after macro instantiations when it breaks some formatting.

davedets:
Could you give me guidance on how to proceed?  Would you be inclinded to accept this PR if I backed out of the semicolon deletions for the macros (leaving the ones following namespace or extern "C" closing braces, and function definitions)?

If so: for the macros, how would you feel about a pattern like
```
    DISABLE_SEMICOLON_WARNINGS;
    MK_ARITH_OP(Z3_mk_add, OP_ADD);
    MK_ARITH_OP(Z3_mk_mul, OP_MUL);
    MK_BINARY_ARITH_OP(Z3_mk_power, OP_POWER);
    MK_BINARY_ARITH_OP(Z3_mk_mod, OP_MOD);
    MK_BINARY_ARITH_OP(Z3_mk_rem, OP_REM);
    ENABLE_SEMICOLON_WARNINGS;
```
for blocks of macro instantiations?  I think I'd have to introduce a dummy struct declaration, with a uniquish name using the line number, I guess, at the end of ENBLE_SEMICOLON_WARNINGS, to make the semicolon after it legal.  I could also do something like this as part of the definitions of the macros, avoiding the enable/disable macros, but that seems like a lot of such dummy declarations.

Would it be better if I broke this up into a series of PRs, like say 10, with 40-ish files changed per?  Or is it OK to do this all in one go?

davedets:
@NikolajBjorner : Gentle ping: as I asked in my last comment, could you give me guidance on how to proceed?  I'm sorry I'm not yet familiar with the customary workflow for Z3 PRs -- what does your closing the PR indicate?  I was reading your comments as being OK with the bulk of the changes, but not the macro ones.  Did I misread that?

PR #9958, opened 2026-06-25
lp: gate Gomory-with-dio on genuine dio failures; separate config from runtime state

levnach:
## Summary

Improves the Diophantine (`dio`) integer-feasibility controller in `int_solver`, and fixes a latent bug where the user's Gomory-cut configuration could be silently overridden at runtime. Also includes the earlier `lia_w` work: randomized hammer gates, the `int_hammer_period` / `random_hammers` parameters, and the linear `dio_calls_period` recovery.

## Motivation

The controller used a **single field** both as the static `lp.dio_cuts_enable_gomory` parameter and as the live "is Gomory running" flag. It started running Gomory (and the gcd test) once `dio_calls_period` crossed a hard-coded `16`. Because `dio_calls_period` is also driven by the randomized hammer gate, on instances where `dio` is only intermittently productive the period could be ratcheted past 16 *by chance*, turning on Gomory + gcd and thrashing — e.g. `dillig/20-14` went from a 100s solve (deterministic) to a 600s timeout (randomized) purely from this spurious activation.

## Changes

- **Separate config from runtime state.** Split the shared field into `m_dio_cuts_enable_gomory` (static config, never mutated) and `m_run_gomory_with_dio` (runtime flag). Toggling the runtime state can no longer clobber the user's `dio_cuts_enable_gomory` parameter.
- **Trigger on genuine dio failures, not the period proxy.** Running Gomory-with-dio now starts after a count of **consecutive `undef` dio returns** (reset on a dio conflict) rather than when the randomization-inflated period crosses a threshold — robust to `random_hammers` gate variance.
- **Parameterize the threshold.** New `lp.dio_gomory_enable_period` (default 16). Set it very large to never auto-start Gomory, so Gomory follows `dio_cuts_enable_gomory` only.
- **Try `dio` before Gomory** in `check()` so a productive dio conflict preempts Gomory on dio-dominated instances.

## Evaluation (QF_LIA, full set, 600s, seed 555 paired)

- Dio-before-Gomory: **+33** problems across the 6 `random_hammers x int_hammer_period` cells (5/6 cells improve).
- New trigger (`dio_gomory_enable_period=32`, random): **6417** vs the period-16 baseline **6409**; no short-cutoff regression.
- Linear `dio_calls_period` recovery: keeping it on is worth ~+20 vs off; `decrease=1` slightly ahead of the default 2.

Default behavior (`dio_gomory_enable_period=16`) is byte-for-byte equivalent to the previous threshold logic.

## Notes

Debug-only tracing used during analysis (the `dio_calls_period_trace` parameter plus per-hammer / period-evolution verbose output) is **not** included.


levnach:
## Benchmark: `lia_w` vs `master` (QF_LIA, `-T:20`, default options, seeds 1/2/3)

Problems solved within N seconds (sat/unsat):

| run | ≤1s | ≤2s | ≤4s | ≤8s | ≤16s | ≤20s | total |
|:--|--:|--:|--:|--:|--:|--:|--:|
| lia_w_rs1 | 8693 | 9383 | 10186 | 10710 | 11070 | 11207 | 11211 |
| lia_w_rs2 | 8703 | 9385 | 10170 | 10704 | 11088 | 11201 | 11204 |
| lia_w_rs3 | 8689 | 9347 | 10175 | 10690 | 11074 | 11200 | 11200 |
| master_rs1 | 8529 | 9216 | 10007 | 10566 | 10948 | 11084 | 11085 |
| master_rs2 | 8538 | 9196 | 9987 | 10546 | 10963 | 11081 | 11084 |
| master_rs3 | 8546 | 9179 | 9996 | 10536 | 10944 | 11074 | 11074 |

Means and delta:

| cutoff | lia_w | master | Δ |
|:--|--:|--:|--:|
| ≤1s | 8695 | 8538 | +157 |
| ≤2s | 9372 | 9197 | +175 |
| ≤4s | 10177 | 9997 | +180 |
| ≤8s | 10701 | 10549 | +152 |
| ≤16s | 11077 | 10952 | +125 |
| ≤20s (total) | 11205 | 11081 | +124 |

**`lia_w` beats `master` at every cutoff** (+124 to +180). At the 20s timeout, lia_w solves **11205** on average vs master **11081** — **+124 (~+1.1%)**. Every lia_w seed (11200–11211) beats every master seed (11074–11085) with no overlap, and the lead is already +157 at ≤1s (not just the slow tail).
The performance gain disappears around 60 seconds as demonstrated by some other tests, but there is no performance lost on longer runs.

PR #9973, opened 2026-06-27
capture row by pointer

levnach:
Capture row as a pointer as lambda strips the reference!


levnach:
image
The effect of the fix. The benchmarks surrounded by the ellipse are solved 2-3 times faster by the new code.

PR #9991, opened 2026-06-28
[coz3-deepfix] nlsat: disable levelwise (lws) by default to fix QF_NIA SIGSEGV

levnach:
## Summary

Fixes a reproducible, non-deterministic **SIGSEGV (signal-11)** in z3 on a QF_NIA benchmark. The crash is in the experimental **levelwise single-cell projection** (`nlsat.lws`, default on). This PR sets its default to `false`; nlsat explanation falls back to the existing `add_all_coeffs` projection, eliminating the crash.

- **Candidate key:** `coz3-https-zenodo.org-records-16740866-files-QF_NIA.tar.zst-dow::non-incremental-QF_NIA-20170427-VeryMax-ITS-From_AProVE_2014__Round3.jar-obl-8__p11898_terminationG_0.smt2`
- **Origin report:** https://github.com/Z3Prover/bench/discussions/2794
- **Stats:** (z3prover.github.io/redacted)
- **Benchmark set:** QF_NIA
- **Query:** `non-incremental/QF_NIA/20170427-VeryMax/ITS/From_AProVE_2014__Round3.jar-obl-8__p11898_terminationG_0.smt2`

## Diagnosis

Reported as `signal-11:2/3` across coz3 runs. Reproduced with a fresh master build at `-T:20 model_validate=true`: ~40% segfault, else timeout/solve. A SIGSEGV-handler backtrace (gdb/valgrind couldn't run: ptrace blocked, too slow) points at:

```
algebraic_numbers::manager::imp::compare
nlsat::levelwise::impl::sort_root_function_partitions   (src/nlsat/levelwise.cpp:977)
nlsat::levelwise::impl::single_cell                     (src/nlsat/levelwise.cpp:1437)
nlsat::explain::imp::levelwise_single_cell              (src/nlsat/nlsat_explain.cpp:944)
nlsat::solver::imp::check
```

The crash is inside the `scoped_anum` comparator during the partition sort. The levelwise path (CADE30 "More is Less") is recent and has a string of heap-corruption fixes; the `scoped_anum` move/sort interactions still corrupt memory. `nlsat_explain` only uses levelwise as an optimization with a `add_all_coeffs` fallback (`nlsat_explain.cpp:1004-1010`).

## Change

`src/nlsat/nlsat_params.pyg`: `lws` default `True` → `False` (one line). Users can re-enable with `nlsat.lws=true`.

## Reproduce / verify (built binary, same query)

[...]

PR #10001, opened 2026-06-29
nlsat: fix levelwise (lws) SIGSEGV instead of disabling it

levnach:
## Summary

Alternative to #9991. Instead of disabling `nlsat.lws` by default, this **fixes the underlying bug** so levelwise single-cell projection stays enabled.

## Root cause

The crash was reproduced on the QF_NIA benchmark from #9991 (`20170427-VeryMax/ITS/From_AProVE_2014__Round3.jar-obl-8__p11898_terminationG_0.smt2`, ~40% SIGSEGV at `-T:20`). A core-dump backtrace points at:

```
mpbq_manager::le                                 (mpbq.cpp:362)
algebraic_numbers::manager::imp::compare         (algebraic_numbers.cpp:1913)  c = 0xea24052d29f2d500  <- wild pointer
algebraic_numbers::manager::imp::compare         (algebraic_numbers.cpp:2128)
nlsat::levelwise::impl::root_function_lt         (levelwise.cpp:949)
... std::__unguarded_linear_insert ...           <- OOB read
std::sort
nlsat::levelwise::impl::sort_root_function_partitions
```

The comparator (`root_function_lt` → `anum_manager::compare`, and `anum_manager::lt`) **refines the isolating intervals of the algebraic numbers it compares** and may **hit the resource limit (throwing)** mid-comparison. Both make the order it induces non-deterministic / not a strict weak ordering across a single `std::sort` — undefined behavior. libstdc++'s *unguarded* insertion pass then walks past `begin()` and dereferences a wild anum cell → SIGSEGV. This only fires when a timeout interrupts levelwise, explaining the non-determinism (`signal-11`).

## Fix

Replace the two affected `std::sort` calls (`sort_root_function_partitions` and `add_adjacent_root_resultants`) with a **bounds-checked insertion sort over an index permutation**. A fully guarded insertion sort can never read out of bounds regardless of comparator consistency, and unwinds cleanly if `compare` throws on cancellation. The partitions sorted here are small, so the O(n²) cost is negligible.

`nlsat.lws` stays `true`.

## Verification

On the Linux repro box (Ubuntu 24.04, g++ 13), RelWithDebInfo:
- **Before:** ~40% SIGSEGV (e.g. 5/16 runs at `-T:20`).
- **After:** **0/30** SIGSEGV; results are `unsat`/`timeout`.
- Sanity batch over 25 QF_NIA/VeryMax/ITS files: no crashes, expected sat/unsat/timeout mix.
- `model_validate=true` full solve still returns `unsat`.

### DIFF
--- a/src/nlsat/levelwise.cpp
+++ b/src/nlsat/levelwise.cpp
@@ -956,34 +956,53 @@ namespace nlsat {
             return m_pm.id(a.ire.p) < m_pm.id(b.ire.p);
         }
 
-        // Sort an index permutation with a bounds-checked insertion sort.
+        // Sort an index permutation with a bounds-safe, mutation-aware merge sort.
         //
-        // We deliberately avoid std::sort here: the comparator (root_function_lt ->
-        // anum_manager::compare) is NOT pure -- it MUTATES the very algebraic numbers
-        // it compares. anum_manager::compare refines their isolating intervals (and may
-        // even collapse a root to a rational), and it can hit the resource limit and
-        // throw mid-comparison. Because the operands change as the sort proceeds, the
-        // order the comparator induces is not a fixed strict weak ordering over a single
-        // sort. std::sort assumes a stable strict weak ordering; violating that is
-        // undefined behavior. In practice libstdc++'s *unguarded* insertion pass then
-        // trusts the comparator to terminate the inner loop, walks past the start of the
-        // range, and dereferences a wild anum cell -> SIGSEGV (this happens via an
-        // out-of-bounds read, independent of whether an exception is thrown, so it
-        // cannot be papered over with a try/catch around the sort).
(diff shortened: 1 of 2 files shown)


src/nlsat/levelwise.cpp
```diff
@@ -956,6 +956,56 @@ namespace nlsat {
             return m_pm.id(a.ire.p) < m_pm.id(b.ire.p);
         }
 
+        // Sort an index permutation with a bounds-safe, mutation-aware merge sort.
```
NikolajBjorner, line 959: isn't this nearly identical AI generated code to the other file?
There has to be some modular approach to deal with sorting vectors?

levnach: Addressed by  https://github.com/Z3Prover/z3/pull/10006

PR #10007, opened 2026-06-30
Improve generation accounting

CanCebeci:
This PR redesigns generation accounting, undoing/rethinking some of the changes from the last two months.

## Changes
* mam.cpp is mostly reverted to its original state (no more equivalence class, `get_f_app`, or cached max generation).
* We now track generations per congruence class, not per enode. We store the generation number in the congruence table. We store the generations of enodes that don't participate in congruence closure in a separate constant generation table.
* Generations change when congruence classes merge (through `merge_cgc_generations`) and when we find a shorter path to an enode (through `update generation`). The latter is sticky and survives backtracking (until the enode itself goes out of scope)
* All equality atoms now have generation 0. Equalities can not be included in patterns, so we don't need to track generations for them.

## Performance
Timeline on Mariposa's unsat core: 
| Version | #Queries with unknowns (/58) | Total # unknowns | 60sec timeouts (/5699) |
| -------- | --------------------------------- | -------------------- | ------------------------- |
| Before May 7th | 28 | 522 | 1818 |
| May 7th  - 'Use the minimum generation number among matching enodes' | 22 | 355 | 1978 |
| May 26th - 'Update generation number of already-internalized enodes' | 16 | 117 | 2040
| June 4   - 'Cleaner fix for initializing and finalizing m_max...' (only on CanCebeci/z3)  | 7 | 56 | 2399 |
| June 29th (current) - 'Debug undo_add_eq, treat eqs as gen0' | 7 | 62 | 1924 |

### Clean-up 
* Fix `enode::get_eq_enode_with_min_gen()`
* Fix/remove case split queue with `CS_RELEVANCY_GOAL`
### Improvements
* The `cg` field in enode should be more or less useless. We should be able to remove it or replace with a boolean.
* Perf: `m_r1_parent_generations` can be a list instead of a map
* Perf: `undo_merge_cgc_generations` can skip updating enodes that are about to be garbage collected
* Perf: Sticky updates can be tagged with decision level. `apply_sticky_updates` only needs to consider the updates above the current level.


src/smt/smt_case_split_queue.cpp
```diff
@@ -692,440 +692,440 @@ namespace {
...
-            m_context(ctx),
-            m_params(p),
-            m_manager(ctx.get_manager()),
-            m_head(0),
-            m_bs_num_bool_vars(UINT_MAX),
-            m_priority_queue2(0, generation_lt(*this)),
-            m_current_goal(nullptr) {
-            set_global_generation();
-        }
+//     /**
```
NikolajBjorner: why is this commented out?
Is the configuration enabling this shadowed?
If it isn't ever used, good to delete it.

src/smt/smt_cg_table.cpp
```diff
@@ -170,7 +170,8 @@ namespace smt {
     void cg_table::display_binary(std::ostream& out, void* t) const {
         binary_table* tb = UNTAG(binary_table*, t);
         out << "b ";
-        for (enode* n : *tb) {
+        for (auto const& kv : *tb) {
```
NikolajBjorner: auto const& [n, v] : *tb) {
...

src/smt/smt_context.h
```diff
@@ -64,7 +65,9 @@ namespace smt {
 
     class model_generator;
     class context;
-    class kernel;
+
+    // Hash table for storing enode -> generation mappings
```
NikolajBjorner, line 69: what is the tradeoff between using an attribute on enode instead of separate tables?

CanCebeci: There are three `enode_generation_table`s. Two of these are supposed store a small number of enodes so I imagine you're referring to the tradeoff re: `m_constant_generations`.

The attribute would be unused for all enodes participating in congruence closure. Using `m_constant_generations` instead saves space (hopefully leading to better cache performance).

The tradeoff is that generations lookups on non-cgc enodes involve a table lookup, so potentially another memory access (if the constant generation table is not in cache).

My assumption was that we have a lot more cgc enodes than non-cgc enodes.

CanCebeci:
## To-do
### Clean-up
* ~Fix enode::get_eq_enode_with_min_gen()~
* ~Fix/remove case split queue with CS_RELEVANCY_GOAL~
### Improvements
* ~The cg field in enode should be more or less useless. We should be able to remove it or replace with a boolean.~ This is not so straighforward. Implemented in a separate [branch](https://github.com/CanCebeci/z3/tree/remove-cg).
* ~Perf: m_r1_parent_generations can be a list instead of a map~
* Perf: undo_merge_cgc_generations can skip updating enodes that are about to be garbage collected
* Perf: Sticky updates can be tagged with decision level. apply_sticky_updates only needs to consider the updates above the current level.

src/smt/smt_case_split_queue.cpp
CanCebeci: It is uncommented back in now. I'm not aware of anyone using it but I can't say confidently that it isn't ever used.

NikolajBjorner:
created branch "generation". Move PR into this branch and we can merge it there for now.

PR #10040, opened 2026-07-04
Fix inconsistent optimization result with unvalidated LP bound (#10028)

### LINKED ISSUE #10028: Z3 incorrect objective with inconsistent bounds
Whilst running Z3 on test instances for the upcoming XCSP3 2026 competition using CPMpy, we encountered an inconsistent returned objective (CPMpy/cpmpy#1036). For the `GolombRuler-a3-10_c26` instance, the objective converged to 20 (lb=ub=20) whilst minimising, but the decision variable that was posted as objective ended up getting value 500 assigned (matching the upper limit of its domain, not at all the reported value of 20). Even more strangely, it has been reported by the competition organisers (and by running with other solvers) that the actual optimal objective value is 55. 

Below is a smaller instances that reproduces the same behavior of the objective variable not falling within the reported bounds:

```python
import z3

n = 10
U = 500
x = [z3.Int(f"x{i}") for i in range(n)]

o = z3.Optimize()
for xi in x:
    o.add(xi >= 0, xi <= U)
o.add(x[0] == 0)
for i in range(n - 1):
    o.add(x[i] < x[i + 1])

ds = [x[j] - x[i] for i in range(n) for j in range(i + 1, n)]
o.add(z3.Distinct(ds))
h = o.minimize(x[n - 1])

print("check:", o.check())
print("lower/upper:", o.lower(h), o.upper(h))
mo = o.model()
print("x9 in model:", mo[x[n - 1]])
```
The output: 
```console
check: sat
lower/upper: 20 20
x9 in model: 500
```

levnach:
Fixes #10028.

## Problem

Minimizing an integer variable over a problem containing a large `distinct` constraint returned an **inconsistent** result: the reported optimum did not match the returned model, and it was not the true optimum.

Reproducer from the issue (a Golomb-ruler problem, true optimum = 55):

```python
import z3
n, U = 10, 500
x = [z3.Int(f"x{i}") for i in range(n)]
o = z3.Optimize()
for xi in x: o.add(xi >= 0, xi <= U)
o.add(x[0] == 0)
for i in range(n - 1): o.add(x[i] < x[i + 1])
o.add(z3.Distinct([x[j] - x[i] for i in range(n) for j in range(i + 1, n)]))
h = o.minimize(x[n - 1])
print(o.check(), o.lower(h), o.upper(h), o.model()[x[n - 1]])
# sat 20 20 500   <-- objective 20, but model has x9 = 500 (and 20 is unsat)
```

## Root cause

A `distinct` with more than 32 arguments is encoded with a fresh uninterpreted sort and function (`smt_internalizer.cpp`), so the objective variable becomes a *shared symbol* whose feasible values depend on EUF as well as arithmetic. The arithmetic relaxation therefore only produces a **hint** for the optimum, which may over-estimate it and be unachievable.

Two combined defects:

- `opt_solver::maximize_objective` committed the hint into `m_objective_values` **before** validating it with `check_bound`, and never rolled it back when validation failed. `update_objective` only ever *raises* the stored value, so the real (achievable) model value was discarded.
- `optsmt::geometric_lex` **ignored** the boolean return value and asserted the blocker derived from the unachievable hint, so the very next `check_sat` was UNSAT and the search terminated prematurely, reporting the bogus bound together with a non-matching model.

## Fix

- `opt_solver.cpp`: do not commit the hint before it is validated. On validation failure, `update_objective` now records the actual achievable model value. The no-model early-return keeps its previous behavior.
- `optsmt.cpp`: `geometric_lex` now honors the validation result. When the hint could not be validated, it discards the poisoned blocker and tightens from the real model value, so the search keeps converging toward the true optimum. When the hint is valid, the condition reduces to the original expression and behavior is unchanged.

After the fix the same reproducer produces consistent, monotonically-improving bounds (325 → 85 → … → 58 → … → 55), and the reported objective always matches the returned model.

## Testing

Exact-optimum, fast-terminating checks (all correct): EUF-forced minimum (= 5), `distinct(x, 0..32)` minimize (= 33), Golomb n=8 (= 34), plus basic min/max, real objective, box, lex, pareto, and weighted soft/maxsat.

Regression suites, rebuilt in **both Release and Debug**:

| Suite | Release | Debug |
|-------|---------|-------|
| `test-z3 /a` | 92 passed, 0 failed | 92 passed, 0 failed |
| z3test `regressions/smt2` (908 files, `model_validate=true`) | 0 failures | 0 failures |

### DIFF
--- a/src/opt/opt_solver.cpp
+++ b/src/opt/opt_solver.cpp
@@ -319,12 +319,28 @@ namespace opt {
             m_models.set(i, m_last_model.get());
 
         TRACE(opt, tout << "maximize " << i << " " << val << " " << m_objective_values[i] << " " << blocker << "\n";);
-        if (val > m_objective_values[i]) {
-            m_objective_values[i] = val;
-        }
+        //
+        // Do NOT commit 'val' to m_objective_values yet: 'val' is only an
+        // optimization hint from the arithmetic relaxation.  When the
+        // objective shares symbols with other theories (e.g. it occurs inside
+        // an uninterpreted function such as the auxiliary function used to
+        // encode large 'distinct' constraints) the hint can over-estimate the
+        // true optimum and may not be achievable by any model.  Committing it
+        // prematurely and then failing validation (check_bound below) would
+        // leave m_objective_values holding an unachievable bound that callers
+        // such as optsmt::geometric_lex report as the optimum, together with a
+        // model that does not attain it (issue #10028).  The value is only
+        // committed after it has been validated, or replaced by the value of
+        // an actual model in update_objective().
+        //
 
-        if (!m_last_model)
+        if (!m_last_model) {
+            // Without a model there is nothing to validate 'val' against; keep
+            // the previous behavior of adopting the (possibly infinite) hint.
+            if (val > m_objective_values[i])
+                m_objective_values[i] = val;
             return true;
+        }
 
         //
         // retrieve value of objective from current model and update 
--- a/src/opt/optsmt.cpp
+++ b/src/opt/optsmt.cpp
@@ -233,7 +233,7 @@ namespace opt {
                   if (is_sat == l_true) m_s->display(tout);
                   );
             if (is_sat == l_true) {                
-                m_s->maximize_objective(obj_index, bound);
+                bool bound_valid = m_s->maximize_objective(obj_index, bound);
                 m_s->get_model(m_model);
                 SASSERT(m_model);
                 inf_eps obj = m_s->saved_objective_value(obj_index);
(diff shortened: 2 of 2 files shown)


NikolajBjorner:
@levnach - thanks for checking on this.
I wasn't confident in the fix when I looked at it.
There is definitely a bug, and the code in opt_solver and the corresponding maximize code in lar_solver isn't great, 
but the way I recall the idea is the following:

If lar_solver can optimize an objective by only changing non-shared variables and maintaining int-satisfiability + nlsat satisfiabiltiy, then it is safe. 
All other cases, local optimization is not safe. 
The objectives tracked in opt_solver are presumably assumed safe.
The fix appears to not rely on this assumption in some case, but probably in other cases.

The idea is that variables mentioned in the distinct predicate are going to be shared because they are arguments of a function that lar_solver doesn't know about. The issue that was filed exposes a bug in tracking this information.

levnach:
@NikolajBjorner , I reverted the change 
[bug.smt2.txt](https://github.com/user-attachments/files/29718102/bug.smt2.txt)
gives a counterexample.

Issue #6158, opened 2022-07-13
[Consolidated] Segfaults


wintered:
[b81f70f](https://github.com/Z3Prover/z3/commit/b81f70f6fcc7850369c3131ebc11a4ffca9da393)
```
$z3release bug.smt2 
[1]    650136 segmentation fault  z3release bug.smt2
$cat bug.smt2 
(assert (forall ((a Bool) (b Bool) (d (_ BitVec 1)) (e (_ BitVec 1)) (f (_ BitVec 1)) (g Bool)) 
        (= (= f (ite a (_ bv0 1) (ite true (_ bv0 1) (ite b e e)))) g)))
(check-sat-using horn)
```

wintered:
```
$z3release bug.smt2
[1]    3178503 segmentation fault  z3release bug.smt2
$cat bug.smt2
(declare-fun a () Bool)  
(set-option :proof true)
(assert (and  (not (not (= 0 0)) ) (not (= 0 0)))) 
(check-sat-using qfufbv_ackr)
```

agurfinkel:
the first example is not `HORN`, please do not use `horn` tactic on it.

wintered:
>the first example is not HORN, please do not use horn tactic on it.
thanks, marked it by 👎 .

levnach:
z3 from the Nightly build crashes on the following input file(s).

**z3 Nightly build:** tag `Nightly`, commit [`241c6211d`](https://github.com/Z3Prover/z3/commit/241c6211d6e0da7dd4ef8904314aa3db0775db30) (`Z3 version 4.17.0`)

**Crashing input(s):**
- [`inputs/issues/iss-6158/bug.smt2`](https://github.com/Z3Prover/bench/blob/main/inputs/issues/iss-6158/bug.smt2)
- [`inputs/issues/iss-6158/bug-4.smt2`](https://github.com/Z3Prover/bench/blob/main/inputs/issues/iss-6158/bug-4.smt2)

z3 crashes when run on the file(s) above.

The crash is observed on Ubuntu 24.04.4 LTS (x86_64).

NikolajBjorner:
fixed

PR #10091, opened 2026-07-12
Fix segfault in horn on formulas with unused quantified variables (#6…

NikolajBjorner:
…158)

rule_manager::mk_query eliminates gaps in de Bruijn indices caused by unused quantified variables via a substitution that renumbers the remaining variables contiguously. This was done in a single pass, but var_subst applies the rewriter, which can simplify away further variable occurrences (e.g. collapsing ite terms such as (ite true a b) or (ite c x x)), introducing new gaps. The leftover null sort was then dereferenced, asserting in debug and segfaulting in release.

Iterate the gap-elimination until the free variables are contiguous. Each iteration either compacts the indices or strictly reduces the set of used variables, so it terminates.

Fixes the crash reported for:
  (assert (forall ((a Bool)(b Bool)(d (_ BitVec 1))(e (_ BitVec 1))(f (_ BitVec 1))(g Bool))
          (= (= f (ite a (_ bv0 1) (ite true (_ bv0 1) (ite b e e)))) g)))
  (check-sat-using horn)

### DIFF
--- a/src/muz/base/dl_rule.cpp
+++ b/src/muz/base/dl_rule.cpp
@@ -276,7 +276,11 @@ namespace datalog {
         // retrieve free variables.
         m_free_vars(q);
         vars.append(m_free_vars.size(), m_free_vars.data());
-        if (vars.contains(static_cast<sort*>(nullptr))) {
+        // Eliminate gaps in the variable indices produced by unused variables.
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
https://github.com/Z3Prover/z3/issues/6158

Issue #2261, opened 2019-04-29
Goal not in unsat core


GerhardSchellhorn:
Sorry for forgetting to add the goal in the last issue. The crash is gone but the problem, that the two axioms + the goal give an unsat-core which does not include the goal still persists.
Here is the file again, this time with the goal:

(set-option :smt.auto-config false) ; disable automatic self configuration
(set-option :smt.mbqi false) ; disable model-based quantifier instantiation
(set-option :smt.ematching true) ; enable ematching
(set-option :smt.macro-finder true)
(set-option :smt.pull_nested_quantifiers false)

;; setting auto-config to true (and removing the rest of the options above)
;; gives the same result

(set-option :unsat_core true)
(set-option :timeout 50)

;;;sorts
(declare-sort elem)

;;;lists
(define-sort list () (Seq elem))

;;;operations
;; membership test
(declare-fun inset (elem list) Bool)

;;;axioms

;; a in x <-> seq.contains(seq.unit(a),x)
(assert (!
  (forall ((a elem) (x list))
           (!
                     (= ; <->
           (inset a x)
           (seq.contains (seq.unit a) x)
        )
           :pattern (inset a x)
           ))
     :named elem-def))

;; x = seq.empty <-> not exists a. a in x
(assert (! (forall ((x list))
                            (= ; <->
           (= x (as seq.empty (Seq elem)))           
           (not (exists ((a elem)) (inset a x)))

        )) :named list--some-element))

;; a in cons(b, x) <-> a = b or a in x
(assert (! (not (forall ((a elem) (b elem) (x list))
        (= ; <->
           (inset a (seq.++ (seq.unit b) x))
           (or (= a b) (inset a x))
        ))) :named goal))

(check-sat)
;(get-unsat-core)

NikolajBjorner:
Note the semantics of "contains", is provided informally on https://rise4fun.com/Z3/tutorial/sequences, an in formaleze in SMT-LIB string proposals:
     
                 (seq.contains s sub) | Does s contain the sub-sequence sub?

Thus, when you write 

            (seq.contains (seq.unit a) x)

It means that (seq.unit a) contains x, not the other way around.

If you flip the order, the unsat core is as expected to contain both elem-def and goal, but not list--some-element.

GerhardSchellhorn:
Ah, I see, apologies for being too dumb to notice this.

PR #10138, opened 2026-07-16
[snapshot-regression-fix] Revert "distribute quantifiers over Booleans" (e-matching completeness regression)

levnach:
## Summary

Fixes a quantifier-reasoning regression detected by the `Z3Prover/bench` snapshot-regression corpus.

- **Originating discussion:** https://github.com/Z3Prover/bench/discussions/3198
- **Benchmark ref:** `iss-2261/bug-1.smt2` (from https://github.com/Z3Prover/z3/issues/2261)

## Divergence

The recorded oracle answers `unsat`, but current `master` answers `unknown`:

```diff
--- bug-1.expected.out (expected)
+++ produced (current z3)
@@ -1 +1 @@
-unsat
+unknown
```

This is **not** a timeout: both the old and the new binary terminate in ~0.01s. Old z3 (4.17.0) proves the goal `unsat`; current z3 returns `unknown`, i.e. it lost the ability to refute the formula. The benchmark uses `smt.mbqi=false` and `smt.ematching=true`, so the proof relies entirely on e-matching instantiation.

## Root cause

Bisection pinpoints commit `e2b9e3a6d` ("distribute quantifiers over Booleans"):

- parent `46b1c68f5` -> `unsat`
- `e2b9e3a6d` -> `unknown`

That commit added `th_rewriter_cfg::distribute_quantifier`, which rewrites a **pattern-less** quantifier over a top-level Boolean connective, e.g. `forall x. (A /\ B)  ->  (forall x. A) /\ (forall x. B)` (plus the De Morgan / exists variants). Splitting the body into separate sub-quantifiers forces z3 to re-infer patterns independently on each fragment. For this benchmark the fragmentation loses the multi-literal instantiation coverage the whole-body quantifier had, so e-matching saturates without deriving the contradiction and z3 gives up with `unknown`. The transformation is logically valid but is an e-matching **completeness** regression.

## Fix

Revert `e2b9e3a6d`, removing the `distribute_quantifier` rewrite and its call site in `reduce_quantifier` (`src/ast/rewriter/th_rewriter.cpp`, 62 deletions). The heuristic is brand-new and unreleased; reverting restores the previously-correct behaviour without affecting released functionality. If the finer-grained instantiation targets are desired, the transformation should be reintroduced behind a guard that preserves e-matching pattern coverage (e.g. only when adequate patterns can be inferred for every fragment).

## Validation

Rebuilt z3 from this branch (CMake/Ninja unavailable in the runner; built with `mk_make.py` + `make`) and re-ran the benchmark exactly as the snapshot capture does (empty `z3_opts`, `-T:20`):

- `z3 -T:20 inputs/issues/iss-2261/bug-1.smt2` -> `unsat` (matches oracle)
- 10 random seeds (`sat.random_seed=K smt.random_seed=K`, K=0..9) -> `unsat` in all 10

Before the revert the same rebuilt binary produced `unknown` for all seeds.

Created as a draft for human review.

> [!WARNING]
> <details>
> <summary>Firewall blocked 2 domains</summary>
>
> The following domains were blocked by the firewall during workflow execution:
>
> - `pypi.org`
> - `releaseassets.githubusercontent.com`
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "pypi.org"
>     - "releaseassets.githubusercontent.com"
> ```
>
>
> </details>

> Generated by [Fix a Z3 snapshot-regression divergence](https://github.com/Z3Prover/bench/actions/runs/29479613574) · 454.4 AIC · ⌖ 20.3 AIC · ⊞ 10.7K · [◷](https://github.com/search?q=repo%3AZ3Prover%2Fz3+%22gh-aw-workflow-id%3A+snapshot-regression-fixer%22&type=pullrequests)

### DIFF
--- a/src/ast/rewriter/th_rewriter.cpp
+++ b/src/ast/rewriter/th_rewriter.cpp
@@ -767,62 +767,6 @@ struct th_rewriter_cfg : public default_rewriter_cfg {
     }
 
 
-    // Distribute a quantifier over the top-level boolean connective of its body:
-    //   forall x . (A /\ B)   -->  (forall x. A) /\ (forall x. B)
-    //   forall x . ~(A \/ B)  -->  (forall x. ~A) /\ (forall x. ~B)
-    //   exists x . (A \/ B)   -->  (exists x. A) \/ (exists x. B)
-    //   exists x . ~(A /\ B)  -->  (exists x. ~A) \/ (exists x. ~B)
-    // Each rule is a validity, and the distributed sub-quantifiers give the
-    // e-matching engine finer-grained instantiation targets.
-    bool distribute_quantifier(quantifier * old_q, expr * new_body, expr_ref & result) {
-        if (old_q->get_kind() == lambda_k)
-            return false;
-        if (old_q->has_patterns())
-            return false;
-        bool is_forall = old_q->get_kind() == forall_k;
-        expr_ref_vector args(m());
-        expr * arg = nullptr;
-        if (is_forall) {
-            if (m().is_and(new_body))
-                args.append(to_app(new_body)->get_num_args(), to_app(new_body)->get_args());
-            else if (m().is_not(new_body, arg) && m().is_or(arg))
-                for (expr * e : *to_app(arg))
-                    args.push_back(m().mk_not(e));
-            else
-                return false;
-        }
-        else {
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
fixed

PR #10154, opened 2026-07-17
Fix macOS build failure blocking z3-5.0.0 NuGet packaging

levnach:
## Problem
The **NuGet Build** for the `z3-5.0.0` release failed (run [29549087366](https://github.com/Z3Prover/z3/actions/runs/29549087366)). Both macOS jobs (`build-macos-arm64`, `build-macos-x64`) failed to compile:

```
../src/util/symbol_table.h:111:18: error: no matching constructor for initialization of 'key_data'
        key_data dummy(key);
../src/util/symbol_table.h:130:22: error: no matching constructor for initialization of 'key_data'
2 errors generated.
make: *** [ast/ast.o] Error 1
FileNotFoundError: 'build-dist/z3'
```

Because macOS never produced `build-dist/z3`, NuGet packaging aborted and `package-nuget-x64` was skipped.

## Root cause
`key_data` is an aggregate (no user constructors). `find()` and `insert()` used C++20 **parenthesized** aggregate initialization `key_data dummy(key);` (P0960), which the Apple Clang on the macOS runners rejects. The same file's `contains()` already uses **brace** init `key_data{key}`, which compiles fine.

## Fix
Switch the two call sites to brace initialization `key_data dummy{key};`.

## Validation
Rebuilt the previously-failing translation unit `src/ast/CMakeFiles/ast.dir/ast.cpp.o` locally with CMake+Ninja — compiles cleanly.

### DIFF
--- a/.github/workflows/nuget-build.yml
+++ b/.github/workflows/nuget-build.yml
@@ -113,7 +113,9 @@ jobs:
           retention-days: 1
 
   build-macos-x64:
-    runs-on: macos-14
+    runs-on: macos-15
+    env:
+      MACOSX_DEPLOYMENT_TARGET: "13.0"
     steps:
       - name: Checkout code
         uses: actions/checkout@v7.0.0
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
@copilot - insetad of changing z3 source files, update nuget-build.yml to use the same macos configuration as release.yml

NikolajBjorner:
let us just remove this workflow. 
We have nuget builds in CI, nightly, release

PR #10157, opened 2026-07-17
Revert "Allow OTP input for WebAssembly npm publish workflow"

NikolajBjorner:
Reverts Z3Prover/z3#10156


bakkot:
The right way to do this is with [trusted publishing](https://docs.npmjs.com/trusted-publishers), incidentally. One set up it require neither a token nor an OTP.

PR #10169, opened 2026-07-20
Test PR for disabling semicolon warnings

davedets:

https://github.com/Z3Prover/z3/pull/10020 deleted a number of unnecessary (and, by a strict interpretation, illegal) semicolons.  These were detected by adding -Wextra-semi to CLANG_ONLY_WARNINGS during testing.  The PR did not eliminate all such unnecessary/illegal semi-colons; Nikolaj Bjorner argued that some IDEs would be confused by doing so for macro invocations, which otherwise look like function calls.  So it left those, and did not include adding -Wextra-semi to CLANG_ONLY_WARNINGS in the PR.

I'm worried that not having -Wextra-semi will allow instances of the semis we'd like to eliminate to creep back in.  This PR introduces a mechanism to disable the warnings for regions with such macro invocations, and uses that mechanism for one file.  If accepted, a subsequent PR would use this mechanism in all the remaining places, so that there is a clean clang build with -Wextra-semi.

I verified that:
* It compiles cleanly for if the enable/disable macros have null definitions, as they would for non-clang compilations.
* It compiles cleanly for a clang build without -Wextra-semis.
* It compiles successfully, albeit with *many* warnings, for a clang build with-Wextra-semis -- but none of those errors are for the regions in ast.h where the warning is disabled.


NikolajBjorner:
@copilot - what are cases where extra semi-colons are really harmful?

NikolajBjorner:
What bugs does -Wextra-semi really catch?

Mostly not solver bugs.
It catches redundant/extension-only semicolons at class or namespace scope.
In Z3’s case, it mainly catches:
semicolons after function definitions
semicolons after namespace { ... }
semicolons after extern "C" { ... }
semicolons after macros that expand to declarations/definitions
What that means in practice:

these are usually hygiene / standards-conformance / “clean build under strict warnings” issues
they can matter when Z3 is embedded in projects that treat warnings as errors
they do not usually indicate a runtime bug
Important distinction:

-Wextra-semi is not the warning that catches the classic “accidental empty statement in control flow” bug.
That is more in the territory of -Wextra-semi-stmt / empty-statement warnings.
Bottom line:

PR #10169 is about making -Wextra-semi usable without forcing removal of semicolons after macro invocations.
The warning mostly catches style/conformance/build-cleanliness problems, not deep semantic bugs.

NikolajBjorner:
so why?

davedets:
My overall goal is to get Z3, when included via CMakes FetchContents feature, to compile cleanly in the ClangTidy build.  (This isn't just an aesthetic concern -- the number of warnings printed is large enough that it significantly slows down the build.)

For each warning, we have the option of either fixing the warnings, or disabling them in the Z3 portion of the larger build (thanks to @Nuno Lopes for pointing out that the latter is an option).  I'm trying to pick the cases where I think fixing the source of the warning adds some value, rather than just doing a PR that disables them all.

In this case, my thought was that there's some value in having the code base be consistent with the usage of major style guides, and especially that it be internally consistent.  Prior to https://github.com/Z3Prover/z3/pull/10020, some files had semis at the end of namespaces and others didn't, and function decls in the same file differed.

I'd thought that acceptance of https://github.com/Z3Prover/z3/pull/10020 was an implicit endorsement of this goal.  In my experience, the best way to ensure future compliance with such conventions is to make a checker part of the build.  You can't add -Wextra-semi for the namespace/function-defn cases if the macro-invocation exceptions are still producing warnings.  Hence this PR, and the implied follow-up to complete the job and add -Wextra-semis to the clang compile command.

But maybe I'm just being OCD :-).  If you don't think this is worth it, well, your opinion matters a lot more than mine for this code base.  I would withdraw this PR, and submit a simpler one that just disables the warning.

For the more global problem, please let me know if you'd rather not see further PRs correcting the sources of further warnings, or if you'd like me to go on presenting them on a case-by -case basis.

Pleaes let me know how to proceed.  Thanks!

PR #10180, opened 2026-07-21
nla: add LP-based nonlinear bound optimization for cross-nested confl…

NikolajBjorner:
…icts

Add core::optimize_nl_bounds() (gated by arith.nl.optimize_bounds) which runs LP max/min over monomial leaf variables inside core::propagate(), analogous to solver=2's max_min_nl_vars, so nla (arith.solver=6) can detect cross-nested conflicts previously missed. Collect improved bounds first, then apply them and re-establish feasibility once; reconcile the core solver via find_feasible_solution before the raw maximize solves to preserve inf_heap_is_correct(). Skip null witnesses in get_dependencies_of_maximum for implied/unconditional bounds.

On FStar-UInt128-divergence solver=6 this yields unsat in 2 final-checks, seed-insensitive (seeds 1-10).

Copilot-Session: ac36bb84-de91-4e6c-86df-6008c7396ceb


NikolajBjorner:
@agurfinkel - I am re-organizing monomial_bounds.
Your two features are now really the only thing happening from the original bounds machinery.
propagate_shared_factor is a candidate to move to bounds tightening - it doesn't use model values.

What are distingishing benchmarks where these features help? It would be good to have them as part of harnessing changes.

Issue #10172, opened 2026-07-21
qe2 leaves a fresh undeclared constant (from the eliminated variable) in the quantifier-eliminated result


daniel-larraz:
### Summary

Applying the `qe2` tactic can leave a **fresh, undeclared constant** in the quantifier-eliminated result. The constant is derived from the eliminated bound variable (it is named `X!0`, from the bound variable `X`) and should not appear in the output at all. The classic `qe` tactic does not do this. The result is logically equivalent to the correct one, so this is not a soundness problem — but the fresh constant is undeclared, so asserting the result back into a solver fails with `unknown constant`.

### Version

Reproduces on **4.15.3**, **4.16.0**, and **5.0.0** (latest release), Linux x64.

### Reproducer

```smtlib
(declare-const x Bool)
(declare-const x2 Bool)
(assert (forall ((X Bool)) (or x2 (distinct 0 (ite (or x X) 0 1)))))
(apply qe2)
```

### Actual output (`qe2`)

```smtlib
(goals
(goal
  (or (not x) (and (not x) (not X!0)) x2)
  x2
  :precision precise :depth 1)
)
```

`X!0` is not declared anywhere; it is a fresh name derived from the eliminated bound variable `X`.

### `qe` does not leave a fresh constant

`(apply qe)` on the same input yields a result mentioning only `x` and `x2`.

### The result is equivalent (not a soundness issue)

Treating `X!0` as a free constant, the `qe2` goal is equivalent to the correct answer `x2` (`X!0` is subsumed):

```smtlib
(set-logic ALL)
(declare-const x Bool)
(declare-const x2 Bool)
(declare-const X!0 Bool)
(define-fun Q () Bool (and (or (not x) (and (not x) (not X!0)) x2) x2))
(assert (distinct Q x2))
(check-sat)
; => unsat
```

NikolajBjorner:
this was fixed

PR #10188, opened 2026-07-22
Test qe2 fresh-constant elimination

1sgtpepper:
## Summary

Add a tactic-level regression for the fresh-constant leak fixed by `caeaaa7d0`. The
test applies `qe2` to the original issue formula, verifies that the projected goal is
quantifier-free and contains only the declared free constants, and proves the result
equivalent to `x2`.

Register `mbp_qel` in the default `test-z3 -a` suite so normal unit-test CI runs the
regression.

Issue: #10172.

## Testing

- [CI](https://github.com/1sgtpepper/z3/actions/runs/31393765639)
- [OCaml Binding CI](https://github.com/1sgtpepper/z3/actions/runs/31393765551)

No local tests were run; validation is delegated to the fork CI workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,7 @@ jobs:
         run: |
           cd build
           ./test-z3 -a
+          ./test-z3 mbp_qel
           cd ..
 
       - name: Clone z3test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,6 @@ jobs:
         run: |
           cd build
           ./test-z3 -a
-          ./test-z3 mbp_qel
           cd ..
 
       - name: Clone z3test
--- a/src/test/main.cpp
+++ b/src/test/main.cpp
@@ -120,6 +120,7 @@
     X(theory_dl) \
     X(model_retrieval) \
     X(model_based_opt) \
+    X(mbp_qel) \
     X(factor_rewriter) \
     X(smt2print_parse) \
     X(substitution) \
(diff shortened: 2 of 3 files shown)


NikolajBjorner:
While this fixes something, the dependency on qel is elided. It could break something else?

1sgtpepper:
Makes sense... please give me some time to iterate. Will update if it is ready for another look.

src/qe/mbp/mbp_qel.cpp
```diff
@@ -186,8 +186,21 @@ class mbp_qel::impl {
...
+        // The term graph persists across qel calls, so a Boolean variable hidden by
+        // non_core can reappear during the final reconstruction. Keep it available
+        // for the model substitution below.
+        expr_sparse_mark remaining_vars, restored_vars;
+        for (auto v : vars)
+            remaining_vars.mark(v);
+        for (auto v : vars_before_non_core)
+            if (m.is_bool(v) && !remaining_vars.is_marked(v)) {
+                restored_vars.mark(v);
+                vars.push_back(v);
```
NikolajBjorner: this always adds all Boolean vars back?

1sgtpepper:
Since #10172 was fixed on `master`, I converted this PR into a regression test for the original issue.

1sgtpepper:
@agurfinkel, could you review this revision? thanks

NikolajBjorner:
> Run the focused mbp_qel target explicitly in CI because it is an extra test
and is not included in test-z3 -a.

we could add more tests to -a?

1sgtpepper:
Implemented exactly that in 79978b0: `mbp_qel` is now registered in `FOR_EACH_ALL_TEST`, so `test-z3 -a` runs it.

PR #10192, opened 2026-07-22
Use macros to disable semi-colon warnings for blocks of macros.

davedets:

This PR completes the job started by https://github.com/Z3Prover/z3/pull/10169.  It adds `-Wextra-semi` to the set of CLANG_ONLY_WARNINGS, and adds 
```
START_DISABLE_EXTRA_SEMI_WARNING;
...macro invocations with trailing semis...
END_DISABLE_WARNING;
```
around all the blocks of macro invocations that provoked warnings.

(Additionally, in realclosure.h, there was one block of macro invocations that did *not* follow the trailing-semi pattern; changed that to look like all the others).


src/util/manage_warnings.h
```diff
@@ -3,7 +3,7 @@ Copyright (c) 2006 Microsoft Corporation
 
 Module Name:
 
-    build_warnings.h
+    manage_warnings.h
```
NikolajBjorner, line 6: good, this slipped previous PR

PR #10195, opened 2026-07-23
Fix some lost Solver.solutions(t)

danielzgtg:
z3-rs correctly used Or. I accidentally forgot that enclosing function call, and had what resulted in And.

Test case:

```python
s = Solver()
x, y, z = Ints("x y z")
s.add(x >= 0, x <= 2, y >= 0, y <= 2, z >= 0, z <= 2, x + y + z == 2)
# I didn't test multivariable constraints last time fearing nondeterminism
# Sorting avoids that
print(sorted(map(lambda x: list(map(lambda x: x.as_long(), x)), s.solutions([x, y, z]))))
```

**Before**: `[[0, 1, 1], [2, 0, 0]]`
**After**: `[[0, 0, 2], [0, 1, 1], [0, 2, 0], [1, 0, 1], [1, 1, 0], [2, 0, 0]]`

Fixes: #8633

### DIFF
--- a/src/api/python/z3/z3.py
+++ b/src/api/python/z3/z3.py
@@ -8090,11 +8090,6 @@ class Solver(Z3PPObject):
         >>> s.add(x >= 0, y < 10, y == 2*x)
         >>> print(list(s.solutions([x, y])))
         [[0, 0], [1, 2], [2, 4], [3, 6], [4, 8]]
-        >>> s.reset()
-        >>> s.add(x >= 0, y < 10, y == 2*x)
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
remove doc sample. It is too brittle across internal changes (hash functions etc)

danielzgtg:
> remove doc sample. It is too brittle across internal changes

I thought I avoided those problems with my `sorted(map(...` dance, but oh well.

Btw, it was infuriating to run into this bug in the middle of the https://stackoverflow.com/beta/challenges/79982305/challenge-21-knights-sudoku contest ... only to realize it was me that introduced the bug.

PR #10200, opened 2026-07-23
Fix assertion failure with smt.ho_matching true (#10196)

### LINKED ISSUE #10196: Assertion failure with smt.ho_matching true
[queries-FStar.Seq.Properties-crash-min.smt2.txt](https://github.com/user-attachments/files/30292421/queries-FStar.Seq.Properties-crash-min.smt2.txt)

Sorry, it's a big file, but it crashes with an assertion failure quickly. 
I am using a source build of Z3 5.0.0 on Ubuntu 24 at commit 0f2bc0c36bb2ad0d556801f2013e48b7a566fbcd

levnach:
Fixes #10196.

## Problem

With `smt.ho_matching true`, a large F*-generated SMT2 file quickly hits `ASSERTION VIOLATION ... ilvl <= m_scope_lvl` in `smt::context::reinit_clauses` (`src/smt/smt_context.cpp`).

## Root cause

The higher-order matcher (`euf::ho_matcher`) is constructed with the solver's **shared** trail stack (`m_context->get_trail_stack()`, `smt_quantifier.cpp`). The solver maintains the invariant `trail.get_num_scopes() == m_scope_lvl`. The matcher pushes/pops scopes on this shared stack during `search()`, but on early-exit paths it leaked a scope:

- `search()` uses `while (m.inc()) { ... }`; the input's `rlimit` makes `m.inc()` return false mid-search, exiting the loop with work items still on the backtrack stack, whose pushed scopes were never popped (the drain only ran in the budget-exhausted branch).
- `refine_ho_match()` / `operator()` push a scope manually; an exception thrown from substitution (`var_subst`) or `search()` (e.g. resource/memory limit) unwound past the matching `pop_scope`.

A leaked scope shifts the trail's scope boundaries, desynchronizing the trail scope count from the solver scope level. The offset persists across later balanced push/pops and eventually causes `pop_scope_core` to undo the wrong trail entries, corrupting clause reinitialization and tripping the assertion.

## Fix (`src/ast/euf/ho_matcher.cpp`)

1. **`search()`**: drain the backtrack stack after the main loop on *all* exit paths (normal, budget exhausted, and `m.inc() == false`), so every scope pushed in `consume_work` is always popped.
2. **`refine_ho_match()` and `operator()`**: restore the shared trail to the entry level on every exit — including exceptions — via `try/catch` and `pop_scope(get_num_scopes() - base_scope)`.
3. Removed the debug-only `UNREACHABLE()` in the sort-mismatch branch; the surrounding comment documents that path as a legitimate "discard candidate" outcome, so aborting there was inconsistent with the intent.

## Verification

- Debug build (assertions enabled): the reproducer previously produced dozens of assertion violations → now **0**; the run completes cleanly with the same result as the non-ho-matching baseline.
- `./test-z3 ho_matcher` unit test passes (all matches, including the sort-mismatch case).

### DIFF
--- a/src/ast/euf/ho_matcher.cpp
+++ b/src/ast/euf/ho_matcher.cpp
@@ -102,13 +102,22 @@ namespace euf {
     }
 
     void ho_matcher::operator()(expr* pat, expr* t, unsigned num_bound, unsigned num_vars) {               
+        unsigned base_scope = m_trail.get_num_scopes();
         m_trail.push_scope();
-        m_subst.resize(0);
-        m_subst.resize(num_vars);
-        m_goals.reset();
-        m_goals.push(0, num_bound, pat, t); 
-        search();
-        m_trail.pop_scope(1);
+        try {
+            m_subst.resize(0);
+            m_subst.resize(num_vars);
+            m_goals.reset();
+            m_goals.push(0, num_bound, pat, t);
+            search();
+        }
+        catch (...) {
+            // Restore the shared trail to the level we entered with, even if the
+            // search was aborted by an exception (e.g. resource/memory limit).
+            m_trail.pop_scope(m_trail.get_num_scopes() - base_scope);
+            throw;
+        }
+        m_trail.pop_scope(m_trail.get_num_scopes() - base_scope);
     }
 
     void ho_matcher::search() {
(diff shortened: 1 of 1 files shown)


NikolajBjorner:
this is band-aid fixes hiding underlying bugs

PR #10216, opened 2026-07-24
Fix floating-point division normalization

### LINKED ISSUE #10175: Solution soundness issue in QF_FP
c9a4a59                                                                         
                                                                                
Goes back to 4.12.3 while #9953 goes back to 4.12.2. Triggers on recent Co-Pilot patch (f00cf75) for #9953                 
                                                                                
```                                                                          
$z3 model_validate=true bug.smt2                                                
sat                                                                             
(error "line 6 column 10: an invalid model was generated")                      
$z3-4.12.3 model_validate=true bug.smt2                                         
sat                                                                             
(error "line 6 column 10: an invalid model was generated")                      
$z3-4.12.2 model_validate=true bug.smt2                                         
unsat                                                                           
$cat bug.smt2                                                                                                          
(define-fun a () (_ FloatingPoint 2 6) (_ -oo 2 6))                             
(declare-fun b () (_ FloatingPoint 2 6))        
(define-fun c () (_ FloatingPoint 2 6) ((_ to_fp 2 6) RTZ -3.25))               
(assert  (and (fp.eq b ((_ to_fp 2 6) RTZ 0.0625)) 
(fp.eq a (fp.div RTZ c b))))
(check-sat)
```

1sgtpepper:
## Summary

Remove `mk_div`'s `ebits > sbits` rejection and preserve symbolic `fp.div`
normalization for both `ebits <= sbits` and `ebits > sbits`. When
`sbits > 2^ebits`, division uses an exact nonzero-significand leading-zero count and
sizes its signed intermediate exponent for
`[3 - 2^ebits - sbits, 2^ebits + sbits - 4]`. When `sbits <= 2^ebits`, it keeps
`unpack`'s existing normalization path.

At `sbits = 2`, the would-be upper quotient extract has high index 7 and low index 8;
`res_sig` already consumes every quotient bit, so `too_large` is false without
constructing that range. Above-range exponents use `round`'s existing infinity or
max-finite selection; below-range exponents retain the shift distance and discarded
bits needed to round subnormal results.

`round` still receives an `ebits + 2`-bit exponent and computes an
`ebits + 2`-bit `sigma`. It caps a negative `sigma` only when `sbits + 2` is
representable in that width. Otherwise `sbits + 2 >= 2^(ebits + 2)`, while
`-sigma <= 2^(ebits + 1)`, so the cap cannot bind. When `sigma` is wider than `sig_ext`'s
`2 * (sbits + 4)`-bit shift operand, `sigma_neg_capped` is at most `sbits + 2` and a
nonnegative `sigma` is at most `sig`'s leading-zero count. Both values are less than
`2 * (sbits + 4)`, so extracting that many low bits preserves them. An FP conversion
regression covers the cap case; separate multiplication, square-root, and conversion
regressions cover the wider-`sigma` branch.

[SMT-LIB FloatingPoint 2.7](https://smt-lib.org/theories-FloatingPoint.shtml) imposes no
ordering between `eb` and `sb`. Z3's AST layer separately caps `eb` at 63.

Fixes #10175.

Regression coverage: Z3Prover/z3test#63.

## Testing

- Fork CI at paired head `86cb28ab4`, merging Z3 master `43ce75ace` with source
  `784328c47` and pinning z3test `8594d141`:
  [full CI](https://github.com/1sgtpepper/z3/actions/runs/33310681302) passed 14/14 jobs and
  [standalone OCaml](https://github.com/1sgtpepper/z3/actions/runs/33310681324) passed 2/2 jobs.
- Mutation controls using the retained regression queries at their earlier filenames:
  [unpatched source](https://github.com/1sgtpepper/z3/actions/runs/32843604222),
  [`sbits = 2` quotient range](https://github.com/1sgtpepper/z3/actions/runs/32843607440),
  [division exponent decrement](https://github.com/1sgtpepper/z3/actions/runs/32843610840), and
  [`round` shift-count cap](https://github.com/1sgtpepper/z3/actions/runs/32843613738) each fail
  all eight jobs that run these regressions; the other six jobs pass.
- No local Z3 build was run; validation used fork CI.

### DIFF
--- a/src/ast/fpa/fpa2bv_converter.cpp
+++ b/src/ast/fpa/fpa2bv_converter.cpp
@@ -962,7 +962,6 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref & 
     const mpz & lz_modulus = m_mpf_manager.m_powers2(ebits);
     SASSERT(ebits <= sbits);
 
-    expr_ref a_sgn(m), a_sig(m), a_exp(m), a_lz(m), b_sgn(m), b_sig(m), b_exp(m), b_lz(m);
     bool needs_wide_lz = m_mpz_manager.lt(lz_modulus, mpz(sbits));
     unsigned exp_bits = ebits + 2;
     if (needs_wide_lz) {
@@ -976,8 +975,31 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref & 
         exp_bits = m_mpz_manager.log2(max_exp) + 2;
     }
     unsigned lz_bits = needs_wide_lz ? exp_bits : ebits;
-    unpack_with_lz_width(x, a_sgn, a_sig, a_exp, a_lz, true, lz_bits);
-    unpack_with_lz_width(y, b_sgn, b_sig, b_exp, b_lz, true, lz_bits);
+
+    expr_ref a_sgn(m), a_sig(m), a_exp(m), a_lz(m), b_sgn(m), b_sig(m), b_exp(m), b_lz(m);
+    unpack(x, a_sgn, a_sig, a_exp, a_lz, false);
+    unpack(y, b_sgn, b_sig, b_exp, b_lz, false);
+
+    // Division needs the exact leading-zero count in its exponent arithmetic.
+    // Keep this local: unpack's ebits-wide count can wrap when sbits > 2^ebits.
+    expr_ref zero_sig(m), a_sig_is_zero(m), b_sig_is_zero(m);
+    zero_sig = m_bv_util.mk_numeral(0, sbits);
+    m_simp.mk_eq(zero_sig, a_sig, a_sig_is_zero);
+    m_simp.mk_eq(zero_sig, b_sig, b_sig_is_zero);
+
+    expr_ref zero_lz(m), a_lz_raw(m), b_lz_raw(m);
+    zero_lz = m_bv_util.mk_numeral(0, lz_bits);
+    mk_leading_zeros(a_sig, lz_bits, a_lz_raw);
+    mk_leading_zeros(b_sig, lz_bits, b_lz_raw);
+    m_simp.mk_ite(a_sig_is_zero, zero_lz, a_lz_raw, a_lz);
+    m_simp.mk_ite(b_sig_is_zero, zero_lz, b_lz_raw, b_lz);
+
+    SASSERT(lz_bits <= sbits);
+    expr_ref a_shift(m), b_shift(m);
+    a_shift = m_bv_util.mk_zero_extend(sbits - lz_bits, a_lz);
+    b_shift = m_bv_util.mk_zero_extend(sbits - lz_bits, b_lz);
+    a_sig = m_bv_util.mk_bv_shl(a_sig, a_shift);
+    b_sig = m_bv_util.mk_bv_shl(b_sig, b_shift);
 
     unsigned extra_bits = sbits+2;
     expr_ref a_sig_ext(m), b_sig_ext(m);
@@ -995,14 +1017,8 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref & 
     res_sgn = m_bv_util.mk_bv_xor({a_sgn, b_sgn});
 
     expr_ref a_lz_ext(m), b_lz_ext(m);
-    if (needs_wide_lz) {
-        a_lz_ext = a_lz;
-        b_lz_ext = b_lz;
-    }
-    else {
-        a_lz_ext = m_bv_util.mk_zero_extend(exp_bits - ebits, a_lz);
-        b_lz_ext = m_bv_util.mk_zero_extend(exp_bits - ebits, b_lz);
-    }
+    a_lz_ext = m_bv_util.mk_zero_extend(exp_bits - lz_bits, a_lz);
+    b_lz_ext = m_bv_util.mk_zero_extend(exp_bits - lz_bits, b_lz);
 
     res_exp = m_bv_util.mk_bv_sub(
         m_bv_util.mk_bv_sub(a_exp_ext, a_lz_ext),
@@ -3838,15 +3854,7 @@ void fpa2bv_converter::mk_unbias(expr * e, expr_ref & result) {
 
 void fpa2bv_converter::unpack(expr * e, expr_ref & sgn, expr_ref & sig, expr_ref & exp, expr_ref & lz, bool normalize) {
     SASSERT(m_util.is_fp(e));
-    sort * srt = to_app(e)->get_decl()->get_range();
-    unpack_with_lz_width(e, sgn, sig, exp, lz, normalize, m_util.get_ebits(srt));
-}
-
-void fpa2bv_converter::unpack_with_lz_width(expr * e, expr_ref & sgn, expr_ref & sig, expr_ref & exp,
-                                            expr_ref & lz, bool normalize, unsigned lz_bits) {
-    SASSERT(m_util.is_fp(e));
     SASSERT(to_app(e)->get_num_args() == 3);
-    SASSERT(lz_bits > 0);
 
     sort * srt = to_app(e)->get_decl()->get_range();
     SASSERT(is_float(srt));
@@ -3877,8 +3885,8 @@ void fpa2bv_converter::unpack_with_lz_width(expr * e, expr_ref & sgn, expr_ref &
     mk_unbias(denormal_exp, denormal_exp);
     dbg_decouple("fpa2bv_unpack_denormal_exp", denormal_exp);
 
-    expr_ref zero_lz(m);
-    zero_lz = m_bv_util.mk_numeral(0, lz_bits);
+    expr_ref zero_e(m);
+    zero_e = m_bv_util.mk_numeral(0, ebits);
 
     if (normalize) {
         expr_ref is_sig_zero(m), zero_s(m);
@@ -3886,20 +3894,20 @@ void fpa2bv_converter::unpack_with_lz_width(expr * e, expr_ref & sgn, expr_ref &
         m_simp.mk_eq(zero_s, denormal_sig, is_sig_zero);
 
         expr_ref lz_d(m), norm_or_zero(m);
-        mk_leading_zeros(denormal_sig, lz_bits, lz_d);
+        mk_leading_zeros(denormal_sig, ebits, lz_d);
         norm_or_zero = m.mk_or(is_normal, is_sig_zero);
-        m_simp.mk_ite(norm_or_zero, zero_lz, lz_d, lz);
+        m_simp.mk_ite(norm_or_zero, zero_e, lz_d, lz);
         dbg_decouple("fpa2bv_unpack_lz", lz);
 
         expr_ref shift(m);
-        m_simp.mk_ite(is_sig_zero, zero_lz, lz, shift);
+        m_simp.mk_ite(is_sig_zero, zero_e, lz, shift);
         dbg_decouple("fpa2bv_unpack_shift", shift);
         SASSERT(is_well_sorted(m, is_sig_zero));
         SASSERT(is_well_sorted(m, shift));
-        SASSERT(m_bv_util.get_bv_size(shift) == lz_bits);
-        if (lz_bits <= sbits) {
+        SASSERT(m_bv_util.get_bv_size(shift) == ebits);
+        if (ebits <= sbits) {
             expr_ref q(m);
-            q = m_bv_util.mk_zero_extend(sbits-lz_bits, shift);
+            q = m_bv_util.mk_zero_extend(sbits-ebits, shift);
             denormal_sig = m_bv_util.mk_bv_shl(denormal_sig, q);
         }
         else {
@@ -3907,9 +3915,9 @@ void fpa2bv_converter::unpack_with_lz_width(expr * e, expr_ref & sgn, expr_ref &
             // would be zero anyways. So we can safely cut the shift variable down,
             // as long as we check the higher bits.
             expr_ref zero_ems(m), sh(m), is_sh_zero(m), sl(m), sbits_s(m), short_shift(m);
-            zero_ems = m_bv_util.mk_numeral(0, lz_bits - sbits);
+            zero_ems = m_bv_util.mk_numeral(0, ebits - sbits);
             sbits_s = m_bv_util.mk_numeral(sbits, sbits);
-            sh = m_bv_util.mk_extract(lz_bits-1, sbits, shift);
+            sh = m_bv_util.mk_extract(ebits-1, sbits, shift);
             m_simp.mk_eq(zero_ems, sh, is_sh_zero);
             short_shift = m_bv_util.mk_extract(sbits-1, 0, shift);
             m_simp.mk_ite(is_sh_zero, short_shift, sbits_s, sl);
@@ -3917,7 +3925,7 @@ void fpa2bv_converter::unpack_with_lz_width(expr * e, expr_ref & sgn, expr_ref &
         }
     }
     else
-        lz = zero_lz;
+        lz = zero_e;
 
     SASSERT(is_well_sorted(m, normal_sig));
     SASSERT(is_well_sorted(m, denormal_sig));
@@ -3932,12 +3940,10 @@ void fpa2bv_converter::unpack_with_lz_width(expr * e, expr_ref & sgn, expr_ref &
     SASSERT(is_well_sorted(m, sgn));
     SASSERT(is_well_sorted(m, sig));
     SASSERT(is_well_sorted(m, exp));
-    SASSERT(is_well_sorted(m, lz));
 
     SASSERT(m_bv_util.get_bv_size(sgn) == 1);
     SASSERT(m_bv_util.get_bv_size(sig) == sbits);
     SASSERT(m_bv_util.get_bv_size(exp) == ebits);
-    SASSERT(m_bv_util.get_bv_size(lz) == lz_bits);
 
     TRACE(fpa2bv_unpack, tout << "UNPACK SGN = " << mk_ismt2_pp(sgn, m) << std::endl; );
     TRACE(fpa2bv_unpack, tout << "UNPACK SIG = " << mk_ismt2_pp(sig, m) << std::endl; );
--- a/src/ast/fpa/fpa2bv_converter.cpp
+++ b/src/ast/fpa/fpa2bv_converter.cpp
@@ -1135,9 +1135,10 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref & 
         sig_ext = m_bv_util.mk_concat(res_sig, m_bv_util.mk_numeral(0, sig_size));
         shifted_sig = m_bv_util.mk_bv_lshr(sig_ext, underflow_shift_sized);
         unsigned sig_extract_low_bit = 2 * sig_size - (sbits + 2);
-        // The shift is at most sbits + 2^(ebits - 1) - 2. If it passes the
-        // zero padding, the leading one enters this discarded range, so
-        // discarded remains true.
+        // Finite operand bounds make underflow_shift at most
+        // sbits + 2^(ebits - 1) - 2. If it exceeds the sbits + 4 appended
+        // zero bits, res_sig's normalized leading one still lands in the range
+        // reduced into discarded, so discarded remains true.
         discarded = m.mk_app(
             m_bv_util.get_fid(), OP_BREDOR,
             m_bv_util.mk_extract(sig_extract_low_bit - 1, 0, shifted_sig));
@@ -1156,8 +1157,10 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref & 
         underflow_retained = m_bv_util.mk_extract(sbits + 1, 2, underflow_sig);
         underflow_inc = mk_rounding_decision(
             rm, res_sgn, underflow_last, underflow_round, underflow_sticky);
-        // This path shifts by at least eight bits, so the retained significand
-        // has a leading zero and rounding cannot carry into minimum normal.
+        // exp_below_round_range implies
+        // underflow_shift >= 2 + 3*2^(ebits - 1), which is at least eight for
+        // ebits >= 2. The retained significand therefore has a leading zero,
+        // so rounding cannot carry into minimum normal.
         underflow_rounded_sig = m_bv_util.mk_bv_add(
             underflow_retained,
             m_bv_util.mk_zero_extend(sbits - 1, underflow_inc));
--- a/src/ast/fpa/fpa2bv_converter.cpp
+++ b/src/ast/fpa/fpa2bv_converter.cpp
@@ -4122,21 +4122,41 @@ void fpa2bv_converter::round(sort * s, expr_ref & rm, expr_ref & sgn, expr_ref &
     SASSERT(m_bv_util.get_bv_size(sigma) == ebits+2);
     unsigned sigma_size = ebits + 2;
 
-    expr_ref sigma_neg(m), sigma_cap(m), sigma_neg_capped(m), sigma_lt_zero(m), sig_ext(m),
-        rs_sig(m), ls_sig(m), big_sh_sig(m), sigma_le_cap(m);
+    expr_ref sigma_neg(m), sigma_neg_capped(m), sigma_lt_zero(m), sig_ext(m),
+        rs_sig(m), ls_sig(m), big_sh_sig(m);
     sigma_neg = m_bv_util.mk_bv_neg(sigma);
-    sigma_cap = m_bv_util.mk_numeral(sbits+2, sigma_size);
-    sigma_le_cap = m_bv_util.mk_ule(sigma_neg, sigma_cap);
-    m_simp.mk_ite(sigma_le_cap, sigma_neg, sigma_cap, sigma_neg_capped);
+    if (log2(sbits + 2) < sigma_size) {
+        expr_ref sigma_cap(m), sigma_le_cap(m);
+        sigma_cap = m_bv_util.mk_numeral(sbits + 2, sigma_size);
+        sigma_le_cap = m_bv_util.mk_ule(sigma_neg, sigma_cap);
+        m_simp.mk_ite(sigma_le_cap, sigma_neg, sigma_cap, sigma_neg_capped);
+        dbg_decouple("fpa2bv_rnd_sigma_cap", sigma_cap);
+    }
+    else {
+        // An unrepresentable cap is at least 2^sigma_size, while negating a
+        // negative sigma yields at most 2^(sigma_size-1), so the cap cannot bind.
+        sigma_neg_capped = sigma_neg;
+    }
     dbg_decouple("fpa2bv_rnd_sigma_neg", sigma_neg);
-    dbg_decouple("fpa2bv_rnd_sigma_cap", sigma_cap);
     dbg_decouple("fpa2bv_rnd_sigma_neg_capped", sigma_neg_capped);
     sigma_lt_zero = m_bv_util.mk_sle(sigma, m_bv_util.mk_numeral(rational(-1), sigma_size));
     dbg_decouple("fpa2bv_rnd_sigma_lt_zero", sigma_lt_zero);
 
     sig_ext = m_bv_util.mk_concat(sig, m_bv_util.mk_numeral(0, sig_size));
-    rs_sig = m_bv_util.mk_bv_lshr(sig_ext, m_bv_util.mk_zero_extend(2*sig_size - sigma_size, sigma_neg_capped));
-    ls_sig = m_bv_util.mk_bv_shl(sig_ext, m_bv_util.mk_zero_extend(2*sig_size - sigma_size, sigma));
+    unsigned shift_width = 2 * sig_size;
+    expr_ref rs_shift(m), ls_shift(m);
+    if (sigma_size <= shift_width) {
+        rs_shift = m_bv_util.mk_zero_extend(shift_width - sigma_size, sigma_neg_capped);
+        ls_shift = m_bv_util.mk_zero_extend(shift_width - sigma_size, sigma);
+    }
+    else {
+        // The right count is capped at sbits+2, and a nonnegative sigma is at
+        // most sig's leading-zero count. Both are below shift_width.
+        rs_shift = m_bv_util.mk_extract(shift_width - 1, 0, sigma_neg_capped);
+        ls_shift = m_bv_util.mk_extract(shift_width - 1, 0, sigma);
+    }
+    rs_sig = m_bv_util.mk_bv_lshr(sig_ext, rs_shift);
+    ls_sig = m_bv_util.mk_bv_shl(sig_ext, ls_shift);
     m_simp.mk_ite(sigma_lt_zero, rs_sig, ls_sig, big_sh_sig);
     SASSERT(m_bv_util.get_bv_size(big_sh_sig) == 2*sig_size);
 
--- a/src/ast/fpa/fpa2bv_converter.cpp
+++ b/src/ast/fpa/fpa2bv_converter.cpp
@@ -1054,9 +1054,9 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref & 
     sticky = m.mk_app(m_bv_util.get_fid(), OP_BREDOR, m_bv_util.mk_extract(extra_bits-2, 0, quotient));
     res_sig = m_bv_util.mk_concat(m_bv_util.mk_extract(extra_bits+sbits+1, extra_bits-1, quotient), sticky);
     if (sbits == 2) {
-        // At sbits == 2, the range [3*sbits+1:2*sbits+4] is [7:8], so there
-        // are no quotient bits above res_sig. Skip the invalid extract and
-        // OP_BREDOR, and use false for too_large.
+        // At sbits == 2, upper would require mk_extract(7, 8, quotient), whose
+        // high index is below its low index. res_sig already consumes every
+        // quotient bit, so too_large is false without constructing upper.
         too_large = m.mk_false();
     }
     else {
@@ -1110,13 +1110,13 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref & 
         round_min_exp_ext = m_bv_util.mk_sign_extend(exp_bits - round_exp_bits, round_min_exp);
         round_max_exp_ext = m_bv_util.mk_sign_extend(exp_bits - round_exp_bits, round_max_exp);
 
-        expr_ref exp_above_round_range(m), exp_in_round_range(m);
+        expr_ref exp_above_round_range(m), narrowed_exp(m);
         exp_below_round_range = m_bv_util.mk_slt(res_exp, round_min_exp_ext);
         exp_above_round_range = m_bv_util.mk_slt(round_max_exp_ext, res_exp);
-        exp_in_round_range = m_bv_util.mk_extract(round_exp_bits - 1, 0, res_exp);
-        // round_max_exp has leading bits 0,1,1, selecting round's existing
-        // overflow result for the current sign and rounding mode.
-        m_simp.mk_ite(exp_above_round_range, round_max_exp, exp_in_round_range, round_exp);
+        narrowed_exp = m_bv_util.mk_extract(round_exp_bits - 1, 0, res_exp);
+        // An above-range res_exp is already an overflow. round_max_exp selects
+        // round's existing infinity or max-finite result for the sign and mode.
+        m_simp.mk_ite(exp_above_round_range, round_max_exp, narrowed_exp, round_exp);
         m_simp.mk_ite(exp_below_round_range, round_min_exp, round_exp, round_exp);
 
         expr_ref min_exp(m), min_exp_ext(m), underflow_shift(m), underflow_shift_sized(m);
@@ -1135,7 +1135,7 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref & 
         sig_ext = m_bv_util.mk_concat(res_sig, m_bv_util.mk_numeral(0, sig_size));
         shifted_sig = m_bv_util.mk_bv_lshr(sig_ext, underflow_shift_sized);
         unsigned sig_extract_low_bit = 2 * sig_size - (sbits + 2);
-        // Finite operand bounds make underflow_shift at most
+        // res_exp >= 3 - 2^ebits - sbits makes underflow_shift at most
         // sbits + 2^(ebits - 1) - 2. If it exceeds the sbits + 4 appended
         // zero bits, res_sig's normalized leading one still lands in the range
         // reduced into discarded, so discarded remains true.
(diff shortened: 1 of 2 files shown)


src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -959,11 +959,18 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
     unsigned sbits = m_util.get_sbits(s);
     if (ebits > sbits)
         throw default_exception("division with ebits > sbits not supported");
+    const mpz & lz_modulus = m_mpf_manager.m_powers2(ebits);
+    // round() uses an ebits+2 signed exponent; normalized division fits iff
+    // sbits <= 2^ebits + 3.
+    if (sbits > 3 && m_mpz_manager.lt(lz_modulus, mpz(sbits - 3)))
+        throw default_exception("division with sbits > 2^ebits + 3 not supported");
```
wintersteiger: I think it's very unlikely that this is correct, and even if it is, we should fix the code below (and possibly the rounder) instead of throwing an exception. For the original problem, it's more likely that the bug is actually in `to_fp`.

1sgtpepper: I don’t think that’s right—the bug reproduces without `to_fp`, using explicit FP constructors. But I agree we should fix the exponent handling instead of throwing an exception.

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -959,11 +959,23 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
     unsigned sbits = m_util.get_sbits(s);
     if (ebits > sbits)
         throw default_exception("division with ebits > sbits not supported");
+    const mpz & lz_modulus = m_mpf_manager.m_powers2(ebits);
```
NikolajBjorner, line 967: this reference is invalid if m_powers2 is called with larger values while lz_modulus is in scope?

1sgtpepper: The reference remains valid: `powers2` separately allocates each cached `mpz` and stores its pointer, so growing the cache does not move the referenced object. In the final code, `lz_modulus` is last used before the two `unpack` calls.

NikolajBjorner:
How do we best validate this change (and changes like these)?
Should we declare victory for now if it fixes a reported bug and passes the CI and fixes the bug, 
or should it have self-contained unit tests or a form of an IEEE spec reference?

NikolajBjorner:
Regression tests mirroring these unit tests have been added (as `.smt2.disabled`) to [z3prover/z3test](https://github.com/Z3Prover/z3test) under `regressions/smt2/` (commit `0c4d428`):

<table>
<thead>
<tr><th>File</th><th>Test</th><th>Expected</th></tr>
</thead>
<tbody>
<tr><td><code>fp-div-norm-1</code></td><td>FP(2,6) bug reproducer: <code>-oo = -3.25/0.0625</code></td><td>unsat</td></tr>
<tr><td><code>fp-div-norm-2</code></td><td>FP(2,6) RTZ overflow clips to &minus;3.9375</td><td>sat</td></tr>
<tr><td><code>fp-div-norm-3</code></td><td>FP(2,6) exact <code>2.0/1.0</code></td><td>sat</td></tr>
<tr><td><code>fp-div-norm-4</code></td><td>FP(2,6) denormal <code>0.0625/0.0625</code></td><td>sat</td></tr>
<tr><td><code>fp-div-norm-5</code></td><td>FP(2,6) max-neg/self</td><td>sat</td></tr>
<tr><td><code>fp-div-norm-6</code></td><td>FP(2,5) basic</td><td>sat</td></tr>
<tr><td><code>fp-div-norm-7</code></td><td>FP(2,5) RNE overflow &rarr; +oo</td><td>sat</td></tr>
<tr><td><code>fp-div-norm-8</code></td><td>Float32 <code>6.0/2.0</code></td><td>sat</td></tr>
<tr><td><code>fp-div-norm-9</code></td><td>Float16 <code>1.0/2.0</code></td><td>sat</td></tr>
<tr><td><code>fp-div-norm-10</code></td><td>FP(3,8) boundary</td><td>sat</td></tr>
</tbody>
</table>

The disabled unit tests are pushed to z3test as SMT-LIB2 regressions; each has a matching `.expected.out` file. All outputs were verified against a fixed z3 build.

1sgtpepper:
@NikolajBjorner, sorry—I may have misunderstood the testing convention and added the regression in z3test#60. Should I also add a `src/test` test?

wintersteiger:
The changes in `unpack` are too extensive for me to review in detail in reasonable time, and that function is used in every single FP function, i.e. potentially breaks lots of stuff. At the very least, we need to run all of the SMT-LIB benchmarks through it. 

IMHO, those changes are not right: `unpack` does not get a special branch for division, anything that is specific to division needs to stay in `mk_div`. Also, I have a hunch that the general fix for `ebits > sbits` could be less complicated while also covering this special case, so perhaps we should go for that right away.

1sgtpepper:
Thanks. I may revert that part, but I’d like to first investigate whether a smaller `mk_div`-local fix can also handle the general `ebits > sbits` case.

wintersteiger:
> Thanks. I may revert that part, but I’d like to first investigate whether a smaller `mk_div`-local fix can also handle the general `ebits > sbits` case.

Absolutely, it would be awesome if we could close this gap!

NikolajBjorner:
> @NikolajBjorner, sorry—I may have misunderstood the testing convention and added the regression in z3test#60. Should I also add a `src/test` test?

you did the right thing. I didnt' see your PR until now. Merged it.

NikolajBjorner:
Can I merge this?

wintersteiger:
I think this is still being worked on; perhaps mark it as draft for now?

1sgtpepper:
Let me take a quick look..... last time I checked I thought it was good to go for review

1sgtpepper:
@wintersteiger, I think this is ready for another look now—could you re-review when you have a chance?

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -957,13 +957,48 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
...
+        // The lower endpoint is -(max_exp + 1), so one sign bit beyond
+        // max_exp's bit-length covers both endpoints.
+        exp_bits = m_mpz_manager.log2(max_exp) + 2;
+    }
+    unsigned lz_bits = needs_wide_lz ? exp_bits : ebits;
 
     expr_ref a_sgn(m), a_sig(m), a_exp(m), a_lz(m), b_sgn(m), b_sig(m), b_exp(m), b_lz(m);
-    unpack(x, a_sgn, a_sig, a_exp, a_lz, true);
-    unpack(y, b_sgn, b_sig, b_exp, b_lz, true);
+    unpack(x, a_sgn, a_sig, a_exp, a_lz, !needs_wide_lz);
```
wintersteiger: The last argument here is very confusing: why would we not want normalization in some but not all cases?

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -995,16 +1030,23 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
...
-    expr_ref sticky(m), upper(m), upper_reduced(m), too_large(m);
+    expr_ref sticky(m), too_large(m);
     sticky = m.mk_app(m_bv_util.get_fid(), OP_BREDOR, m_bv_util.mk_extract(extra_bits-2, 0, quotient));
     res_sig = m_bv_util.mk_concat(m_bv_util.mk_extract(extra_bits+sbits+1, extra_bits-1, quotient), sticky);
-    upper = m_bv_util.mk_extract(sbits + sbits + extra_bits-1, extra_bits+sbits+2, quotient);
-    upper_reduced = m.mk_app(m_bv_util.get_fid(), OP_BREDOR, upper.get());
-    too_large = m.mk_eq(upper_reduced, m_bv_util.mk_numeral(1, 1));
+    if (sbits == 2) {
+        // There are no quotient bits above the retained slice.
+        too_large = m.mk_false();
```
wintersteiger, line 1060: Was this a pre-existing bug, or should the branch condition be tighter?

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -1019,7 +1061,13 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
     shift_cond = m_bv_util.mk_ule(res_sig_lz, m_bv_util.mk_numeral(1, sbits + 4));
     expr_ref res_sig_shifted(m), res_exp_shifted(m);
     res_sig_shifted = m_bv_util.mk_bv_shl(res_sig, res_sig_shift_amount);
-    res_exp_shifted = m_bv_util.mk_bv_sub(res_exp, m_bv_util.mk_extract(ebits + 1, 0, res_sig_shift_amount));
+    expr_ref exp_shift_amount(m);
+    // The selected unsigned correction is at most sbits + 3.
+    if (exp_bits <= sbits + 4)
```
wintersteiger, line 1089: The condition here does not line up with the comment. I think +4 is correct, but the comment says +3. Also: what's an "unsigned correction"? We have no unsigned formats here and nothing needs to be "corrected".

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -4046,52 +4095,63 @@ void fpa2bv_converter::round(sort * s, expr_ref & rm, expr_ref & sgn, expr_ref &
...
     expr_ref ext_emax(m), t_sig(m);
-    ext_emax = m_bv_util.mk_zero_extend(2, e_max);
+    ext_emax = m_bv_util.mk_zero_extend(exp_bits - ebits, e_max);
     t_sig = m_bv_util.mk_extract(sbits+3, sbits+3, sig);
     m_simp.mk_eq(ext_emax, exp, e_eq_emax);
     m_simp.mk_eq(t_sig, one_1, sigm1);
     m_simp.mk_and(e_eq_emax, sigm1, e_eq_emax_and_sigm1);
-    m_simp.mk_or(e_top_three, e_eq_emax_and_sigm1, OVF1);
+    // Preserve the existing formula shape for legacy-width callers; the signed
+    // comparison below is needed only when division widens the exponent.
```
wintersteiger: Division does not "widen". The result is still the same width as the inputs.

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -4037,7 +4085,8 @@ void fpa2bv_converter::round(sort * s, expr_ref & rm, expr_ref & sgn, expr_ref &
     SASSERT(m_bv_util.is_bv(exp) && m_bv_util.get_bv_size(exp) >= 4);
 
     SASSERT(m_bv_util.get_bv_size(sig) == sbits+4);
-    SASSERT(m_bv_util.get_bv_size(exp) == ebits+2);
+    unsigned exp_bits = m_bv_util.get_bv_size(exp);
+    SASSERT(exp_bits >= ebits + 2);
```
wintersteiger: `round` does not need special cases for division, it always rounds the same way for all operations. If the exponent is wider than required, we need to shorten it before handing it to `round` (like all the other functions do). Rounding decisions never depend on these extra bits. If necessary, that can indeed mean mapping to infinity before rounding (or not rounding at all in such cases), or computing a sticky bit (which would surprise me for an exponent, but it's possible that it's needed here).

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -4131,8 +4191,19 @@ void fpa2bv_converter::round(sort * s, expr_ref & rm, expr_ref & sgn, expr_ref &
...
     sig_ext = m_bv_util.mk_concat(sig, m_bv_util.mk_numeral(0, sig_size));
-    rs_sig = m_bv_util.mk_bv_lshr(sig_ext, m_bv_util.mk_zero_extend(2*sig_size - sigma_size, sigma_neg_capped));
-    ls_sig = m_bv_util.mk_bv_shl(sig_ext, m_bv_util.mk_zero_extend(2*sig_size - sigma_size, sigma));
+    expr_ref rs_shift(m), ls_shift(m);
+    if (sigma_size <= 2 * sig_size) {
+        rs_shift = m_bv_util.mk_zero_extend(2*sig_size - sigma_size, sigma_neg_capped);
+        ls_shift = m_bv_util.mk_zero_extend(2*sig_size - sigma_size, sigma);
+    }
+    else {
+        // The selected right count is capped at sbits + 2; the selected
```
wintersteiger: What's a "selected right count"? Unclear why we can assume `sigma_size == 1` here

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -957,13 +957,48 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
     // else comes the actual division.
     unsigned ebits = m_util.get_ebits(s);
     unsigned sbits = m_util.get_sbits(s);
-    if (ebits > sbits)
-        throw default_exception("division with ebits > sbits not supported");
-    SASSERT(ebits <= sbits);
+    const mpz & lz_modulus = m_mpf_manager.m_powers2(ebits);
```
wintersteiger, line 967: Are there any tests for `ebits > sbits`?

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -957,13 +957,48 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
...
 
     expr_ref a_sgn(m), a_sig(m), a_exp(m), a_lz(m), b_sgn(m), b_sig(m), b_exp(m), b_lz(m);
-    unpack(x, a_sgn, a_sig, a_exp, a_lz, true);
-    unpack(y, b_sgn, b_sig, b_exp, b_lz, true);
+    unpack(x, a_sgn, a_sig, a_exp, a_lz, !needs_wide_lz);
+    unpack(y, b_sgn, b_sig, b_exp, b_lz, !needs_wide_lz);
+
+    if (needs_wide_lz) {
+        // Division needs the exact leading-zero count in its exponent arithmetic.
+        // Keep this local: unpack's ebits-wide count can wrap when sbits > 2^ebits.
```
wintersteiger: Would this be fixed by a `+1` in unpack's call to `mk_leading_zeroes`? It's quite possible that it's just off by one in some cases and that we should fix it. (That's a one-line fix in unpack, but hundreds of fixes in all other functions; we can only do that if we're 100% sure).

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -957,13 +957,48 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
...
+        mk_leading_zeros(a_sig, lz_bits, a_lz_raw);
+        mk_leading_zeros(b_sig, lz_bits, b_lz_raw);
+        m_simp.mk_ite(a_sig_is_zero, zero_lz, a_lz_raw, a_lz);
+        m_simp.mk_ite(b_sig_is_zero, zero_lz, b_lz_raw, b_lz);
+
+        SASSERT(lz_bits <= sbits);
+        expr_ref a_shift(m), b_shift(m);
+        a_shift = m_bv_util.mk_zero_extend(sbits - lz_bits, a_lz);
+        b_shift = m_bv_util.mk_zero_extend(sbits - lz_bits, b_lz);
+        a_sig = m_bv_util.mk_bv_shl(a_sig, a_shift);
```
wintersteiger, line 1019: If division needs something relating to leading zeroes, why are the significands overwritten here? I would expect a single call to `mk_leading_zeroes` with a different number of `lz_bits`, nothing more. Are you saying that unpacking delivers incorrect significands in cases of `ebits > sbits`? (If it's just about an additional bit, just make the input to `unpack` wider and this would be a one-line fix.)

NikolajBjorner:
@wintersteiger - thanks for reviewing!

1sgtpepper:
Thanks for the review. Marking this draft for now while I think this through, will get back to you

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: Only formats with `sbits > 2^ebits` defer normalization: `unpack`’s count would wrap, so `mk_div` computes and applies it locally. Other formats keep the existing path.

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: The slice has width `sbits - 2`, so it is empty only at the valid minimum `sbits == 2`; `false` is its exact reduction. The minimum-precision regression covers this.

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: The shift is at most `sbits + 3`, but its vector is `sbits + 4` bits wide. I updated the comment and removed the unclear ‘unsigned correction’ wording.

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: Agreed; I removed that wording. Only the intermediate workspace widens, not the result sort.

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: Agreed. `round` keeps its fixed interface; `mk_div` narrows in-range exponents and handles overflow or underflow before calling it.

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: Removed. The cap now has the required width, `sigma` is widened before negation, and the bounded count is resized for the shift.

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: Yes. z3test#63 covers `ebits > sbits` across rounding modes and boundary cases; the supplemental check covers all 1,891 format pairs.

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: No: FP(2,16) can require a count of 15, so one extra bit is still insufficient. `unpack` remains unchanged and division computes the exact count locally.

src/ast/fpa/fpa2bv_converter.cpp
1sgtpepper: `unpack(false)` returns the raw significand, so `mk_div` must shift it and apply the same count to the exponent. Other formats keep the existing normalized path.

NikolajBjorner:
thanks for the update

NikolajBjorner:
@wintersteiger - appropriate to merge at this point?

NikolajBjorner:
Fairly indirect and a poor-man's one-shot prompt investigation.
But additional evidence that isn't entirely in the "trust me" mode for this PR with cross reference to IEEE spec and code.

```
(*
   FloatsIEEE754Division.fst

   Formal model of the width-sensitive fp.div transformation in Z3 PR #10216,
   source head 51b80b21ac59a35f474860ed2e63cc502098e1f1.

   IEEE 754-2019 connections
   -------------------------
   * 3.3 and 3.4: finite binary values and subnormal encodings.
   * 4.3.1-4.3.3: nearest and directed rounding attributes.
   * 5.4.1: division computes the infinitely precise quotient and rounds it
     once to the destination format.
   * 7.5: tiny inexact results use gradual underflow; discarded nonzero bits
     therefore remain relevant to rounding.

   Source connections (line numbers are at the PR head above)
   -----------------------------------------------------------
   * 960-1014: exact local leading-zero counts and operand normalization.
   * 1025-1038: widened signed exponent arithmetic.
   * 1073-1093: quotient normalization and matching exponent correction.
   * 1101-1175: local deep-underflow shift, sticky bit, and rounding.
   * 4108-4149: shared rounding-decision truth table.
   * 4265-4319: widened/capped shared-round shift counts.

   The module deliberately models mathematical integers rather than F* machine
   integers.  The lemmas below prove the range conditions under which the C++
   fixed-width bit-vector operations denote these mathematical operations.
   No axiom or assume is introduced by this module.
*)
module FloatsIEEE754Division

open FloatsIEEE754Spec

#push-options "--fuel 2 --ifuel 1 --z3rlimit 80"

(* ------------------------------------------------------------------------- *)
(* 1. Width predicates and exact leading-zero transport.                     *)
(* ------------------------------------------------------------------------- *)

let unsigned_fits (w:pos) (n:nat) : prop = n < pow2n w

let signed_fits (w:pos{w >= 2}) (n:int) : prop =
  -(pow2n (w - 1)) <= n /\ n < pow2n (w - 1)

(* mk_leading_zeros returns the exact count modulo 2^w.  A nonzero s-bit
   significand has at most s-1 leading zeroes, so s <= 2^w is precisely the
   condition that prevents wrapping.  This is the invariant used at
   fpa2bv_converter.cpp:962-984 and 1002-1007. *)
val leading_zero_width_is_exact (s w:pos) (lz:nat)
  : Lemma
      (requires lz < s /\ s <= pow2n w)
      (ensures unsigned_fits w lz)
let leading_zero_width_is_exact s w lz = ()

(* A dyadic pair (coefficient, exponent) denotes coefficient * 2^exponent.
   Equality is expressed without division or negative powers. *)
type dyadic = {
  coefficient : nat;
  exponent : int;
}

let dyadic_equal (a b:dyadic) : prop =
  if a.exponent <= b.exponent
  then a.coefficient == b.coefficient * pow2n (b.exponent - a.exponent)
  else b.coefficient == a.coefficient * pow2n (a.exponent - b.exponent)

let normalize (x:dyadic) (lz:nat) : dyadic =
  { coefficient = x.coefficient * pow2n lz;
    exponent = x.exponent - lz }

(* The source shifts the significand left by lz and subtracts lz from the
   exponent (lines 1011-1014 and 1036-1038).  This preserves the represented
   exact dyadic value. *)
val normalization_preserves_value (x:dyadic) (lz:nat)
  : Lemma (ensures dyadic_equal (normalize x lz) x)
let normalization_preserves_value x lz = ()

val pow2n_add (a b:nat)
  : Lemma (ensures pow2n (a + b) == pow2n a * pow2n b)
           (decreases a)
let rec pow2n_add a b =
  if a = 0 then ()
  else pow2n_add (a - 1) b

type exact_quotient = {
  numerator : nat;
  denominator : nat;
  scale : int;
}

let quotient_of (a b:dyadic) : exact_quotient =
  { numerator = a.coefficient;
    denominator = b.coefficient;
    scale = a.exponent - b.exponent }

(* Equality of (n/d)*2^scale is reduced to equality of two cross-multiplied
   dyadics, avoiding real division and any nonzero side condition. *)
let quotient_equal (a b:exact_quotient) : prop =
  dyadic_equal
    { coefficient = a.numerator * b.denominator; exponent = a.scale }
    { coefficient = b.numerator * a.denominator; exponent = b.scale }

(* This composes the two source transformations at lines 1011-1014 and
   1033-1038: applying independent exact leading-zero normalizations to the
   dividend and divisor leaves their infinitely precise quotient unchanged. *)
val operand_normalization_preserves_quotient
      (a b:dyadic) (a_lz b_lz:nat)
  : Lemma
      (ensures
        quotient_equal
          (quotient_of (normalize a a_lz) (normalize b b_lz))
          (quotient_of a b))
let operand_normalization_preserves_quotient a b a_lz b_lz =
  if a_lz <= b_lz
  then pow2n_add a_lz (b_lz - a_lz)
  else pow2n_add b_lz (a_lz - b_lz)

(* ------------------------------------------------------------------------- *)
(* 2. Exact exponent bounds and absence of bit-vector wrap.                  *)
(* ------------------------------------------------------------------------- *)

val pow2n_step (k:pos)
  : Lemma (ensures pow2n k == 2 * pow2n (k - 1))
let pow2n_step k = ()

let ieee_emin (ebits:pos{ebits >= 2}) : int =
  2 - pow2n (ebits - 1)

let ieee_emax (ebits:pos{ebits >= 2}) : int =
  pow2n (ebits - 1) - 1

let division_exp
      (a_exp b_exp:int) (a_lz b_lz:nat) (quotient_correction:nat)
    : int =
  (a_exp - a_lz) - (b_exp - b_lz) - quotient_correction

let division_exp_min (ebits sbits:pos) : int =
  3 - pow2n ebits - sbits

let division_exp_max (ebits sbits:pos) : int =
  pow2n ebits + sbits - 4

(* Unpacked finite operands have exponent in [emin, emax], each exact
   leading-zero count is in [0,sbits-1], and the normalized quotient needs
   either zero or one further left shift.  These facts imply the bounds stated
   by the source at lines 968-975. *)
val division_exponent_bounds
      (ebits:pos{ebits >= 2}) (sbits:pos{sbits >= 2})
      (a_exp b_exp:int) (a_lz b_lz quotient_correction:nat)
  : Lemma
      (requires
        ieee_emin ebits <= a_exp /\ a_exp <= ieee_emax ebits /\
        ieee_emin ebits <= b_exp /\ b_exp <= ieee_emax ebits /\
        a_lz < sbits /\ b_lz < sbits /\ quotient_correction <= 1)
      (ensures
        division_exp_min ebits sbits <=
          division_exp a_exp b_exp a_lz b_lz quotient_correction /\
        division_exp a_exp b_exp a_lz b_lz quotient_correction <=
          division_exp_max ebits sbits)
let division_exponent_bounds ebits sbits a_exp b_exp a_lz b_lz quotient_correction =
  pow2n_step ebits

(* C++ chooses one sign bit beyond floor(log2(max_exp)).  The equivalent
   arithmetic contract is max_exp < 2^(w-1).  Since the negative endpoint is
   -(max_exp+1), that single condition covers both endpoints. *)
val widened_exponent_does_not_wrap
      (ebits:pos{ebits >= 2}) (sbits:pos{sbits >= 2})
      (w:pos{w >= 2}) (result_exp:int)
  : Lemma
      (requires
        division_exp_min ebits sbits <= result_exp /\
        result_exp <= division_exp_max ebits sbits /\
        division_exp_max ebits sbits < pow2n (w - 1))
      (ensures signed_fits w result_exp)
let widened_exponent_does_not_wrap ebits sbits w result_exp = ()

(* The quotient correction at lines 1087-1091 is transported by either
   extraction or zero extension.  Its mathematical value is unchanged when it
   fits the destination width. *)
val quotient_shift_transport_is_exact (w:pos) (shift:nat)
  : Lemma
      (requires shift < pow2n w)
      (ensures unsigned_fits w shift)
let quotient_shift_transport_is_exact w shift = ()

(* ------------------------------------------------------------------------- *)
(* 3. Sticky preservation for the local deep-underflow path.                 *)
(* ------------------------------------------------------------------------- *)

let bit (b:bool) : nat = if b then 1 else 0

(* The ideal sticky bit after two consecutive discarded regions is the OR of
   both regions.  C++ can only inspect the second region after the physical
   right shift. *)
let ideal_sticky (first_discarded second_discarded:bool) : bool =
  first_discarded || second_discarded

let cpp_sticky (second_discarded:bool) : bool = second_discarded

(* Lines 1139-1144 establish the key invariant: either the shift remains in
   the appended zero padding (nothing was discarded by the first shift), or
   the normalized leading one remains in the low slice reduced by BREDOR
   (second_discarded is true).  Under exactly that invariant the local sticky
   bit equals the infinitely precise sticky bit. *)
val deep_underflow_sticky_is_exact
      (first_discarded second_discarded:bool)
  : Lemma
      (requires (not first_discarded) \/ second_discarded)
      (ensures cpp_sticky second_discarded ==
               ideal_sticky first_discarded second_discarded)
let deep_underflow_sticky_is_exact first_discarded second_discarded = ()

(* The source's stated numeric bounds imply that a deep-underflow shift remains
   inside the doubled significand.  This is what makes the second-region test
   above available rather than shifting the entire source away. *)
val deep_underflow_shift_stays_in_workspace
      (ebits:pos{ebits >= 2}) (sbits:pos{sbits > pow2n ebits})
      (shift:nat)
  : Lemma
      (requires shift <= sbits + pow2n (ebits - 1) - 2)
      (ensures shift < 2 * (sbits + 4))
let deep_underflow_shift_stays_in_workspace ebits sbits shift =
  pow2n_step ebits

(* res_sig is normalized with its leading one at bit sbits+2.  Appending
   sbits+4 zeroes puts that bit at 2*sbits+6.  If a shift passes the padding,
   the bounds at lines 1139-1141 prove that this one survives and lands below
   bit sbits+6, inside the slice reduced to sticky. *)
val normalized_one_remains_in_sticky_slice
      (ebits:pos{ebits >= 2}) (sbits:pos{sbits > pow2n ebits})
      (shift:nat)
  : Lemma
      (requires
        sbits + 4 < shift /\
        shift <= sbits + pow2n (ebits - 1) - 2)
      (ensures
        shift <= 2 * sbits + 6 /\
        2 * sbits + 6 - shift < sbits + 6)
let normalized_one_remains_in_sticky_slice ebits sbits shift =
  pow2n_step ebits

(* Being below the ebits+2 signed rounder range forces at least an eight-bit
   right shift.  Hence the retained significand has a leading zero and adding
   one for rounding cannot carry into the minimum normal encoding
   (lines 1160-1164). *)
val deep_underflow_has_no_normal_carry
      (ebits:pos{ebits >= 2}) (shift:nat)
  : Lemma
      (requires 2 + 3 * pow2n (ebits - 1) <= shift)
      (ensures 8 <= shift)
let deep_underflow_has_no_normal_carry ebits shift = ()

(* ------------------------------------------------------------------------- *)
(* 4. IEEE 754 rounding decision and source truth-table equivalence.         *)
(* ------------------------------------------------------------------------- *)

(* Boolean transcription of fpa2bv_converter.cpp:4130-4146.  sign=true means
   a negative result; last is the least retained bit; round is the half-ulp
   bit; sticky is the OR of all less-significant discarded bits. *)
let cpp_rounding_increment
      (rm:rmode) (sign last round sticky:bool)
    : bool =
  match rm with
  | RNE -> round && (last || sticky)
  | RNA -> round
  | RTP -> not sign && (round || sticky)
  | RTN -> sign && (round || sticky)
  | RTZ -> false

(* IEEE 754-2019 4.3, expressed as the increment decision on a nonnegative
   retained magnitude. *)
let ieee_rounding_increment
      (rm:rmode) (sign last round sticky:bool)
    : bool =
  let inexact = round || sticky in
  match rm with
  | RNE -> round && (sticky || last)       (* nearest, ties to even *)
  | RNA -> round                           (* nearest, ties away *)
  | RTP -> not sign && inexact             (* toward +infinity *)
  | RTN -> sign && inexact                 (* toward -infinity *)
  | RTZ -> false                           (* toward zero *)

val rounding_decision_matches_ieee
      (rm:rmode) (sign last round sticky:bool)
  : Lemma (ensures cpp_rounding_increment rm sign last round sticky ==
                   ieee_rounding_increment rm sign last round sticky)
let rounding_decision_matches_ieee rm sign last round sticky = ()

(* Replacing the ideal sticky by the locally computed sticky therefore leaves
   every IEEE rounding mode unchanged. *)
val deep_underflow_rounding_is_equivalent
      (rm:rmode) (sign last round:bool)
      (first_discarded second_discarded:bool)
  : Lemma
      (requires (not first_discarded) \/ second_discarded)
      (ensures
        cpp_rounding_increment rm sign last round
          (cpp_sticky second_discarded)
        ==
        ieee_rounding_increment rm sign last round
          (ideal_sticky first_discarded second_discarded))
let deep_underflow_rounding_is_equivalent
      rm sign last round first_discarded second_discarded =
  deep_underflow_sticky_is_exact first_discarded second_discarded;
  rounding_decision_matches_ieee
    rm sign last round (ideal_sticky first_discarded second_discarded)

(* ------------------------------------------------------------------------- *)
(* 5. Shared-round shift cap.                                                *)
(* ------------------------------------------------------------------------- *)

let bounded_right_shift (sigma:int) (cap:nat) : nat =
  if sigma < 0
  then let magnitude = -sigma in if magnitude <= cap then magnitude else cap
  else 0

(* Lines 4265-4305 widen sigma before negation and represent cap in a width
   where it cannot wrap.  At the mathematical level this is exactly
   min(-sigma, cap), and it never exceeds cap. *)
val bounded_round_shift_is_safe (sigma:int) (cap:nat)
  : Lemma (ensures bounded_right_shift sigma cap <= cap)
let bounded_round_shift_is_safe sigma cap = ()

(* Concrete regression at FP(2,16): the old four-bit cap encoded 18 as 2,
   while the widened comparison keeps the required shift 6. *)
val fp_2_16_cap_regression (u:unit)
  : Lemma
      (ensures
        18 % pow2n 4 == 2 /\
        bounded_right_shift (-6) 18 == 6 /\
        bounded_right_shift (-6) 18 < 2 * (16 + 4))
let fp_2_16_cap_regression () =
  pow2n_step 4;
  pow2n_step 3;
  pow2n_step 2;
  pow2n_step 1

#pop-options
```

wintersteiger:
I haven't had any time too look at this yet, but from a cursory glance I'll have to go through the details of the division changes and I'm still unhappy with the changes to `round`. There is nothing here requires a change of rounding behavior and if we do change the rounder, we will have to run at least all SMT-LIB benchmarks because rounding is used literally in every single operation.

@NikolajBjorner: careful with IEEE 754: this PR is specifically about very weird formats, e.g. tiny width, or ebits > sbits, for which 754-assumptions and properties don't necessarily apply.

src/ast/fpa/fpa2bv_converter.cpp
wintersteiger: Did this throw an exception on any tests before these changes were made?

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -4070,7 +4216,9 @@ void fpa2bv_converter::round(sort * s, expr_ref & rm, expr_ref & sgn, expr_ref &
     SASSERT(is_well_sorted(m, OVF1));
 
     expr_ref lz(m);
-    mk_leading_zeros(sig, ebits+2, lz); // CMW: is this always large enough?
+    // Keep round's established ebits+2 leading-zero workspace. Division
```
wintersteiger: Unnecessary comment change; no need to comment on what you're keeping.

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -4114,22 +4262,60 @@ void fpa2bv_converter::round(sort * s, expr_ref & rm, expr_ref & sgn, expr_ref &
     SASSERT(sig_size == sbits + 4);
     SASSERT(m_bv_util.get_bv_size(sigma) == ebits+2);
     unsigned sigma_size = ebits + 2;
-
-    expr_ref sigma_neg(m), sigma_cap(m), sigma_neg_capped(m), sigma_lt_zero(m), sig_ext(m),
+    // The exponent workspace is not necessarily wide enough to represent the
```
wintersteiger: What's a "workspace"?

The exponent is always wide enough by definition, because we can't change the output width and the additional two bits in the input are enough to make any rounding decision. Is there an example of an operation different from division that requires these changes?

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -4114,22 +4262,60 @@ void fpa2bv_converter::round(sort * s, expr_ref & rm, expr_ref & sgn, expr_ref &
...
-    m_simp.mk_ite(sigma_le_cap, sigma_neg, sigma_cap, sigma_neg_capped);
-    dbg_decouple("fpa2bv_rnd_sigma_neg", sigma_neg);
+    if (sigma_count_size == sigma_size)
+        sigma_neg_ext = m_bv_util.mk_bv_neg(sigma);
+    else {
+        // Widen sigma before negating it. Zero-extending the narrow negation
+        // would turn a positive sigma into a small unsigned count instead of
+        // preserving its negative two's-complement value for the comparison.
+        sigma_ext = m_bv_util.mk_sign_extend(sigma_count_size - sigma_size, sigma);
+        sigma_neg_ext = m_bv_util.mk_bv_neg(sigma_ext);
```
wintersteiger: Is this about the potential corner-case during negation of sigma? I think that was covered before, but it's of course possible that I'm wrong. If so however, that would be fixed by making sigma one bit wider instead of this complicated construction here. I would like to see a concrete example of this though, ideally without involving division.

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -1002,16 +1051,24 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
...
-    upper_reduced = m.mk_app(m_bv_util.get_fid(), OP_BREDOR, upper.get());
-    too_large = m.mk_eq(upper_reduced, m_bv_util.mk_numeral(1, 1));
+    if (sbits == 2) {
+        // The upper quotient slice has width sbits - 2, so it is empty here.
+        // The sort constructor guarantees that sbits is never below 2.
+        too_large = m.mk_false();
+    }
+    else {
+        expr_ref upper(m), upper_reduced(m);
+        upper = m_bv_util.mk_extract(sbits + sbits + extra_bits-1, extra_bits+sbits+2, quotient);
```
wintersteiger, line 1064: Sorry, but I still don't get it. What is "the slice"? Does `OP_BREDOR` produce something other than `false` on a zero-size bit-vector?

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -1026,14 +1083,94 @@ void fpa2bv_converter::mk_div(sort * s, expr_ref & rm, expr_ref & x, expr_ref &
...
+    res_exp_shifted = m_bv_util.mk_bv_sub(res_exp, exp_shift_amount);
     m_simp.mk_ite(shift_cond, res_sig, res_sig_shifted, res_sig);
     m_simp.mk_ite(shift_cond, res_exp, res_exp_shifted, res_exp);
 
     dbg_decouple("fpa2bv_div_res_sig", res_sig);
     dbg_decouple("fpa2bv_div_res_exp", res_exp);
 
-    round(s, rm, res_sgn, res_sig, res_exp, v9);
+    if (exp_bits > ebits + 2) {
+        // The rounder consumes an ebits+2 exponent. Keep ordinary exponents
```
wintersteiger: This is good - that's a safe branch and any new bugs will be localized here. I'm slightly suspicious because underflows are handled, but not overflows. Shouldn't this be symmetric?

wintersteiger:
There are a number of comments that I received by email, but they are not on this PR page. It took me over an hour to decipher them. Please use more precise language, e.g. what is the "shifted operand"? Which one is the "shared commit"? What is the "already bounded count"? I still don't know what "host-side" is supposed to mean.

But yes, you're right, the size of the bit-vector for sigma can be wrong and it throws an exception. I would have fixed this differently (e.g. by computing a `sigma_size` that's always correct first), but I believe the current fix is at least not unsound.

NikolajBjorner:
I was unaware of the ieee-p3109 project that has been ongoing. I wonder what (non-trivial) effort it would be to ingest it into certified code within z3. I poked at using reflection in F* to extract code, but it alone appears useless without understanding the spec.

NikolajBjorner:
so regarding this PR, I take it takes care of the identified bug and doesn't regress elsewhere.
THen, can I merge it?

wintersteiger:
Yes, I think it's safe enough to merge it. I was waiting for an answer on the sbits=2 issues, but that's a rare enough corner case that we can ignore for now.

P3109 is almost ready for prime time and some companies have started developing hardware. There are two formalizations, one in lean and one (by me) in Imandra modelling language. The latter one does provide semantics for all operators, but not all of them are exact (i.e. not "correctly rounded"). Basically, for everything involving irrational numbers like in square roots and trigonometry, we didn't bother with proofs of precision.

NikolajBjorner:
good material for a tokenmaxxer to replace hand-written code in fpa_rewriter and fpa2bv/bv2fpa converters :-).

1sgtpepper:
Sorry I didn't close the loop on these questions sooner. I had planned to
answer them after marking the PR ready for another review.

For `sbits == 2`, `res_sig` uses `quotient[7:3]` and `sticky` reduces
`quotient[2:0]`, so every quotient bit is already accounted for. The remaining
`upper` range would be `quotient[7:8]`, which is an invalid extract. This branch
therefore sets `too_large` to `false` without constructing `upper` or applying
`OP_BREDOR` to it.

Overflow and underflow require different handling because only underflow
depends on the exact distance outside `round`'s exponent range. An exponent
above the signed `ebits + 2` range is already an overflow. Clamping it to
`round_max_exp` still allows `round` to choose infinity or the maximum finite
value according to the sign and rounding mode. For an exponent below that
range, the exact distance determines the right shift and the round and sticky
bits, which can change the result between signed zero and the minimum
subnormal. The underflow path therefore preserves that distance and performs
the shift locally.

wintersteiger:
> I had planned to answer them after marking the PR ready for another review.

Sounds good, thanks for the additional info, that makes sense! If you have more changes, feel free to open a new PR for them, improvements are always welcome!

PR #10284, opened 2026-07-28
Make implicit switch case fall-throughs explicit

davedets:

This PR enable the "-Wimplicit-fallthrough" warning, then fix all the warnings this gets
in the clang build, by:
* Augmenting UNREACHABLE to add __builtin_unreachable(), which suppresses
warnings for fallthrough in that cse.
* Adding Z3_fallthrough in many cases, to make it clear that fallthroughs are intentional.
* Adding [[noreturn]] to functions that throw, so the compiler knows they don't fall through to the next case.
* In a couple of cases, there's a fall-through to a default case, which does "break", or "return nullptr".  In those cases, I duplicated the action in the preceding case, to make it more self-contained, and robust in the face of change.

In some cases, I am concerned about whether the warnings are identifying real bugs.  For example, the fall-throughs in these files seem at least a little suspect:
nnf.cpp
seq_rewriter.cpp
lar_solver.cpp 

while very probably correct, also seem at least a tiny bit suspect.  However, this PR does *not* attempt to change any behavior, only to silence the warnings.  It would be great if somebody with more knowledge of the code could vet these cases.  If vetted, the explicit presence of the Z3_fallthrough would reassure future readers of the code that the fall-through is intentional, not accidental.


NikolajBjorner:
fallthroughs in "suspect" places seem right.
There is a common pattern for recursive traversal of expressions to have an index of argument being processed and if recursive call isn't needed, use fallthrough to process next arguments. The case label is associated with the index.

davedets:
> There is a common pattern for recursive traversal of expressions to have an index of argument being processed and if recursive call isn't needed, use fallthrough to process next arguments. The case label is associated with the index.

Ah, kind of an un unrolled loop -- I see.  I was thinking it was trying to be a sort of "Duff's Device" idiom, and was confused that the indices were ascending rather than descending.  Thanks for checking, and for merging the PR.  I hope that the explicit presence of the Z3_fallthrough's will prevent future readers from wondering whether the fallthrough is intentional.

PR #10296, opened 2026-07-29
seq_monadic: self-contained monadic-decomposition regex membership solver (generic elements, witnesses, Boolean combinations)

veanes:
## Summary

Adds `seq_monadic` (`src/ast/rewriter/seq_monadic.{h,cpp}`), a self-contained,
rewriter-level decision procedure for regex membership of a term that is a
concatenation of sequence variables and constant elements — e.g. `x·a·x ∈ R`,
including repeated and multiple variables. It uses a whole-language *monadic
decomposition* plus automaton product-reachability; it is minterm-free and does
**not** use Nielsen word-equation splitting or `seq_split`.

The component is **purely additive**: it is not wired into any solver path, so
default behavior is unchanged. It ships with a unit test and an opt-in benchmark
harness that is inert unless `Z3_SEQ_BENCH_DIR` is set.

## What it does

- `x·u ∈ R  ⇔  ⋁_q ( x reaches q in A_R  ∧  u ∈ q )` over the derivative
  automaton; `reach(q)` is never materialized as a regex (avoids the
  state-elimination blowup). A variable's constraint is decided by a lazy
  product-reachability search over tuples of derivative states, with transitions
  = the product of `brz_derivative_cofactors` branches and pairwise-conjoined
  `seq::range_predicate` guards.
- **Generic in the element sort**: characters use the exact `range_predicate`
  algebra; any other element sort uses a candidate-basis over the element values
  the guards mention (sound and complete for the `{true,false,=,<=,and,or,not}`
  guard grammar the derivatives emit).
- **Concrete witnesses**: on sat it reconstructs a witness value (a sequence of
  concrete elements, not predicates) per variable from the accepting product path.
- **Boolean combinations**: `solve_and` decides a conjunction of memberships
  jointly, so a variable shared across memberships is constrained consistently.
  This is the natural extension since `¬(t∈R) ≡ t∈~R`, `∨` = union of DNFs, and
  `∧` = product of DNFs.

Also de-duplicates the char-guard → `range_predicate` translator into a single
public `seq::guard_to_range_predicate` in `seq_range_collapse` (it was previously
duplicated there and in `seq_monadic`).

## Testing

- `tst_seq_monadic`: single / multiple / repeated variables, nested complement,
  bounded loops, per-variable constraints, a generic `(Seq Int)` section, witness
  verification (substitute the model back and re-decide membership), and
  `solve_and` cases that are individually sat but jointly unsat.
- Full unit suite `test-z3 /a` passes (93/93).

## Evaluation (offline harness, not part of CI)

On a regex-membership benchmark corpus, restricted to files carrying a genuine
`(set-info :status)`, the solver decides 318 and 316 are correct (99.4%); the
only 2 disagreements are a length limitation (`|x|=2k`) that is outside the
membership fragment.

## Known limitations / follow-ups

- Pathological deeply-nested, high-multiplicity regexes can overflow the
  recursive derivative stack; a recursion-depth guard to degrade to `unknown` is
  a natural follow-up.
- Out-of-fragment constraints (word equations, length / Parikh) are not handled,
  by design — this decides regex membership only.


src/ast/rewriter/seq_monadic.cpp
```diff
@@ -0,0 +1,609 @@
...
+
+    // element values compared to v0 by the equalities in `g`.
+    void collect_consts(expr* g, ptr_vector<expr>& out) const {
+        expr* a = nullptr, * b = nullptr;
+        if (m.is_and(g) || m.is_or(g)) {
+            for (expr* arg : *to_app(g)) collect_consts(arg, out);
+            return;
+        }
+        if (m.is_not(g, a)) { collect_consts(a, out); return; }
+        if (m.is_eq(g, a, b)) {
```
NikolajBjorner, line 68: To be comprehensive: should check that m_v0 does not occur?

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -0,0 +1,609 @@
...
+    if (u().str.is_unit(t)) {                     // seq.unit of a constant element
+        expr* elem = to_app(t)->get_arg(0);
+        if (m.is_value(elem)) {
+            atoms.push_back(atom{ false, nullptr, elem });
+            return true;
+        }
+        return false;                             // symbolic (non-constant) unit: unsupported
+    }
+    // uninterpreted 0-ary constant of sequence sort => a sequence variable
+    if (is_app(t) && to_app(t)->get_num_args() == 0 &&
```
NikolajBjorner, line 419: is_uninterp_const

PR #10301, opened 2026-07-30
Add -V:level command-line option

rod-chapman:
This PR adds a new top-level command-line option "-V:level"

Same as -v:level but suppresses reporting of wall-clock timing and memory consumption data, and data that are the result of a floating-point division, so resulting logs are amenable to cross-platform diff.

The intent is that, following PRs 10163 and 10165, logs created by "-V:10" should be byte-for-byte identical on all host platforms, for example, on macOS vs Linux.

Some file committed here show changes in white-space, since my editor automatically replaced tab characters with spaces. Hope that's OK.

Running on our mlkem-native library, I now see identical behaviour and logs for all 191 SMT files between macOS and Linux. This is a huge UX improvement for our developers, where we develop and maintain proofs on macOS, but run CI on Linux machines.

### DIFF
--- a/src/sat/sat_asymm_branch.cpp
+++ b/src/sat/sat_asymm_branch.cpp
@@ -49,12 +49,12 @@ namespace sat {
         report(asymm_branch & a):
             m_asymm_branch(a),
             m_elim_literals(a.m_elim_literals),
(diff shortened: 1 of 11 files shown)


NikolajBjorner:
https://github.com/Z3Prover/z3/pull/10319

PR #10302, opened 2026-07-30
Choose array solver

juliusz-pham-huawei:
Adding a possiblity to choose solver for array theory independently from selected logic


NikolajBjorner:
Is there a need for this?
There is essentially only one array solver at the moment. 
The setup has to do with what functions are supported.

PR #10319, opened 2026-07-31
Add global suppress_platform_verbose parameter

NikolajBjorner:
## Summary

Adds a global configuration parameter **`suppress_platform_verbose`** (bool, default `false`).

When set to `true`, memory and time information is suppressed from verbose output — the same behavior as the `-V` flag proposed in #10301, but driven by a global parameter instead of a dedicated command-line switch. This yields deterministic, platform-independent verbose output that is useful for testing and diffing.

Usage:
```
z3 -v:2 suppress_platform_verbose=true problem.smt2
```

## Approach

Instead of introducing a `-V` command-line flag (#10301), this registers `suppress_platform_verbose` as a global parameter via `env_params`, wired to a lightweight `get_suppress_platform_verbose()` flag in `util`.

Verbose output sites guard only their **memory** (`mem_stat`) and **time** (`m_watch` / `:time` / `:before-memory` / `:after-memory`) fields on this flag. Platform-independent counters (e.g. `:cost`, `:threshold`, `:elim-vars`) are always emitted.

Affected sites:
- SAT: asymm-branch, probing, scc, simplifier (subsumer / blocked-clauses / resolution), solver stats (`sat.stats`, `:memory`, `mk_stat`)
- SMT: context stats (`smt.stats`)
- Tactic report (`:time`, `:before-memory`, `:after-memory`)

## Validation

- Release build succeeds.
- `suppress_platform_verbose` appears in `z3 -pd` with the correct type/default, and is accepted on the command line.
- All 93 `test-z3 /a` unit tests pass.

Closes the need for the `-V` switch in #10301.

### DIFF
--- a/src/sat/sat_asymm_branch.cpp
+++ b/src/sat/sat_asymm_branch.cpp
@@ -69,8 +69,11 @@ namespace sat {
                        if (num_units > 0)   verbose_stream() << " :units " << num_units;
                        if (tr > 0)          verbose_stream() << " :hte " << tr;
                        verbose_stream() << " :cost " << m_asymm_branch.m_counter;
-                       verbose_stream() << mem_stat();
-                       verbose_stream() << m_watch  << ")\n";);
+                       if (!get_suppress_platform_verbose()) {
+                           verbose_stream() << mem_stat();
+                           verbose_stream() << m_watch;
+                       }
+                       verbose_stream() << ")\n";);
         }
     };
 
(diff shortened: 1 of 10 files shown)


rod-chapman:
Thanks for this. Your solution is more elegant for sure.

rod-chapman:
I will re-run tests on the 191 files arising from mlkem-native on both macOS and Linux and confirm results.

rod-chapman:
I still see a small diff in two of my test cases. For example - macOS vs Linux:
```
<  (sat-cleaner :elim-literals 1134 :elim-clauses 6 :cost 1551596 :time 0.00)
---
>  (sat-cleaner :elim-literals 1134 :elim-clauses 6 :cost 1551596 :time 0.01)
```
I think that's generated on line 169 of sat/sat_cleaner.cpp

rod-chapman:
Could we get a tagged build (e.g. v5.0.1) that incorporates this PR plus 10163 and 10165 please? This would be enormously helpful for our developers working on both macOS and Linux.

rod-chapman:
I also see those diffs in the "sat-cleaner" entries when -v:2 is active as well as -v:10

PR #10323, opened 2026-07-31
seq_monadic: add light Antimirov cofactor mode

veanes:
Light-weight Antimirov cofactors for seq_monadic
================================================

Overview
--------

The seq_monadic solver explores symbolic regex derivatives when computing live
states and product transitions. Its original transition representation used
Brzozowski cofactors.

This change adds a light-weight Antimirov representation. It first computes the
existing Brzozowski cofactors and then decomposes targets whose outer shape is

    s1 | ... | sn

or

    (s1 | ... | sn) . tail

into separate transitions. Concatenations are maintained in right-associative
form, so a distributable union occurs as the head of the concatenation.
Targets are reconstructed with mk_regex_concat to preserve normalization.

After decomposition, transitions with the same target are merged by disjoining
their guards. The existing path-aware cofactor traversal and range predicates
are therefore retained.

Example
-------

For

    r = .* a .{k}

the Brzozowski cofactors have the shape

    [a,  r | .{k}]
    [^a, r]

The light-weight Antimirov transformation produces

    [., r]
    [a, .{k}]

This preserves the language while avoiding the deterministic subset states
that grow exponentially on this family.

Modes
-----

seq_monadic exposes two transition modes:

    light_antimirov   default
    brzozowski        retained as an explicit option

The implementation is seq_rewriter::light_ant_derivative_cofactors.

Correctness
-----------

The seq_monadic unit tests run in both modes. They cover character and generic
element sequences, multiple and repeated variables, variable constraints,
bounded loops, conjunctions of memberships, and witness construction. A
focused test checks the cofactor transformation above. The complete 94-test
unit suite used during evaluation passed. The benchmark harness is not
registered as a normal unit test; after detaching it, all 93 registered tests
pass.

Benchmark evaluation
--------------------

The final optimized comparison used all 1,545 SMT2 files under
C:\git\bench\inputs\regexes. Each file was run in a separate process with a
15-second timeout. There were 1,513 cases where both modes completed without a
process failure or timeout.

                                    Brzozowski       Light-Ant
    paired solver time                 94.98 s          63.26 s
    median solver time                  1.734 ms         0.710 ms
    derivative calls                    5.70 M           3.39 M
    cofactors                           13.33 M           6.99 M
    live states                          2.74 M           0.44 M
    product states                     652.8 K         642.8 K

Light-Ant reduced paired solver time by 33.4%, derivative calls by 40.5%,
cofactors by 47.5%, and live states by 84.0%. It was faster on 1,091 cases;
Brzozowski was faster on 421 cases.

Light-Ant changed 131 Brzozowski undef results to sat. There were no reverse
verdict changes and no mismatches against known sat/unsat statuses. Both modes
had two timeouts. Process failures decreased from 30 to 27.

By corpus, paired solver time improved by 39.1% on ClemensRegex and by 11.9% on
MargusRegex.

Alternatives considered
-----------------------

Direct use of full Antimirov derivatives was also evaluated. It introduced
large performance outliers, particularly around intersections, and produced
additional undef results and timeouts. Disabling intersection-over-union
distribution improved some of these cases but remained slower and less robust
than the light-weight transformation. The direct full-Ant mode was therefore
removed from this change.

### DIFF
--- a/src/ast/rewriter/seq_rewriter.cpp
+++ b/src/ast/rewriter/seq_rewriter.cpp
@@ -2931,6 +2931,66 @@ expr_ref seq_rewriter::mk_derivative(expr* ele, expr* r) {
     return result;
 }
 
+void seq_rewriter::light_ant_derivative_cofactors(expr* r, expr_ref_pair_vector& result) {
+    expr_ref_pair_vector brz(m());
+    m_derive.derivative_cofactors(r, brz);
+
+    obj_map<expr, unsigned> target_index;
+    expr_ref_vector guards(m());
+    expr_ref_vector targets(m());
+
+    auto add = [&](expr* guard, expr* target) {
+        unsigned index = 0;
+        if (target_index.find(target, index)) {
+            expr_ref merged(m());
+            m_br.mk_or(guards.get(index), guard, merged);
+            guards.set(index, merged);
+        }
+        else {
+            target_index.insert(target, targets.size());
+            targets.push_back(target);
+            guards.push_back(guard);
+        }
+    };
+
+    for (auto const& [guard, target] : brz) {
+        ptr_vector<expr> pending;
+        pending.push_back(target);
+        while (!pending.empty()) {
+            expr* t = pending.back();
+            pending.pop_back();
+            expr* left = nullptr, * right = nullptr;
+            if (re().is_union(t, left, right)) {
+                pending.push_back(right);
+                pending.push_back(left);
+            }
+            else if (re().is_concat(t, left, right) && re().is_union(left)) {
+                ptr_vector<expr> heads;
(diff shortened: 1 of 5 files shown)


src/ast/rewriter/seq_rewriter.cpp
```diff
@@ -2931,6 +2931,66 @@ expr_ref seq_rewriter::mk_derivative(expr* ele, expr* r) {
     return result;
 }
 
+void seq_rewriter::light_ant_derivative_cofactors(expr* r, expr_ref_pair_vector& result) {
```
NikolajBjorner, line 2934: should we just move these co-factor features into seq_derive to unbloat seq_rewriter.

PR #10332, opened 2026-07-31
Disable "-Wnoctad-maybe-unsupported", add and fix "-Wdeprecated-copy-with-user-provided-copy".

davedets:
This is another PR towards the goal of getting Z3 to compile cleanly when included via FetchContents into clang-tidy, which uses a pretty strict set of warnings.  (2 more flags after this!)

The first of these was -Wctad-maybe-unsupported.  That has to do with "class template argument deduction" -- the flag requires template deduction guides to be explicitly provided if templated types are used in situations that requires argument deduction.  This fired for various uses of templated types in the utils directory.

I decided that this should be a case of if it ain't broke, don't fix it, and explicitly disabled the warning in the Z3 build (which will override the setting if the flag is enabled in a larger build including Z3, like clang).

-----

The second flag has to do with the C++ "rule of 3".  Here is Google's AI summary (inf_s_integer is a class in Z3 that triggered the warning):

_This warning means your inf_s_integer class defines a custom copy assignment operator but lacks a user-defined copy constructor, which the C++ standard deprecates to encourage the "Rule of Three".  To fix this, explicitly declare and = default the copy constructor in your class definition._

_...example of how to fix..._

_This updates your code to modern C++ standards, cleanly silencing the warning._

This seemed like a good standard to follow, and didn't require too many changes, so I propose them.


NikolajBjorner:
It is more reliable to have a CI that does this check.
Towards that, the following agentic workflow: https://github.com/Z3Prover/z3/pull/10336

PR #10361, opened 2026-08-02
guard_set: cache guard -> range predicate translation

veanes:
`guard_set::conjoin` re-translated its guard expression into a `seq::range_predicate` on every call. It sits in the innermost loop of `seq_monadic::product_nonempty`'s product-transition enumeration, where the same handful of cofactor guards recurs on every branch of the product search — a single regex benchmark was measured at **4.4M translations of a few dozen distinct guards**.

## Change

Add an optional `guard -> range_predicate` cache to `guard_set`. It is owned by the caller so its lifetime can be tied to the lifetime of the guard expressions themselves; `seq_monadic` keys it off `m_cofactor_cache` (which owns the guards) and resets both together. A null cache entry records an unsupported guard, preserving the `m_ok = false` behaviour without retranslating.

The cache is used only on the character-sort path, and is valid only while every `guard_set` sharing it uses the same element variable `v0`. That holds in `seq_monadic`: `v0` is `m.mk_var(0, m_elem_sort)`, which is hash-consed, and the element sort is fixed for the duration of a solve. The parameter defaults to `nullptr`, so any other caller keeps the previous behaviour unchanged.

No behavioural change.

## Validation

`test-z3 seq_monadic`: ALL PASS (0 fail) in both `brz` and `light-ant` modes.

1545-file regex corpus (`light-ant`, 20s timeout per file):

| | master | this PR |
|---|---|---|
| sat / unsat / undef | 1166 / 255 / 121 | 1166 / 255 / 121 |
| timeouts | 3 | 2 |
| solve time, 1421 files decided by both | 74.7 s | **8.4 s (~9x)** |

Verdicts are identical to master on every file; one benchmark that previously hit the 20s timeout now terminates.

Full QF_S corpus (22,172 files) on the follow-up branch that builds on this one: 0 crashes, 0 timeouts, and on the 4,089 benchmarks where the monadic solver is complete and the expected status is known, 0 mismatches against the declared status.

### DIFF
--- a/src/ast/rewriter/guard_set.cpp
+++ b/src/ast/rewriter/guard_set.cpp
@@ -19,12 +19,19 @@ Author:
 #include "ast/arith_decl_plugin.h"
 #include "ast/bv_decl_plugin.h"
 
-guard_set::guard_set(ast_manager& _m, seq_util& _u, sort* elem_sort, expr* v0)
+guard_set::guard_set(ast_manager& _m, seq_util& _u, sort* elem_sort, expr* v0,
+                     rp_cache* cache)
     : m(_m), u(_u), m_sort(elem_sort), m_v0(v0),
-      m_is_char(_u.is_char(elem_sort)),
+      m_is_char(_u.is_char(elem_sort)), m_rp_cache(cache),
       m_rp(_u.max_char()), m_guard(_m) {
     if (m_is_char) m_rp = seq::range_predicate::top(u.max_char());
     else           m_guard = m.mk_true();
+}
+
+void guard_set::dealloc_cache(rp_cache& c) {
+    for (auto const& [k, v] : c)
+        dealloc(v);
+    c.reset();
 }
 
 void guard_set::collect_consts(expr* g, ptr_vector<expr>& out) const {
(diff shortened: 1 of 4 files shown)


src/ast/rewriter/guard_set.cpp
```diff
@@ -132,13 +139,30 @@ lbool guard_set::generic_eval(expr_ref* witness) const {
...
+    }
+    if (m_rp_cache) {
+        // Translating a guard expression into a range predicate is the inner loop of the
+        // product-transition enumeration; the same handful of guards recurs on every
+        // branch, so translate each one once.
+        seq::range_predicate* s = nullptr;
+        if (!m_rp_cache->find(g, s)) {
+            s = alloc(seq::range_predicate, u.max_char());
+            if (!seq::guard_to_range_predicate(u, m_v0, g, *s)) {
+                dealloc(s);
```
NikolajBjorner, line 154: we can recycle the last allocated range predicate for next insertion?

NikolajBjorner: it would require passing more than a hash-table, but a cache structure.

src/ast/rewriter/guard_set.h
```diff
@@ -31,12 +31,21 @@ Module Name:
 #include "ast/rewriter/seq_range_collapse.h"
 
 class guard_set {
+public:
```
NikolajBjorner, line 34: create class 

guard_set_cache {
   obj_map<expr, seq::range_predicate*> m_cache;
   seq::range_predicate* m_fresh = nullptr; // if non-null, then it can be used to insert., assume seq::range_predicate has reset method.
   void reset(); // clears memory.

  
   
public:
    guard_set_cache() {}
    ~guard_set_cache() { reset(); }
    seq::range_predicate* find(expr*);
    void insert(expr*, seq::range_predicate* p);
   // we could here assume expr* life-time is managed outside of guard_set_cache, otherwise include the usual trail and ast_manager& m in constructor.
};

PR #10366, opened 2026-08-02
seq_monadic: replace DNF expansion with depth-first search

NikolajBjorner:
Materializing the monadic decomposition as a DNF is the dominant cost in the
solver.  The product over per-position split degrees (and, for a conjunction of
memberships, over memberships) is exponential, and the `DNF_CAP` of 2^14
disjuncts meant the solver mostly gave up *structurally* rather than because the
problem was hard — it paid the full product cost before testing anything.

This decides the same disjunction by depth-first search, without ever
materializing it.

## How it works

`decide()` -> `prepare()` -> `dfs_membership(0)`.

`prepare()` parses each membership into atoms and records every variable's
**last occurrence** as a packed `(membership_idx << 32) | atom_idx`.  Positions
compare lexicographically in exactly the order the search visits them, so
"have I seen this variable for the last time?" is an integer equality.

`dfs_atoms(mi, i, R)` walks one membership:
- end of atoms: the remainder is epsilon, so the branch survives iff `R` is
  nullable (an undecidable nullability propagates as `l_undef` rather than being
  guessed);
- constant element: consumed by a derivative, `re.empty` prunes immediately;
- variable: the last atom is a plain membership, otherwise the variable drives
  the derivative automaton from `R` to some live state `q`, one child per target.

Components are accumulated per variable in `m_groups[vi]` along the current
branch — pushed on entry, popped on backtracking.  The per-variable emptiness
test runs as soon as the group is complete (the search just passed the
variable's last occurrence) or holds more than one component, which is the
earliest point an inconsistency can exist.  That test has to happen anyway;
running it there prunes the whole remaining subtree instead of after the product
has been built.  The search stops at the first satisfying leaf.

A variable shared by several memberships accumulates several components in the
same branch and they are intersected, so the joint solve falls out of the search
rather than needing the DNF multiplication.

## Supporting changes

- `group_nonempty` memoizes on the sorted, deduplicated `(state, target)`
  signature of the group, and collapses duplicated components before the product
  search (which is exponential in component count);
- memoize live split states per regex, and `der_elem` per `(regex, element)`;
- memoize nullability locally — `seq_rewriter`'s own cache is capped at 10000 and
  `cleanup()` flushes the entire table;
- hoist the cofactor vectors out of the product loop and reference them instead
  of re-materializing `expr_ref` pairs on every pop;
- `m_budget` becomes a global node budget, and `m_giveup` unwinds the whole
  search instead of letting sibling branches keep expanding.

Removed: `build_membership_dnf`, `decide_dnf`, `simplify_dnf`, the `disjunct`
type and `DNF_CAP`.  The persistent cofactor cache and the separate
`guard_set::cache` are kept as-is; both own their pins, so they survive the
per-`decide` `m_pin` reset.

## Results

Measured against master (4b2e69c66), same bench harness.

**regexes (1545 files), light-antimirov:** 41 benchmarks newly decided
(22 sat, 19 unsat), **no verdict regressions**, and 8.8s -> 2.2s on the 1421
files both versions decide.  Unchanged: 2 timeouts, 0 crashes, 0 mismatches
against declared status.  Brzozowski mode agrees: 0 mismatches, 0 disagreements
with light-antimirov.

**QF_S (22172 files):** 0 crashes, 0 timeouts, 0 mismatches on the 4089
benchmarks that are complete and have a declared status.  Verdicts are identical
to master.

Unit tests pass in both transition modes.

## Caveat

On the 81 files neither version decides, DFS costs more (62s -> 149s): it
explores until the 200000-node budget is exhausted, whereas the DNF path bailed
immediately at its structural 2^14 cap.  `m_budget` has not been retuned since
the per-node cost dropped, so there is likely room to recover most of that
without losing the 41 newly decided benchmarks.

### DIFF
--- a/src/ast/rewriter/seq_monadic.cpp
+++ b/src/ast/rewriter/seq_monadic.cpp
@@ -8,7 +8,9 @@ Module Name:
 Abstract:
 
     Whole-language monadic decomposition for regex membership.  See seq_monadic.h.
-    Automaton-based (product-reachability); reach(q) is never materialized as a regex.
+    Automaton-based (product-reachability); reach(q) is never materialized as a regex, and
+    the disjunction produced by the decomposition is never materialized as a DNF: it is
+    explored as a depth-first search tree with per-variable emptiness pruning.
 
     Generic in the element sort.  The decomposition, liveness and product-reachability
     are element-agnostic; only the *guard algebra* over the derivative cofactor guards
@@ -19,7 +21,6 @@ Abstract:
     yields the concrete element used to build a witness sequence.
 
 TODOs:
-- if perf suffers: use DFS backtracking search instead of DNF expansion (space overhead)
 - create a validation harness: expose certificates for correctness that can be checked.
 - consider using expr_ref as alternative to pinned expressions
 - revisit parse_term and "the_var" condition. A sequence of units should be allowed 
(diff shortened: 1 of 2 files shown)


src/ast/rewriter/seq_monadic.cpp
```diff
@@ -51,17 +52,37 @@ Module Name:
...
+    m_pin.push_back(r);                        // keep the cache keys and value alive
+    m_pin.push_back(elem);
+    m_pin.push_back(d2);
+    m_der_cache.insert(r, elem, d2);
     return d2;
 }
 
+lbool seq_monadic::nullable(expr* r) {
+    char v = 0;
+    if (m_nullable_cache.find(r, v))
```
NikolajBjorner, line 85: isn't this mostly redundant? 
We cache nullable on the info field which is the common case for ground regexes.

veanes: right, this is indeed redundant, missed that

veanes: Will remove this code.

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -51,17 +52,45 @@ Module Name:
...
+    expr* cached = nullptr;
+    if (m_der_cache.find(r, elem, cached))
+        return expr_ref(cached, m);
     expr_ref d = m_rw.mk_derivative(elem, r);   // mk_derivative(element, regex)
     // Normalize: for a general element sort the derivative by a non-matching constant can
     // leave a ground guard (e.g. (= 1 2)) unfolded; simplifying collapses such dead
     // branches to re.empty so nullability/emptiness stay decidable.
     expr_ref d2(m);
     m_thrw(d, d2);
+    m_pin.push_back(r);                        // keep the cache keys and value alive
```
NikolajBjorner, line 68: there is also a cache in seq_derive.
Does this not get hit?

PR #10376, opened 2026-08-04
nla: re-linearize violated monomials with fixed factors

levnach:
## Summary

- re-linearize violated monomials with at most one non-fixed factor on entry to nonlinear final checks
- restrict defining-row installation to currently violated monomials with wide fixed-factor products, avoiding tableau growth from branch-enumerated factors
- add `smt.arith.nl.linearize_violated_monomials`, enabled by default

## Evaluation

Compared the option enabled and disabled using `build/release/z3` over the 19 SMT2 files in `~/dev/z3test/regressions/fstar`:

| Setting | Total time | Results |
|---|---:|---|
| `true` | 4.881s | 69 `unsat`, 3 `unknown` |
| `false` | 5.510s | 69 `unsat`, 3 `unknown` |

There were no result mismatches or timeouts. Disabling the option increased total time by 12.9%, rlimit usage by 10.2%, arithmetic conflicts by 15.5%, NLA lemmas by 26.1%, and final checks by 13.0%.

Five-run medians for the largest movers:

| Benchmark | `true` | `false` | Winner |
|---|---:|---:|---|
| `FStar.UInt128-1.smt2` | 188ms | 532ms | `true` 2.83x |
| `queries-FStar.UInt128-10.smt2` | 613ms | 968ms | `true` 1.58x |
| `FStar-UInt128-1-reduced-core.smt2` | 229ms | 71ms | `false` 3.23x |


src/math/lp/nla_core.cpp
```diff
@@ -1545,7 +1545,10 @@ bool core::propagate() {
         propagated = true;
     if (m_monomial_bounds.tighten_lp_bounds())
 		propagated = true;
-    if (m_monomial_bounds.propagate_changed_bounds()) 
+    if (m_monomial_bounds.propagate_changed_bounds())
+        propagated = true;
+    if (params().arith_nl_linearize_violated_monomials() &&
```
NikolajBjorner: move parameter check inside of monomial bounds

levnach:
## Post-merge validation notes

Independent A/B evaluation of `smt.arith.nl.linearize_violated_monomials` (on vs. off) run against the rebased branch. **No soundness issues found — merge confirmed sound.** Two methodology corrections are worth recording for future work on this code.

### Soundness: clean

Zero `sat`/`unsat` disagreements across ~4,300 A/B runs. Full per-query result *sequences* were compared, not just the first result:

| Suite | Files | Mismatches |
|---|---:|---:|
| `z3test/regressions/{smt2,issues,Everest}` | 933 | 0 |
| QF_NIA sample (default portfolio) | 1200 | 0 |
| QF_NIA sample (forced SMT core) | 1200 | 0 |
| `z3test/regressions/{nl,fstar}` | 31 | 0 |

`test-z3 /a`: 94/94 pass, no new compiler warnings.

### The QF_NIA regression this PR guards against does not occur

The `bitsize <= 16` gate does its job. With the SMT core forced, rlimit is *bit-identical* on 95.5% of both-solved pairs; 18 instances better vs. 19 worse (>5%); totals within 0.6%. Solve counts are neutral under the default portfolio (875 on / 874 off).

An apparent 837-vs-844 solve-count gap at a 10s timeout was **timeout-boundary noise**: re-running all 27 discordant instances at 60s gave 24 solved in *both* modes, 1 only-on, 1 only-off.

### Correction 1: QF_NIA A/B numbers on this code path are not reproducible by default

`qfnia_tactic.cpp` dispatches through wall-clock branches:

```cpp
or_else(mk_qfnia_sat_solver(m, p),
        try_for(mk_qfnia_smt_solver(m, p), 2000),
        mk_qfnia_nlsat_solver(m, p),
        mk_qfnia_smt_solver(m, p))
```

Three *identical* invocations on `From_T2__Prim_4.t2__p7985_terminationG_0.smt2` produced rlimit 6,852,919 / 9,140,920 / 6,853,707 — the 2000ms `try_for` boundary is crossed non-deterministically, so a run can land in a completely different solver. Forcing `tactic.default_tactic=smt` gives 15,766,695 / 15,766,695 / 15,759,446 (stable).

**Any future A/B measurement of this option on QF_NIA needs `tactic.default_tactic=smt`**, otherwise the signal is dominated by tactic-dispatch noise rather than by the change under test.

### Correction 2: the F* speedups in the PR description are overstated by roughly 2x

Re-measured sequentially on an idle machine, 5 repeats per configuration, medians:

| Measurement | PR description | Re-measured |
|---|---:|---:|
| F* suite total | 12.9% | **7.7%** (5.054s on / 5.442s off) |
| `FStar.UInt128-1.smt2` | 2.83x | **1.21x** (0.214s / 0.260s) |
| `queries-FStar.UInt128-10.smt2` | 1.58x | **1.23x** (0.661s / 0.813s) |
| `FStar-UInt128-1-reduced-core.smt2` | 3.23x slower | **1.64x slower** (0.104s / 0.064s) |

The direction is right and the win is real — `FStar-UInt128-axiom.smt2` is a solid 2.15x (0.193s / 0.416s) — but the magnitudes in the description look like they predate the `bitsize` gate. The original figures were likely collected under parallel load; parallel runs in this evaluation inflated small-file deltas by up to 2.4x relative to sequential re-measurement.

### The design decision to *not* gate on scope level is correct

The comment block explains that rows installed at deep decision levels can perturb the search, so I tested the obvious alternative — skipping the row-installing path when `lra.get_scope_level() > 0`:

```cpp
if (fixed_to_zero == null_lpvar && w != null_lpvar) {
    rational k = fixed_var_product(m, w);
    if (k.is_int() && k.bitsize() <= 16)
        continue;
    if (c().lra.get_scope_level() > 0)   // <-- tested, do not do this
        continue;
}
```

This **erases every F\* win** (`FStar-UInt128-axiom` collapses 2.15x → 1.01x, the whole suite goes flat). The benefit genuinely comes from linearizing at non-zero decision levels, so the current formulation is the right one.

The tableau-growth mechanism the comment warns about is nonetheless real, just bounded — on `From_T2__s1.t2_fixed__p17087_safety_0.smt2` (a ~9x outlier): max-columns 2001 vs 912, max-rows 897 vs 362, `arith-nla-propagate-eq` 74203 vs 12487. Worth keeping in mind if the `bitsize` threshold is ever revisited.

### Minor follow-ups

- The `16` in `k.bitsize() <= 16` is a magic number; a named constant (or a parameter) would make the enumeration heuristic easier to tune.
- `fixed_var_product(m, w)` is computed twice per candidate — once for the gate, then again inside `linearize`.
- The gate only applies when `k.is_int()`, so a non-integer `k` reaches the row-installing path ungated. Probably intentional, but it is not what the comment describes.
- Stray double blank line added in `monomial_bounds.h`.

PR #10390, opened 2026-08-04
CI: collect F* test suite failure artifacts

### LINKED ISSUE #10375: F* test suite doesn't collect artifacts
The scheduled builds run the F* test suite on
https://github.com/Z3Prover/z3/actions/runs/30786942559

They have failures, but the action does not collect artifacts the way that the basic build failures would do.
F* has to be configured with options to ensure that smt2 files are created.
For the basic build, it is ensured by hard-wiring default values for scheduled runs.
Note that the parameters passed in by manual runs are not used in scheduled runs.

levnach:
Fixes #10375.

## Analysis of [run 30786942559](https://github.com/Z3Prover/z3/actions/runs/30786942559)

The job is green, but `make test` exited 2. Every failing test is an **expected-output mismatch**, and all 70 of them differ only by the `--proof_recovery` banner:

```
   - The SMT solver could not prove the query.
+  - This query was retried due to the --proof_recovery option, yet it still
+    failed on all attempts.
```

That text comes from `--proof_recovery` in the workflow's own `fstar_otherflags` default, which F*'s `.expected` files do not carry — so these are configuration mismatches, not Z3 regressions. There are also failures that never reach the solver (`hello.__all`, `dune`, extraction diffs on `RemoveUnusedTypars_B.fs` / `Bug3865.out`).

None of that was reconstructible from the artifact: the run uploaded 526 `.smt2` files, almost all logged failing queries from negative tests that are *supposed* to fail, and nothing else. No logs, no produced output, no `.expected` oracle, no diff.

## Changes

* **`Resolve FStar options`** — computes `OTHERFLAGS` once (removing the duplicated `--z3version` extraction) and enforces `--log_failing_queries`. Scheduled runs get no `workflow_dispatch` inputs, so the options that make F* emit `.smt2` files are hard-wired instead of assumed to come from the inputs. The effective flags are recorded in the artifact, the job summary and the discussion.
* **`Build FStar`** now tees to a log, as the test step already did.
* **`Collect FStar failure artifacts`** replaces `Collect generated SMT2 files` and produces:
  * `logs/` — build log, test log, versions, commit, effective flags, and a `failing-tests.txt` summary of mismatched outputs plus failed make targets;
  * `smt2/` — the logged failing queries, as before;
  * `test-output/` — for every expected-output test whose result differs from its oracle: `<name>.actual`, `<name>.expected` and a unified `<name>.diff`.
* **Upload always runs.** Previously, if no `.smt2` file existed the collect step short-circuited and the upload was skipped, so the hardest failures produced no artifact at all.
* The failure summary is surfaced in the **job summary** and in the **discussion**, so a mismatch is visible without downloading anything.

Applied to the run above, `test-output/` would hold the actual/expected/diff triple for each of the 70 mismatches and `failing-tests.txt` would list them alongside `hello.__all` and the other non-SMT failures.

## Validation

The workflow parses as YAML; every `run:` block passes `bash -n` and the `github-script` body passes `node --check`. The new steps were executed locally against a fixture reproducing the run's failure shapes:

* mismatched `.output` / `.json_output` / `.ideout` / `.fs` / `.out` files are collected with correct diffs; the generated diff for `Basic.fst.output` reproduces the annotation in the run exactly;
* matching outputs, `_output` files with no `.expected`, and files over 4 MB are correctly skipped;
* failed make targets are parsed from both logs;
* with no F* tree at all (build failed before clone) the collector still exits 0 and the artifact still contains the logs;
* option resolution was checked with the default flags, with `--log_failing_queries` absent, with empty flags, and with an unparseable `z3 --version`;
* the rendered discussion body is 53 040 characters in the worst case, below GitHub's 65 536 limit.

Behaviour deliberately unchanged: the build and test steps keep `continue-on-error`, so a broken F* master still does not block the report.

### DIFF
--- a/.github/workflows/fstar-master-build.yml
+++ b/.github/workflows/fstar-master-build.yml
@@ -90,6 +90,25 @@ jobs:
           ln -sf "$GITHUB_WORKSPACE/build/release/z3" /tmp/gh-aw/agent/z3-bin/z3-4.13.3
           /tmp/gh-aw/agent/z3-bin/z3 --version
 
+      # Scheduled runs receive no workflow_dispatch inputs, so the effective F* options
+      # come from the hard-wired env defaults above. Resolve them once here, and make sure
+      # the options that cause F* to emit .smt2 files are always present: without them the
+      # build and the test suite fail without leaving any query behind to analyse.
+      - name: Resolve FStar options
+        run: |
+          set -euo pipefail
+          Z3_VERSION="$(sed -E -n 's/^Z3 version ([0-9]+\.[0-9]+\.[0-9]+).*/\1/p' /tmp/gh-aw/agent/z3-version.txt | head -1)"
+          test -n "$Z3_VERSION" || { echo "Error: Failed to extract Z3 version from /tmp/gh-aw/agent/z3-version.txt (expected: 'Z3 version X.Y.Z')"; cat /tmp/gh-aw/agent/z3-version.txt || true; exit 1; }
+
+          OTHERFLAGS="--z3version $Z3_VERSION $FSTAR_OTHERFLAGS"
+          case " $OTHERFLAGS " in
+            *" --log_failing_queries "*) ;;
+            *) OTHERFLAGS="$OTHERFLAGS --log_failing_queries" ;;
+          esac
+
+          echo "FSTAR_EFFECTIVE_OTHERFLAGS=$OTHERFLAGS" >> "$GITHUB_ENV"
+          printf '%s\n' "$OTHERFLAGS" | tee /tmp/gh-aw/agent/fstar-otherflags.txt
+
       - name: Build FStar
         id: build_fstar
         continue-on-error: true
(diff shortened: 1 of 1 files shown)


levnach:
End-to-end validation run on this branch: https://github.com/Z3Prover/z3/actions/runs/30955996012

It exercises the new `Resolve FStar options` step, the broader `Collect FStar failure artifacts` step and the reworked discussion body against real F* master output.

levnach:
## Validation run finished ✅

[Run 30955996012](https://github.com/Z3Prover/z3/actions/runs/30955996012) completed with all steps green, including the four new/changed ones.

```
Collected 526 .smt2 file(s) and 100 mismatched test output(s).
```

The old workflow captured **0** of those 100 mismatches. Artifact `fstar-artifacts-30955996012` (3.97 MB) now contains:

| directory | files |
|---|---|
| `logs/` | 9 — build log, test log, versions, commit, effective flags, `failing-tests.txt`, `mismatched-outputs.txt` |
| `smt2/` | 526 — logged failing queries, as before |
| `test-output/` | 300 — `.actual` / `.expected` / `.diff` for each of the 100 mismatches |

Mismatches by area:

```
 58 tests/error-messages      19 pulse/test/error_messages
  6 tests/interfaces           6 pulse/test/bug-reports
  5 tests/ide/emacs            3 pulse/test
  2 pulse/test/nolib           1 tests/bug-reports/closed
```

30 of these are under `pulse/`, which never surfaced as annotations in the run cited in #10375.

The captured diff for `tests/error-messages/Basic.fst.output` reproduces the annotation from [run 30786942559](https://github.com/Z3Prover/z3/actions/runs/30786942559) byte for byte:

```diff
@@ -6,6 +6,8 @@
   - Expected failure:
   - Assertion failed
   - The SMT solver could not prove the query.
+  - This query was retried due to the --proof_recovery option, yet it still
+    failed on all attempts.
   - Env =
   - VC = Prims.l_False
```

`logs/fstar-otherflags.txt` confirms the enforced options reached both the build and the test suite:

```
--z3version 5.0.0 --split_queries on_failure --log_failing_queries --ext higher_order_smt --proof_recovery
```

The generated discussion is [#10395](https://github.com/Z3Prover/z3/discussions/10395); its body is 52 891 characters, below GitHub's 65 536 limit.

Note for triage: all 100 mismatches differ *only* by the `--proof_recovery` banner, which comes from this workflow's own `fstar_otherflags` default and is absent from F*'s `.expected` files. They are configuration mismatches, not Z3 regressions — which is now visible directly from the artifact and the discussion rather than requiring a log dig.

PR #10407, opened 2026-08-05
Add a noodler-bench workflow comparing z3 against z3-noodler

veanes:
## What

A manually dispatchable workflow that compares [VeriFIT/z3-noodler](https://github.com/VeriFIT/z3-noodler) against this repository's z3 on the regex corpus, in two configurations: the stock seq solver, and the monadic regex end-game solver (`smt.seq.regex_monadic=true`).

## Why it needs a token

The benchmarks live in the **private** `z3prover/bench` repository, so the default `GITHUB_TOKEN` cannot read them. The workflow mints an installation token for the z3prover CI app exactly as `nightly.yml` already does, and sparse-checks-out only `inputs/regexes` — about 2 MB of a ~1 GB repository.

## What it reports

The decided counts matter less than the two soundness checks the driver performs:

- any pair of solvers answering `sat` and `unsat` on the same benchmark, and
- any answer contradicting the benchmark's own `(set-info :status ...)` annotation.

Both are surfaced in the run summary. Neither fails the job — this is a measurement workflow, and a bug in another solver should not turn our CI red. Full per-benchmark timings are uploaded as a CSV artifact.

## Validation

A full run over all 1476 benchmarks at a 10 s timeout on a 4-vCPU runner ([run 30991409264](https://github.com/z3prover/z3/actions/runs/30991409264)):

| configuration | decided | % | total time (s) |
|---|---:|---:|---:|
| `noodler` | 1407 | 95.3% | 664 |
| `z3-seq` | 1163 | 78.8% | 3907 |
| `z3-monadic` | 1367 | 92.6% | 1131 |

| family | benchmarks | `noodler` | `z3-seq` | `z3-monadic` |
|---|---:|---:|---:|---:|
| ClemensRegex | 1178 | 1109 | 945 | 1071 |
| MargusRegex | 298 | 298 | 218 | 296 |

No `sat`/`unsat` disagreements, and no answer contradicting a `:status` annotation.

Incidental findings from that run, for the record: the monadic solver decides 10 benchmarks noodler does not (noodler segfaults on 11 benchmarks overall and times out on 58), and noodler decides 50 that the monadic solver does not, 26 of them in `ClemensRegex/Automatic/generated`.

### DIFF
--- a/.github/workflows/noodler-bench.yml
+++ b/.github/workflows/noodler-bench.yml
@@ -1,0 +1,121 @@
+name: noodler-bench
+
+# Compare VeriFIT/z3-noodler against this branch's z3 (stock seq solver, and the
+# monadic regex end-game solver) on the regex corpus in the private z3prover/bench
+# repository. Measurement only: the job stays green even if a solver disagrees,
+# but any sat/unsat disagreement is reported prominently in the run summary.
+
+on:
+  workflow_dispatch:
+    inputs:
+      timeout:
+        description: "per-benchmark timeout in seconds"
+        default: "10"
+      corpus:
+        description: "subdirectory of bench/inputs/regexes to run (empty = all)"
+        default: ""
+      limit:
+        description: "sample only N benchmarks, spread evenly (0 = all)"
+        default: "0"
+      noodler_tag:
+        description: "z3-noodler release tag"
+        default: "v1.6.1"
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
move this to the bench repository

NikolajBjorner:
added to bench

PR #10453, opened 2026-08-07
Round-robin arithmetic final checks

NikolajBjorner:
## Summary
- add a persistent round-robin final-check loop over LIA, shared equalities, and NLA in `theory_lra`
- add a persistent nonlinear strategy loop over bound propagation, Horner, Gröbner, and branching in `nla_core::check`
- move full nonlinear bound propagation from `theory_lra` into the NLA strategy cycle
- move nonlinear bound optimization from Horner into `nla_core::check`
- request another final-check round after patching or bound optimization changes the LP model
- leave free integer zero splits to LIA and guard Gröbner quotient propagation against non-integral intermediate values

## Performance
On `C:\z3test\regressions\fstar\FStar.UInt128-1.smt2` with a Release build:

| arithmetic solver | median |
|---|---:|
| solver 2 | 156 ms |
| solver 6 | 636 ms |

Solver 6 previously took roughly 156 seconds in the Debug baseline. Both solvers now complete the benchmark subsecond in Release and return `unsat`.

## Validation
- `nl17.smt2` and `10220.smt2` return `unsat`
- Debug and Release builds succeed
- 94/94 Debug unit tests pass
- 94/94 Release unit tests pass


src/math/lp/nla_core.cpp
```diff
@@ -1279,7 +1279,7 @@ void core::add_bounds() {
         lpvar i = m_to_refine[(k + r) % sz];
         auto const& m = m_emons[i];
         for (lpvar j : m.vars()) {
-            if (!var_is_free(j))
+            if (!var_is_free(j) || lra.var_is_int(j))
```
NikolajBjorner: what?

src/math/lp/nla_core.cpp
```diff
@@ -1303,20 +1303,25 @@ lbool core::check(unsigned level) {
...
     patch_monomials();
-    set_use_nra_model(false);    
     if (m_to_refine.empty())
-        return l_true;    
+        return l_false;
     init_search();
-    m_nla_satisfied = false;
+
+    m_monomial_bounds.optimize_nl_bounds();
+    if (m_to_refine.empty())
```
NikolajBjorner, line 1325: fragile? Values have been updated by optimize_nl_bounds.

src/math/lp/nla_core.cpp
```diff
@@ -1303,20 +1303,25 @@ lbool core::check(unsigned level) {
...
 
+    set_use_nra_model(false);
     init_to_refine();
+    if (m_to_refine.empty())
+        return l_true;
     patch_monomials();
-    set_use_nra_model(false);    
     if (m_to_refine.empty())
-        return l_true;    
+        return l_false;
```
NikolajBjorner, line 1316: what?

src/math/lp/nla_core.cpp
NikolajBjorner: solver=2 has this down to a stable approach.
This sketch needs revisions

src/math/lp/nla_grobner.cpp
```diff
@@ -469,9 +469,8 @@ namespace nla {
                                 << "lc: " << lc << " " << lc_value << "\n"
                                 << "r: " << r << " " << r_value << "\n";
             );
-            SASSERT(v_value.is_int());
-            SASSERT(r_value.is_int());
-            SASSERT(lc_value.is_int());
+            if (!v_value.is_int() || !r_value.is_int() || !lc_value.is_int())
```
NikolajBjorner: fragile

src/math/lp/nla_core.cpp
NikolajBjorner: Removed the integer exclusion. The round-robin now runs LIA before NLA whenever the tableau is not integer-feasible, so integer zero-splitting remains enabled without letting NLA inspect a fractional integer assignment.

src/math/lp/nla_core.cpp
NikolajBjorner: Made this contract explicit: patch_monomials now reports whether it changed the LP assignment. A successful patch requests another final-check round so LIA and the remaining theory/model state are revalidated.

src/math/lp/nla_core.cpp
NikolajBjorner: Revised. optimize_nl_bounds now reports that it may have changed the model; core::check recalibrates m_to_refine afterward. The optimizer asserts both tableau feasibility and integer-column integrality after every model-moving phase.

src/math/lp/nla_grobner.cpp
NikolajBjorner: Removed the guard and restored the integer assertions. The scheduler and optimize_nl_bounds now enforce the integral-model invariant before Grobner quotient propagation.

src/math/lp/nla_core.cpp
NikolajBjorner: The revised flow follows the solver=2 contract more closely: LIA establishes an integral assignment before NLA, model-changing patch/optimization requests continuation, and NLA is rerun if a later LIA pass repairs the assignment.

PR #10454, opened 2026-08-07
Complete work of https://github.com/Z3Prover/z3/pull/10445: eliminate default switch cases shown unnecessary by clang checks

davedets:
This is another -- almost the last! -- PR towards the goal of getting Z3 to compile cleanly when included via FetchContents into clang-tidy, which uses a pretty strict set of warnings.  (This is the last compiler flag.  After this, I just need to clean up a couple of "naked" throws in z3++.)

Adds -Wcovered-switch-default to CLANG_ONLY_WARNINGS, and then deletes all the dfault: classes for which it complains.  (To recap, this flag gives a warning when the type of the argument of switch is an enum, and case statements cover all members of the enum, but there is still a default case.  The default can't be reached except via undefined behavior, casting an int to a the enum type, where the int is not a value of the enum).

I examined the changes pretty carefully to ensure that the preceding case ended in break, return or throw.  There are a couple of exceptions, where some other case falls through intentionally to the body of the default; in that case, just deleting the default works fine.

Nothing in the build that I can find exercises z3++.h (please advise if there is something; that would be useful to know!).  The change there is because I've seen the error when including z3++.h in the clang-tidy build.  Here it's valided "by eye", comparing the rounding_mode enum in that file with the switch that uses it.


NikolajBjorner:
some build errors

NikolajBjorner:
```
  7%] Building CXX object src/math/simplex/CMakeFiles/simplex.dir/simplex.cpp.o
INFO:root:Using /tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/src/math/polynomial/algebraic_params.pyg
INFO:root:Generated "/tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/build/src/math/polynomial/algebraic_params.hpp"
[  7%] Building CXX object src/math/polynomial/CMakeFiles/polynomial.dir/algebraic_numbers.cpp.o
In file included from /tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/src/math/interval/interval.h:22,
                 from /tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/src/math/interval/interval_def.h:21,
                 from /tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/src/math/interval/interval_mpq.cpp:19:
/tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/src/util/ext_numeral.h: In function ‘bool lt(numeral_manager&, const typename numeral_manager::numeral&, ext_numeral_kind, const typename numeral_manager::numeral&, ext_numeral_kind) [with numeral_manager = mpq_manager<false>]’:
/tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/src/util/ext_numeral.h:287:1: error: control reaches end of non-void function [-Werror=return-type]
  287 | }
      | ^
/tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/src/math/dd/dd_bdd.cpp: In member function ‘double dd::bdd_manager::current_cost()’:
/tmp/build-via-sdist-kw8sx8o9/z3_solver-5.0.0.0/core/src/math/dd/dd_bdd.cpp:289:5: error: control reaches end of non-void function [-Werror=return-type]
  289 |     }
      |     ^
cc1plus: some warnings being treated as errors
make[2]: *** [sr
```

davedets:
Interesting.  This seems like a bug, or at least an incompleteness, in the mingw compiler.  The ext_numeral_kind has only 3 elements; the switch statement at line 272 of ext_numeral.h has cases for each of the elements, as does the nested switch for EN_NUMERAL.

I'm able to repro a simplified version of this in godbolt.  I don't know why it's complaining in this case, but not in similar cases.  It shuts up if I add back the default on the inner switch.

I restored the default on the inner switch, with a pragma disabling the clang warning around it, and an explanatory comment.

davedets:
I resolved some merge conflicts, in which your commit:

https://github.com/Z3Prover/z3/commit/4e0ec5d01bbb8042026ffecf39225cd5763d54d7

had fixed some warnings occurring in default clauses that this PR eliminated.  I resolved in favor of my PR, that is, eliminating the default clauses.  I'm assuming that just deleting those default clauses will silence the other warnings (that Clang is consistent in it's reasoning about whether switch statements are "complete").  We'll see...

davedets:
I just noticed build breaks on my last set of changes, for non-Clang compilers.  Fixing.

NikolajBjorner:
created a branch and PR #10460 from that branch into master. In this way I can run local fixes for other platforms.
You can PR into the Detlefs branch.

NikolajBjorner:
merged Detlefs branch

davedets:
I'm sorry it's taking me a while to get the hang of reading the github actions output, and finding the errors.  I fixed one more, and I changed the way I fixed these problems.  I went back to removing the unnecessary default clause, but put the UNREACHABLE () and the return statement at the end of the functions in question (in ext_numeral.h, and dd_bdd.cpp).  I think this is cleaner, and more clearly addresses the GCC warning.

davedets:
> You can PR into the Detlefs branch.

I'm sorry I'm a github noob.  I pushed the changes described in my last comment into the Detlefs branch; please let me know if there's something further I need to do.  Thanks.

PR #10472, opened 2026-08-10
Semilinear length abstraction for regexes

veanes:
Regular languages have semilinear length sets, but `rex::info` only carried
`min_length` / `max_length`. Two bounds cannot express a *phase*, so a goal like
`z in (a^4)* /\ w in (b^6)* /\ |z| + 1 = |w|` reaches arithmetic as `0 <= |z|`,
`0 <= |w|`, `|z| + 1 = |w|` -- satisfiable -- and the solver has to discover the
contradiction by unrolling. This PR gives the length set an ultimately periodic
abstraction and hands the periodicity to the arithmetic solver, so `6n - 4m = 1`
is closed by the gcd test instead.

## What is here

**`src/ast/seq_len_abs.{h,cpp}` -- a new self-contained module.** The value is

```
Lambda(r)  subseteq  { n : lo <= n <= hi, (n mod period) in residues }
```

with `period <= 64` and `residues` a 64-bit mask, plus the lattice operations
(`concat`, `unite`, `meet`, `star`, `plus`, `opt`, `loop`, `residues_mod`, `gcd`).
Only the containment is required, so every rule may weaken its result: precision
can be lost, soundness cannot. Complement degrades to top.

Keeping a residue *set* rather than a single progression `d + g*N` is what makes
the abstraction closed under union: for `(aa)* | (aaa)*` a single progression
degrades to `gcd(2,3) = 1` and says nothing, while this representation records
period 6 with residues `{0,2,3,4}` and still refutes lengths 1 and 5 mod 6.

The module is producer-agnostic -- the abstraction can equally be computed
syntactically over a regex or semantically from a derivative automaton via a BFS
potential function (`progression()`), and two sound abstractions of the same
language combine with `meet()`.

**`rex::info` stores and forwards it** (`seq_decl_plugin`), so the structural
recursion that already computed the bounds now also computes the residues.

**`theory_seq`'s regex solver consumes it** (`seq_regex.cpp`). This is where the
care is: the natural encoding `|s| mod p in residues` needs a length term, and
`th.mk_len(s)` on a goal that never mentioned `str.len` cascades into
theory_seq's length axioms and then into arithmetic. Measured on
`t08-periodic-abc-xyzyxz-sat`, that turned 3 decisions / 4 arithmetic columns
into 2513 columns / 109 arithmetic conflicts. So `propagate_length_residue`
takes whichever path avoids materializing a length term:

1. a **local refutation with no arithmetic at all** -- relax every non-constant
   part of the concatenation to an unconstrained length, giving `|s| in cst + g*N`
   (a superset of the true length set, hence sound for refutation); if no
   admissible residue is reachable there, emit `~lit` directly;
2. otherwise emit the `mod` axiom only when the lengths are already live, i.e.
   `|s|` is internalized or every non-constant part of `s` is (theory_seq derives
   `|xy| = |x| + |y|`, so no new length variable appears).

That split is what separates the families exactly: `x."e".x in intersect_k (Sigma^k)*`
gives `1 + 2k = 0 (mod even)`, unreachable, refuted locally; `xyzyxz in ("abc")+`
gives `2k = 0 (mod 3)` with `gcd(2,3) = 1`, reachable, so nothing is emitted and
no length term is created.

**Two `seq_monadic` commits ride along** because they are measured together below
and touch disjoint files: the work budget is recalibrated from 200k to 1M (a
product pop costs roughly an order of magnitude less since the
interval-refinement product replaced the cartesian one, so the old value bailed
to the legacy fallback on goals the monadic path finishes outright), and an
unread `m_closed` bit vector is dropped from `live_states`.

## Results

1476 regex benchmarks, `-T:10`, inputs pre-warmed, job order shuffled under a
fixed seed, each (file, binary) scored by the minimum of 2 runs.

`smt.seq.regex_monadic=true`:

```
decided 1376 -> 1389    gained 13   lost 0   CONTRADICTIONS 0
commonly decided 1376:  76.3s -> 66.8s  (-12.5%)
  [1000, inf)  n=7      18.3s ->  8.1s  (-56.0%)
```

Default (legacy) mode, evaluated on the length-abstraction commits: **27 gains,
0 losses**, every verdict change re-run 3x on both binaries to filter the
families that flip between repetitions on the *same* binary. The local
refutation is what makes the default mode benefit at all -- it bypasses
arithmetic, which legacy unrolling otherwise drowns.

Gains are coherent families rather than scattered flips: `hard_length_2/3/4`,
`hard_len_nonprim_8`, `noodler_killer_2/5/7` (length/residue refutations,
10 s timeout -> `unsat` in 20-30 ms) and `split_membership_medium_*` (budget).

The two halves are orthogonal: built separately, every length-abstraction gain
survives the budget change and every budget gain survives the length
abstraction.

## Soundness

- 0 contradictions across every sweep (~3000 verdict pairs per run).
- The `split_membership` regressions seen with an earlier, ungated version were
  separated from a possible abstraction bug by pinning each file's stated witness:
  with the witness pinned the branch answered `sat` in 95-137 ms, matching master,
  so the cost was arithmetic, not unsoundness.
- New unit tests in `src/test/seq_len_abs.cpp` pin the lattice arithmetic itself
  (no ast_manager), including a generic "every operation over-approximates" check;
  `src/test/seq_rewriter.cpp` covers how the abstraction is derived from a regex.


NikolajBjorner:
basic arithmetic operators are in util.h or use std::gcd if it can be used.

seq_len_abs should be renamed to not have "seq", not use the seq namespace and moved to util (or math). It doesn't depend on expressions or seq. call it len_abs? or sl_set or sl_abstraction or basic_sl_abs or ...

veanes:
Addressed in 63cdf4b2a:

- Moved `src/ast/seq_len_abs.{h,cpp}` to `src/util/len_abs.{h,cpp}`, dropped the `seq` namespace and the `seq_` prefix — the class is now just `len_abs`. It only includes `util/util.h`, so it has no dependency on expressions or on seq. The unit test moved with it and is registered as `tst_len_abs`.
- Replaced the hand-written `u_gcd` with `std::gcd`.
- The lcm helper stays local, renamed `bounded_lcm`. `std::lcm` is not usable there: its behaviour is undefined precisely on the overflow the helper needs to detect, since it reports overflow by returning 0 to make the caller degrade to the gcd.

No functional change — the moved files are otherwise identical, all seq unit suites pass, and the 13 benchmarks this branch newly decides still decide with the same verdicts.

PR #10499, opened 2026-08-13
Move some_string_in_re into a self-contained seq::regex_witness class

NikolajBjorner:
## Summary

Refactors seq_rewriter::some_string_in_re into a self-contained seq::regex_witness class in a new module st/rewriter/seq_regex_witness.h/.cpp.

- The witness search is now a Brzozowski/Antimirov derivative-cofactor reachability search, using the same general purpose guard_set cofactor algebra that seq_monadic.cpp uses to find non-emptiness witnesses for monadic decomposition components (seq_monadic::product_nonempty), replacing the old manual ITE/exclude-range walker.
- Exposes two entry points:
  - get_witness(expr* r, expr_ref& witness): general purpose - returns a concatenation of seq.unit terms (or seq.empty) for any sequence element sort.
  - get_witness(expr* r, zstring& s): specialization that decodes the same search into a zstring when the regex ranges over strings.
- seq_rewriter::some_string_in_re now constructs a seq::regex_witness locally and delegates to it. The public API is unchanged, so existing call sites (xpr_inverter.cpp, sls_seq_plugin.cpp, lim_uncnstr_tactic.cpp) require no changes.

## Testing
- Full CMake build (libz3, 	est-z3) succeeds with no errors/warnings.
- 	est-z3 /a: 96/96 unit tests pass.
- Manual smoke tests via z3 -in on str.in_re/str.contains queries (including empty-witness and quantified-regex cases) produce correct sat/unsat results and models.


NikolajBjorner:
@copilot
Make bool regex_witness::decode_string(seq_util& u, expr* e, zstring& out) iterative in the last argument of the concatentation.
That is, if e is a concatentation, then call decode_string recursively on the num_args() - 1 first arguments, and set e = to_app(e)->get_arg(num_args-1); and loop.

NikolajBjorner:
@copilot 
Make bool regex_witness::decode_string(seq_util& u, expr* e, zstring& out) iterative in the last argument of the concatentation.
That is, if e is a concatentation, then call decode_string recursively on the num_args() - 1 first arguments, and set e = to_app(e)->get_arg(num_args-1); and loop.

Issue #10417, opened 2026-08-06
`(pop 1)` after an rlimit-cancelled check runs unboundedly inside `lp::dioph_eq` undo — neither timeout nor rlimit applies (≥59 min on a 26 KB file; `lp.dio=false` exits in 23 s)


headinthebox:
### Summary

We run z3 as the backend of an SMT-based program verifier with **both** cancellation mechanisms set
on every query (`rlimit=40000000`, `timeout=1800000`). We keep hitting solves that run minutes to
hours past both. Reducing them produced something more specific than "z3 is slow": **the same
bignum-rational inner loops appear at several sites with three different cancellation outcomes** —
polled-late, polled-never, and structurally un-pollable:

| site | phase | `timeout` | `rlimit` |
|---|---|---|---|
| `lp::dioph_eq::imp::check → rewrite_eqs → normalize_e_by_gcd` | final_check | honoured **~9× late** (10 s wall → 88 s) | honoured at a sparse poll (~25–48 s in) |
| `grobner::compute_basis_step → merge_monomials` (arith.solver=2) | final_check | honoured at an op boundary (30-min wall → +4.8 s once; in other instances never before an external kill) | **never fired across a 30-min grind with a 40 M cap set** |
| `smt::context::internalize_assertions` (bit-blast) | pre-search | latency Θ(goal): 1-s wall returns at 5–14 s as the goal doubles | phase charges <5 M for a 28-s / 9-GB build — a budget that lets normal queries through can never bind it (**your #4192, closed wontfix — cited for completeness, not re-litigation**) |
| **`smt::context::pop_scope_core → theory_lra::pop_scope_eh → lp::lar_solver::pop → lp::dioph_eq::imp::undo_add_term_method`** | **backtracking / `(pop 1)`** | **cannot apply** | **cannot apply** |

The last row is the reportable core. A `(pop 1)` executed by the SMT2 front end after an
rlimit-cancelled check spends **≥59 minutes** (we killed it at a 1-hour external wall; it had
answered everything it was ever going to answer within the first half-minute) doing dense rational
matrix elimination:

```
smt2::parser::parse_pop → cmd_context::pop → solver_na2as::pop → smt_solver::pop_core
→ smt::context::pop → smt::context::pop_scope_core → smt::theory_lra::pop_scope_eh
→ lp::lar_solver::pop → lp::dioph_eq::imp::undo_add_term_method → shrink_matrices
[...]

PR #10419, opened 2026-08-06
Avoid eager Diophantine matrix elimination on pop

NikolajBjorner:
fix #10417

Defer rebuilding Diophantine derived matrices until the next check instead of performing dense rational elimination while popping scopes. Rebuild only callback-tracked surviving terms and restore term-tightening state. Adds scoped activation/pop/recheck coverage.

PR #10503, opened 2026-08-13
Bound the work done when a term is retired from dioph_eq (#10417)

levnach:
Retiring a term makes undo_add_term_method() eliminate the column of that term from m_l_matrix by dense Gaussian elimination: the last row is added to every other row of the column. The cost is a function of the accumulated state rather than of the increment being undone, and the rational coefficients blow up along the way. On the reported 26 KB file a single (pop 1) reached 400 rows per elimination, pivot rows of 445 entries holding 423000 bits, and ran for over an hour without finishing.

The path runs under the trail, either from the parser reading (pop 1) or from resolve_conflict(), so neither the timeout nor the rlimit applies to it, and a half finished elimination would leave m_l_matrix and m_e_matrix inconsistent, so it cannot be made interruptible either.

Bound it instead. Every entry is a linear combination of the definitions of the active terms, so when a pop exceeds the budget the derived state is dropped and the still active terms are put back into m_added_terms; init() refills the matrices on the next check(). The rebuilt state is the one a fresh solver would have for the same terms, and the rebuilding happens inside check(), where the rlimit and the timeout are polled. The same is done when the cancel flag is already set, since nothing reads the derived state before it is rebuilt.

The budget is lp.dio_undo_max_work, in machine words, 0 disables it and restores the previous behavior. Resets are reported as arith-dio-state-resets.

The reported file now finishes in 17-33 seconds across seeds instead of never, matching the lp.dio=false control.


NikolajBjorner:
see also, https://github.com/Z3Prover/z3/pull/10419

PR #10519, opened 2026-08-14
Let the monadic regex solver choose the direction it reads memberships in

veanes:
Adds an option to the monadic regex solver to read memberships backwards, and a
policy that picks the direction by observation rather than up front.

`w ∈ R` iff `rev(w) ∈ rev(R)`, and the two directions can need very differently
sized derivative automata: `.*a.{k}` needs `2^(k+1)` states read forwards and
`k+2` read backwards. Nothing in the search depends on the direction, so the
direction becomes a choice.

**All three commits are inert by default.** `seq.regex_orientation` defaults to
`forward`, which is bit-identical to master in verdict and rlimit count. Nothing
here changes behaviour unless it is asked for.

## Results

1476 regex benchmarks (`z3prover/bench`, `inputs/regexes`), 10s timeout, run
serially:

| | solved | time |
|---|---|---|
| `forward` (default, = master) | 1395 / 1476 | 894.6s |
| `retry` | **1407** / 1476 | **782.2s** |

**+12 solved, 0 lost, 0 verdict disagreements.** The solved set of `forward` is
a strict subset of `retry`'s.

All twelve gains are files where forward times out. Each contains a `.*a.{k}`
under a complement — exactly the shape whose forward automaton is exponential
and whose reverse is linear. `k` ranges 3–12, and none of the twelve contains
any arithmetic (`str.len`, Int), so these gains look orthogonal to the
arithmetic-rewrite work landing separately.

This is a tail fix, not a general speedup. Of the 69 files neither direction
solves, most are ones whose reverse has the same shape as the original, which
reversal cannot help.

## The three commits

1. **Expose `seq_monadic` budget and transition mode as parameters** —
   `seq.regex_budget` and `seq.regex_transition_mode`, previously hard-coded.
   Needed to calibrate anything here.

2. **Bound the per-state expansion in the product search** — independent
   completeness fix, worth reviewing on its own merits. `product_nonempty`
   charged one budget unit per product state, but expanding a single state is
   itself unbounded: the interval sweep and cartesian branch enumeration are
   lambdas called from inside the polled outer loop, and neither polled
   anything. A state whose components have many branches could run arbitrarily
   long without returning to the budget *or the resource limit*. Each state now
   gets a fixed expansion allowance. Verified verdict- and rlimit-identical to
   its parent over 200 files.

3. **Choose the direction** — `seq.regex_orientation` = `forward` (default) |
   `reversed` | `retry`.

## Why `retry` rather than just `reversed`

Committing to the reversed reading up front is a bad bet: over a 200-file
sample `reversed` solves one file *fewer* than forward and spends 6% more
rlimit, even though it is 500x better on the cases it wins. The direction that
pays off is not predictable from the input.

`retry` reads forwards and, only when a decision runs out of its work budget,
re-reads that same decision backwards. Bailing is the right trigger: the other
ways of giving up (unsupported shape, undecidable nullability, a guard the
range solver cannot evaluate) are properties of the problem rather than of the
direction, and reversing cannot fix them.

Two details matter for this not to lose ground:

- **The second attempt gets a full budget, not half of a split one.** Halving
  costs decisions plain forward solves outright:
  `split_membership_easy_unsat_0013` is unsat at 15155 rlimit with a budget of
  3000 and no bails at all, but bails 52 times at 1500 and then dies on a
  reversed search that never terminates.
- **One failed retry disables retry for the rest of the query.** A bail is not
  in itself bad — it hands the problem back to `theory_seq`, which has its own
  way of making progress — so a reversed attempt that *also* fails is pure
  loss, and the same regexes recur at every decision.

## Implementation notes

Reversal is **all or nothing**: a system where some memberships read forwards
and others backwards constrains a mixture of `w` and `rev(w)` and is not the
original problem, so if any regex cannot be reversed the whole problem stays
forwards (counted under a distinct `not reversible` bail; it never fires on
this corpus).

A reversed variable is named `rev(x)` for a fresh uninterpreted marker rather
than silently reinterpreting `x`, so a forward and a reversed occurrence of the
same variable are distinct terms and cannot be collected into one group by
mistake. The marker is internal — `leaf()` strips it and turns the witness back
around — so `get_model()` only ever mentions the caller's own terms, and **no
reverse operator on sequences is required**.

Lengths are preserved by reversal, so the bounds added by `add_lo`/`add_hi`/
`add_len` need no special handling. Cores are unaffected, being keyed on the
caller's dependencies.

## Validation

- 96/96 unit tests.
- 200 files × 3 orientations with `model_validate=true`: zero verdict
  contradictions, zero invalid models — the un-reversal of witnesses is sound.
- `forward` bit-identical to master across the corpus.
- Each commit builds in isolation (bisectable).

## Open question

The data supports making `retry` the default (+12, 0 losses, −13% time), but
this PR deliberately leaves the default at `forward` so it can be merged
without changing anyone's results. Happy to flip it in a follow-up if that is
preferred.

One known limitation if the default is ever changed: the trigger observes
budget exhaustion, not slowness. A forward search that is merely slow without
ever bailing never triggers a retry — a minimal `re.comp(.*a.{16})` reproduces
this. On this corpus the default budget gives 12/12 recall, so it is latent,
but it means `seq.regex_budget` also controls trigger sensitivity.


src/test/seq_monadic_bench.cpp
```diff
@@ -63,6 +63,24 @@ seq::transition_mode get_mode() {
...
+unsigned get_budget() {
+    char const* budget = getenv("Z3_SEQ_MONADIC_BUDGET");
+    if (budget)
+        return static_cast<unsigned>(strtoul(budget, nullptr, 10));
+    return 1000000;
+}
+
+// Mirrors the seq.regex_orientation parameter so the bench can compare directions.
+seq_monadic::orientation get_orientation() {
+    char const* o = getenv("Z3_SEQ_MONADIC_ORIENTATION");
```
NikolajBjorner, line 76: this is something agents add. You can use configuration parameters

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -534,8 +586,86 @@ void seq_monadic::reset_search() {
...
 }
 
-bool seq_monadic::prepare(membership_vec const& memberships) {
+bool seq_monadic::reverse_regex(expr* r, expr_ref& result) {
+    result = expr_ref(re().mk_reverse(r), m);
+    m_thrw(result, result);
+    // seq_rewriter pushes re.reverse through every constructor it knows, so a surviving
+    // occurrence means some subterm was opaque to it and the language is not the one the
+    // search would then explore.
+    ptr_vector<expr> todo;
```
NikolajBjorner, line 595: use instead a standard iterator over subterms.

for (auto e : subterms::ground(result)) 
    if (is_reverse(e))
         return false;

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -315,6 +330,34 @@ lbool seq_monadic::product_nonempty(svector<component> const& comps, expr_ref* w
     key st_key;
     bool bail = false;
 
+    // Expanding one product state is not a bounded amount of work: the interval sweep and
```
NikolajBjorner, line 333: skip the agent verbiage

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -534,8 +586,86 @@ void seq_monadic::reset_search() {
...
+            continue;
+        }
+        if (u().str.is_empty(e))
+            continue;
+        units.push_back(e);
+    }
+    expr_ref_vector es(m);
+    for (unsigned i = units.size(); i-- > 0; )
+        es.push_back(units[i]);
+    if (es.empty())
```
NikolajBjorner, line 631: you don't need this special case. mk_concat handles it.

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -534,8 +586,86 @@ void seq_monadic::reset_search() {
...
+            continue;
+        if (re().is_reverse(e))
+            return false;
+        for (expr* arg : *to_app(e))
+            todo.push_back(arg);
+    }
+    return true;
+}
+
+expr_ref seq_monadic::reverse_word(expr* w) {
```
NikolajBjorner, line 611: does this really belong in seq_monadic?
seq_rewriter instead?

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -983,12 +1166,18 @@ void seq_monadic::collect_statistics(::statistics& st) const {
         "seq monadic bail state cap",
         "seq monadic bail dnf cap",
```
NikolajBjorner, line 1167: BTW, dnf cap is probably no longer used

src/ast/rewriter/seq_monadic.h
```diff
@@ -183,6 +207,14 @@ class seq_monadic {
     std::unordered_map<group_sig, lbool, group_sig_hash> m_group_cache;
     seq::live_states m_live_states;
 
+    // Names the reversed reading of a sequence variable.  When the search runs backwards a
```
NikolajBjorner, line 210: skip the AI verbiage

PR #10527, opened 2026-08-15
Fix non-ground regex fallback relevance

### LINKED ISSUE #10511: [Solution Soundness Bug] Incorrect SAT with `check-sat-assuming` across multiple regex operators
Z3 incorrectly reports `sat` when a constraint is supplied through `check-sat-assuming`.

### Input:

```smt2
(declare-fun y () String)
(declare-fun o () String)
(assert
  (not
    (str.in_re
      (str.substr y 0 1)
      (re.+ (str.to_re (str.substr y (str.len o) 1))))))

(check-sat-assuming ((= (str.len o) 0)))
```

### Output:

```text
$ z3 test.smt2 model_validate=true
sat
(error "line 9 column 39: an invalid model was generated")
$ cvc5 test.smt2 -q
unsat
```

The same constraint added with a regular `assert` is correctly reported as `unsat`.

Under the assumption `(= (str.len o) 0)`, both `str.substr` terms denote the same string `s`. Since `s` always belongs to `(re.+ (str.to_re s))`, including the case `s = ""`, the asserted negation is false.

Adding `(get-model)` after the incorrect `sat` gives, for example:

```smt2
(define-fun o () String
  "")
(define-fun y () String
  "A")
```

which also violates the assertion under the assumption.

Replacing the assumption with a regular assertion:

```smt2
(assert (= (str.len o) 0))
(check-sat)
```

correctly returns `unsat`.

Here, `R = (str.to_re (str.substr y (str.len o) 1))`. Replacing the second argument of `str.in_re` with the following regex terms gives:

| Regex term | Z3 | Expected |
|---|---|---|
| `(re.+ R)` | `sat` | `unsat` |
| `(re.* R)` | `sat` | `unsat` |
| `((_ re.loop 0 2) R)` | `sat` | `unsat` |
| `((_ re.loop 1 3) R)` | `sat` | `unsat` |
| `(re.++ R (re.* R))` | `sat` | `unsat` |
| `(re.union R (str.to_re "Z"))` | `sat` | `unsat` |
| `(re.inter (re.+ R) (re.* R))` | `sat` | `unsat` |

Possibly related to #10492 (`str.at` / `re.*`). The `str.at` variant of this pattern is answered correctly after that fix, while the `str.substr` variants above still reproduce the issue. I also tried to explore the boundary of this soundness issue. There may be other related cases in nearby regex paths, so it may be worth checking them as well.

Commit: 388af165c5459f27890c06f3718927b1ca07d11d

NikolajBjorner:
Route all regex memberships through the monadic solver and mark fallback guard atoms relevant so their theory assignments are enforced.

Fixes #10511

### DIFF
--- a/src/smt/seq_regex.cpp
+++ b/src/smt/seq_regex.cpp
@@ -65,6 +65,14 @@ namespace smt {
     arith_util& seq_regex::a() { return th.m_autil; }
     void seq_regex::rewrite(expr_ref& e) { th.m_rewrite(e); }
 
+    void seq_regex::mark_guard_relevant(expr* e) {
+        auto mark_atom = [&](expr* n) {
+            if (is_atom(m, n) && !m.is_true(n) && !m.is_false(n))
+                ctx.mark_as_relevant(th.mk_literal(n));
+        };
+        for_each_expr(mark_atom, e);
+    }
+
     expr_ref seq_regex::expand_shallow(expr* e, void*& deps, unsigned depth) {
         expr* elem = nullptr;
         if (depth > 40)
@@ -522,8 +530,7 @@ namespace smt {
 
         propagate_length_residue(lit, s, r);
 
-        bool is_ground = re().is_ground(r);
-        if (th.use_monadic_regex() && is_ground)
+        if (th.use_monadic_regex())
             add_monadic_membership(lit, s, r);
         else
             propagate_accept_legacy(lit, s, r);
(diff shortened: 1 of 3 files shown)


veanes:
Heads-up: this landed a ~15x slowdown on regex benchmarks that fall back from the monadic solver to the legacy accept path. Bisected to this commit, and narrowed within it to the two `mark_guard_relevant(...)` calls rather than the routing change -- removing just those on current master takes `split_membership_medium_unsat_0034.smt2` from timeout back to 5.6s (it was 3.08s before this commit, 46.7s now). Details and the per-half measurements are in #10543. The fix looks correct, so this is a report rather than a revert request.

PR #10532, opened 2026-08-15
Use the PA view constraint for seq_monadic as well

CEisenhofer:


src/ast/rewriter/seq_view.h
```diff
@@ -0,0 +1,108 @@
...
+#include "util/uint_set.h"
+#include "util/vector.h"
+
+namespace seq {
+
+    struct view {
+        expr*           m_state = nullptr;
+        expr*           m_target = nullptr;     // null: membership view
+        uint_set const* m_region = nullptr;     // null: whole automaton
+        bool            m_complemented = false;
```
NikolajBjorner: is this field used?

src/ast/rewriter/seq_view.h
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    inline bool operator==(view::sig const& a, view::sig const& b) {
+        return a.state == b.state && a.target == b.target
+            && a.region == b.region && a.complemented == b.complemented;
+    }
+
+    using view_vector = svector<view>;
+
+    // Uncached reference semantics; an engine with its own caches has to agree with these.
```
NikolajBjorner, line 82: these appear not to be currently used.
Don't we want steps over an entire view vector, which is what the non-emptiness check internal to seq_monadic accomplishes.

src/ast/rewriter/seq_monadic.h
```diff
@@ -317,15 +263,21 @@ class seq_monadic {
 
     seq::transition_mode mode() const { return m_config.m_mode; }
 
-    // Enable/disable model generation (default: enabled).  When enabled, a successful
-    // solve()/check() extracts a feasible model retrievable via get_model().
-    void set_gen_model(bool b) { m_config.m_model = b; }
+    // Record the solution (default: enabled).  Callers that only want the verdict
+    // switch it off; the search itself is unaffected.
+    void set_gen_solution(bool b) { m_config.m_solution = b; }
```
NikolajBjorner, line 320: It is fine to (have an agent) rename this to m_solution. 
It can be surfaced in a self-contained PR to declutter from other merge conflicts.

src/ast/rewriter/seq_monadic.h
```diff
@@ -317,15 +263,21 @@ class seq_monadic {
...
-    // where each witness is a concrete sequence term (over the element sort) giving one
-    // satisfying assignment.  Witness terms are pinned by the solver and remain valid
-    // until the next solve()/check().  Only valid when model generation is enabled.
-    obj_map<expr, expr*> const& get_model() const { return m_model; }
+    // Collapse `var`'s views into one value: a word driving all of them to acceptance
+    // at once -- the first the product search finds, not the shortest.  The only place
+    // a word gets built.  l_true = `word` set, l_false = empty intersection,
+    // l_undef = no recorded solution, or the search gave up.
+    lbool materialize(expr* var, expr_ref& word);
+    lbool materialize_all(obj_map<expr, expr*>& model, expr_ref_vector& pin);
```
NikolajBjorner: a bit odd to have pin vector be a separate argument. 
The pin vector maintains life-time of range of map.
Could we use expr_substitution or expr_map instead as these structures manage life-times of elements?

NikolajBjorner:
looks like the right direction: surfacing views, or products of views, and operations on these. 
Additional in comments

NikolajBjorner:
Merged current master into this branch and resolved the seq_monadic/seq_regex conflicts by keeping the view-based seq_monadic representation from this PR while also preserving master's newer search behavior (out_of_budget(), retry/reversed orientation handling, and the updated bail reasons including state-expansion and not-reversible).

Addressing review comments:
1. seq_view.h field question: kept `m_region`. It is still used by `view::in_region()`, participates in `view::key()`/equality for dedup+memos, and is consulted by the reference semantics in seq_view.cpp and by seq_monadic's conservative `product_nonempty` guard.
2. seq_view.h stepping API: removed the unused per-view `step(...)` helper from seq_view.h/.cpp. The remaining view helpers are the reference semantics; whole-view-vector stepping stays inside seq_monadic's product emptiness search, as suggested.
3. seq_monadic.h rename to `m_solution`: intentionally deferred per review guidance so it can land separately in a self-contained cleanup PR.
4. seq_monadic.h pin-vector lifetime management: replaced the `materialize_all(..., pin)` API with `materialize_all(expr_substitution&)`, so the substitution object itself owns refs for both keys and witnesses instead of maintaining a parallel pin vector. Updated seq_regex and the seq_monadic tests accordingly.

Validation:
- `ninja test-z3` (existing Ninja build in `build`)
- `build\\test-z3.exe seq_monadic`

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -706,22 +712,63 @@ lbool seq_monadic::leaf() {
...
         }
+        m_solution.insert(m_vars[vi], m_groups[vi]);
+    }
+    return l_true;
+}
+
+lbool seq_monadic::materialize(expr* var, expr_ref& word) {
+    // without a recorded solution m_solution is empty, and an empty word would pass
+    // for a satisfying assignment
+    if (m_last_result != l_true || !m_config.m_solution || !m_seq_sort)
```
NikolajBjorner: do we depend on m_seq_sort to be globally consistent among all membership constraints?
This will not work if there are different types of sequences mixed.

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -766,9 +813,9 @@ lbool seq_monadic::dfs_atoms(unsigned mi, unsigned i, expr* R) {
 
     // Explores one split target; the caller stops at the first l_true.
     auto explore = [&](expr* target) -> lbool {
-        m_groups[vi].push_back(component{ a.var.get(), R, target });
+        m_groups[vi].push_back(target ? seq::view::reach(R, target) : seq::view::membership(R));
```
NikolajBjorner, line 825: just use seq::view::reach(R, target). membership is a special case when target == nullptr

src/ast/rewriter/seq_view.h
CEisenhofer: Nope; artifact - removed 👍

src/ast/rewriter/seq_monadic.h
CEisenhofer: Yup; done

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -619,10 +632,18 @@ bool seq_monadic::prepare(membership_vec const& memberships, bool reversed) {
     }
     unsigned mi = 0;
     for (auto const& [term, regex, d] : memberships) {
-        if (!u().is_re(regex, m_seq_sort)) {
+        sort* seq_sort = nullptr;
+        if (!u().is_re(regex, seq_sort)) {
+            m_stats.inc_bail(bail_reason::unsupported);
+            return false;
+        }
+        // The memberships are decided jointly over one guard algebra, so they all have to
```
NikolajBjorner, line 640: ok for now. Can be fixed later

PR #10534, opened 2026-08-15
Fix incorrect strict optimization objective reporting

NikolajBjorner:
## Summary
- track active delta-rational optimization validation constraints
- suppress implied-bound projection only when its explanation depends on such a constraint
- add a regression for the disjunctive strict-real optimization case

Fixes #10356

Issue #10505, opened 2026-08-14
z3 performance issue: QF_NRA degree-80 with perfect-square bound


NicholasEhsanRoy:
## Summary

z3 5.0.0 takes >15s (and often times out) on a trivial QF_NRA formula
involving `y^80 = x` with `x <= 100`, while degree 79 and 81 complete
in <200ms. Changing the upper bound to 99 or 101 also makes degree 80
fast (<500ms). cvc5 handles all cases in <200ms.

## Reproduction

```python
import z3, time

x, y = z3.Real('x'), z3.Real('y')
s = z3.Solver()
s.set('timeout', 15000)
s.add(x >= 1, x <= 100, y >= 0, y**80 == x, y < 1)
t0 = time.time()
print(s.check(), f'{time.time()-t0:.1f}s')
# Result: unknown 15.2s (timeout; should be unsat in <200ms)
```

### SMT-LIB2 Equivalent

```smt2
(set-logic QF_NRA)
(declare-const x Real)
(declare-const y Real)
(assert (>= x 1.0))
(assert (<= x 100.0))
(assert (>= y 0.0))
(assert (= (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y (* y y))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))) x))
(assert (< y 1.0))
(check-sat)
; Expected: unsat in <200ms (consistent with degrees 79, 81)
; Actual (z3 5.0.0): unknown after 15s timeout
```

### Degree sensitivity (bounds [1, 100])

| Degree | Time | Result |
|--------|------|--------|
| 79 | 0.196s | unsat |
| **80** | **>15s** | **unknown (timeout)** |
| 81 | 0.127s | unsat |

### Bound sensitivity (degree 80)

| Bounds | Time | Result |
|--------|------|--------|
| [1, 99] | 0.419s | unsat |
[...]

PR #10552, opened 2026-08-17
Decompose regex intersections in the monadic solver

veanes:
A membership `t ∈ (R₁ ∩ … ∩ R_k)` forces the derivative search to explore the product of all *k* regexes at once. That is where the monadic solver spends its budget on the intersection-heavy benchmarks, even when a couple of the conjuncts already contradict each other and the rest are irrelevant to the refutation.

This PR adds an opt-in CEGAR loop that decides a **relaxation** keeping only some of the intersected regexes, and grows it on demand.

## Why it is sound

Dropping conjuncts from an intersection only enlarges the language, so the relaxation is implied by the original:

- relaxation `unsat` ⟹ original `unsat` — return immediately;
- relaxation `sat` with model *w* ⟹ original `sat` **iff** every dropped conjunct also accepts *w*, which is checked by running *w* through its derivatives.

Otherwise one violated conjunct is added back and the loop repeats. Running out of rounds returns `l_undef`, which is what the caller already handles.

## Which conjunct to add back

This is the whole game — it decides how fast the loop converges.

Prefer a conjunct that **pins word lengths to a residue class** (a `re.loop` with `lo == hi > 1`). The search refutes a membership set by exhausting the reachable product, and a length constraint shrinks that product across every branch at once, where a conjunct that merely forbids a rare infix leaves it essentially unchanged.

Among equals prefer the **cheapest**, measured by the work a short lookahead probe leaves unspent: a candidate that exhausts its probe has rebuilt the product this is trying to avoid. A probe that refutes within its budget is an answer to the whole query, not just to the comparison, so it is returned directly.

Ranking by cost alone is actively misleading, because the counting conjuncts carrying the refutation are precisely the expensive ones it avoids. Measured in isolation with `test-z3 seq_monadic_bench` over 28 unsat files:

| ranking | decided |
|---|---|
| decomposition off | 2 / 28 |
| cheapest first | 9 / 28 |
| **length-constraining first, then cheapest** | **14 / 28** |
| widest first | 3 / 28 |

## When it runs, and what it costs

The decomposition is only worth its price for a decision the undivided search cannot make, and the cheapest way to find that out is to give the undivided search a fraction of its budget first. A decision reached within that fraction is one the decomposition could only have slowed down; and when the decomposition fails, nothing is lost, because the fraction is re-spent as the prefix of the full attempt that follows. So `decide()` becomes:

1. undivided search at `limit/128` — today's policy, small budget;
2. if that *ran out of work* (as opposed to meeting something undecidable), the decomposition at `limit/16` per relaxation;
3. otherwise the undivided search at the full `limit` — exactly today's behaviour.

The decomposition as a whole is held to one undivided budget, so a query it cannot decide costs no more than the search it stands in for. It is also suppressed inside `minimize_core`, which calls `decide()` once per deletion candidate and would otherwise pay for it quadratically.

There is a free structural gate before any of this: if the conjunct count does not exceed the membership count, nothing is intersected and the loop returns immediately.

## Measurements

`smt.seq.regex_split` is the number of refinement rounds, **default 0 (off)**. 1476-file regex corpus, 10s timeout, same binary both columns:

| | `regex_split=0` | `regex_split=10` |
|---|---|---|
| decided | 1407 | **1430** |
| gains | — | **23** |
| losses | — | **0** |
| sat/unsat disagreements | — | **0** |
| wall clock | 794.7s | **602.0s** (−24%) |

`regex_split=0` reproduces master file for file. Verified against a *pristine build of current master* over the 131 files that cover every file the decomposition engages on: 84 decided on both sides, **0 differences**. Against that same pristine master, `regex_split=10` gains 23 and loses 0.

The decomposition engages on 79 files (5.4%), all intersection-heavy: 78 in `ClemensRegex/Automatic`, one in `ClemensRegex/Manual`. The other 1397 pay only a conjunct count. All 23 newly decided files were 10s timeouts before; 15 of them now finish in under 2.5s.

`rounds=10` was chosen against 20 and 40, which reintroduce a sat regression: enough budget goes into relaxations that a file the full-budget undivided search solves no longer gets there.
## Caveats

`constrains_length` is a syntactic test, and the benchmark family that benefits is built around modular saturation factors, so the criterion is to some extent tuned to it. It is a heuristic for *ordering* refinements, though — it cannot make an answer wrong, only make the loop converge slower on a family it does not fit.

## Interaction with the seq::view refactor (#10532)

Rebased onto it. The refinement loop reads the solution of a search it ran itself, so two things had to change: `materialize()`'s body is factored into a private `materialize_recorded()` without the precondition on `m_last_result` (which only a top-level `solve()`/`check()` sets), and the `gen_solution()` check moves into that helper — without a recorded solution every variable materializes to the empty word, which would pass for a satisfying assignment.

Collapsing views also runs a product search where the old model was a hash lookup, so witnesses are cached per refinement round rather than rebuilt once per dropped conjunct.

## Tests

`test-z3 /a` = 96/96. Adds an "intersection decomposition" case to `src/test/seq_monadic.cpp`: two length-incompatible cycles behind eight camouflage infixes, which is `l_undef` with the decomposition off and `l_false` with it on, in both transition modes. `src/test/seq_monadic_bench.cpp` honours `smt.seq.regex_split` so the loop can be exercised without the surrounding SMT search.


NikolajBjorner:
@levnach - we should fix the unit tests so they don't fail sporadically. Some of the unit tests set global parameters. If the parameters are set in different threads they can interfere. We could single out unit tests that have to be serialized or we can remove brittle tests.

levnach:
Fix in #10557.

The flake is the `api` test, not thread-shared global parameters — `run_test_child` runs each test in its own `popen` child process, so nothing leaks between tests.

`test_qfnra_degree80_square_bound` asserted `Z3_solver_check` refutes the degree-80 formula within a 2M rlimit. The resource count is machine independent, but the path reaching it is not: `Z3_mk_solver` lands in `mk_qfnra_mixed_solver`, which picks among five sub-solvers using `mk_memory_probe()` (`memory::get_allocation_size()`, a process-global counter) against a 30MB threshold, and each candidate runs under a `try_for()` wall-clock deadline. Debug measures `:max-memory 33.09` — just past the threshold — and spends 6.2s working through those deadlines. Under CI load the deadlines expire after less work, more candidates run, and the total crosses 2M.

It was also already removed once in f9c99afb3; #10534 reintroduced it from a branch that predated the removal.

#10557 drops it and covers the #10505 fix where the measurement is deterministic: factoring `y^80 - 100` costs 107,425 resource units with the default of 4 finite-field trials versus 7,455,518 with 1, so a 1M limit separates them by 9x/7x. Pure computation, so the count is identical across runs and across debug and release. Verified by reintroducing the regression, and `test-z3 /a` is 96/96 in both build types.

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -892,32 +899,264 @@ lbool seq_monadic::decide_oriented(membership_vec const& memberships, bool rever
     return r;
 }
 
+bool seq_monadic::constrains_length(expr* r) {
```
NikolajBjorner, line 902: use instead the 

return any_of(subterms::ground(expr_ref(r, m))), [&](expr* t) {
     return re().is_loop(t, body, lo, hi) && lo == hi && lo > 1;
      });

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -892,32 +899,264 @@ lbool seq_monadic::decide_oriented(membership_vec const& memberships, bool rever
...
+    }
+    return nullable(state);
+}
+
+lbool seq_monadic::decide_split(membership_vec const& memberships, unsigned budget,
+                                unsigned allowance) {
+    // Flatten every membership into the regexes its top-level intersection conjoins.
+    vector<ptr_vector<expr>> conjuncts;
+    unsigned total = 0;
+    for (auto const& [term, regex, d] : memberships) {
```
NikolajBjorner, line 975: should we normalize memberships on creation time to not have top-level intersection?

NikolajBjorner:
I will check against QF_S and then merge

src/ast/rewriter/seq_monadic.cpp
veanes: I think flattening here is fine, low overhead.

PR #10565, opened 2026-08-18
Rule to eliminate word equations based on their conflicting regex over-approximations

CEisenhofer:


src/ast/rewriter/seq_eq_approx.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+    Clemens Eisenhofer 2026
+
+--*/
+
+#include "ast/rewriter/seq_eq_approx.h"
+#include "ast/ast_pp.h"
+
+namespace {
+    char const* mode_name(seq::transition_mode mode) {
```
NikolajBjorner: this kind of code doesnt belong here.
It should be part of seq_derive.

src/ast/rewriter/seq_eq_approx.cpp
```diff
@@ -0,0 +1,171 @@
...
+        switch (mode) {
+        case seq::transition_mode::brzozowski_tm: return "brzozowski";
+        case seq::transition_mode::light_antimirov_tm: return "light-antimirov";
+        }
+        return "unknown";
+    }
+
+    char const* result_name(lbool r) {
```
NikolajBjorner: can we not have special case code like this?
Just one thing to display lbool, defined in lbool.h

src/ast/rewriter/seq_eq_approx.cpp
```diff
@@ -0,0 +1,171 @@
...
+    m_h_pin.reset();
+}
+
+expr* seq_eq_approx::get_regex(expr* t) const {
+    expr* r = nullptr;
+    m_h.find(t, r);
+    return r;
+}
+
+bool seq_eq_approx::abstract_rec(expr* t, expr_ref_vector& parts, sort* re_sort) {
```
NikolajBjorner: can this ever be used in a context where it returns false?

src/ast/rewriter/seq_eq_approx.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+bool seq_eq_approx::abstract(expr* term, expr_ref& result) {
+    sort* seq_sort = term->get_sort();
+    if (!u().is_seq(seq_sort))
+        return false;
+    sort* re_sort = re().mk_re(seq_sort);
+    expr_ref_vector parts(m);
+    if (!abstract_rec(term, parts, re_sort))
+        return false;
+    if (parts.empty())
```
NikolajBjorner: There is an overload of re().mk_concat that works directly without special cases for empty vectors and N-ary.

src/ast/rewriter/seq_eq_approx.h
```diff
@@ -0,0 +1,104 @@
...
+    seq_eq_approx.h
+
+Abstract:
+
+    Regular over-approximation of a word equation.
+
+    Both sides of L = R are lifted to regexes: constants map to themselves, any other
+    part to its image under a mapping h, or to Sigma^* when h does not mention it.
+
+        L = x.a.y   =>   h(L) = h(x) . (str.to_re "a") . h(y)
```
NikolajBjorner: isn't the requirement to be able to approximate variables by the intersections from seq_monadic?
So the approximations would require view vectors?
Or is this plain version enough for you?

src/ast/rewriter/seq_eq_approx.cpp
CEisenhofer: I am not sure what you mean

NikolajBjorner: this function returns false only if r->get_sort() != re_sort.
But can this ever happen? We do assume that m_h is type consistent: it maps sequences t of type (Seq X) to regexes of sort (RegEx (Seq X)). Right?

src/ast/rewriter/seq_eq_approx.cpp
```diff
@@ -0,0 +1,171 @@
...
+    ++m_stats.m_checks;
+    m_last_result = l_undef;
+    m_lhs_image = nullptr;
+    m_rhs_image = nullptr;
+    if (lhs->get_sort() != rhs->get_sort() || !u().is_seq(lhs->get_sort()) ||
+        !abstract(lhs, m_lhs_image) || !abstract(rhs, m_rhs_image)) {
+        ++m_stats.m_unsupported;
+        return l_undef;
+    }
+    m_last_result = m_search.intersect_nonempty(m_lhs_image, m_rhs_image);
```
NikolajBjorner: Reduction to intersect_nonempty is perhaps a good direction, but it is not fully compatible with "views".
Suppose you make m_lhs_image and m_rhs_image to be concatentations of view vectors.
Then we can check non-emptiness for concatenatations of view vectors using an extension of intersect_nonempty. It would cover more cases, right?

The algorithm for checking non-emptiness of this intersection can use derivatives and backtracking search on the view vectors. seq_monadic implements a special case where it checks non-emptiness of a single view vector. What is needed is to continue from terminal states of a view vector to the next element in the concatentation of view vectors. 

Abstract example: m_lhs_image is  "a <(R1, R2), R3> ab < R4, (R5, R6)>", where "a", "b" are characters and R1, .., R6 are regexes.
The first character has to be "a". Then consume the vector of two views (R1, R2) and R3. Here (R1, R2) means a regex that starts with R1 and with accepting state R2. R3 is just a normal regex (the view (R3, null)).
Any accepting state of this view vector can continue to consuming ab, and then <R4, (R5, R6)>.

Thus, you have a method to set_regex(expr* t, expr* regex) for a term.
What could be supplied is access to seq_monadic that maintains the map from t to a view_vector.

CEisenhofer:
In the end we should make a view again a regex expr - that would simplify a lot

NikolajBjorner:
@veanes - what are the options for scheduling the call to approx-set that should be considered?
before/after final check on seq_monadic? Based on partial saturation of seq-monadic?

veanes:
> Margus Veanes (@veanes) - what are the options for scheduling the call to approx-set that should be considered? before/after final check on seq_monadic? Based on partial saturation of seq-monadic?

I would say running before, with lightweight syntactic guard that both sides of the approx have at least one non-`.*` regex (otherwise rather pointless). Also individual conditions such as `a.*` are handled well elsewhere and might just disguise word equations (such as starts or ends with a certain letter or concrete word). Based on the fact that there are only 57 benchmarks currently in the regex bench with any word equations and only 22 have useful approximations out of which only 6 refute anything. The 6 cases I found refute early, running them after would not run them at all as monadic times out. The cost is (near-nothing) bounded with the guard in place. (Running after also ties into the partial saturation idea -- tricky -- and we need more benchmarks to evaluate this, which we don't really have.)

veanes:
There is also the question of later C3 integration which would make it possible to expose more of the internal structure of seq_monadic (wrt. partial saturation).

veanes:
Some additional notes, to clarify:

Regarding the guard: any regex that is a concatenation of literals and  .* , so  a.* ,  .*"@".* ,  "b".* etc is equivalent to a word-equation pattern with fresh unconstrained variables.

Regarding partial saturation: seq_monadic decomposes a conjunction of memberships into a disjunction over branches. Views taken from one branch are sound only inside that branch: emptiness of  equation ∧ views(branch_i)  refutes branch_i, not the equation. To use them you need one of

1. every branch is refuted unconditionally;
2. keeping it conditional and emitting a blocking clause that names the branch, instead of a conflict.

2 is not expressible in master currently because the branches are internal to seq_monadic. It only becomes available with the landing/split machinery, where a pulled branch is an actual case split. So "based on partial saturation" is a question about the C3 integration (rather than about the placement in the current PR).

PR #10599, opened 2026-08-21
Iterative iterator for seq_monadic

CEisenhofer:


src/ast/rewriter/seq_monadic.cpp
```diff
@@ -139,8 +138,9 @@ expr_ref_pair_vector const& seq_monadic::derivative_cofactors(expr* r) {
 }
 
 void seq_monadic::reset_ivl_cache() {
-    for (auto& kv : m_ivl_cache)
+    for (auto& kv : m_ivl_cache){
```
NikolajBjorner: skip brackets on single statements

src/ast/rewriter/seq_monadic.cpp
```diff
@@ -257,8 +258,9 @@ lbool seq_monadic::product_nonempty(expr* var, seq::view_vector const& comps, ex
     struct key_hash {
         size_t operator()(key const& k) const {
             uint64_t h = 1469598103934665603ull;
-            for (unsigned x : k)
+            for (unsigned x : k) {
```
NikolajBjorner: skip bracket

NikolajBjorner:
apart from lipstick, should be the right direction.

veanes (approved):
Checked that Behaviorally same as master, no regressions. 

Two remarks:

1. Drop the two brace hunks: line 203 and 1522 in seq_monadic.cpp.

2. Note in the PR that C3's  mon_state  holds a  seq_monadic::iterator by value, one per landing node, so it depends on exactly the property being removed. The C3 merge then has a known, scoped follow-up: either one-iterator-at-a-time discipline in the Nielsen search, or park m_stack  +  m_groups  in  mon_state  — which recovers the old ownership.

veanes:
On the 1476-file bench, PR head vs its exact merge-base: 1422 --> 1422 decided, 0 gained, 0 lost, 0 contradictions, +1.0% time (noise). The additional 6 (i.e., 1428), where 2 are just timeout noise but 4 are due to #10565 as a genuine win precisely at the word-equation and regex-membership interface: 10 s -->  0.01 s by a single refutation. (That is the same junction C3's  mk_mon_state  currently refuses to fire on (the  eq_vars  gate).)

veanes:
Note that this PR branched before #10565 landed but the shipping config is master with #10565 plus the current PR, and **both touch  seq_monadic.cpp**. That combination was untested. I've tested it now, merge is clean, no conflict, and the merged diff stat is identical to the PR's (the two changes never touch the same lines). Full corpus A/B, master vs master+PR: 1426--> 1425, 0 gained, 1 lost, 0 contradictions, (actually -2% time). The one loss is noise (right at 10s time boundary). 

So ready to merge in my opinion, as it enables seq_monadic intergation into C3.

PR #10622, opened 2026-08-22
Fix fp.rem significand truncation with a subnormal divisor

### LINKED ISSUE #10609: fp.rem drops significand bits with a subnormal divisor (shift amount widened more than the significand)
Similar to #10608.

In `mk_rem` the dividend significand and the shift amount are widened by different amounts:

```cpp
m_mpz_manager.set(remaining, max_exp_diff);          // 2^ebits - 3   (line 1180)
    a_sig_ext_l = m_bv_util.mk_zero_extend(rn, a_sig_ext_l);          // significand
...
m_mpz_manager.add(max_exp_diff, sbits, max_exp_diff_adj);             // (line 1197)
m_mpz_manager.sub(max_exp_diff_adj, ebits, max_exp_diff_adj);
m_mpz_manager.add(max_exp_diff_adj, 1, max_exp_diff_adj);
    edr_tmp = m_bv_util.mk_zero_extend(..., edr_tmp);                 // shift amount
...
shifted = m.mk_ite(exp_diff_is_neg, ... mk_bv_shl(a_sig_ext, lshift));
```

`exp_diff` is computed from the *normalized* exponents (`a_exp - a_lz` minus `b_exp - b_lz`). With a subnormal divisor, `b_exp` sits at the minimum while `b_lz` is non-zero, so `exp_diff` reaches `(2^ebits - 3) + (sbits - 1)`. The shift amount is wide enough to represent that, but the significand was only extended by `2^ebits - 3`, so `mk_bv_shl` pushes the high significand bits off the end and the remainder is computed from a truncated dividend.

binary16: the significand vector is `sbits + (2^5-3) + 3 = 43` bits and `exp_diff` can reach 39, so anything above 29 overflows.

## Reproducers

Found by exhaustively cross-checking `fpa2bv` against Z3's own native FP solver over all binary16 inputs: build the operation both ways over shared symbolic input bits, assert the two results differ, and let UNSAT prove equivalence over every input rather than a sample.

`x = #b1110100000101010` (-2132.0), `y = #b1000000000010101` (-1.2516975402832031e-06, subnormal):

```smt2
(set-option :produce-models true)
(declare-const x (_ FloatingPoint 5 11))
(declare-const y (_ FloatingPoint 5 11))
(declare-const xb (_ BitVec 16))
(declare-const yb (_ BitVec 16))
(declare-const rb (_ BitVec 16))
(assert (= x ((_ to_fp 5 11) xb)))
(assert (= y ((_ to_fp 5 11) yb)))
(assert (= xb #b1110100000101010))
(assert (= yb #b1000000000010101))
(assert (= rb ((_ fp.to_ieee_bv 16) (fp.rem x y))))
(check-sat-using (then fpa2bv simplify bit-blast smt))
(get-value (rb))
```

Z3 answers `#x8000` (-0.0). The exact remainder is `5/8388608` = 5.960464477539062e-07 = `#x000a`: `x/y` is 1703286881.52..., the nearest integer quotient is 1703286882, and `x - 1703286882*y` is exactly that. Here `exp_diff = 31`, two past the limit.

The -0.0 is the signature of the truncation. Once the high significand bits are shifted out, the dividend becomes an exact multiple of the divisor, so the remainder collapses to zero.

A binary32 witness of the same shape: `fp.rem(3.0e38, 7.0e-39)`, where the host's `std::remainder` is exact and Z3 is about 116 ulp out.

As with fp.fma, the explicit tactic is required, because `propagate-values` folds constant operands before `fpa2bv` runs.

## Fix

The significands need to be extended by the largest shift they can receive, `max_exp_diff + sbits - 1`. Note that `max_exp_diff_adj` (`max_exp_diff + sbits - ebits + 1`, already computed for the shift amount) is *not* enough: it falls short by `ebits - 2`, 3 bits at binary16 and 9 at binary64, so reusing it would fix the witness above but still truncate at larger exponent differences.

For what it is worth as a cross-check rather than as a patch, camada replaced this shift-and-divide encoding with square-and-multiply modular arithmetic over `2*sbits`-bit values, to avoid the huge bit-vectors the comment above `mk_rem` already notes (`mkFPRemImpl` in https://github.com/mikhailramalho/camada/blob/master/src/theories/camadafp.cpp). That encoding returns the correct remainder on every witness here and cross-checks equivalent to native FP over all binary16 inputs, which is what confirmed the expected values above.

mikhailramalho:
`fp.rem` in the fpa2bv converter could return a wrong remainder, often zero, when the divisor is subnormal. The dividend significand is extended by `max_exp_diff`, but the left shift applied to it is by `exp_diff`, which is computed from the normalized exponents and reaches `max_exp_diff + sbits - 1` once the divisor is subnormal. The shift then pushes the significand's high bits off the end and the remainder is computed from a truncated dividend.

This patch extends the significands by the largest shift they can receive:
  - Use `max_exp_diff + sbits - 1` rather than `max_exp_diff`.
  - Grow `max_exp_diff_adj` by the same amount, so `lshift`/`rshift` stay the width `bvshl` and `bvlshr` require.

Fixes #10609

### DIFF
--- a/src/ast/fpa/fpa2bv_converter.cpp
+++ b/src/ast/fpa/fpa2bv_converter.cpp
@@ -1176,8 +1176,16 @@ void fpa2bv_converter::mk_rem(sort * s, expr_ref & x, expr_ref & y, expr_ref & r
     // way to go in the long run (using the lazy bit-blaster helps on simple instances).
     expr_ref lshift(m), rshift(m), shifted(m), huge_rem(m), huge_div(m), huge_div_is_even(m);
     expr_ref a_sig_ext_l = a_sig, b_sig_ext_l = b_sig;
+
+    // exp_diff is computed from the normalized exponents, so a subnormal
+    // divisor (b_exp at the minimum with a non-zero b_lz) drives it up to
+    // max_exp_diff + sbits - 1.
+    scoped_mpz max_sig_ext(m_mpz_manager);
+    m_mpz_manager.add(max_exp_diff, sbits, max_sig_ext);
+    m_mpz_manager.sub(max_sig_ext, 1, max_sig_ext);
+
     scoped_mpz remaining(m_mpz_manager);
-    m_mpz_manager.set(remaining, max_exp_diff);
+    m_mpz_manager.set(remaining, max_sig_ext);
     while (m_mpz_manager.gt(remaining, INT32_MAX)) {
         SASSERT(false); // zero-extend ast's expect int params which would overflow.
         a_sig_ext_l = m_bv_util.mk_zero_extend(INT32_MAX, a_sig_ext_l);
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
see also the copilot extracted https://github.com/Z3Prover/z3/pull/10610

src/ast/fpa/fpa2bv_converter.cpp
```diff
@@ -1193,10 +1201,14 @@ void fpa2bv_converter::mk_rem(sort * s, expr_ref & x, expr_ref & y, expr_ref & r
...
     b_sig_ext = m_bv_util.mk_concat(b_sig_ext_l, m_bv_util.mk_numeral(0, 3));
 
+    // Sizes lshift/rshift to match a_sig_ext, which bvshl and bvlshr
+    // require, so it grows with the significand extension above.
     scoped_mpz max_exp_diff_adj(m_mpz_manager);
     m_mpz_manager.add(max_exp_diff, sbits, max_exp_diff_adj);
     m_mpz_manager.sub(max_exp_diff_adj, ebits, max_exp_diff_adj);
     m_mpz_manager.add(max_exp_diff_adj, 1, max_exp_diff_adj);
+    m_mpz_manager.add(max_exp_diff_adj, sbits, max_exp_diff_adj);
+    m_mpz_manager.sub(max_exp_diff_adj, 1, max_exp_diff_adj);
```
NikolajBjorner, line 1211: there is already a +1 two lines above.

NikolajBjorner:
Let's keep one AI PR in play. They are fairly similar and likely to introduce bugs.

Issue #9319, opened 2026-04-16
Consolidated invalid model bugs


lmaltsis:
```bash
$ z3 model_validate=true /tmp/bug.smt2
sat
(error "line 6 column 10: an invalid model was generated")

$ cat /tmp/bug.smt2
(declare-const x Real)
(assert (= x 0.0))
(assert (not (= (^ (* x 2.0) (/ 1.0 x))
                (^ (* 0.0 (/ 1.0 x))
                   (* x (/ 1.0 2.0))))))
(check-sat)
```

Commit: [daf2506b](https://github.com/Z3Prover/z3/commit/daf2506b6002149d531cb6c9d8ba976a040cd64)

lmaltsis:
```bash
$ z3 model_validate=true /tmp/bug.smt2
sat
(error "line 5 column 10: an invalid model was generated")

$ cat /tmp/bug.smt2
(declare-fun +zero () (_ FloatingPoint 3 2))
(declare-fun r () (_ FloatingPoint 3 2))
(assert (fp.eq +zero (fp #b0 #b000 #b1)))
(assert (fp.eq r (fp.fma RTN r +zero (_ +zero 3 2))))
(check-sat)
```

Commit: [daf2506b](https://github.com/Z3Prover/z3/commit/daf2506b6002149d531cb6c9d8ba976a040cd64)

fabiomadge:
Related, but actually produces the wrong result:
```
; fma(RNE, 1.0, 1.0, +0) should equal 1.0, so this should be unsat.                                                                     
; Z3 folds the fma to 2.0 and returns sat.                                                                               
(assert
  (not (=
    (fp.fma RNE
      (fp #b0 #b011 #b0)   ; x = 1.0  in FP(3,2)
      (fp #b0 #b011 #b0)   ; y = 1.0
      (fp #b0 #b000 #b0))  ; z = +0
    (fp #b0 #b011 #b0))))  ; != 1.0 ?
(check-sat)
```

lmaltsis:
```sh
$ cat /tmp/bug.smt2
(set-option :sat.euf true)
(set-option :tactic.default_tactic smt)
(assert (distinct false (= ((_ to_fp 11 53) RTZ #xFFFF) (fp #b1 #b10000000000 #x0000000000000))))
(check-sat)

$ z3 model_validate=true /tmp/bug.smt2
[...]

PR #10628, opened 2026-08-22
Fix FMA results for small floating-point precisions

NikolajBjorner:
## Summary

- fix `mpf_manager::fma` result extraction for formats with two- or three-bit significands
- remove the incorrect small-precision branch that scaled the significand in the wrong direction
- add unit coverage for exact multiply-add, addition, and cancellation at both affected precisions

For `(_ FloatingPoint 3 2)`, `fma(RNE, 1, 1, +0)` incorrectly folded to `2`; with a three-bit significand it folded to `4`. Both now produce `1`.

This fixes the constant-FMA sub-issue reported in #9319.

## Validation

- original `(_ FloatingPoint 3 2)` reproducer returns `unsat`
- exact FMA checks across significand widths 2 through 8
- `test-z3 /a`

### DIFF
--- a/src/util/mpf.cpp
+++ b/src/util/mpf.cpp
@@ -915,14 +915,8 @@ void mpf_manager::fma(mpf_rounding_mode rm, mpf const & x, mpf const & y, mpf co
 
         set(o, x.ebits, x.sbits, res.sign(), res.exponent(), mpz(0));
 
-        if (x.sbits >= 4) {
-            m_mpz_manager.machine_div_rem(res.significand(), m_powers2(x.sbits - 4 + 3), o.significand, sticky_rem);
-            renorm_sticky |= !m_mpz_manager.is_zero(sticky_rem);
-        }
-        else {
-            m_mpz_manager.mul2k(res.significand(), 4 - x.sbits + 3, o.significand);
-            o.exponent -= 4 - x.sbits + 3;
-        }
+        m_mpz_manager.machine_div_rem(res.significand(), m_powers2(x.sbits - 1), o.significand, sticky_rem);
+        renorm_sticky |= !m_mpz_manager.is_zero(sticky_rem);
 
         if (renorm_sticky && m_mpz_manager.is_even(o.significand))
             m_mpz_manager.inc(o.significand);
--- a/src/test/mpf.cpp
+++ b/src/test/mpf.cpp
@@ -111,9 +111,31 @@ static void test_to_sbv_mpq() {
     ENSURE(fm.mpq_manager().is_zero(r));
 }
 
+static void test_fma_small_precision() {
+    mpf_manager fm;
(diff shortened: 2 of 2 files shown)


NikolajBjorner:
> ## Pull request overview
> Fixes constant-folding behavior in `mpf_manager::fma` for very small floating-point significand widths by correcting how the post-addition significand is extracted/rounded, and adds regression coverage for the affected precisions. This change sits in Z3’s core floating-point constant evaluation path (used by the FPA rewriter), so it directly impacts correctness of simplification/model construction for FP terms.
> 
> **Changes:**
> 
> * Simplify/standardize result extraction in `mpf_manager::fma` by removing the small-precision special-case that scaled the significand incorrectly.
> * Ensure sticky-bit handling is applied consistently via `machine_div_rem` for the extraction step.
> * Add unit tests covering exact FMA and exact cancellation for `sbits ∈ {2,3}`.
> 
> ### Reviewed changes
> Copilot reviewed 2 out of 2 changed files in this pull request and generated no comments.
> 
> File	Description
> src/util/mpf.cpp	Removes the incorrect small-precision extraction branch and consistently extracts the result significand with sticky remainder tracking.
> src/test/mpf.cpp	Adds regression tests for FMA correctness at `sbits=2` and `sbits=3` (exact multiply-add and cancellation to +0).
> 💡 [Add a `code-review` agent skill](/Z3Prover/z3/new/master?filename=.github/skills/code-review/SKILL.md) or configure MCP servers for context-aware, tailored reviews. [Learn more in the docs.](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review#mcp-servers-and-agent-skills)

pretty lame review

src/util/mpf.cpp
```diff
@@ -915,14 +915,8 @@ void mpf_manager::fma(mpf_rounding_mode rm, mpf const & x, mpf const & y, mpf co
...
 
-        if (x.sbits >= 4) {
-            m_mpz_manager.machine_div_rem(res.significand(), m_powers2(x.sbits - 4 + 3), o.significand, sticky_rem);
-            renorm_sticky |= !m_mpz_manager.is_zero(sticky_rem);
-        }
-        else {
-            m_mpz_manager.mul2k(res.significand(), 4 - x.sbits + 3, o.significand);
-            o.exponent -= 4 - x.sbits + 3;
-        }
+        m_mpz_manager.machine_div_rem(res.significand(), m_powers2(x.sbits - 1), o.significand, sticky_rem);
```
wintersteiger, line 918: I think this loses the extra bits that the rounder needs to make the correct rounding decision. IIRC, the other branch just adds zeroes in the right places so that there are enough bits for the rounder. It's quite possible that something's off in that branch though, sbits < 4 is not very common.

Issue #10645, opened 2026-08-24
Performance regression in 5.1.0 (since 8th July)


rod-chapman:
Regression testing Z3 5.1.0 on our Cryptographic code reveals one performance regression on the attached SMT file.

At commit 965db5 (7th July), z3 reports "unsat" on this problem in about 1.25 seconds (AArch64/macOS)

At the next commit c56b2cb ("fix pattern inference to deal with binders properly", 8th July), performance on that problem regresses to "unsat" in 47 seconds.

At commit 82a0d42 ("Improve hash mixing..." PR 10120), z3 reports "timeout" after 5 minutes.
Releases 5.0.0 and 5.1.0 retain that behaviour and do not return a useful result with any tolerable timeout.

[t.smt2.gz](https://github.com/user-attachments/files/31372705/t.smt2.gz)

NikolajBjorner:
There are related and unrelated changes that affect this.
Requires some baby sitting to tweak coding agents to come up with a draft fix. They are probably a bit off.

NikolajBjorner:
Finding

- use pi.pull_nested_quantifiers=true to deal with instantiation guidance.

The legacy hash function was so weak that it would consistently hit a stable configuration. 
Pattern inference is back to the same it was before the regression.

See closed PR for details.

PR #10665, opened 2026-08-26
Add legacy nested-binder pattern inference mode

NikolajBjorner:
## Summary
- add an opt-in `pi.legacy_nested_binders` mode that reproduces pre-`c56b2cbaa` nested-binder candidate handling
- preserve current pattern inference behavior by default
- retain current master's `combine_hash` implementation and its low-bit dispersion regression coverage

## Regression analysis
Chronological replay from `965db51b5` identified `c56b2cbaa` and `aa6dddbdf` as transient pattern-inference regressions; `7c8c6a4df` later superseded those changes and restored the useful patterns. The legacy option permits direct A/B comparison with the earlier binder handling without changing the default strategy.

Further testing showed that the current hash exposes an order dependency between independently inferred patterns for nested binders. Pulling the nested quantifiers creates a coupled multi-pattern and solves all tested seeds in 3.5–5.6 seconds while retaining the improved hash dispersion. Those findings are documented in the PR discussion; this change does not revert the hash function.

## Validation
- 99/99 unit tests pass
- clean-build chained-hashtable tests pass with the current hash


NikolajBjorner:
Additional investigation shows that the hash change exposes an E-matching order dependency, but reverting `combine_hash` is not the only way to recover stable performance.

### Seed sensitivity

Release builds on issue #10645:

| Seed | Restored old hash | Current hash |
|---:|---:|---:|
| 0 | `unsat`, 3.16s | timeout, >30s |
| 1 | `unsat`, 4.27s | `unsat`, 5.33s |
| 2 | `unsat`, 4.42s | timeout, >30s |
| 3 | `unsat`, 3.41s | `unsat`, 2.62s |
| 10 | `unsat`, 3.36s | `unsat`, 4.64s |
| 100 | `unsat`, 4.47s | `unsat`, 4.75s |

For seed 0, the current hash stalls with only 388 quantifier instantiations, 151,653 decisions, and 244 conflicts. The old hash completes with 22,292 instantiations, 425,697 decisions, and 658 conflicts. The stalled run enters repeated final checks and array refinement instead of producing the useful E-matching instances.

### Pattern structure

Default inference assigns separate unary triggers to the nested binders:

```smt2
(concat #x00000000 outer-index)
(concat #x00000000 inner-index)
```

With `smt.pull_nested_quantifiers=true`, inference instead creates a multi-pattern coupling both indices:

```smt2
((concat #x00000000 outer-index)
 (concat #x00000000 inner-index))
```

This removes the observed hash-order instability while retaining the current, better-dispersed hash function:

| Seed | Old hash + pull | Current hash + pull |
|---:|---:|---:|
| 0 | 4.12s | 3.52s |
| 1 | 3.65s | 3.75s |
| 2 | 3.13s | 3.82s |
| 3 | 2.68s | 3.91s |
| 10 | 3.16s | 4.64s |
| 100 | 4.41s | 5.63s |

All twelve runs return `unsat`.

Increasing `qi.eager_threshold`, setting `qi.cost=0`, increasing `qi.lazy_threshold`, enabling either quick-checker mode, and disabling MBQI all leave the failing seed at the same 388 instances and still time out. The issue is therefore match production/connectivity, not delayed entries in `qi_queue`.

A potentially narrower long-term fix is to extend the existing nested-quantifier fallback in pattern inference: tentatively pull nested binders when inferred patterns bind each level independently, and prefer the pulled form when it creates a multi-pattern connecting variables across levels. This would retain the current hash's collision improvements without relying on hash-table discovery order for productive E-matching.

PR #10671, opened 2026-08-27
Remove obsolete maximize shared flag

NikolajBjorner:
## Summary
- remove the unused `has_shared` output parameter from `theory_opt::maximize`
- update LRA, legacy arithmetic, and both difference-logic implementations
- simplify `opt_solver` around unconditional model post-processing and bound validation

`opt_solver::maximize_objective` has unconditionally set `has_shared = true` since arithmetic optimization hints began requiring validation against theory combination. The values computed by theory implementations were therefore never observed.

The internal shared-variable tracking in legacy arithmetic `max_min` remains intact because it is also used to gate implied-bound propagation outside the optimizer interface.

## Testing
- `ninja -C build test-z3 shell`
- `build\test-z3.exe api`
- mixed integer/real optimization smoke test

### DIFF
--- a/src/opt/opt_solver.cpp
+++ b/src/opt/opt_solver.cpp
@@ -288,19 +288,14 @@ namespace opt {
        Return a predicate that blocks the current maximal value.
        
        The result of 'maximize' is post-processed. 
-       When maximization involves shared symbols the model produced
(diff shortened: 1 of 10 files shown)


levnach (approved):
It is a good change removing dead code.

PR #10672, opened 2026-08-27
fix heap overflow

### LINKED ISSUE #3508: `unsat` sequence benchmark, now can't be handled by z3
Again, about a month ago version old of z3 was able to answer this benchmark `unsat` pretty quickly. Right now it seems to be taking forever:

```
(set-logic ALL) 
(define-fun s3 () Int 0)
(define-fun s6 () Int 1)
(declare-fun s0 () (Seq Int))
(declare-fun s1 () Int)
(declare-fun s2 () Int)
(define-fun s4 () Bool (>= s1 s3))
(define-fun s5 () Int (seq.len s0))
(define-fun s7 () Int (- s5 s6))
(define-fun s8 () Bool (<= s1 s7))
(define-fun s9 () Bool (and s4 s8))
(define-fun s10 () Int (seq.nth s0 s1))
(define-fun s11 () Bool (= s2 s10))
(define-fun s12 () Bool (and s9 s11))
(define-fun s13 () Bool (not s12))
(define-fun s14 () (Seq Int) (seq.unit s2))
(define-fun s15 () Int (seq.indexof s0 s14 s3))
(define-fun s16 () Bool (<= s15 s1))
(define-fun s17 () Bool (or s13 s16))
(assert (not s17))
(check-sat)
```

NikolajBjorner:
Fixes #3508.

## Problem
Benchmarks combining `seq.nth`/`seq.extract` with `seq.indexof` on the same sequence, which used to solve quickly, started timing out. The axiomatization only connects `nth`/`extract` to `indexof` indirectly (through `at`/`pre`/`post`/`first`/`last` skolems and `tightest_prefix`), requiring many rounds of ground instantiation before the solver derives that an occurrence of `s` at position `i` bounds `indexof(t, s, 0) <= i`. This causes huge `:added-eqs`/`:seq-add-axiom` counts and non-termination.

## Fix
Add direct bridging axioms in `src/ast/rewriter/seq_axioms.cpp`:
- `nth_axiom`: `0 <= i < len(s) => indexof(s, unit(nth(s,i)), 0) <= i`
- `extract_axiom` (both the general path and the `small_segment_axiom` numeral-length shortcut): `0 <= i, i + l <= len(s), l >= 0 => indexof(s, extract(s,i,l), 0) <= i`

These let the solver directly discharge the `indexof` upper bound instead of deriving it through many rounds of skolemization.

## Testing
- Confirmed the reported repro (Int sequences, `seq.nth` + `seq.indexof`) now returns `unsat` quickly instead of timing out.
- Full unit test suite (`test-z3 /a`): 99 passed, 0 failed.
- Note: the BitVector/`seq.extract` repro from the issue's comment still times out; it appears to hit a related but deeper interaction (confirmed the extract-based bridging axiom alone isn't sufficient for that case) and may need follow-up.

### DIFF
--- a/src/ast/rewriter/seq_axioms.cpp
+++ b/src/ast/rewriter/seq_axioms.cpp
@@ -316,7 +316,7 @@ namespace seq {
             expr_ref zero(a.mk_int(0), m);
             expr_ref i_ge_0 = mk_ge(i, 0);
             expr_ref i_l_le_len_s = mk_le(mk_sub(mk_sub(mk_len(s), i), l), 0);
-            expr_ref idx(seq.str.mk_index(s, e, zero), m);
+            expr_ref idx(seq.str.mk_index(e, s, zero), m);
             add_clause(~i_ge_0, ~i_l_le_len_s, mk_le(mk_sub(idx, i), 0));
             return true;
         }
@@ -650,7 +650,7 @@ namespace seq {
             nth = es.back();
             es.push_back(m_sk.mk_tail(s, i));
             add_clause(~i_ge_0, i_ge_len_s, mk_seq_eq(s, seq.str.mk_concat(es, e->get_sort())));
-            add_clause(~i_ge_0, i_ge_len_s, mk_seq_eq(nth, e));                
+            add_clause(~i_ge_0, i_ge_len_s, mk_seq_eq(nth, e));
         }
         else {
             expr_ref x =     m_sk.mk_pre(s, i);
@@ -659,6 +659,15 @@ namespace seq {
             expr_ref len_x = mk_len(x);
             add_clause(~i_ge_0, i_ge_len_s, mk_seq_eq(s, xey));
             add_clause(~i_ge_0, i_ge_len_s, mk_eq(i, len_x));
+
+            // 0 <= i < len(s) => indexof(s, e, 0) <= i
+            // e = at(s, i) occurs in s at position i, so its first occurrence
+            // is at most i. seq.extract(s, i, 1) is rewritten to at(s, i) by
+            // the rewriter, so this bridges seq.extract/seq.at with
+            // seq.indexof (issue #3508), mirroring nth_axiom's bridging
+            // clause below.
+            expr_ref idx(seq.str.mk_index(s, e, zero), m);
+            add_clause(~i_ge_0, i_ge_len_s, mk_le(mk_sub(idx, i), 0));
         }
 
         add_clause(i_ge_0, mk_eq(e, emp));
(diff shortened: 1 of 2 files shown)


NikolajBjorner:
Follow-up commit addresses two issues found while extending this fix to the BitVector \seq.extract\+\seq.indexof\ repro from #3508 (comment):

1. **Fixed the comment repro**: length-1 \seq.extract(s,i,1)\ is rewritten to \seq.at(s,i)\ by the rewriter before axiomatization, bypassing the \xtract_axiom\/\small_segment_axiom\ bridging clauses. Added the equivalent bridging clause to \t_axiom\'s symbolic-index branch (restricted there to avoid an instantiation blowup observed in generic/symbolic \t()\ usage, e.g. seq_monadic-style benchmarks).
2. **Fixed a haystack/needle swap bug** in \small_segment_axiom\'s bridging clause (\mk_index(s, e, zero)\ should be \mk_index(e, s, zero)\, since \\ is the base sequence and \s\ is the extract result inside this function's parameter naming).
3. **Fixed a real, unrelated regression this PR exposed**: \	est-z3 get_consequences\ crashed with a heap-buffer-overflow (found via AddressSanitizer) in \inc_sat_solver::get_assertion\, which didn't call \convert_internalized()\ before indexing (unlike \get_num_assertions()\), causing \simplifier_solver::translate\ to read past the end of \m_internalized_fmls\ once axiom changes altered internalization timing. Fixed by adding the missing \convert_internalized()\ call.

Verified: all issue #3508 repros (original + BitVector comment repro) now solve quickly, full \	est-z3 /a\ suite passes cleanly (99/99) across multiple runs, and ASan confirms no more heap-buffer-overflow.

NikolajBjorner:
CI caught broad regressions from the bridging axioms in this PR (run: https://github.com/Z3Prover/z3/actions/runs/33045166552/job/98427369511). Benchmarks unrelated to #3508 (e.g. \3226.smt2\, \3889.smt2\, \2447.smt2\, \
e.smt2\, \string-concat.smt2\, \issue-1677.smt2\) went from solving in well under a second to timing out (>15s).

Tried narrowing the bridging clauses (numeral-only guards on \i\, restricting which branch adds the clause) but every variant I tried still regressed at least one of the CI benchmarks. Reverted all three bridging-axiom additions (\
th_axiom\, \xtract_axiom\'s general path, \small_segment_axiom\, \t_axiom\) to restore master's behavior on the CI suite.

Kept the \inc_sat_solver::get_assertion\ fix (independent, correct bug found via ASan — missing \convert_internalized()\ call causing a heap-buffer-overflow / \get_consequences\ crash).

Verified: all 6 previously-timing-out CI benchmarks now solve in <0.2s (matching master), \get_consequences\ still passes reliably, and the full \	est-z3 /a\ suite passes 99/99.

Issue #3508 itself remains open — a more targeted fix (e.g. lazily firing the bridging axiom only when \seq.indexof\ is already asserted/relevant, rather than eagerly for every \
th\/\xtract\/\t\ term) would be needed to fix it without these regressions.

PR #10701, opened 2026-08-30
Introduce generalized Parikh engine

CEisenhofer:


src/ast/rewriter/seq_parikh.h
```diff
@@ -0,0 +1,120 @@
...
+    Clemens Eisenhofer 2026
+
+--*/
+#pragma once
+
+#include "ast/arith_decl_plugin.h"
+#include "ast/seq_decl_plugin.h"
+#include "util/obj_hashtable.h"
+
+class seq_parikh {
```
NikolajBjorner: Use namespace seq and call the class parikh

NikolajBjorner:
@copilot implement review ask to use seq namespace and calls the class parikh

NikolajBjorner:
@copilot resolve the merge conflicts in this pull request

NikolajBjorner:
@copilot

change this:

// numerals are handed around as raw pointers, so keep them alive for the whole encoding
expr* parikh::num(int i) {
    expr* e = m_autil.mk_int(i);
    m_pinned.push_back(e);
    return e;
}
have it return an expr_ref.
Change the places where it is used to first create the expr_ref then pass as arguments to the term constructors.
Examples

auto one = num(1);
auto zero = num(0);
auto e = mk_app("f", 2, one.get(), zero.get(), seq_sort);

instead of

auto e = mk_app("f", 2, num(1), num(0), seq_sort);

it takes care of C/C++ semantics of evaluating nested functions.

NikolajBjorner:
@copilot 
do we need this:

```
expr_ref parikh::conj(expr_ref_vector const& args) {
    expr_ref_vector as(m);
    for (expr* e : args) {
        if (m.is_false(e))
            return expr_ref(m.mk_false(), m);
        if (!m.is_true(e))
            as.push_back(e);
    }
    return mk_and(as);
}
```
Doesn't mk_and in ast_util take care of simplifiation?

NikolajBjorner:
@copilot

can we use rationals and power operator?

unsigned parikh::num_grams(unsigned level) const {
    unsigned n = 1;
    for (unsigned i = 0; i < level; ++i) {
        n *= m_p;
    }
    return n;
}
This is m_p^level
what if unsigned overflows?

NikolajBjorner:
@copilot

bool parikh::has_char(expr_ref_vector const& side) const {
    expr* c = nullptr;
    for (expr* e : side) {
        if (m_util.str.is_unit(e, c) && m_util.is_const_char(c))
            return true;
    }
    return false;
}

can you use any_of comprehension instead?

NikolajBjorner:
@copilot - when does m_chars get reset? Should it get reset when invoking

bool parikh::operator()(expr_ref_vector const& l, expr_ref_vector const& r,
                              expr_ref_vector& defs, expr_ref_vector& eqs) {
                              
                              ?
                              
                              Does m_chars refer to character within one equation or in the accumulation of equations that were processed?

NikolajBjorner:
@copilot - have you considered to use uint_set for the m_chars datatype?

NikolajBjorner:
@copilot - instead of uint_set you can use tracked_uint set or other abstractions available in uint_set.h. Some of these provide indices into positions based on when they were inserted.

src/ast/rewriter/seq_parikh.cpp
```diff
@@ -0,0 +1,527 @@
...
+    m_defined.insert(key);
+    m_pinned.push_back(key);
+    return true;
+}
+
+rational parikh::num_grams(unsigned level) const {
+    return power(rational(m_p), level);
+}
+
+unsigned parikh::char_index(unsigned ch) const {
```
NikolajBjorner: Can this be simplified using tracked 8ndex set function?

NikolajBjorner:
@copilot - add math comments with references to paper.
- insread of bound at is_unsigned set bound to m_config.max_size which is set at 25.

NikolajBjorner:
@copilot revert m_chars back to unsigned_vector

NikolajBjorner:
@copilot
Simplify this code:
    // a modulus is redundant as soon as a proper multiple of it is in range as well
    unsigned_vector moduli;
    for (unsigned n = 1; n <= m_config.m_n; ++n) {
        if (2 * n > m_config.m_n)
            moduli.push_back(n);
    }
    
    start with position where you know 2*n > m_config.m_n instead of 1.

PR #10718, opened 2026-09-01
Prove unboundedness of real objectives over nonlinear constraints with one quantified nlsat query

levnach:
When maximizing a real objective under nonlinear constraints, the optimizer's climb loop can improve the model forever without ever refuting an upper bound, so it never terminates on its own. Example: maximize x under x*y = 1 and y > 0 — every bound x >= c stays satisfiable, so the climb runs until the timeout. The same happens on nonconvex feasible sets such as x^2 >= 4.

This PR decides the question once instead: after a streak of improving rounds with no refuted bound, it asks the quantified query "for every c there is a point of the feasible set with objective > c" and solves it with the nlsat-based quantified solver (qe/nlqsat). On a positive answer the objective is reported as oo (or -oo for minimize, which goes through negation), keeping the best model found as a witness. Both examples above now finish in milliseconds instead of timing out.

Changes:
- `opt/opt_nlsat`: new `nlsat_opt::prove_unbounded` builds the query from the engine's preprocessed goal (purification removes division and other non-polynomial arithmetic that nlqsat rejects) and runs nlqsat on it. The `maximize` entry point also uses it after a few rounds of unbounded progress when no upper hint was given.
- `opt/optsmt`: `prove_unbounded_above` triggers the check from the climb loop after 8 rounds of improving models with no refuted arithmetic bound; on success it commits oo with the current model. Gated by `opt.optsmt_nlsat` (default true).
- `nlsat/nlsat_solver`: the cached maximal-variable flags are replaced by an on-demand query `max_var_sup(sup, attained)` that reads the infeasible-interval set of the maximal variable; `prove_supremum` now takes the supremum as a parameter.
- `test/opt_nlsat.cpp`: unit tests for the new query — proven unbounded (both the polynomial form x*y = 1 and the division form x = 1/y), proven bounded, and outside the supported fragment — plus unbounded detection inside `maximize`.

End-to-end SMT-LIB regressions accompany this in a z3test PR.


NikolajBjorner:
is prove_unbounded best solved as a quantified formula or is there a short-cut by directly producing a certificate for unboundedness?
qsat itself reduces search to a sequence of projections and cores.

levnach:
> is prove_unbounded best solved as a quantified formula or is there a short-cut by directly producing a certificate for unboundedness? qsat itself reduces search to a sequence of projections and cores.

I, together with AI, do not see a shortcut that would not repeat the loop of nlqsat. It seems to reuse nlqsat is the cheapest version here. When we prove the unboundedness with nlqsat we get an infinite to the right cell - the certificate.

NikolajBjorner:
so nlqsat is the penultimate solver for this?
Flattering, but I somewhat doubt given that there should be also some completely different methods possible that are not based on the projection/core loop. For example, you can search for a certificate over a template space of positiv-stellensatz proofs? https://arxiv.org/abs/2412.16226

levnach:
For this PR,  nlqsat  is the safer decision engine: after purification, the unboundedness check is a quantified NRA sentence, and  nlqsat  is *complete* on supported inputs. It lazily builds a CAD. From the other side, the paper's template method is only semi-complete and cannot prove  maximize x  subject to  x*y = 1  and  y > 0 : polynomial Skolem functions would require  x(c)*y(c) = 1  identically, forcing both to be constants, which is incompatible with  x(c) > c . It could be an optional fast path, but not the completeness fallback.

NikolajBjorner:
1. it is true that it would be for a "fast path".
2. completeness: you solve for both F and ~F using the same r.e. approach. Templates are enumerable.
3. fine to merge

PR #10728, opened 2026-09-03
[CMake] Add CI coverage for co-installed static and shared packages

alexreinking:
A common pitfall with CMake is that declaring `cmake_minimum_required` does not guarantee compatibility with the stated version. The reason is that it's a _backwards-compatibility_ setting, not a _forwards-compatibility_ one. For instance, you can set `cmake_minimum_required(VERSION 3.16)` and then use a generator expression introduced in a newer version, and it will silently break compatibility with 3.16. Newer versions can run builds authored with older versions, but not vice versa. Therefore, one must test the build with the stated version in CI to ensure compatibility.

This PR adds a workflow to test package building at the promised minimum (currently 3.16) and consuming at the consumer minimum (via `find_package`, currently 3.5). It also tests the shared and static package variants to ensure the loading logic is correct.

I had wanted to add this to #10717 but it was merged too quickly 🙂

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -459,6 +459,18 @@ jobs:
         with:
           cmake-version: ${{ matrix.consumer_cmake }}
 
+      # jwlawson's old CMake/ctest binaries are linked against libidn.so.11,
+      # which Ubuntu dropped after focal; pull just that runtime lib from the
+      # focal archive since noble no longer ships it.
+      - name: Install libidn11 for old ctest binary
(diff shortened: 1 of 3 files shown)


alexreinking:
Note: I think I might need to backport some changes from my branch to 3.16 to make this test work.

alexreinking:
I'm running this on my local fork. It's too slow right now, so I'm thinking of ways to improve the situation. I'll mark "Ready for review soon", hopefully.

PR #10733, opened 2026-09-04
[CMake] Use z3::libz3 as the preferred way to consume Z3

alexreinking:
In #10717 we enhanced the CMake package to support multiple build configurations, so that Z3 could be installed and consumed as either a static or shared package, or both side by side. This PR dog-foods that work: the Python bindings and the example projects now consume Z3 through the same installed CMake package that external users rely on, rather than through separate in-tree plumbing.

To make that possible, we introduce a `z3::libz3` ALIAS that mirrors the installed target in-tree. In-tree consumers (like the examples and the Python bindings) and out-of-tree consumers of the installed package now refer to Z3 the same way, so now there's one blessed interface. The examples and package-consumer test were updated accordingly, both to exercise this path and to serve as a model for how downstream projects should consume Z3 going forward.

Finally, we add test coverage for the reworked Python build, including a CI job that installs a shared build of libz3 and then builds the Python bindings against that installed package, confirming the Python-only build path works end to end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -454,6 +454,15 @@ jobs:
           cmake --build build --target libz3 shell
           cmake --install build --prefix "$PWD/install"
 
+      - name: Build Python bindings from installed shared package
+        run: |
+          cmake -S . -B build-python-only -G Ninja \
+            -DZ3_BUILD_LIBZ3_CORE=OFF \
+            -DZ3_BUILD_PYTHON_BINDINGS=ON \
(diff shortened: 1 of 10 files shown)


NikolajBjorner:
build failure is benign to this change: tests were updated for a new feature committed independently

alexreinking:
> build failure is benign to this change: tests were updated for a new feature committed independently

I was looking at that and wondering how I could have broken it! No library code changes here, and that workflow doesn't even build with CMake!

PR #10740, opened 2026-09-05
[CMake] Miscellaneous cleanup

alexreinking:
These are the last of the changes I have that are compatible with 3.16. There are a few things here:

1. A small correction to the resolution chain from #10717 plus a test. The package is now more permissive when only `BUILD_SHARED_LIBS` is set.
2. Use the `SHELL:` argument form to `target_link_options` in emscripten to avoid dropping `-s` flags.
3. Use CMake's native IPO support rather than hand-rolled detection.

After this, I will open a large PR that bumps the minimum version to 3.30 and reworks the `z3_add_component` function. It will remove 300+ lines of CMake code.


NikolajBjorner:
Some say "links in bio" other say "build systems masochist" in bio.

PR #10741, opened 2026-09-05
[CMake] Rework the CMake component graph

alexreinking:
This PR substantially modernizes Z3's CMake build, replacing a significant amount of hand-rolled dependency bookkeeping with CMake's native target-graph features.

The main benefits are:

* **Less custom build logic.** The existing component system in `cmake/z3_add_component.cmake` manually walks and records the component-dependency graph. This PR moves that information onto CMake targets themselves, eliminating much of the `Z3_COMPONENT_*` / `Z3_DEPENDENT_*` variable machinery.
* **A more accurate target graph.** Component dependencies, generated registration headers, and static-component linking are now modeled directly in CMake rather than reconstructed separately.
* **More natural source-tree organization.** `add_subdirectory()` calls no longer need to be kept in a global topological order or reach deeply into the source tree. Forward component dependencies are allowed, so each directory can add its own children in a natural, alphabetical order.
* **Better dependency propagation.** External dependencies such as GMP, Threads, and `atomic` are available to the components that actually need them. In particular, non-system GMP installations now work correctly because component targets can see GMP's include path.
* **Cleaner package consumption.** Build-only compiler/linker policy is separated from the dependencies exported to static-package consumers, and Z3 behaves more cleanly when included via `add_subdirectory()` or `FetchContent`.
* **Less code with the same observable package behavior.** The PR changes 34 files with +566/-902 lines, for a net reduction of roughly 336 lines, while preserving the existing targets, public headers, install layout, and registration behavior.

The main enabler is raising the minimum CMake version from 3.16 to **3.30**.

CMake 3.30 introduced **custom transitive properties**, which let targets propagate the registered module headers, tactic headers, and memory headers needed for generated registration code. This replaces the custom graph-walking logic previously used to accumulate those headers.

As a result, the three functions `z3_add_install_tactic_rule`, `z3_add_memory_initializer_rule`, and `z3_add_gparams_register_modules_rule` are replaced by a single, substantially simpler function, `z3_generate_registration`, which generates the required headers from properties propagated through the target graph.

Newer CMake also lets us model chains of `OBJECT` libraries correctly. Historically, only `OBJECT` libraries named directly in `target_link_libraries()` contributed their object files to the final link, which made transitive static-component dependencies awkward to represent. Using `INTERFACE_LINK_LIBRARIES_DIRECT`, those chains can now be expressed directly in the target graph.

Every Z3 component also links to a new `INTERFACE` target, `z3_common`. This target owns the external library dependencies that consumers of a component may need, including GMP, Threads, `atomic`, and similar dependencies. Previously, these dependencies were attached only to the final `libz3` target; among other things, that meant individual components could not see the include path for a non-system GMP installation.

`z3_common` is exported for static-package consumers, but Z3's own build policy is not. Internal compiler and linker flags now live on a separate `z3_internal_options` target used throughout the build. A possible follow-up is to replace `z3_internal_options` with toolchain files and/or CMake presets, but that would be a more disruptive change and is intentionally left out of this PR.

Once CMake itself has an accurate model of the component graph, several pieces of surrounding machinery become unnecessary. We no longer need a topologically sorted list of `add_subdirectory()` calls or the previous requirement that child components be declared before their dependents. Each directory now adds its own subdirectories, in alphabetical order. Public headers are likewise installed through CMake file sets instead of hand-written `install(FILES ...)` rules.

The CMake version bump also enables a few smaller cleanups:

* use `GENERATOR_IS_MULTI_CONFIG` instead of inferring that from whether `CMAKE_CONFIGURATION_TYPES` is defined;
* use `PROJECT_IS_TOP_LEVEL` guards so Z3 behaves correctly when consumed as a subdirectory or through `FetchContent`;
* use `block()` instead of manually saving and restoring state around the Python `find_package()` call.

One unrelated fix is included alongside these changes: the .NET example now stages its `csproj` and `Program.cs` into the build tree with `add_custom_command()` rather than `configure_file(..., COPYONLY)`, so the `dotnet` build target correctly notices source changes.

Overall, the PR moves responsibility for dependency propagation and graph structure out of Z3's custom CMake code and back into CMake itself. The result is a smaller build system with fewer ordering constraints and less duplicated state, while keeping the package's externally visible behavior unchanged.


alexreinking:
One thing worth calling out when weighing the 3.16-to-3.30 minimum-version bump is that this only raises the CMake requirement for building Z3 itself. The minimum version for consuming an installed Z3 package via `find_package(Z3)` remains CMake 3.5+.

For users who do build Z3 from source, obtaining CMake 3.30+ is also relatively low-friction on the platforms downstreams are likely to be using.

In many environments, it is just a package-manager command:

* **Windows:** `winget install -e --id Kitware.CMake` (latest)
* **macOS:** `brew install cmake` (latest)
* **Linux:** distribution packages can lag substantially — Debian stable and older Ubuntu/RHEL releases, for example, ship CMake versions below 3.30. However, Kitware also publishes official `cmake` wheels on PyPI with broad manylinux support, so users are not dependent on the system package:
  * `pip install cmake`
  * `pipx install cmake`
  * `uv tool install cmake`
* **FreeBSD:** `pkg install cmake` (currently 3.31)

The latter two are particularly convenient when an isolated CMake CLI is preferable to modifying the system Python environment. In practice, the higher version requirement should mostly affect how CMake is obtained on older Linux distributions rather than whether a sufficiently recent CMake is available at all.

NikolajBjorner:
> For users who do build Z3 from source, obtaining CMake 3.30+ is also relatively low-friction on the platforms downstreams are likely to be using.

I hope so. In the past I see users file issues about cmake invoked from python installation because their platform wasn't supported. setup.py will compile z3 in these cases. Some of these gaps are filled with consistent ARM support and down to MacOS 13.

Nevertheless, it isn't an objective for z3 to support PDP-11 or other downstream architectures.

alexreinking:
> In the past I see users file issues about cmake invoked from python installation because their platform wasn't supported

The matrix of supported architectures is quite impressive these days: there are prebuilt binaries for Windows (x86, x64, ARM64), macOS (universal), and Linux with glibc or musl (x86-64, s390x, riscv64, ppc64le, i686, ARMv7l, ARM64).

Anyone downstream who can't use these will probably be quite accustomed to building their own tools 🙂

NikolajBjorner:
should I (or we or the first one to have the need) fix up scripts to publish for risc64 and the likes?

alexreinking:
@NikolajBjorner — I just looked through the GitHub Actions workflows and found that `cross-build.yml` and `fstar-master-build.yml‎` needed a version bump (they aren't run on PRs). I think everything else should be set. Did you have something else in mind?

NikolajBjorner:
> @NikolajBjorner — I just looked through the GitHub Actions workflows and found that `cross-build.yml` and `fstar-master-build.yml‎` needed a version bump (they aren't run on PRs). I think everything else should be set. Did you have something else in mind?

What i have in mind is to add workflow triggers and conditions on releases that trigger these workflows and publish to pypi automatically. There is now a workflow that publishes wasm/pyodide. It is not part of release.yml so is managed independently. This template could be applied elsewhere if it makes sense. riscv is maybe (still) emerging.

NikolajBjorner:
Well, after another second of thinking about this: let's say if someone is brave enough to use riscv today (this is verifiably a non-empty set) and want pypi convenience, DMs are open for pull requests.

alexreinking:
> Well, after another second of thinking about this: let's say if someone is brave enough to use riscv today (this is verifiably a non-empty set) and want pypi convenience, DMs are open for pull requests.

Sure. Happy to defer that work until someone demonstrates a need for it.

src/ast/CMakeLists.txt
```diff
@@ -58,3 +58,15 @@ z3_add_component(ast
   PYG_FILES
     pp_params.pyg
 )
+
```
NikolajBjorner, line 61: we will soon need add_subdirectory(seq)
There is a separate PR for disentangling growing seq functionality.

alexreinking: Cool, that should just work!

alexreinking: Did you want to land this PR or that one first?

alexreinking:
The Windows failure is real. I will try to reproduce locally.

src/ast/CMakeLists.txt
NikolajBjorner: merged & updated this PR

NikolajBjorner:
I assigned to copilot too

alexreinking:
> I assigned to copilot too

I think I got to it first 🙂

alexreinking:
Once this lands, I should really update `README-cmake.md` ...

PR #10753, opened 2026-09-07
[CMake] Refresh build documentation and invocation style

alexreinking:
This PR updates Z3’s documentation and automation to reflect the recently modernized CMake build.

The CMake build is now Z3’s most capable build system, including support for all language bindings, so the main README presents it first and recommends it for general use.

### Documentation

`README-CMake.md` has been rewritten as a downstream-facing usage guide, covering configuration, building, installation, build types, generators, static and shared libraries, language bindings, and package consumption. It documents how to install shared and static libraries into the same prefix by reconfiguring a build tree in place, and recommends `Z3_ROOT`, as either a CMake cache variable or an environment variable, for locating installed Z3 packages. It also removes the outdated claim that CMake cannot build the OCaml bindings.

### Build and automation updates

`mkdir`, `cd`, and `cmake ..` sequences have been replaced with explicit source and build directories:

```sh
cmake -S . -B build
cmake --build build
```

Explicit generators are now specified up front, e.g. `cmake -G Ninja -S . -B build`, and generator-independent `cmake --build` and `cmake --install` commands are used where appropriate. The standard `BUILD_SHARED_LIBS` option replaces the legacy `Z3_BUILD_LIBZ3_SHARED` spelling, and explicit parallelism has been removed from Ninja builds, since Ninja already uses the available cores by default.

Repeated compiler and flag environment variables have been moved into appropriately scoped GitHub Actions `env:` blocks, and the coverage workflow now uses `CMAKE_BUILD_TYPE` directly as an environment variable. The cross-build workflow now explicitly selects the Ninja generator it already installs; this also fixes a pre-existing failure introduced when the workflow stopped installing Ubuntu’s `cmake` package, and consequently stopped acquiring `make` incidentally (oops). Python packaging and distribution helpers have also been updated to use explicit CMake source and build directories.

I have run all modified workflows on my fork and can confirm they succeed there.


build_z3.bat
```diff
@@ -1,26 +1,15 @@
 @echo off
 REM Z3 Build Script
```
NikolajBjorner, line 2: this file looks like AI junk left behind.

NikolajBjorner: will remove separately, not sure how to use this UI to remove files from PR

NikolajBjorner:
killing the python build has been a dream for ages. Should we :-).

alexreinking:
> killing the python build has been a dream for ages. Should we :-).

Personally, I think we can. Having multiple build systems is always a hassle.

alexreinking:
> killing the python build has been a dream for ages. Should we :-).

@NikolajBjorner Would you like me to work on this? I took a closer look at the generated registration files and I think they could be effectively replaced by X-macros and the Google Test/LLVM-style static registration trick. Then we could remove quite a few things:

1. The whole Makefile-generation system in Python
2. The `pyg2hpp` system, including mk_gparams_register_modules_cpp.py, mk_mem_initializer_cpp.py, and mk_install_tactic_cpp.py.
3. The custom transitive properties in CMake; this should let us relax the minimum requirement to 3.26. Good that we now have testing in place to confirm this 🙂

Also, I was wondering if you'd be willing to add me to the Z3Prover org so that I can work from branches on this (the upstream) repository? Mostly, I am interested in being able to [_stack_ pull requests](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) so that I can open chains of dependent PRs without needing to wait for a review/merge cycle each time. It would also let CI run on my PRs without your intervention.

alexreinking:
@NikolajBjorner — I just wanted to ping you about my question above.

PR #10765, opened 2026-09-09
opt: reuse nlsat state across Pareto checks

levnach:
## Summary

Follow-up to Z3Prover/z3#10736. Reuse nlsat state across exact Pareto candidate and dominance checks instead of starting each check from scratch, without carrying temporary dominance constraints into later front points.

- Add a persistent nlsat-backed solver with stable expression mappings and incremental goal translation. Tag scoped assertions and their learned consequences so popping a scope removes dependent clauses while preserving independent lemmas.
- Enable reuse for supported polynomial Real problems with `opt.pareto_nlsat_reuse=true`. Preserve the fresh exact solver with `false`, and the general SMT fallback for Optimize assumptions and unsupported fragments.
- Cap independent learned clauses retained between climbs with `opt.pareto_nlsat_max_lemmas` (default 128, preferring shorter clauses). Setting it to 0 drops learned clauses without rebuilding the solver.
- Make conflict-explanation and conflict-resolution mark cleanup safe on cancellation. This prevents stale duplicate marks from suppressing required literals on a direct retry or after scope retraction. Document partial-result handling on exceptions.

## Validation

- Built `z3` and `test-z3` in Release and Debug; `test-z3 /j:2 nlsat opt_nlsat opt_pareto 13` passes in both.
- Added regressions that cancel after an explanation literal has been emitted, compare recovered explanations with uninterrupted results, and cover retries with and without retraction. A solver-level resource-budget sweep covers `unknown -> pop -> retry`. The pre-fix implementation reproduced a debug assertion and missing explanation literals in Release.
- Both existing algebraic Pareto golden-output regressions pass unchanged in Release and Debug with model validation.
- Verified exactly the same 91 unique expected grid triples, followed by UNSAT, with fresh solving, default lemma reuse, and a zero retained-lemma cap.

The earlier companion regression updates in Z3Prover/z3test#68 are already merged; no new z3test changes are required. The continuous-front benchmark remains a known baseline timeout, also observed with reuse disabled.

### DIFF
--- a/src/nlsat/nlsat_explain.cpp
+++ b/src/nlsat/nlsat_explain.cpp
@@ -24,6 +24,7 @@ Revision History:
 #include "nlsat/nlsat_common.h"
 #include "util/ref_buffer.h"
 #include "util/mpq.h"
+#include "util/util.h"
 
 namespace nlsat {
 
@@ -129,6 +130,15 @@ namespace nlsat {
             for (literal lit : *m_result) 
                 m_already_added_literal[lit.index()] = false;
             SASSERT(check_already_added());
+        }
+
+        // Clear duplicate marks before cancellation lets the caller discard the
+        // result buffer; otherwise a retry can silently omit required literals.
+        void reset_result() {
+            if (m_result) {
+                reset_already_added();
+                m_result = nullptr;
+            }
         }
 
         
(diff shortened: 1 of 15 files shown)


levnach:
@NikolajBjorner , @shuvendu-lahiri : the PR it failed because z3prover-ci-bot could not start a workflow in bench: this is . I created a PR at https://github.com/microsoft/github-operations/pull/1796 to fix it.

PR #10801, opened 2026-09-14
Fix FP overflow with large significands

NikolajBjorner:
## Summary

- compute integer-to-FP infinity thresholds without narrowing `sbits` to signed `int`
- use the exact integer cutoff when significand precision exceeds the maximum exponent
- detect constant rational overflow before large-precision scaling can wrap
- add constant and symbolic regressions for the reproducer from Z3Prover/bench#3893

This addresses the surviving audit finding in Z3Prover/bench#3890 for the optimization introduced by #9038.

## NFV audit evidence

**Dashboard record:** [NFV Formal Audit Dashboard](https://z3prover.github.io/coz3/nfv_dashboard.html), audited September 14, 2026. The #9038 audit evaluated 6 propositions: 2 proved, 3 abstained, and 1 refuted with 1 counterexample surviving adjudication. The full finding is published in Z3Prover/bench#3890.

The NFV report identified the property `integer_conversion_infinity_thresholds` in `fpa_rewriter::mk_is_inf_of_int`. The audited contract assumes a valid floating-point format, an arbitrary integer `x`, a concrete rounding mode, the exact maximum finite value `M`, and the ULP-derived half-step `H`. For nearest rounding, the generated predicate is intended to classify values with `x >= M + H` or `x <= -(M + H)` as infinite.

Dafny proved the negated translated postcondition for:

- `rm = RNE`
- `ebits = 2`
- `sbits = 0xffffffff`
- `x = 4`

The report traces the mismatch to converting the unsigned `sbits` value to signed 32-bit `int`. The cast yields `-1`, so the implementation derives `ulp_exp = 3` and `half_ulp = 4`, classifying `4` as finite. The exact nearest-rounding boundary is below `4`, making the integer cutoff `4`, so the conversion must produce infinity.

The report initially noted that reachability through Z3 still required human confirmation. Z3Prover/bench#3893 supplies that confirmation: the public SMT-LIB input returns `sat` before this fix despite its expected `unsat` status. The same issue also reaches constant folding before `mk_is_inf_of_int`, where unsigned precision scaling wraps; this PR fixes both paths.

The NFV proof relies on its stated `MaxFinite` and `MaxExp` declarations, while the checked-in SMT-LIB regression validates the behavior end to end in Z3.

## Validation

- `ninja -C build shell test-z3`
- `build\test-z3.exe fpa`
- exact SMT-LIB reproducer: `unsat`
- `build\test-z3.exe /a`: 105 passed, 0 failed
- basic SMT shell and command-line help smoke tests

## Dafny verification artifact

The compact Dafny refutation harness published by the NFV audit is included below. It models the unsigned-to-signed narrowing in the original implementation and machine-checks the concrete witness `rm = RNE`, `ebits = 2`, `sbits = 0xffffffff`, and `x = 4`. `MaxFinite` and `MaxExp` are the explicitly trusted declarations described by the audit.

<details>
<summary>Dafny refutation for `integer_conversion_infinity_thresholds`</summary>

```dafny
// synth: rounding-mode representation
datatype RoundingMode = RNE | RNA | RTP | RTN | RTZ

// synth: signed interpretation of a 32-bit C++ int
function Signed32(value: bv32): int
// synth: signed interpretation definition
{
  // synth: two's-complement interpretation
  if value < 0x80000000 then value as int else (value as int) - 0x100000000
}

// synth: signed interpretation of a 64-bit mpf_exp_t
function Signed64(value: bv64): int
// synth: signed interpretation definition
{
  // synth: two's-complement interpretation
  if value < 0x8000000000000000 then value as int else (value as int) - 0x10000000000000000
}

// synth: exact largest finite value postulated by the documented environment
function {:axiom} MaxFinite(ebits: bv32, sbits: bv32): real
// synth: documented largest finite value for the refutation format
  ensures ebits == 2 && sbits == 0xffffffff ==>
    MaxFinite(ebits, sbits) == 4.0 - PowerOfTwo(-4294967293)

// synth: exact maximum unbiased exponent postulated by the documented environment
function {:axiom} MaxExp(ebits: bv32): bv64
// synth: documented maximum exponent for the refutation format
  ensures ebits == 2 ==> MaxExp(ebits) == 1

// synth: exact rational power of two
function PowerOfTwo(exponent: int): real
// synth: termination measure for exact rational power of two
  decreases if exponent >= 0 then exponent else -exponent
// synth: exact rational power-of-two definition
{
  // synth: recursive exact rational arithmetic
  if exponent == 0 then 1.0
  else if exponent > 0 then 2.0 * PowerOfTwo(exponent - 1)
  else PowerOfTwo(exponent + 1) / 2.0
}

// synth: bounds for nonpositive exact powers of two
lemma PowerOfTwoNonpositiveBounds(exponent: int)
  requires exponent <= 0
  ensures 0.0 < PowerOfTwo(exponent)
  ensures PowerOfTwo(exponent) <= 1.0
  decreases -exponent
{
  // synth: inductive step toward exponent zero
  if exponent < 0 {
    // synth: recurse on the next larger exponent
    PowerOfTwoNonpositiveBounds(exponent + 1);
    // synth: expose the negative-exponent defining equation
    assert PowerOfTwo(exponent) == PowerOfTwo(exponent + 1) / 2.0;
  } else {
    // synth: base case
    assert exponent == 0;
    // synth: expose the zero-exponent defining equation
    assert PowerOfTwo(exponent) == 1.0;
  }
}

// synth: C++ signed-int subtraction and addition, with arbitrary result on signed overflow
method CIntSubAddOne(left: bv32, right: bv32) returns (value: bv32)
// synth: the defined signed result must be representable by the frozen bv32 cast
  requires 0 <= Signed32(left) - Signed32(right) + 1
// synth: exact result when signed arithmetic does not overflow
  ensures Signed32(left) - Signed32(right) + 1 <= 0x7fffffff ==>
    value == (Signed32(left) - Signed32(right) + 1) as bv32
// synth: executable model of defined and undefined signed arithmetic
{
  // synth: mathematical value used to detect signed overflow
  var mathematical := Signed32(left) - Signed32(right) + 1;
  // synth: defined signed-int arithmetic case
  if -0x80000000 <= mathematical && mathematical <= 0x7fffffff {
    // synth: cast lower bound supplied by the helper precondition
    assert 0 <= mathematical;
    // synth: exact 32-bit representation in the defined case
    value := mathematical as bv32;
  } else {
    // synth: signed overflow is reachable and has an arbitrary result
    value :| true;
  }
}

// src: source.cpp:1 | expr_ref fpa_rewriter::mk_is_inf_of_int(mpf_rounding_mode rm, unsigned ebits, unsigned sbits, expr * int_expr) {
method fpa_rewriter_mk_is_inf_of_int(rm: RoundingMode, ebits: bv32, sbits: bv32, x: int) returns (result: bool)
// synth: frozen precondition
  requires ebits >= 2
// synth: frozen precondition
  requires sbits >= 2
// synth: frozen precondition
  requires rm == RNE || rm == RNA || rm == RTP || rm == RTN || rm == RTZ
// synth: refutation goal: negation of the translated postcondition
  ensures !((result == (if rm == RNE || rm == RNA then (x as real) >= MaxFinite(ebits, sbits) + Power

... truncated at 4000 characters; the full artifact is in the run's output directory.
```

</details>


NikolajBjorner:
Investigated the Ubuntu Python-make Debug failure. The new constant-path overflow check reached `mpf`, whose `sbits` field is 16 bits; `4294967295` was truncated to `65535`, so constant folding returned an FP numeral with the wrong sort and triggered `rewriter_def.h`'s sort-preservation assertion. Commit `204238ff5` makes `mk_to_fp` leave formats above the `mpf` storage limit symbolic, allowing the enclosing integer `fp.isInfinite` rewrite to decide the formula without constructing a truncated numeral. Verified with the exact NFV reproducer (`unsat`), focused Debug `fpa`, and the full Debug suite (105/105).

PR #10814, opened 2026-09-15
macro_manager: do not create macros for functions used in recursive definitions

### LINKED ISSUE #10813: smt.macro_finder gives false sat when the macro head occurs inside a define-fun-rec body
## Summary

With `smt.macro_finder=true`, z3 answers `sat` on unsatisfiable inputs when a universally quantified "bridge" axiom `(forall (x) (= (f x) (g x)))` defines an uninterpreted `f` in terms of a `define-fun-rec` `g` whose body refers back to `f`. The model z3 returns does not satisfy the assertions: `(eval (f n))` on z3's own model is `false` while `(f n)` is asserted.

Reproduced on the 4.15.2 release and on current `master` (2d2fb04, built from source).

## Reproducer

```smt2
(set-option :auto_config false)
(set-option :smt.mbqi true)
(set-option :smt.macro_finder true)
(declare-fun f (Int) Bool)
; "mirror" of f: a define-fun-rec whose body recurses through the uninterpreted f, not through g itself
(define-fun-rec g ((x Int)) Bool (ite (> x 0) (f (- x 1)) false))
(assert (forall ((x Int)) (! (= (f x) (g x)) :pattern ((f x)))))
(declare-const n Int) (declare-const m Int)
(assert (= n (+ m 1)))
(assert (<= m 0))
(assert (f n))        ; f n = g n = (n>0 ? f m : false) = f 0 or false = false  => must be unsat
(check-sat)
(get-model)
(eval (f n))
```

Output:

```
sat
(
  (define-fun n () Int 1)
  (define-fun m () Int 0)
  (define-fun f ((x!0 Int)) Bool (g x!0))
)
false
```

Expected: `unsat`. Any of the following gives `unsat`: `smt.macro_finder=false`; writing `g` as a plain `define-fun`; or making `g` call itself instead of `f`.

(Using literal arguments, e.g. `(assert (= n 1))`, hides the bug because the rewriter unfolds the recursive definition on values at preprocessing time.)

## Cause

Two behaviours combine:

1. `g`'s body contains no call to a recursive function, so the recfun plugin classifies it as a *macro* definition (`:recfun-macro-expansion` in `-st` output) and it is inlined lazily by the recfun theory during search, not at preprocessing.
2. `macro_finder` recognizes the bridge axiom as a macro `f := g` (the RHS is a recfun application, so `f` does not occur in it), rewrites `f` in the *asserted formulas*, and drops the axiom. The occurrence of `f` inside `g`'s definition body is not an asserted formula and is never rewritten.

When the theory later inlines `g n`, it exposes `f m`, an uninterpreted term with no remaining axiom, which z3 is free to set to `true`. The model converter then re-adds `f := g`, which is why `eval` contradicts the verdict. The same happens for mutually recursive definitions where one body refers to the bridged twin of another.

`eliminate_predicates` (the newer simplifier) already documents "forbidden from macros: uninterpreted functions in recursive definitions"; the legacy `macro_manager` used by `smt.macro_finder` / `asserted_formulas` has no such guard.

## Real-world context

The reproducer is a minimized version of an encoding produced by a verification toolchain: every recursive function `F` is emitted as an uninterpreted `F`, a `define-fun-rec F_LRec` whose body calls the uninterpreted twins (so definitions can be emitted in any order without cyclic dependencies), and a bridge axiom `forall x. F x = F_LRec x`. A valid goal about list equality (needing induction, so `unknown`/timeout expected) is reported `sat` with a counterexample that does not satisfy the premises.

## Fix

PR follows: make `macro_manager::insert` refuse to create a macro for a function symbol that occurs in the body of any recursive function definition.

etiennejf:
Fixes #10813.

## Problem

With `smt.macro_finder=true`, a bridge axiom `(forall (x) (= (f x) (g x)))`, where `g` is a `define-fun-rec` whose body mentions the uninterpreted `f`, is accepted as the macro `f := g`. Macros are expanded in the asserted formulas only; the body of `g` is not an asserted formula and is unfolded lazily by the recfun theory. Once the axiom is dropped, the occurrences of `f` inside `g`'s body have no defining axiom left. z3 then answers `sat` on unsatisfiable inputs and returns a model that does not satisfy the assertions (its own `(eval (f n))` is `false` while `(f n)` is asserted). Reproducer in #10813; the same happens for mutually recursive definitions where one body refers to the bridged twin of another.

## Fix

`macro_manager::insert` now refuses to create a macro for a function symbol that occurs in the body of any recursive function definition. This mirrors the rule `eliminate_predicates` already applies ("forbidden from macros: uninterpreted functions in recursive definitions"). The check is a no-op when the recfun plugin is absent or has no definitions, so inputs without `define-fun-rec` are unaffected.

## Testing

Built from source on Linux (`mk_make.py`, g++).

| input | before | after |
|---|---|---|
| reproducer from #10813 | `sat` (model violates assertions) | `unsat` |
| list-equality encoding it was minimized from (quantifier-free instance) | `sat` | `unsat` |
| same encoding, valid goal needing induction | `sat` (bogus) | timeout, as expected |
| macro axiom `f x = x+1`, no recursive definition | `unsat`, quantifier eliminated | unchanged |
| macro axiom `f x = x+1` next to an unrelated `define-fun-rec` | `unsat`, quantifier eliminated | unchanged |

Also ran the ~30 SMT-LIB queries of the verification toolchain this came from (recursive datatypes, define-fun-rec, macro_finder on): the only verdict changes are the two reproducers (`sat` → `unsat`) and the original goal (`sat` → timeout). One query that previously returned `unknown` after unsound macro elimination now keeps its quantifiers and times out at 10s instead.


etiennejf:

PR #10818, opened 2026-09-15
opt: replace NLQSAT limit checks with native NLSAT projection

levnach:
## Summary

Replace the NLQSAT-based optimization checks with native `nlsat::solver::check_limit`, shared by finite open-supremum and unboundedness queries.

The checker selects a satisfying arithmetic cube, projects away all variables except the objective, and checks whether the resulting feasible region approaches the requested limit or is unbounded above. Regions that do not establish the limit are blocked and the search continues. This removes the quantified-formula construction and NLQSAT tactic calls from `opt_nlsat`.

Keep exact rational/algebraic endpoints, feasible model witnesses, proof budgets, cancellation, and conservative `unknown` results. Finite open suprema still require a separate strict-upper-bound proof. Integer variables remain unsupported by the limit checker; unattained non-final lex objectives still return `unknown`.

## Coverage

Native NLSAT/optimization and public API tests cover finite and infinite limits, Boolean branches, varying real witnesses, algebraic endpoints, scopes, resource limits, and lex/box results.

Companion regression updates: Z3Prover/z3test#69. Regression CI needs the matching `z3test` expectations.

## Publication status

The native replacement is implemented in the local working tree but is not yet committed or pushed. The currently published commits still use NLQSAT.


levnach:
Superseded by Z3Prover/z3#10847, which contains the final native NLSAT limit checker and optimizer/API integration in a single commit (ff4b22106). The replacement goes directly from the existing NLQSAT queries to native NLSAT projection, without the intermediate implementation history. Companion regression updates remain in Z3Prover/z3test#69.

Issue #10825, opened 2026-09-16
Option to detect (mutually) recursive function definitions given as axioms (enables MBQI counterexamples)


etiennejf:
### Motivation

Tools that translate a source language into SMT often cannot use `define-fun-rec` / `define-funs-rec` directly: to avoid ordering and cyclic-dependency problems between many generated definitions, they emit each recursive function `f` as an *uninterpreted* symbol plus a universally quantified **defining axiom**

```smt2
(assert (forall ((x ...)) (! (= (f x) body[x, f]) :pattern ((f x)))))
```

where `body` refers back to `f` (or, for mutual recursion, to the other functions' uninterpreted symbols). Semantically this is the same function, but z3 treats the axiom as an ordinary quantified assertion: it is instantiated by E-matching / MBQI rather than unfolded by the recursive-function theory.

For **unsat** goals this is usually fine. For **sat** goals — i.e. when we want a **counterexample** to a conjecture about a recursively defined function — it is not: to answer `sat`, MBQI has to build a model of the defining axiom, and for a genuinely recursive `body` the instantiation does not terminate, so z3 diverges (`unknown` / timeout) even when a small counterexample exists.

Recognising that such an axiom **is** a recursive definition and registering it as one (as if it had been written with `define-funs-rec`) fixes this: the function is unfolded lazily and depth-bounded by the recfun theory, and **MBQI can then build a finite model and return the counterexample**. In short, detecting recursive functions at the axiom level is what makes MBQI usable for counterexample detection on these encodings.

### Proposal

A preprocessing option `smt.recfun_finder` (off by default) that, before solving:

1. finds axioms of the form `(= (f X) body)` (and the `iff`, negated-Boolean and arithmetic-normal-form variants) where `f` is uninterpreted and applied to exactly the bound variables, with a single defining axiom, and where `body` may contain `f` (the recursion) and be any expression — not only `ite`/arithmetic;
2. groups them by the strongly connected components of their call graph (so mutually recursive functions are found together), including the case where `body` is the "mirror" `g(X)` of an already-declared recursive helper;
3. registers each recursive component as a group of recursive functions, removes the axiom, replaces `f` by the fresh recursive symbol, and keeps `f(X) = f'(X)` as a macro so models still interpret `f`.

### Soundness

[...]

PR #10826, opened 2026-09-16
Add smt.recfun_finder: detect (mutually) recursive definitions given as axioms

etiennejf:
Implements the feature requested in #10825.

## What this adds

A preprocessing option **`smt.recfun_finder`** (off by default). When on, `asserted_formulas::find_recfuns_core` runs first in `reduce()` and turns quantified **defining axioms** into recursive function definitions, as if the functions had been declared with `define-funs-rec`.

Translators that generate SMT often emit each recursive function `f` as an uninterpreted symbol plus an axiom `(forall (X) (! (= (f X) body[X,f]) :pattern ((f X))))` (to avoid ordering/cyclic dependencies between many definitions). z3 then treats the axiom as an ordinary quantified assertion. **Detecting that this is a recursive definition and registering it as one favours MBQI for counterexample detection**: the function is unfolded lazily by the recfun theory, so MBQI can build a finite model and return a `sat` counterexample instead of diverging on unbounded axiom instantiation. See #10825 for the three worked examples (single-recursive, mutually recursive, and a no-`ite` arithmetic body); each times out with the option off and returns `sat` with a concrete counterexample with the option on.

## How it works

- **Detect** `(= (f X) body)` / `(= body (f X))`, the negated-Boolean forms, and the arithmetic normal form `(= (+ (f X) t) c)` (via `macro_util::is_arith_macro`) that the rewriter produces for a numeric body without `ite`. `f` must be uninterpreted, applied to exactly the bound variables, with a single defining axiom. Unlike `is_simple_macro` the head **may** occur in the body (the recursion), and the body can be any expression, not only `ite`/arithmetic.
- **Mirror** case: if `body` is `g(X)` for an existing recursive `g` whose own body refers back to uninterpreted candidates, `g` is inlined one step so recursion that only closes through the axioms becomes visible; `g` is then redefined as the alias `g(X) = f'(X)`.
- **Group** candidates by the SCCs of their call graph (mutual recursion found together); each cyclic component becomes a group of fresh recfun symbols. `f` is replaced by `f'` everywhere, the axiom removed, and `f(X) = f'(X)` kept as a macro so models still interpret `f`.
- **Guards**: Boolean `ite` that the simplifier turned into `or`/`and` is rebuilt around the recursive calls. A component is dropped if its unguarded recursive calls form a cycle (an unconditional recursive macro is unfolded eagerly and would loop), if a head occurs in a surviving non-macro recursive body (the macro `f = f'` could not reach it), or if the body contains a lambda (unsupported by the recfun theory).
- `macro_manager::occurs_in_recursive_definition` now ignores `define-fun` macro definitions (their applications are expanded by the parser). `smt_setup` registers `theory_recfun` when the option is on. `-v:10` reports the number of definitions; `-v:11` traces detection decisions.

## Soundness

Always sound for **`unsat`** (every unfolding is an instance of the axiom). For **`sat` / counterexamples the option assumes that every detected definition terminates**, exactly the assumption behind `define-funs-rec`; z3 does not check termination. Hence it is opt-in and off by default, and the guard/lambda restrictions above keep the transformation from turning a divergent axiom into a wrong `sat`.

## Testing

Built from source (Linux, g++). Verified on the three examples from #10825 (timeout → `sat` + valid counterexample; counterexamples re-checked with `eval`: `len = 3`, `even k` with `k = 2`, `sizeT = 5`), on mirror-encoded mutually-recursive predicates (a real translator output: timeout → `sat` in ~1s with a correct model), and that an axiom whose body contains a lambda is skipped without aborting. Existing behaviour with the option off is unchanged. I have **not** yet run the full z3 regression suite; guidance on the preferred place for regression tests welcome.


NikolajBjorner:
this should be a simplifier. I don't want to add code to asserted_formulas. 
You can chain simplifiers as pre-processing by setting them in SMTLIB and over the API.
See the z3 tutorial microsoft.github.io/z3guide for material on the simplifier.

NikolajBjorner:
https://microsoft.github.io/z3guide/docs/strategies/simplifiers

PR #10828, opened 2026-09-16
Avoid quadratic blowup on long fixed-length strings (fixes #10827)

### LINKED ISSUE #7745: Incorrect SAT result for string query
Z3 returns SAT for a query that should be UNSAT. 

## To reproduce
Run z3 on the following query:
```smt
(set-logic QF_S)
(declare-const xs String)
(declare-const ys String)
(assert (= (str.++ (str.substr ys 0 18) (str.substr xs 0 18)) (str.++ "It")))
(assert (> (str.len ys) 18))
(check-sat)
```
**Expected**: unsat  
**Actual**: sat  

A string concatenation of 18 characters should not equal a 2 character string "It".

### LINKED ISSUE #10827: Quadratic blowup on long fixed-length strings since 4.15.3 (20 s / 3–4.7 GB for (= 256 (str.len s)), :timeout not honored)
Since 4.15.3 a single fixed-length constraint on a string of length 256 takes 20–33 s and 3–4.7 GB (4.15.2: 4 s, 0.46 GB), and `:timeout` is not honored while it runs.

### Reproducer 1 — `fixed_len_256.smt2`

```smt2
(declare-const s String)
(assert (= 256 (str.len s)))
(check-sat)
```

| version | result | wall | max RSS |
|---|---|---|---|
| 4.15.2 | sat | 4.1 s | 0.46 GB |
| 4.15.3 | sat | 32.7 s | 4.67 GB |
| 4.15.4 | sat | 33.1 s | 4.68 GB |
| 4.16.0 | sat | 32.7 s | 4.68 GB |
| 5.0.0 | sat | 21.7 s | 3.19 GB |
| 5.1.0 | sat | 20.9 s | 3.19 GB |
| master d5d9266 | sat | 21.2 s | 3.18 GB |
| master + #10828 | sat | **2.1 s** | **0.98 GB** |

`(>= (str.len s) 256)` behaves the same. Length 20 is instant, 64 takes 0.4 s.

### Reproducer 2 — `hash_len_timeout.smt2` (`:timeout` not honored)

```smt2
(set-option :smt.mbqi true)
(set-option :auto_config false)
(set-option :timeout 5000)
(declare-fun hash (String) String)
(declare-const x String)
(assert (forall ((a String) (b String))
  (! (=> (= (hash a) (hash b)) (= a b)) :pattern ((hash a) (hash b)))))
(assert (forall ((s String)) (! (= 256 (str.len (hash s))) :pattern ((hash s)))))
(assert (not (< (str.len (hash x)) 256)))
(check-sat)
```

The expected answer is `unknown` (the conjecture is false and the premises are not refutable), but it should arrive at ~5 s. Instead:

| version | result | wall | max RSS |
|---|---|---|---|
| 4.15.2 | unknown | 5.9 s | 0.80 GB |
| 4.15.3 / 4.15.4 / 4.16.0 | unknown | 23 s | 3.44 GB |
| 5.0.0 / 5.1.0 / master | unknown | 19 s | 1.84 GB |
| master + #10828 | unknown | **5.3 s** | 1.06 GB |

A variant with two such hash functions takes 117 s and 3.7 GB on 5.1.0 before reporting `unknown`. The hard timeout (`-T:5`) works; the soft one only takes effect once the current round ends.

### Analysis

`theory_seq::fixed_length` expands a string of fixed length n into `unit(nth 0) ++ ... ++ unit(nth n-1)`; the equation solver then creates one concatenation per suffix, n(n+1)/2 = 32 896 terms for n = 256 (`:seq-num-reductions 32898`), each with its own `str.len` term, length axiom and arithmetic bounds. This happens in the initial base-level propagation of `context::check` (before `search()`), where nothing checks for cancellation — hence the ignored soft timeout.

Bisecting 4.15.2..4.15.3 gives **0528c8690 (fix #7745)** as the first bad commit (parent: 6.5 s; commit: 26.6 s on reproducer 2). Two contributing causes:

1. `seq::axioms::length_axiom` previously rewrote `len(x)` for a concatenation `x`, which flattens nested concatenations and folds unit lengths into a numeral. Since 0528c8690 it emits `len(a) + len(b)` literally. For the 33 152 binary concatenations `unit ++ suffix` created above, each gets a fresh arithmetic length term and row: `arith-max-columns` 66k → 133k, `arith-eq-adapter` 35k → 67k, memory ×4.
2. `theory_lra::imp::flush_bound_axioms` groups the bounds queued before search by popping one and rescanning the whole pending list for its variable. Here 131 328 bounds on 67 071 distinct variables are pending in a single call, so the grouping is quadratic: `perf` attributes 83 % (5.1.0) / 94 % (4.15.2) of the run to this function's own loop, ~12 s on master, with no cancellation check inside.

Fixing (1) alone brings reproducer 1 back to 2 s; fixing (2) alone removes the 12 s but not the memory. The PR referenced below does both; the #7745 reproducer stays `unsat`.

etiennejf:
Fixes #10827: since 4.15.3 (commit 0528c8690, fix #7745) a fixed-length constraint on a 256-character string takes 20–33 s and 3–4.7 GB, and `:timeout` is not honored during that time.

Two independent changes:

**`seq::axioms::length_axiom`** — for a concatenation, flatten nested concatenations (`get_concat`) and fold components of known length (units, string literals, empty) into one numeral, as the rewriter did before 0528c8690. Other components still become `len(e)` terms with their own axioms, so the #7745 fix (explicit `len(extract ...)` axioms instead of nested rewriting) is preserved: its reproducer stays `unsat`. The expansion of a fixed-length string into n characters creates one `unit ++ suffix` concatenation per suffix; emitting `len(unit) + len(suffix)` for each of them added a fresh arithmetic term and row per suffix (`arith-max-columns` 66k → 133k, memory ×4).

**`theory_lra::imp::flush_bound_axioms`** — group the bounds queued before search by sorting them by variable instead of rescanning the pending list once per variable. On these problems 131k bounds over 67k variables are pending in one call, so the grouping was quadratic (~12 s, 83–94 % of the run in `perf`) and ran without any cancellation check.

### Measurements (Linux x64, same tree, unloaded; both files are in #10827)

`fixed_len_256.smt2` — `(assert (= 256 (str.len s)))`, no timeout:

| build | result | wall | max RSS |
|---|---|---|---|
| master d5d9266 | sat | 21.2 s | 3.18 GB |
| `flush_bound_axioms` change only | sat | 6.6 s | 1.8 GB |
| `length_axiom` change only | sat | 2.0 s | 0.98 GB |
| both (this PR) | sat | **2.1 s** | **0.98 GB** |
| 4.15.2, for reference | sat | 4.1 s | 0.46 GB |

`hash_len_timeout.smt2` — quantified length/injectivity axioms on a hash function, `:timeout 5000`, false conjecture (expected `unknown`, at the timeout):

| build | result | wall | max RSS |
|---|---|---|---|
| master d5d9266 | unknown | 18.7 s | 1.83 GB |
| `flush_bound_axioms` change only | unknown | 6.7 s | 1.9 GB |
| `length_axiom` change only | unknown | 8.4 s | 1.1 GB |
| both (this PR) | unknown | **5.3 s** | **1.06 GB** |

The remaining cost is the O(n²) suffix expansion itself (2 s / 1 GB for n = 256), which this PR does not touch.

Checks: #7745 reproducer `unsat` before and after; 37 local SMT files (strings, quantifiers, datatypes, recursive functions) give identical verdicts before and after, except the three affected by this issue which now stop at their timeout.


src/ast/seq/seq_axioms.cpp
```diff
@@ -1375,9 +1375,30 @@ namespace seq {
...
+                else if (seq.str.is_string(e, str))
+                    k += rational(str.length());
+                else if (seq.str.is_empty(e))
+                    ;
+                else 
+                    args.push_back(seq.str.mk_length(e));
+            }
+            if (!k.is_zero() || args.empty())
+                args.push_back(a.mk_int(k));
             expr_ref len(a.mk_add(args), m);
```
NikolajBjorner: we can call the rewriter on len, instead of performing ad-hoc rewriting.

etiennejf: Done in d56261150: the sum `len(a) + len(b) + ...` is now built and passed to `m_rewrite`, no hand folding. Same effect on the fixed-length expansion (`arith-max-columns` 66k as before, 2.3 s / 0.98 GB on `fixed_len_256.smt2`, 5.8 s on `hash_len_timeout.smt2`), and the #7745 reproducer plus a few substring-length variants (offsets, symbolic lengths, unit concatenations) keep their verdicts.

NikolajBjorner: I am asking copilot to split this PR into two

src/smt/theory_lra.cpp
```diff
@@ -2965,18 +2965,20 @@ class theory_lra::imp {
...
 
         while (!m_new_bounds.empty()) {
+            // Group the pending bounds by variable with one sort instead of
+            // rescanning the pending list once per variable: the list can hold
+            // every bound atom created before search starts (over 100k on
+            // string problems with long fixed-length strings), which made the
+            // grouping quadratic and uninterruptible.
+            lp_bounds pending(m_new_bounds);
+            m_new_bounds.reset();
+            std::sort(pending.begin(), pending.end(),
```
NikolajBjorner: std::stable_sort

PR #10830, opened 2026-09-16
Kill the legacy Python build system

### LINKED ISSUE #10812: Should we remove the Python build?
In the discussion of PR #10753, @NikolajBjorner asked:

> killing the python build has been a dream for ages. Should we :-).

I think we could indeed remove it as CMake now builds a _superset_ of what the generated Makefiles can build.

But we can go further. I took a closer look at the tactic, memory-initializer, and gparams registration generators and I think they can be fully replaced with C preprocessor tricks (namely X-macros) and safe (non-throwing) C++ static initializers.

Then we could remove quite a few things:

1. The whole Makefile-generation system in Python
2. The pyg2hpp system, including mk_gparams_register_modules_cpp.py, mk_mem_initializer_cpp.py, and mk_install_tactic_cpp.py.
3. The custom transitive properties in CMake; this should let us relax the minimum requirement to 3.26. Good that we now have testing in place to confirm this 🙂

How far do we want to go?

alexreinking:
This removes the legacy `mk_make.py`/`mk_util.py`/`mk_project.py` Python build system (and its associated test suite and helper scripts), now that CMake is the sole build system for Z3.

`scripts/mk_unix_dist.py` and `scripts/mk_win_dist.py` are rewritten to drive the CMake build directly for producing release distributions (folding in the logic that previously lived in the separate `mk_unix_dist_cmake.py`/`mk_win_dist_cmake.py` variants, which are removed), with their shared helpers factored into a new `scripts/mk_dist_util.py`.

CI workflows, READMEs, and Dockerfiles are updated to configure and build with `cmake -S . -B build` / `cmake --build build --parallel $(nproc)` instead of `mk_make.py`/`make`. The `examples/dotnet` CMake target is now marked `EXCLUDE_FROM_ALL`, like the other example targets, so it's no longer built by default outside of jobs that explicitly ask for it; the OCaml findlib install step in CI now installs via a directory glob against the CMake build tree; and `scripts/build-win-signed.yml` is updated to match the `build-dist/<arch>/dist/...` output layout produced by the rewritten `mk_win_dist.py`. `examples/java/README` is corrected to describe how to actually build and run the Java example under CMake.

Fixes #10812


NikolajBjorner:
@copilot - examine the ci tasks to remove what may be essentially duplicates: some of the build steps using python build system may have been there just to test the python build system.

NikolajBjorner:
@copilot - do the builds also retain running examples: 
In other words, is there any related to : make -j$(nproc) examples

PR #10839, opened 2026-09-17
Replace .pyg/pyg2hpp.py parameter codegen with hand-written headers

alexreinking:
Each `<module>_params.hpp` was generated at build time by evaluating a `.pyg` file (a Python snippet) through `pyg2hpp.py`. Replace this with a plain X-macro parameter list per module, expanded via a single `Z3_DEFINE_MODULE_PARAMS(class, module, LIST)` macro in `util/params.h` that generates the struct, `collect_param_descrs`, and accessors. This removes the .pyg files, scripts/pyg2hpp.py, and the PYG_FILES/`add_custom_command` machinery in `z3_add_component.cmake` entirely.

These headers are now ordinary, hand-maintained C++ sources, but are very similar in content and structure to the old `.pyg` files.

---


NikolajBjorner:
this is a welcome change. Great to omit level of indirection with pyg.

alexreinking:
> this is a welcome change. Great to omit level of indirection with pyg.

Shall I merge, then?

PR #10840, opened 2026-09-17
Replace generated registration with C++ self-registration

alexreinking:
This PR removes CMake/Python-generated registration sources in favor of C++ self-registration for Z3’s built-in extension points.

- Self-register parameter modules and global parameter descriptions, fixing missing factor parameter registration.
- Self-register memory initializers and finalizers while preserving initializer priority ordering.
- Self-register tactics, probes, and simplifiers through lazy factories, restoring registrations from unlinked components.
- Remove the associated CMake metadata plumbing and three obsolete generator scripts.

---


alexreinking:
> Also fixed a bug: the `ufbv-rewriter` tactic was inadvertently registered using the `quasi-macros` tactic’s factory. This change gives it its own factory registration, so invoking `ufbv-rewriter` constructs the intended tactic.

It seems fixing this apparent bug actually broke a test. @NikolajBjorner — what's the right move here? Should `z3test/regressions/smt2/3688.smt2` be updated? Currently, it's

```smt
(set-option :proof true) 
(declare-fun a (Int Int Real) Real) 
(assert (= 0 0)) 
(assert (forall ((?b Real)) (= (a 0 0 ?b) 0.0))) 
(check-sat-using ufbv-rewriter)
```

Should that be `(check-sat-using ufbv)` instead?

src/tactic/ufbv/ufbv_rewriter_tactic.h
```diff
@@ -19,12 +19,11 @@ Module Name:
...
+#include "tactic/tactic.h"
 class ast_manager;
 class tactic;
 
 tactic * mk_ufbv_rewriter_tactic(ast_manager & m, params_ref const & p = params_ref());
 
-/*
-  ADD_TACTIC("ufbv-rewriter",  "Applies UFBV-specific rewriting rules, mainly demodulation.", "mk_quasi_macros_tactic(m, p)")
-*/
+Z3_ADD_TACTIC(ufbv_rewriter, "ufbv-rewriter", "Applies UFBV-specific rewriting rules, mainly demodulation.", mk_ufbv_rewriter_tactic(m, p));
```
alexreinking, line 28: This was the buggy tactic registration

alexreinking:
I moved the `ufbv-rewriter` bugfix to #10885 to validate this PR has no _other_ unintended functional regressions.

PR #10885, opened 2026-09-21
Fix ufbv-rewriter to use its own tactic factory

alexreinking:
The `ufbv-rewriter` tactic was inadvertently registered using the `quasi-macros` tactic's factory. This gives it its own factory registration, so invoking `ufbv-rewriter` constructs the intended tactic.

It seems fixing this apparent bug actually broke a test. @NikolajBjorner — what's the right move here? Should `z3test/regressions/smt2/3688.smt2` be updated? Currently, it's

```smt
(set-option :proof true) 
(declare-fun a (Int Int Real) Real) 
(assert (= 0 0)) 
(assert (forall ((?b Real)) (= (a 0 0 ?b) 0.0))) 
(check-sat-using ufbv-rewriter)
```

Should that be `(check-sat-using ufbv)` instead?

---


NikolajBjorner:
hmm, quasi_macros were developed at same time as ufbv features.

NikolajBjorner:
> The `ufbv-rewriter` tactic was inadvertently registered using the `quasi-macros` tactic's factory. This gives it its own factory registration, so invoking `ufbv-rewriter` constructs the intended tactic.
> 
> It seems fixing this apparent bug actually broke a test. Nikolaj Bjorner (@NikolajBjorner) — what's the right move here? Should `z3test/regressions/smt2/3688.smt2` be updated? Currently, it's
> 
> ```smt
> (set-option :proof true) 
> (declare-fun a (Int Int Real) Real) 
> (assert (= 0 0)) 
> (assert (forall ((?b Real)) (= (a 0 0 ?b) 0.0))) 
> (check-sat-using ufbv-rewriter)
> ```
> 
> Should that be `(check-sat-using ufbv)` instead?
> 
> Stack created with [GitHub Stacks CLI](https://github.com/github/gh-stack) • [Give Feedback 💬](https://gh.io/stacks-feedback)

Looks to me it should be using a quai-macro rewriter tactic if available.
There is nothing ufbv in this example so out of place to rely on ufbv functionality for this.

NikolajBjorner:
@copilot - can you fnd a substitute for ufbv-rewriter tactic, manybe something that use quasi-macros in a tactic?

NikolajBjorner:
updated regression tests, rebuilding

